PR #60, opened 2016-09-15
Reg free COM for DIA

AbhitejJohn:
Fix to make sure DIASession does not require COM registration in the registry. Using Reg Free Com.

Todo:
- Will wire in the msdia dll's as a nuget package and change the build script.


saikrishnav (commented):
👍

PR #87, opened 2016-09-26
Handle VS Test console crash

smadala:
- Add ProcessExited Event to IProcessManager
- Cache pathToExtensions in VSTestConsoleWrapper to restart vstest.console 
- Make Async read from stream using signal
- Add Unittests for above functionalities


codito (changes requested):

saikrishnav (changes requested):
I am not approving this PR yet. I don't like the way the changes are inside VsTestConsoleRequestSender especially. I would like a PR which does not touch  VsTestConsoleRequestSender  too much. There is lot of duplicate code in places. 

Also, I thought we decided on not caching the "pathToExtensions". Why is it still there?

I believe we can achieve the experience we are looking for with a less complex looking PR.

codito (approved):

PR #171, opened 2016-10-30
Enable unix workspace

codito:
- Fix sign issue with unix build. Workaround for dotnet-cli #2862.
- Fix case sensitive file system operations


Faizan2304 (approved):
We should bring case insensitive search in practice because our code will run in Linux as well.

PR #185, opened 2016-11-06
Fix path to dotnet executable.

codito:
In case installation from msi, dotnet executable path may contain whitespaces.
Wrap the path in whitespaces before spwaning the process.


singhsarab (commented):
👍

PR #316, opened 2017-01-05
Convert TestPlatform.vsix from V2 to V3 format.

Faizan2304:
Due to limitation with dotnet core tooling, we are going through zip approach.


singhsarab (approved):
👍

PR #327, opened 2017-01-09
Added thirt party notice to nuget packages using third party package

Faizan2304:


AbhitejJohn (approved):
Just check if we can include newtonsoft.dll in a package we ship using the same notices file.

PR #328, opened 2017-01-09
Supporting Multiple TestProperty with the same key value.

navin22:
- Changed the way of deserialization using custom TraitObject.
- Added UnitTests for the same.


codito (commented):
Approach looks fine. Request few changes.

PR #332, opened 2017-01-12
Fix assembly loading in symbol loader.

codito:
Instead of attempting to load the test assembly, try to query for it. It should
have been already loaded by the test adapter.

In netcoreapp1.1, attempting to load the same executable within itself throws.


singhsarab (approved):
👍

PR #335, opened 2017-01-13
relative path for test dlls as arguments

mayankbansal018:


codito (commented):
Approach looks good to me. We should add tests in this PR.

PR #342, opened 2017-01-18
Find vstest.console.dll from the same directory where Microsoft.TestPlatform.targets reside

Faizan2304:
Issue: dotnet test is failing because the it is not able to find vstest.console.dll. The reason why it is not able to find vstest.console.dll  is that we are passing VSTestConsolePath ( whose current value is "vatest.console.dll") to VSTestTask which uses it as it is to create a processThis process Current working is the directory from where we are running test, hence it is not able to resolve vstets.console.dll.

Fix: Find full path of vstest.console.dll and pass it to VSTestTask.

Test Plan:
1) Manually validated the scenario by replacing Microsoft.TestPlatform.targets and Micrsosoft.Testplatform.build.dll
2) Manually enable trace log and verified that right path is getting pass to VstestTask.


singhsarab (approved):
Add the test plan please.

PR #347, opened 2017-01-18
Update ci script for public repo

codito:


tannergooding (approved):
This looks correct to me (I don't think we can validate the changes until after everything gets merged though).

PR #378, opened 2017-01-25
Localize TestPlatform.TranslationLayer nuget package 

Faizan2304:


src/TestPlatform.TranslationLayer.nuspec
```diff
@@ -50,12 +48,124 @@
     <file src="net46\$Runtime$\Microsoft.TestPlatform.CommunicationUtilities.dll" target="lib\net46\" />
     <file src="net46\$Runtime$\Microsoft.TestPlatform.CoreUtilities.dll" target="lib\net46\" />
```
codito: We can remove coreutilities as well? OM package will pull it in? Or may be make coreutilities a separate nuget package?

codito:
This change is for RTM+.

Tagging @KevinH-MS @tannergooding for any comments w.r.t usage in LUT repo.

KevinH-MS (approved):
@tannergooding should probably take a peek as well, but this looks fine to me.  We already reference the ```Microsoft.TestPlatform.ObjectModel``` package, so I ***think*** everything should just continue to work.

singhsarab (approved):

PR #381, opened 2017-01-25
Fix TL-DesignMode Thread-safety issue

saikrishnav:
Add Lock to the RunTests section (Can get triggered by TL)
Use local variables to handle registrations

TL can get the "RunComplete" event faster than vstest.console can cleanup due to asynchronous nature of the flow. TL thinks that run is complete and allows another run request. Vstest.console tries to run the next request while cleaning up the previous one. This is causing a thread-safety issue.

[PENDING UnitTests]


ManishJayaswal:
@saikrishnav please let us know if you want us to represent this fix in shiproom tomorrow if you have the bits ready to merge into d15prerel by 1p ( PST)

PR #412, opened 2017-01-31
Executor Thread ApartmentState Default to STA

saikrishnav:


smadala:
We should use PlatformAbstractions for thread operation.

sbaid:
This PR has been abandoned since STA fix is coming via mstestv2 sta extension

PR #426, opened 2017-02-01
Make adapter thread as background thread

saikrishnav:
Foreground threads block the process from terminating themselves whereas background thread do not. TPV1 sets this too.


sbaid:
This PR has been abandoned since STA fix is coming via mstestv2 sta extension

PR #429, opened 2017-02-01
Show IFrameworkHandle.SendMessage messages from test adapters.

### LINKED ISSUE #300: Tests silently fail when custom test cases fail to deserialize/serialize. (.net core)
## Description
This is in connection to https://github.com/Microsoft/vstest/issues/291. The issue ended up being that our custom test cases weren't able to be serialized/deserialized. In the end this manifested as tests not running at all without any easily consumable errors.

I debugged through multiple code bases to fix the issue; is there a route I could have taken to see the deserialization error? If not that information should definitely be made available via a warning or something of the like when running tests.

codito:
Fix #300.


codito:
@Faizan2304 @smadala @singhsarab @harshjain2 review please

singhsarab (approved):

Faizan2304 (approved):

PR #437, opened 2017-02-03
Replacing invalid xml character by unicode escape character

Faizan2304:
Issue: 

Fix: 	Replace invalid xml character by unicode escape character

Testing:
1) UT to cover changes.
2) Manually validated by creating trx file with invalid xml


src/Microsoft.TestPlatform.Extensions.TrxLogger/XML/XmlPersistence.cs
```diff
@@ -700,13 +700,23 @@ private static string RemoveInvalidXmlChar(string str)
                 // From xml spec (http://www.w3.org/TR/xml/#charsets) valid chars: 
                 // #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]  
 
-                string invalidChar = @"[^\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD\u10000-\u10FFFF]";
-                return Regex.Replace(str, invalidChar, TrxLoggerResources.PlaceHolderForInvalidChar);
+                // we are handling only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD]
```
sbaid: add unit test string to include Unicode character valid & invalid in all of these ranges

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
```diff
@@ -13,14 +13,15 @@ public class XmlPersistenceTests
         [TestMethod]
         public void SaveObjectShouldRemoveInvalidCharacter()
         {
+            System.Diagnostics.Debugger.Launch();
```
sbaid: need to remove this

src/Microsoft.TestPlatform.Extensions.TrxLogger/XML/XmlPersistence.cs
```diff
@@ -700,13 +700,23 @@ private static string RemoveInvalidXmlChar(string str)
...
+                string invalidChar = @"[^\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD]";
+                return Regex.Replace(str, invalidChar, evaluator);
             }
 
             return str;
         }
 
+        private static string ReplaceInvalidCharacterWithUniCodeEscapeSequence(Match match)
+        {
+            char x = match.Value.ToArray()[0];
```
sbaid: match.Value[0] can be used

src/Microsoft.TestPlatform.Extensions.TrxLogger/XML/XmlPersistence.cs
sbaid: do a perf run on a 10k, 1L tests on utperf06 to assess the impact.

Faizan2304: addressed in next commit

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
Faizan2304: Removed!!

src/Microsoft.TestPlatform.Extensions.TrxLogger/XML/XmlPersistence.cs
Faizan2304: Yes we can use this.

Faizan2304:
Testing:
1) Covered code change with unit test.
2) Manually validated that invalid xml char are getting replace by escape sequence and trx is getting loaded using XmlDocument.

Faizan2304:
By default invalid xml character will not get replace. To replace it with escape sequence use `` --logger:trx;EscapeInvalidChar=true``

AbhitejJohn:
This fix looks more of a workaround that is offloaded to the customer. There would also have discoverability issues trying to figure out this setting. Should we not do this by default if its known to be an invalid character?
@Faizan2304 : Please share the perf impact.

Faizan2304:
image


From the data it seems that replacing invalid xml is not affecting much on perf.

src/Microsoft.TestPlatform.Extensions.TrxLogger/XML/XmlPersistence.cs
Faizan2304: Done. Please find the result below comment.

Issue #394, opened 2017-01-30
Test projects do not get the test project icon


davkean:
_From @jnm2 on January 28, 2017 18:5_

I've been used to scanning along the row of icons to pick out the test projects in a long list of similarly-named projects. With the new .csproj format there is nothing marking the project as a test project.

Could you either:

 * Change the icon when tests are first discovered and cache the fact that it's a test project so that on subsequent opening of the project it's immediately assumed to still be a test project. When dependencies change it could be invalidated. It could also be driven off dependencies to testing frameworks maybe.

Or

 * Provide some attribute to specify which projects are intended to be run as tests.
  I strongly prefer this option because it's possible to have a project that references a testing framework or even contains tests which should not be treated as a test assembly and tests (if any) should not be discovered.

   For example, if I'm building an extension or utility library to extend a testing framework, I'll need to reference the testing framework but it's still not a test library and shouldn't get the icon.

   For another example, if you're developing a testing framework or an extension or a utility library, it's common to have a mock test assembly which references the testing framework and contains discoverable tests, but which the IDE should not discover because those mock tests are intended to fail and timeout, etc. The mock tests are used when some real tests (in an _actual_ test project which the IDE _should_ discover) point the testing framework at the mock tests to verify that certain failures are reported and certain successes are reported, timeouts are handled, and so on.

_Copied from original issue: dotnet/roslyn-project-system#1361_

davkean:
I think this should be based on the TestContainer capability @tmat suggested was to be added. The targets files of the packages that "add" test support (such as xunit, nunit, mstest, etc) should carry it and then we have a component that sets the icon based on that.

codito:
Addition of test marker from Microsoft.NET.Test.Sdk should also fix #409.

tmat:
@codito What do you meam by "test marker"? Test capability?

tmat:
@davkean Just to double-check - the projects that reference test framework nuget that you don't want to be treated as test projects could remove the TestContainer capability, correct?

codito:
Not sure yet, "capability" is one option, second option is "service guid" used by test explorer.

Are there any docs to read more on capabilities in project system? Is this just an msbuild property?

tmat:
@codito https://github.com/tmat/VSProjectSystem/blob/9be02ba5c9539adac2f6bdd56fcd72362bb40e18/doc/overview/about_project_capabilities.md
[...]

PR #440, opened 2017-02-03
Add UnitTestContainer project capability

tmat:
This capability is used to detect test projects.

Related issue: https://github.com/Microsoft/vstest/issues/394


tmat:
@srivatsn @codito

src/Microsoft.NET.Test.Sdk_props
```diff
@@ -11,6 +11,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
```
srivatsn: I believe this props file is only dropped into the netcoreapp folder of the package - so if the unit test is targeting net46, it won't have this capability. There's a different issue for which I think there should just be one props file - so maybe this is fine if that issue is fixed.

tmat:
OK to merge?

sbaid:
@tmat we can add this capability into the mstestv2 nuget & raise PR for xunit as well. That is better since it will address other projects as well - like u said. comments?

secondly - there is another build conditional support for test project that @natemcmaster is referring in #441 - we should rationalize whether to add separate property like IsTestProject or simply depend on the project capability.

tmat:
@sbaid Actually, as I learned the legacy projects doesn't support capabilities. Long term it's better but I believe we have telemetry reasons to add it to the test sdk. @srivatsn for details.

natemcmaster:
As I originally proposed in #441, would you also consider adding a global static property in addition to the addition of the ProjectCapability?

ProjectCapability is fine for VS purposes, but it's a bit cumbersome to program with. MSBuild developers would have to know MSBuild and the expected values in ProjectCapability well enough to know they need to write something like this to determine if a project is a test project:

```xml
<ItemGroup>
  <_FilteredCapabilities Include="ProjectCapabilities" Condition="'%(Identity)'=='UnitTestContainer'" />
</ItemGroup>
<PropertyGroup>
  <IsTestProject Condition="@(_FilteredCapabilities -> Count()) &gt; 0">true</IsTestProject>
</PropertyGroup>
```

It would be nice the SDK could just define `<IsTestProject>true</IsTestProject>`.

codito (approved):

srivatsn:
@natemcmaster's point is valid - it's easier to do a property check. However VS also needs the capability as that's the common mechanism that the project system uses to light up components. I think having both is fine as well.

sbaid:
@srivatsn @tmat to close on the project capability - if we close on the name as `TestContainer` 1) [This link](https://github.com/Microsoft/VSProjectSystem/blob/master/doc/overview/about_project_capabilities.md) seem to suggest a process to add new project capability. Can you pl add the right folks here?  2) will this work only for CPS projects and hence only for .NET core projects? 3) if yes - for desktop projects - is there a way to address it - It will LUT & in general scoping test discovery to not scan all projects @genlu 

pl note that test.sdk is only included in .net core templates - not in desktop templates.

sbaid:
@tmat @codito do u have any comments on adding `IsTestProject` to sdk props?

tmat:
@sbaid 
1) The name is actually ```UnitTestContainer```, not ```TestContainer```. The capability has already been added. 
2) IsHierarchy API works only for CPS projects, but the capability item can be added to any project.
3) We check for capability and also for well-known test framework references, to make LUT work for xunit, nunit and mstest in v1.0.

tmat:
```IsTestProject``` is fine.

**Comment by sbaid** [comment]
should we not name it `TestContainer` because test projects in vs work for unit testing as well as functional testing?

**Comment by tmat** [comment]
We are already using the name in LUT. I wonder if it would be useful to have both UnitTestContainer and TestContainer. The distinction is interesting for LUT, since we would likely ignore functional tests. Of course it's completely up to the user to make the distinction.

**Comment by tmat** [comment]
If we add both we should also have IsTestProject and IsUnitTestProject.

**Comment by tmat** [comment]
@ManishJayaswal For opinion.

**Comment by sbaid** [comment]
@pvlakshm @piotrpMSFT looping in

**Comment by sbaid** [comment]
I think we should add `TestContainer` & `IsTestProject` by default and let the user flag `unit test` capability by adding himself and not add by default. We are shipping only one set of templates and marking them by default `unit test` will not help since user may right functional tests in them & not change the defaults.

**Comment by tmat** [comment]
I guess it would be ok to change the name to TestContainer, since LUT doesn't work for CPS based projects yet anyways. So we can change the code in LUT that checks for the capability without breaking anything.

**Comment by bradwilson** [comment]
@sbaid said:

> we can add this capability into the mstestv2 nuget & raise PR for xunit as well.

This seems like it belongs in `Microsoft.NET.Test.Sdk`, not the test frameworks. We don't know what this capability is, and we certainly don't directly provide it. I would reject the PR to xUnit.net for this.

**Comment by codito** [comment]
@sbaid wrote:
> I think we should add TestContainer & IsTestProject by default 

+1. It will be consistent, `TestContainer` term is already used in Testwindow extensibility.

The types of test projects are numerous (functional test, load test, generic test, ordered test etc..). Each individual template should provide additional distinctions (e.g. CUIT should provide `IsFunctionalTest` etc.). Default unit testing templates don't need to add anything.

**Comment by sbaid** [comment]
@tmat @bradwilson  we can add this in `Microsoft.NET.Test.Sdk` for .net core. Do u have a suggestion how to address this for desktop test projects? currently there is no consistent way to identify test projects. 
1. There is test guid for projects created using inbox UT templates. 
2. user can create classlibrary project & add his framework of choice mstest,xunit, nunit - now this is a test project.

**Comment by smadala** [comment]
**Validation matrix** 
validation_matrix

**Review by Faizan2304 — APPROVED** [review]

File: src/Microsoft.NET.Test.Sdk.targets [hunk 2]
```diff
@@ -23,6 +23,12 @@
     <GeneratedProgramFile Condition="'$(GeneratedProgramFile)' ==''">$(IntermediateOutputPath)$(MSBuildProjectName).Program$(DefaultLanguageSourceExtension)</GeneratedProgramFile>
     <GenerateProgramFile Condition="'$(GenerateProgramFile)' == ''">true</GenerateProgramFile>
   </PropertyGroup>
+
+  <!-- Output type for dotnet core test project should be exe. For full CLR it is Library by default. -->
```
codito: This is a workaround for sdk bug. Can you please mention the bug here for tracking?

I guess the only caveat is all comparisons in the test csproj will return $(OutputType) to be Library since that's the default, and the targets are imported at the end.

codito:
LGTM

src/Microsoft.NET.Test.Sdk.targets
smadala: Added issues link.

src/Microsoft.NET.Test.Sdk_props
smadala: Made changes to copy same props file for net45 and netcore.

src/Microsoft.NET.Test.Sdk.targets
```diff
@@ -23,6 +23,12 @@
     <GeneratedProgramFile Condition="'$(GeneratedProgramFile)' ==''">$(IntermediateOutputPath)$(MSBuildProjectName).Program$(DefaultLanguageSourceExtension)</GeneratedProgramFile>
     <GenerateProgramFile Condition="'$(GenerateProgramFile)' == ''">true</GenerateProgramFile>
   </PropertyGroup>
+
```
Faizan2304: Please add the bug reference (https://github.com/dotnet/roslyn-project-system/issues/268) here. It will tell us why did we moved OutputType from props file target file. In future once the bug will get fixed we will revert it back.

smadala: Done.

src/Microsoft.NET.Test.Sdk.props
```diff
@@ -11,8 +11,10 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <ProjectCapability Include="TestContainer" />
```
davkean: I think you should be putting this in the packages themselves, I can install xUnit in a random project that doesn't have to be test project (in fact Brad mentioned they are enabling that in their new dotnet xunit runner).

We should get xUnit, NUnit, etc to all mark themselves as a TestContainer

sbaid: @davkean this was discussed earlier. Pl refer comments on feb 12th. 
@bradwilson it will be good to add this projectcapability in the xunit.runner.visualstudio nuget package. Can this be done for 2.3 release?

PR #441, opened 2017-02-03
Minor usability improvements to Microsoft.NET.Sdk.Test targets

### LINKED ISSUE #430: Generated program file on netcoreapp1.0 is in global namespace
On .NET Core, the **GenerateProgramFile** target generates a Program.cs file to ensure the app builds as OutputType=exe. This class is created in the global namespace. This means it takes precedence over any other class by that name.

Example: when converting this test: 
```cs
using Mvc.RenderViewToString;

namespace EntropyTests
{
    public class RenderViewToStringTest
    {
        [Fact]
        public  void RenderView_ReturnsContents()
        {
             Program.RenderView();
        }
   }
}
```
Program is resolved to `::Program`, not `Mvc.RenderViewToString.Program`.

**Suggestions**

- Use a less common class name, such as  `class MicrosoftNetTestSdkProgram { public static void Main() {} }`
- Or put it in a namespace to avoid collisions with user code.

**Workarounds**

- Add `using` alias `using MyProgram = Mvc.RenderViewToString.Program`. 
- `<GenerateProgramFile>false</GenerateProgramFile>` and add your own static void main.

natemcmaster:
~~1. Add a property 'IsTestProject'. This is useful for changing build behavior conditioned on this property.~~
2. Change the typename of the generates program file on .NET Core. Resolves #430 

cc @codito


singhsarab:
Changes look good.
What is the expected change in build behavior for test projects ?

Faizan2304:
@dotnet-bot test Windows / Release Build please

natemcmaster:
I've updated the PR to ensure this works for both .NET Framework and .NET Core test projects.

sbaid:
@natemcmaster thanks for the contribution! can we scope this PR to the name change of automated generated `Program`  and merge? Suggest name to be `_AutoGeneratedProgram` or `AutoGeneratedProgram` which tools don't complain.

And follow up on the build conditional support on #440

sbaid:
@tmat  fyi

natemcmaster:
@sbaid I've rescoped the PR. I left in the rename of the props file to include TFM and other little elements of code cleanup (trim trailing whitespace and add final newlines), but can back those out too if you'd prefer.

sbaid:
@natemcmaster pl remove the rename of the props file related change and that will essentially remove the other space cleanup (goodness). There is another props file change @Faizan2304 is doing for [this bug](https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems?_a=edit&id=375688&triage=true)

natemcmaster:
@sbaid I stripped out all other changes.

codito (approved):

sbaid:
thanks @natemcmaster

PR #456, opened 2017-02-08
Remove redirection to NewtonSoft.Json

mayankbansal018:


codito:
Can we fix `TestHost` and `TranslationLayer` nuspec dependencies?

codito (commented):
Looks good with suggestion. Please fix the build break.

PR #469, opened 2017-02-11
Script update: Execution policy bypass

AbhitejJohn:
Updating the scripts to by pass execution policy not requiring devs to global bypass.
Will edit the docs as well.


codito (commented):
Does this require build.cmd to run only in Elevated command prompts?

codito (approved):
Please merge after we have confirmed that build.cmd will continue to work in non-admin command prompts.

AbhitejJohn:
@codito : Nope. Don't need an admin command prompt.

PR #470, opened 2017-02-11
Honor cache timeout for discovery.

AbhitejJohn:
Nit fix to honor stats timeout instead of just passing max value. Since discovery is a very fast operation the batch size should be sufficient to trigger the discovery cache but just in case the batch size is made configurable and for some strange reason discovery is slow, this would help clear up the cache sooner.


codito (commented):
Please add a test.

codito:
Since the variable is now used for both Discovery and Execution, we should rename it appropriately.

codito (approved):

Issue #309, opened 2016-12-31
data collection support in vstest


sbaid:
## Description
data collectors are an extensibility point for the test platform. What is the CLI experience for the data collection support? 

    vstest.console /collect:coverage;filename=foo.coverage
    vstest.console /collect:events;filename=bar.log

I am not sure whether integrating opencover as a data collector make sense. I see that it is currently has a [command line tool](https://github.com/opencover/opencover/wiki/Usage)  with rich set of arguments.

Also what will be the way to pick the data collector? like logger.

codito:
Some open questions we need to decide on.

## Acquisition of data collectors
Choices:
1. Nuget packages.
    * Similar to test adapters, all `*.datacollectors.dll` available in nuget dependency of a test project are marked as data collector extensions.
    * This means datacollectors will load/unload for every test project (at the cost of perf). Can the datacollectors merge the reports etc. if multiple projects are passed for a run?
    * Ease of managing these extensions = add/remove nuget packages.
    * Works for any IDE/Editor/CLI since the configuration is with the project itself.
2. `/extensions:<path>` command line argument
    * `*.datacollectors.dll` are picked from user provided path. They are loaded for the entire run.
    * It is difficult to find the path of such packages in netcore (since they are not in dependency of any project)
    * For IDE/Editors this argument needs to be available in runsettings.
3. Exact assembly specification provided in `runsettings`.
    * Run settings provides the fullyqualified assembly name of a datacollector. These are loaded with default CLR assembly resolution logic, and a special assembly resolver to load inbox collectors
    * Datacollectors can ship as nuget packages which copy dlls as content to test project output directory.
    * For IDE/Editors this needs to be provided along with `TestRunRequest`.

## Enable/disable of data collectors
Choices:
1. All data collectors specified in runsettings are enabled by default. User needs to modify runsettings to disable.
2. Acquisition of datacollectors is a separate. The data collectors to be enabled need to be passed in as `/collect:<friendly name>` in command line arguments. Here IDE/Editors need an extensibility in `TestRunRequest`.

## Configuring a data collector
A data collector can be configured with runsettings or with CLI arguments passed in with `-- codecoverage.filename=foo.log`

## Authoring own datacollector
1. User can author any .net based data collector. They will load with default host.
2. The data collector protocol should be open/published as part of SDK.
3. Any IDE/Editor/Extension author can plugin their own datacollector host. `vstest.console` will launch the host and communicate with json (similar to testhost extensibility).
[...]

PR #477, opened 2017-02-13
Implementation of TestCaseEventSender.

harshjain2:
Implementation of TestCaseEventSender.


src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventSender.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (EqtTrace.IsInfoEnabled)
+            {
+                EqtTrace.Info("Closing the connection !");
+            }
+        }
+
+        /// <summary>
+        /// Disposes communication manager.
+        /// </summary>
+        public void Dispose()
```
codito: Contract for caller is not clear?
`using (var x = new DataCollectionTestCaseEventSender()) { // don't call Close here, it may lead to dispose twice? }`?

codito (approved):

Issue #224, opened 2016-11-15
Few unit tests are taking infinite time and requires cancelling the test run


harshjain2:
## Description
> After migrating from xproj to csproj, few UTs are taking infinite time for execution, test run has to be stopped. 

## Steps to reproduce
> Tests showing this behaviour :
> 1. EnableDiagArgumentProcessorExecutorShouldCreateDirectoryOfLogFileIfNotExists

> Clone repo https://github.com/harshjain2/vstest and checkout branch migrate2.
> Remove tag [Ignore] from any of the above mentioned tests. 
> Run command dotnet test [ProjectName].csproj.

## Expected behavior
> Tests should complete successfully.

## Actual behavior
> Test execution is stuck. 

## Environment
> Please share additional details about the test environment.
> Operating system, Build version of vstest.console

Product Information:
 Version:            1.0.0-preview3-004056
 Commit SHA-1 hash:  ccc4968bc3

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64

sbaid:
@harshjain2  pl update for the latest build

harshjain2:
@sbaid Same behavior is observed when running this command :
`.\tools\dotnet\dotnet.exe test .\test\vstest.console.UnitTests\vstest.console.UnitTests.csproj `

PR #483, opened 2017-02-14
Added dummy testcase.

harshjain2:
Added dummy test case as at least one dotnet core test case is required for test script to succeed.

Related to : https://github.com/Microsoft/vstest/issues/224


scripts/test.ps1
```diff
@@ -59,7 +59,6 @@ $Script:TPT_Configuration = $Configuration
 $Script:TPT_SourceFolders =  @("test")
 $Script:TPT_TargetFrameworks =@($TPT_TargetFrameworkCore, $TPT_TargetFrameworkFullCLR)
 $Script:TPT_TargetRuntime = $TargetRuntime
-$Script:TPT_SkipProjects = @("testhost.UnitTests")
```
codito: Please don't delete TPT_SkipProjects functionality. Let's keep it as 911 :) We can remove testhost.UnitTests from the list.

codito (approved):

scripts/test.ps1
harshjain2: done.

PR #495, opened 2017-02-15
Separating vsmanproj in separate file

Faizan2304:


codito (approved):
We will change location for vsix once we've unblocked build.

PR #496, opened 2017-02-15
Add xunit perf test project.

codito:
Extract common dependencies to a props.


AbhitejJohn:
@codito : Should we start moving some of these assets out of TestPlatform.sln to reduce build time?

AbhitejJohn (approved):

navin22 (approved):

scripts/build/TestPlatform.Settings.targets
```diff
@@ -31,6 +31,9 @@
     <DefineConstants>$(DefineConstants);CODE_ANALYSIS</DefineConstants>
   </PropertyGroup> 
 
+  <!-- Package dependency versions -->
+  <Import Project="$(MSBuildThisFileDirectory)TestPlatform.Dependencies.props" />
```
Faizan2304: I am not able to find TestPlatform.Dependencies.props. From where we are getting this.

codito:
@AbhitejJohn we don't want to differentiate VS build vs command line. TestPlatform.sln is built everywhere. These assets are required for running Performance Tests, it is an overhead to require running build.cmd if someone wants to run perf tests.

test/TestAssets/SimpleTestProject/SimpleTestProject.csproj
```diff
@@ -1,19 +1,22 @@
...
     <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netcoreapp1.0' Or '$(TargetFramework)' == 'netcoreapp1.1' ">$(PackageTargetFallback);dnxcore50;portable-net45+win8</PackageTargetFallback>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="MSTest.TestFramework">
-      <Version>1.0.6-preview</Version>
+      <Version>$(MSTestFrameworkVersion)</Version>
     </PackageReference>
     <PackageReference Include="MSTest.TestAdapter">
-      <Version>1.1.6-preview</Version>
+      <Version>$(MSTestAdapterVersion)</Version>
```
smadala: This should be sync with  [acceptance test reference](https://github.com/Microsoft/vstest/blob/master/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs#L31).
Is there way we can read reference from `prop` file directly in acceptance or we have to sync both manually always?

codito: Good catch. We should pull that as a $env variable? Since integration are a child process of test.cmd?

Faizan2304: +1

scripts/build/TestPlatform.Settings.targets
codito: Added now. It was ignore by gitignore earlier, fixed that as well.

smadala (approved):
Nice cleanup 👍.

AbhitejJohn:
@codito: Was referring to pulling it into another solution and build.cmd would build these two. But yeah, if it isn't adding much time no point. Just wanted to see if we can reduce build time for normal development.

PR #502, opened 2017-02-15
Fix for the TypeDescriptor issue

singhsarab:
Instead of registering the custom converters with TypeDescriptor, use them as static fields.

TestPlan:
IDE, CLI and LUT for
- MSTest based decorated tests.
- Xunit based decorated tests.


codito (approved):
When porting to master, please add comments with motivation for not using dynamic TypeDescriptor.

PR #506, opened 2017-02-16
OutputType based on TargetFramework

Faizan2304:
Issue: TestPlatform side fix for issue https://devdiv.visualstudio.com/DevDiv/_workitems?id=375688&_a=edit

Fix: To fix this issue we need to fix in two places
1) CPS side fix, Which already has been done. 
2) Fix in Microsoft.Net.Test nuget package to conclude OutputType type based on TargetFramework.  We cant do it in props file because that get included before we conclude TargetFramework, hence adding that logic in target file and deleting props file as it is not needed now.

Test:
Manually validated the fix.


Faizan2304:
@srivatsn : Can you please point me the bug which CPS is tracking for this issue. As per my understanding this fix wont be needed once CPS will fix it completely.

scripts/build.ps1
```diff
@@ -328,9 +328,6 @@ function Create-NugetPackages
         Copy-Item $tpSrcDir\$file $stagingDir -Force
     }
 
-    # Copy and rename props file.
-    Copy-Item $tpSrcDir\"Microsoft.Net.Test.Sdk_props" $stagingDir\"Microsoft.Net.Test.Sdk.props" -Force
```
harshjain2: Comment out as we are going to add it again in future.

singhsarab (approved):

src/Microsoft.NET.Test.Sdk_props
```diff
@@ -1,18 +0,0 @@
...
-          created a backup copy.  Incorrect changes to this file will make it
-          impossible to load or build your projects from the command-line or the IDE.
-
-Copyright (c) .NET Foundation. All rights reserved. 
-***********************************************************************************************
--->
-
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
-    <!-- Output type for dotnet core test project should be exe. For full CLR it is Library by default. -->
```
harshjain2: Comment out with appropriate bug url because of which we doing this temporary fix.

harshjain2 (approved):
Approved with few suggestions.

codito (approved):

smadala (changes requested):
Moved these changes to #440. Please ignore this PR.

srivatsn:
@Faizan2304 - https://github.com/dotnet/roslyn-project-system/issues/268 is the issue tracking reloading when targets change.

PR #528, opened 2017-02-21
For full CLR TargetFrameworks make DebugType is 'Full'

smadala:
- In IDE scenario for full CLR projects default test platform is v1, which expects full PDB for source information.

Tested Manually.

Related bugs: #373 & #523

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/vstest/528)

PR #531, opened 2017-02-22
For full CLR TargetFrameworks make DebugType is 'Full' (#528)

codito:
* For full CLR TargetFrameworks make DebugType is 'Full'

* Use TargetFrameworkIdentifier instead TargetFramework.

### DIFF
--- a/src/Microsoft.NET.Test.Sdk.targets
+++ b/src/Microsoft.NET.Test.Sdk.targets
@@ -24,13 +24,22 @@
     <GenerateProgramFile Condition="'$(GenerateProgramFile)' == ''">true</GenerateProgramFile>
   </PropertyGroup>
 
-  <!-- Output type for dotnet core test project should be exe. For full CLR it is Library by default.
-       This should be added in props file.
-       Issues: https://github.com/dotnet/roslyn-project-system/issues/268,
-       https://devdiv.visualstudio.com/DevDiv/_workitems?id=375688&_a=edit
+  <!--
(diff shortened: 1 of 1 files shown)


smadala (approved):

harshjain2 (approved):

src/Microsoft.NET.Test.Sdk.targets
```diff
@@ -24,16 +24,25 @@
...
   <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
   <!--
+     In IDE scenario for full CLR projects default test platform is v1, which expects full pdbs for source information.
+     This can be removed once TPv2 is default for full CLR. Related issue https://github.com/Microsoft/vstest/issues/373.
+   -->
+  <PropertyGroup>
+    <DebugType Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">Full</DebugType>
```
harshjain2: From compatibility perspective between v1 and v2, is it a good idea to keep this forever?

harshjain2 (approved):

src/Microsoft.NET.Test.Sdk.targets
codito: It is not a good idea, as @smadala mentioned in comments we should remove it once we've TPv2 OM on test explorer as well :)

PR #533, opened 2017-02-22
Fixing nunit inconclusive tests reported as failure

AbhitejJohn:
Fixing up the mappings of NotRun tests in trx logger and printing it out on console. This is a fix ported from TPV1.


singhsarab (approved):
LGTM

PR #538, opened 2017-02-23
For empty --version-suffix build.cmd should work

smadala:
- For rtm releases, --version-suffix is empty, build.ps1 should handle --version-suffix empty scenario
- Change vsix version to 15.0.4 from 15.0.3, otherwise vsix won't upgrade because we have already published 15.0.3.2017XXXXXX


smadala:
@codito approved offline.

PR #553, opened 2017-03-01
Implemented AreTestCaseEventsSubscribed

harshjain2:
Implemented a flag to find out whether testcase events are subscribed by datacollectors.
Based on this flag, it can be determined whether or not send testcase events from execution process to data collection process.

Related to : https://github.com/microsoft/vstest/issues/309


src/Microsoft.TestPlatform.Common/DataCollection/Interfaces/IDataCollectionManager.cs
```diff
@@ -42,7 +42,7 @@ internal interface IDataCollectionManager : IDisposable
         /// <summary>
         /// Raises SessionStart event to all data collectors configured for run.
         /// </summary>
-        /// <returns>Are test case level events required.</returns>
+        /// <returns>bolean value specifying whether test case events are subscribed by datacollectors. Based on this execution process will decide whether to send TestCaseStart and TestCaseEnd events to dataCollector process.</returns>
```
codito: * Out of proc data collector: can we reliably find if the testcase start or end events are actually subscribed?
* Inproc data collector: is the cost of sending an event high?

codito: > Out of proc
vstest.console waits until the `SessionStarted` call returns from the datacollector exe. In that call return, `false` indicates that no events are subscribed.

harshjain2: `Out of proc`
If no events are subscribed, then socket communication for communicating with Execution process is not established, dataCollectionPort value is returned as 0 to the vstest.console.exe process signifying that testcase level events are not required for this data collection session.

`In proc`
Don't think there would be a significant cost of sending the event for In Proc data collectors.

PR #567, opened 2017-03-06
Supported Debugging through VSTEST_DATACOLLECTOR_DEBUG env variable.

harshjain2:
Added support for debugging DataCollector process through VSTEST_DATACOLLECTOR_DEBUG env variable.


src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/ProxyDataCollectionManager.cs
```diff
@@ -130,6 +133,13 @@ public void Initialize()
         {
             var port = this.dataCollectionRequestSender.InitializeCommunication();
 
+            // Warn the user that execution will wait for debugger attach.
+            var dataCollectorDebugEnabled = Environment.GetEnvironmentVariable("VSTEST_DATACOLLECTOR_DEBUG");
```
codito: Can we move this line after 143, we could print the name and PID of data collector in that case.

codito (approved):
Looks good with small suggestion.

PR #573, opened 2017-03-07
Formatting changes for logs coming from datacollector process.

harshjain2:
Formatting changes for logs coming from datacollector process.


src/Microsoft.TestPlatform.CommunicationUtilities/Resources/Resources.resx
```diff
@@ -126,6 +126,9 @@
   <data name="ConnectionClosed" xml:space="preserve">
     <value>An existing connection was forcibly closed by the remote host.</value>
   </data>
+  <data name="DataCollectorUriForLogMessage" xml:space="preserve">
+    <value>Diagnostic data adapter ('{0}') message: {1}</value>
```
codito: Should we not use the known terminology (`Data collector` instead of Diagnostic data adapter)?

codito (approved):
Suggest to use known terms.

PR #574, opened 2017-03-07
Removed unnecessary catch blocks.

harshjain2:
Removed unnecessary catch blocks.

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
@@ -57,19 +57,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client
         /// </summary>
         public override void Initialize()
         {
-            try
-            {
-                this.ProxyDataCollectionManager.Initialize();
-            }
-            catch (Exception ex)
-            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
```diff
@@ -84,22 +72,11 @@ public override void Initialize()
...
-                }
-                finally
-                {
-                    // There is an issue with Data Collector, skipping data collection and continuing with test run.
-                    if (EqtTrace.IsErrorEnabled)
-                    {
-                        EqtTrace.Error("ProxyExecutionManagerWithDataCollection: Error occured while communicating with DataCollection Process: {0}", ex.Message);
-                    }
-                }
+                // On failure in calling BeforeTestRunStart, call AfterTestRunEnd to end DataCollectionProcess
```
codito: Why not push this to `finally`? We can remove this catch and throw?

codito (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
harshjain2: This is conditional, if there is any exception, then only we will call the AfterTestRunComplete with cancel as true, otherwise not.

PR #580, opened 2017-03-09
Add props file to buildMultiTargeting in Microsoft.NET.Test.Sdk nuspec

natemcmaster:
By adding this property file to buildMultiTargeting, NuGet will import this as a top-level file in test projects with multiple target frameworks. Currently, the properties in from Microsoft.NET.Test.Sdk are only available during the inner build.

Example of the effect:
```xml
  <PropertyGroup>
    <TargetFrameworks>netcoreapp1.1;net46</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
  </ItemGroup>
  <Target Name="Peek">
    <Message Text="IsTestProject = $(IsTestProject)" Importance="high" />
  </Target>
</Project>
```
**Before**
Executing `dotnet msbuild /t:Peek` shows:
```
Microsoft (R) Build Engine version 15.1.548.43366
Copyright (C) Microsoft Corporation. All rights reserved.

  IsTestProject = 
```
**After**
```
Microsoft (R) Build Engine version 15.1.548.43366
Copyright (C) Microsoft Corporation. All rights reserved.

  IsTestProject = true
```


singhsarab (approved):

natemcmaster:
👍 I don't have push rights, so someone will have to merge this for me.

src/package/nuspec/Microsoft.NET.Test.Sdk.nuspec
```diff
@@ -17,9 +17,10 @@
     </dependencies>
   </metadata>
   <files>
+    <file src="Microsoft.NET.Test.Sdk.props" target="buildMultiTargeting\" />
```
singhsarab: Do we need to include the targets file as well ?

natemcmaster: You shouldn't need to. NuGet will still import and execute the targets file on the "inner" build. It looks like the current targets file only generates a "Program.cs" file, and which depends on `TargetFrameworkIdentifier` being set. It probably will work incorrectly when running in the "outer" build.

Btw, if you're not familiar with inner vs outer, here is _conceptually_ what MSBuild is doing in multi-targeting projects:
```xml
<!-- this executes when `msbuild.exe /t:Build` runs on the project -->
<Target Name="Build" Condition="'$(TargetFrameworks)' != '' AND '$(TargetFramework)' == ''">
    <!-- import buildMultiTargeting/Microsoft.NET.Test.Sdk.props  -->
    <ItemGroup>
      <_TargetFramework Include="$(TargetFrameworks)" />
    </ItemGroup>
    <!-- starts a new "build" of the project, once per target framework -->
    <MSBuild Projects="$(MSBuildProjectFile)"
             Targets="InnerBuild"
             Properties="TargetFramework=%(_TargetFramework.Identity)" />
</Target>

<Target Name="InnerBuild" Condition="'$(TargetFramework)' != ''">
   <!-- import build/$(TargetFramework)/Microsoft.NET.Test.Sdk.{props, targets} -->
   <!-- gather *.cs files, run compiler, etc. -->
</Target>
```

PR #583, opened 2017-03-09
Downgrade assembly version

smadala:
- Keep AssemblyVersion 15.0.0 always, to avoid protocol breaking.
- Fix exception if package folder already exists


AbhitejJohn:
@smadala : Just curious how does this break the protocol? We do not carry assembly versions in the json.

smadala:
TestOutcome had assembly version.

codito (approved):

PR #601, opened 2017-03-11
Normalize line endings for Microsoft.TestPlatform.Build.csproj.

codito:
Treat csproj files as text.


codito:
Merging this since this is a blocker. `Microsoft.TestPlatform.Build.csproj` always shows up in `unstaged` files and doesn't allow to switch branches.

PR #610, opened 2017-03-14
Fix versioning.

codito:
AssemblyVersion should be 15.0.0 in both VS and CLI build.
AssemblyInformationVersion should have the required suffix.

Ignore data collector platform tests since they are flaky.


smadala (approved):

singhsarab (approved):
👍

PR #647, opened 2017-03-22
Socket platform tests

codito:
Tests for SocketCommunicationManager functionality.


test/Microsoft.TestPlatform.CommunicationUtilities.PlatformTests/Program.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.TestPlatform.CommunicationUtilities.PlatformTests
+{    
+    public static class Program
```
AbhitejJohn: Needed?

AbhitejJohn (approved):

harshjain2:
More tests could be added as per these specification of TcpClient and TcpServer :
[TcpClient](https://msdn.microsoft.com/en-us/library/system.net.sockets.tcpclient(v=vs.110).aspx)
1. ExclusiveAddressUse : Communication protocol currently support only one client. How will the system behave if more than one clients are connected. 
2. LingerState : For scenarios like support UAP apps, LingerState is of importance. How will the communication layer behave in case server and client are not on same machine(device, emulator, IoT, etc.)
3. SendBufferSize, ReceiveBufferSize : How will the system behave if the data sent exceeds the default buffer size, 8192 bytes. 

NoDelay : This field sounds interesting as it can significantly improve the efficiency of communication layer, specifically in remote scenarios.

[TcpListener](https://msdn.microsoft.com/en-us/library/system.net.sockets.tcplistener(v=vs.110).aspx)
1. ExclusiveAddressUse : Same as for TcpClient.

test/Microsoft.TestPlatform.CommunicationUtilities.PlatformTests/Program.cs
codito: Yes, for every test project that targets netcoreapp. We don't include NET.Test.Sdk directly.

codito:
> ExclusiveAddressUse : Communication protocol currently support only one client. How will the system behave if more than one clients are connected.
We don't allow multiple client use. One SCM per ProxyExecutionManager.

> LingerState : For scenarios like support UAP apps, LingerState is of importance. How will the communication layer behave in case server and client are not on same machine(device, emulator, IoT, etc.)
This is a choice of the implementation. In our implementation we use the `default`; consumers of ISocketCommunicationManager don't have a way to modify this. I believe for UAP/Emulator (anywhere .NET is, LingerState will be default false (as per spec in msdn). There's no timeout. I don't think we require tests for this.

> SendBufferSize, ReceiveBufferSize : How will the system behave if the data sent exceeds the default buffer size, 8192 bytes.
This is internal implementation detail. Consumers of SocketCommunicationManager cannot change this. We use a length-prefixed buffer (BinaryReader/Writer), this shouldn't create a problem. Are you suggesting we add happy-path tests?

NoDelay - is on our backlog for next PR (refactors SCM).

AbhitejJohn:
@codito: Are these tests running in the PR today?

codito:
Yes, they are running in the PR.

PR #658, opened 2017-03-24
Fix for #657, where dotnet vstest fails after publish and deploy

### LINKED ISSUE #657: After dotnet publish, Could not find testhost.dll for source
## Description
After running `dotnet publish` on a test project that includes a reference to Microsoft.TestPlatform.TestHost, and then deploying the published binaries to another host, running `dotnet vstest <project.dll>` throws the error `Could not find testhost.dll for source` even though `testhost.dll` is in the published directory. 

Deleting `<project>.runtimeconfig.dev.json` makes `dotnet vstest <project.dll>` work fine.

This is because of a probing bug in `GetTestHostPath` in https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CrossPlatEngine/Hosting/DotnetTestHostManager.cs 

On line 320, if `<project>.runtimeconfig.dev.json` is present, it probes for the relative path at a number of locations including `/root/.nuget/packages` folder. However, when the published directory is deployed on a different host (which does not have a .nuget cache), the probing fails.

However, line 368 is meant to handle this scenario of testing after publish. Except the if condition is skipped because the package reference was found, and `testHostPath` was set in line 344. 

## Fix
Add `testHostPath = null;` to Line 366.

saikatguha:
Fixes #657

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Hosting/DotnetTestHostManager.cs
@@ -363,6 +363,9 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting
                         }
                     }
                 }
+                
+                // Probing failed to find full path. Fallback to alternative options.
+                testHostPath = string.Empty;
             }
 
             if (string.IsNullOrEmpty(testHostPath))


codito:
@saikatguha thanks for catching (and fixing) this issue! Can you please add a test for this code change?

Unit tests are here: https://github.com/Microsoft/vstest/blob/master/test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Hosting/DotnetTestHostManagerTests.cs. Steps to run tests are available in [contribution guide](https://github.com/Microsoft/vstest-docs/blob/master/docs/contribute.md#running-tests-visual-studio). `test -p crossplatengine` will do the job.

codito (approved):
Leave a comment

Faizan2304:
@saikatguha : Thanks for this PR. We have fixed this bug slightly different way. Please take a look if you got time https://github.com/Microsoft/vstest/pull/690

Issue #608, opened 2017-03-13
Program.cs is getting added to full clr test project.


harshjain2:
## Description
> Program.cs is getting added to full clr test project. Program.cs is required in case of netcoreapp tests, but not for full clr test projects.

## Steps to reproduce
> Create a new .net core test project. 
> Change the Target Framework to net46.
> Build the solution. 
> Open the `obj\net46\` folder.

## Expected behavior
> {ProjectName}.Program.cs should not be present. 

## Actual behavior
> {ProjectName}.Program.cs is present. 

Root Cause Analysis : We are now adding this (https://github.com/Microsoft/vstest/blob/master/src/package/nuspec/Microsoft.NET.Test.Sdk.targets) targets file for both netcoreapp and net46 tfms. Earlier we used it only for netcoreapp tfm. 

Conditionally adding the program.cs based on tfm should fix this.

PR #665, opened 2017-03-27
Inject entry point only for project targeting netcoreapp

Faizan2304:
Issue: https://github.com/Microsoft/vstest/issues/608
Fix: Inject entry point only for project targeting netcoreapp

### DIFF
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.targets
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.targets
@@ -18,7 +18,7 @@
      Generates Program file which contains the Main entry point
      ============================================================
   -->
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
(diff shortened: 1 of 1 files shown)


harshjain2 (approved):
Suggestion : Maintain two separate copies for Microsoft.NET.Test.Sdk.targets for dotnet core and full clr Target Frameworks, This will avoid issues like https://github.com/Microsoft/vstest/issues/608 in future.

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
```diff
@@ -18,7 +18,7 @@
      Generates Program file which contains the Main entry point
      ============================================================
   -->
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
```
codito: A general thought (not specific to this PR):
For every build of a netcore test project, we will create this file. Does this imply that the test project will be compiled again always? (even if no actual code has changed)

If this is true, can it be avoided?

codito (approved):

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
Faizan2304: No, It will create only first time. Subsequent build and rebuild will use the same

PR #670, opened 2017-03-28
Remove binding redirect of Newtonsoft.Json from testhost config file

Faizan2304:


Faizan2304:
It was a dummy pull request for some testing

PR #681, opened 2017-04-03
Latest MSTest framework and assert extension.

AbhitejJohn:
Updating to the latest mstest framework version and using an extension based on the new assert extensibility that it provides.


smadala (approved):

scripts/build/TestPlatform.Settings.targets
```diff
@@ -85,6 +85,9 @@
         <PackageReference Include="Moq">
           <Version>4.6.38-*</Version>
         </PackageReference>
+        <PackageReference Include="MSTest.Assert.Extensions">
+          <Version>1.0.2-preview</Version>
```
codito: Should we pull out the dependency version to props?

codito (approved):

scripts/build/TestPlatform.Settings.targets
AbhitejJohn: Done. Treated it similar to moq earlier.

Issue #532, opened 2017-02-22
dotnet vstest throws an exception when running Full CLR assemblies


AbhitejJohn:
## Description
On a test project targeting net46, if dotnet vstest is performed on the test assembly using the same arguments as one would use for netcoreapp targeted assembly it throws an exception. The workaround is to provide a /Framework:Framework45 switch to it.

## Steps to reproduce
1. Create a vanilla .Net Core unit test project targeting net46 and build. 
2. Perform dotnet vstest on the unit test assembly.

## Expected behavior
The default test should run.

## Actual behavior
The following exception is thrown out:
```
Error: Could not find testhost.dll for source 'C:\SomeRandomPath\UnitTestProject4\UnitTestProject4\bin\Debug\net46\UnitTestProject4.dll'. Make sure test project has a nuget reference of package "microsoft.testplatform.testhost".
```

## Environment
dotnet --version `1.0.0-rc4-004812`

AbhitejJohn:
There seems to be two issues here:
1. The exception message doesn't help me identify the actual cause of the problem: This assembly cannot be loaded unless its target framework is specified. 
2. It would be great to not mandate the specification of the Framework. The test platform could smartly figure this information out.

PR #691, opened 2017-04-05
Improve testplatform message

Faizan2304:
Fix for issue:
1) https://github.com/Microsoft/vstest/issues/341
2) https://github.com/Microsoft/vstest/issues/532


smadala:
From https://github.com/Microsoft/vstest/issues/532#issuecomment-281622592
> 2. It would be great to not mandate the specification of the Framework. The test platform could smartly figure this information out.

We should fix this too.

src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+  <file datatype="xml" source-language="en" target-language="cs" original="../Resources.resx" build-num="1496413687">
+    <body>
+      <trans-unit id="NoTestHostFileExist">
+        <source>No test is available in {0}. Make sure test project has a nuget reference of package "Microsoft.NET.Test.Sdk" and framework version settings are appropriate and try again.</source>
+        <target state="new">Could not find testhost.dll for source '{0}'. Make sure test project has a nuget reference of package "Microsoft.NET.Test.Sdkt". And also check the framework passed when running test.</target>
```
smadala: `Microsoft.NET.Test.Sdkt` is this expected?

smadala (approved):
Close #532 only when both parts addressed.

src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
Faizan2304: No it shouldn't. It has new tag associated with it. The source is correct. Actually in the beginning I had wrong string which I fixed in next commit. It updated the source tag in xlf file but the target tag remained the same. There is no issue with it as it will get updated by translated string.

Issue #452, opened 2017-02-07
Diag switch : Exception not getting caught, eventually crashing the process


singhsarab:
## Description
> UnauthorizedAccessException not getting handled while creating the log file when diag is enabled.

## Steps to reproduce
>Start dotnet vstest with diag switch and give any path which requires elevated privileges.
>> dotnet vstest /port:123 /diag:C:\log.txt

## Expected behavior
> The error should be handled properly and process should not crash. 

## Actual behavior
> The process exits with following exception

```
Unhandled Exception: System.UnauthorizedAccessException: Access to the path 'c:\foo.txt' is denied.
   at System.IO.Win32FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
   at System.IO.Win32FileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.File.Open(String path, FileMode mode, FileAccess access, FileShare share)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.RollingFileTraceListener.OpenTextWriter(String fileName)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.RollingFileTraceListener..ctor(String fileName, String name, Int32 rollSizeKB)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace.EnsureTraceIsInitialized()
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace.WriteLine(TraceLevel level, String message)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace.Error(String format, Object[] args)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.ArgumentProcessorFactory.<>c__DisplayClass20_0.<WrapLazyProcessorToInitializeOnInstantiation>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.GetArgumentProcessors(String[] args, List`1& processors)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.Execute(String[] args)
[...]

Issue #448, opened 2017-02-06
Improve the console message related to /diag switch


singhsarab:
## Description
>While running tests via dotnet vstest and using the diag switch, there are 2 files getting generated, but only one is being reported.
> Add the information for the other file as well. 

## Steps to reproduce
> Just run dotnet vstest for given testcontainer with diag switch.
dotnet vstest <path_to_dll> /diag:foo.txt 

## Expected behavior
`Information:`
`Logging VsTest Diagnostics in file: foo.txt`
`Logging TestHost Diagnostics in file: foo.host.17-02-06_04-21-41_75662_4.txt`

## Actual behavior
Starting test execution, please wait...
`Information: Logging TestHost Diagnostics in file: foo.host.17-02-06_04-21-41_75662_4.txt`

Faizan2304:
Fixed!!

PR #707, opened 2017-04-11
Fix diag related issue

Faizan2304:
Issue:
1) https://github.com/Microsoft/vstest/issues/452
2) https://github.com/Microsoft/vstest/issues/448

Fix:
1) Catch all exception while creating diag file and exit process peacefully.
2) Show diag log file to user at the beginning.


src/Microsoft.TestPlatform.CoreUtilities/Helpers/FileHelper.cs
```diff
@@ -58,5 +58,18 @@ public FileAttributes GetFileAttributes(string path)
         {
             return new FileInfo(path).Attributes;
         }
+
+        /// <summary>
```
codito: `inheritdoc`

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs
```diff
@@ -80,11 +81,26 @@ public virtual void SetupChannel(IEnumerable<string> sources)
 
                 var portNumber = this.RequestSender.InitializeCommunication();
                 var processId = this.processHelper.GetCurrentProcessId();
-                var connectionInfo = new TestRunnerConnectionInfo { Port = portNumber, RunnerProcessId = processId, LogFile = this.GetTimestampedLogFile(EqtTrace.LogFile) };
+
+                var logFileName = this.GetTimestampedLogFile(EqtTrace.LogFile);
```
codito: Do we expect a consumer of EqtTrace to check for permissions? E.g vstest.console may still fail today if there is no write access to current directory? Or file system has no space etc?

Faizan2304: You are right but there is one scenario where consumer of EqtTrace has to check for permission.
1) Let say our command to run test is "vstest.console.exe  test.dll /diag:logfile.txt"
2) Suppose we are running test from "c:\Program File" and this location already has a file with name logfile.txt.

In this case vstest will be able open and update logfile.txt but when testhost will try to create logfile.host.*.txt file it will fail with unauthorized exception.

src/Microsoft.TestPlatform.CoreUtilities/Helpers/FileHelper.cs
Faizan2304: Fixed in next commit

smadala (approved):

src/vstest.console/Processors/ListTestsArgumentProcessor.cs
```diff
@@ -208,6 +209,10 @@ public ArgumentProcessorResult Execute()
             }
 
             this.output.WriteLine(CommandLineResources.ListTestsHeaderMessage, OutputLevel.Information);
+            if (!string.IsNullOrEmpty(EqtTrace.LogFile))
+            {
+                this.output.WriteLine(string.Format(CommandLineResources.VstestDiagLogOutputPath, EqtTrace.LogFile), OutputLevel.Information);
```
Faizan2304: Fixed in next commit

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs
```diff
@@ -80,11 +81,14 @@ public virtual void SetupChannel(IEnumerable<string> sources)
...
```
codito: Extract method

PR #723, opened 2017-04-17
Use a dev build of stylecop temporarily.

codito:
1.0.0 is broken with dotnet-cli 2.0, 1.0.1 fixes it but results in another error.


codito:
This may not be required.

Issue #761, opened 2017-04-26
running netcoreapp2.0 test in container: Could not find testhost.dll


tmds:
I am trying to run a netcoreapp2.0 xunit project.
I have installed the `2.0.0-preview1-005783` cli.

I create a `dotnet new xunit` project and adjust the NuGet.Config file and csproj file to use the `15.1.0-preview-20170414-10` Sdk.

When I run this on a RHEL7 machine, tests are discovered.
When I run this in a RHEL7 container, I get a message:

> Could not find testhost.dll for source '/tmp/test/bin/Debug/netcoreapp2.0/test1.dll'. Make sure test project has a nuget reference of package "microsoft.testplatform.testhost".

In the container, the restore completes successfully and contained this line:
```
  Installing Microsoft.TestPlatform.TestHost 15.1.0-preview-20170414-10.
```

When I manually start the vstest.console in the container and set the logging verbosity to normal the error message changes:

```
$ dotnet exec /opt/dotnet/sdk/2.0.0-preview1-005783/vstest.console.dll --framework:.NETCoreApp,Version=v2.0 --logger:console;verbosity=normal  /tmp/testapp/bin/Debug/netcoreapp2.0/testapp.dll
```

> No test source files were specified.

dotnet info:
```
$ dotnet --info
.NET Command Line Tools (2.0.0-preview1-005783)

Product Information:
 Version:            2.0.0-preview1-005783
 Commit SHA-1 hash:  8bd15c7ec0

Runtime Environment:
 OS Name:     rhel
 OS Version:  7
 OS Platform: Linux
 RID:         rhel.7-x64
 Base Path:   /opt/dotnet/sdk/2.0.0-preview1-005783/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview1-001961-00
  Build    : 884c69a4ac461a5e1fad31f60f2e01871371c288
```

cc @smadala

smadala:
@tmds Can you please check `microsoft.testplatform.testhost` exists in values specified for additionalProbingPaths in bin/Debug/netcoreapp2.0/<test_project_name>.runtimeconfig.dev.json.

> $ dotnet exec /opt/dotnet/sdk/2.0.0-preview1-005783/vstest.console.dll --framework:.NETCoreApp,Version=v2.0 --logger:console;verbosity=normal  /tmp/testapp/bin/Debug/netcoreapp2.0/testapp.dll

[...]

PR #767, opened 2017-04-27
Find testhost.dll using StringComparison.OrdinalIgnoreCase.

tmds:
Using CurrentCultureIgnoreCase causes the lookup to fail depending on culture.

Fixes https://github.com/Microsoft/vstest/issues/761

CC @codito @smadala @Faizan2304


Faizan2304 (approved):
👍

PR #772, opened 2017-04-28
Wait for testhost stderr to be available if connection is broken.

codito:
This allows to provide user information if the test host has crashed.

Don't send stderr to console runner output if it's null or empty.

PR #774, opened 2017-04-29
Wait for testhost stderr to be available if connection is broken. (#772)

sbaid:
This allows to provide user information if the test host has crashed.

Don't send stderr to console runner output if it's null or empty.

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
@@ -396,10 +396,34 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities
 
             if (message == null)
             {
-                var reason = string.IsNullOrWhiteSpace(this.clientExitErrorMessage)
-                    ? CommonResources.UnableToCommunicateToTestHost
-                    : this.clientExitErrorMessage;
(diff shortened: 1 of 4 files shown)


codito:
@dotnet-bot test Ubuntu16.04 please

codito:
Looking into the build failure.

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -396,10 +396,34 @@ private string TryReceiveRawMessage()
...
+                }
+                else
+                {
+                    // Test host process has not exited yet. We will wait for exit to allow us gather
+                    // standard error
+                    var processWaitEvent = new ManualResetEventSlim();
+                    try
+                    {
+                        EqtTrace.Info("TestRequestSender: Waiting for test host to exit.");
+                        processWaitEvent.Wait(TimeSpan.FromSeconds(10), this.clientExitCancellationSource.Token);
```
Faizan2304: If in 10 sec testhost doesn't exit, what will happen to testhost process? Shouldn't we kill testhost explicitly?

codito: Yes, we should. Will bring this in a future PR. cc: @smadala

PR #781, opened 2017-05-03
Monitoring Data Collector Launch and Exit events.

harshjain2:
Monitoring Data Collector Launch and Exit events, logging error in case data collector crashes.


src/Microsoft.TestPlatform.Common/ProcessCallbacks.cs
```diff
@@ -1,17 +1,18 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.TestPlatform.TestHostProvider.Hosting
+namespace Microsoft.VisualStudio.TestPlatform.Common
```
codito: I am not sure if Common should include utility code. Isn't coreutilities/platformabstractions the component?

src/Microsoft.TestPlatform.Common/ProcessCallbacks.cs
```diff
@@ -1,17 +1,18 @@
...
+
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Host;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
     using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces;
 
-    internal class TestHostManagerCallbacks
+    internal class ProcessCallbacks
     {
         public static void ErrorReceivedCallback(StringBuilder testHostProcessStdError, string data)
```
codito: If it's a generic API, the names should change.

src/Microsoft.TestPlatform.Common/ProcessCallbacks.cs
harshjain2: This class has dependency on `HostProviderEventArgs` which is in object model, therefore, I intentionally kept it in Common. 
Ideally it should go to platform abstractions, could be along with ProcessHelper

codito: HostProviderEventArgs is interesting for DataCollectors? Can that not stay along side TestHostProvider implementation?

codito (changes requested):
We need to figure a better place for the callbacks.

src/Microsoft.TestPlatform.Common/ProcessCallbacks.cs
harshjain2: @codito I don't think, except for the name there is anything in `HostProviderEventArgs` that is specific to TestHostProvider.

PR #788, opened 2017-05-08
Removed reference of Common from TestHostProvider

harshjain2:
1. Removed reference of Microsoft.TestPlatform.Common project from Microsoft.TestPlatform.TestHostProvider.
2. Removed argument messageLogger from ExitCallBack as it is no more required.


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -44,6 +44,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider
     {
         private const string DotnetTestHostUri = "HostProvider://DotnetTestHost";
         private const string DotnetTestHostFriendltName = "DotnetTestHost";
+        private const string TestAdapterRegexPattern = @".*.TestAdapter.dll";
```
codito: Since this is publicly documented, should we pick it up from OM? Please note it may require an update in vsualm.

codito: cc @Faizan2304

codito (approved):

Issue #646, opened 2017-03-22
running dotnet test --logger:trx with docker does not always produce trx file


schwamster:
## Description
running dotnet test --logger:trx in a docker image where the respective folder is mounted (Win 7 Docker Toolbox) will not produce a trx file since the produced filename is not valid on windows.

## Steps to reproduce
Win7 Docker Toolbox
```
docker run -it --rm  -v //c/Users/myuser/myapp:/app microsoft/aspnetcore-build:latest
cd /app
dotnet new xunit
dotnet restore
dotnet test --logger:trx
```

The reason for this can be observed running the following sequence:

```
docker run -it --rm  -v //c/Users/myuser/myapp:/app microsoft/aspnetcore-build:latest
mkdir /app-test && cd /app-test
dotnet new xunit
dotnet restore
dotnet test --logger:trx
cp -R TestResults /app/
```
Now you can see that the trx file is created (since its done outside of the mounted volume) but you cant copy it - protocol error
![capture](https://cloud.githubusercontent.com/assets/10597594/24186260/c65de9d2-0ed7-11e7-8cd8-ca4d18b479c9.PNG)

## Expected behavior
- runnig dotnet test --logger:trx should work when running within docker in a mounted folder from Win7
- resulting trx file should have the same filename format as on windows

## Actual behavior
- trx file is not produced since the produced filename is not valid on windows
- filenames for trx reports have a different format on windows and linux

## Environment
Win 7,
Docker Toolbox

## Other
See also original Issue: https://github.com/aspnet/aspnet-docker/issues/211

harshjain2:
@schwamster You can provide custom filename for trx logger file. 

`dotnet test --logger "trx;LogFileName=abc.trx"`

Will this solve your issue?

schwamster:
valid workaround. still would be nice if it worked the same way ( plus compatible) not matter what operating system.

harshjain2:
@schwamster Thanks for the feedback.

This issue seems to be specific to Win7 only. 
I tried the below command by installing docker in Win10, it is working.

>docker run -it --rm  -v //c/Users/myuser/myapp:/app microsoft/aspnetcore-build:latest
mkdir /app-test && cd /app-test
dotnet new xunit
dotnet restore
dotnet test --logger:trx
cp -R TestResults /app/

schwamster:
[...]

PR #789, opened 2017-05-08
Extend list of invalid file path characters 

natemcmaster:
Extend the list to include all Windows characters, even when running on Unix-based systems.

Path.GetInvalidFileNameChars on Linux only contains { \0, / }, which means the TRX logger and other code may produce files that would be invalid on Windows systems. This causes issues when running tests in WSL (Bash for Windows) or Docker containers that mount Windows volumes.

Resolves https://github.com/Microsoft/vstest/issues/646.

cc @harshjain2


codito (approved):

codito:
Thank you!

Issue #798, opened 2017-05-10
Cannot easily launch the feature for debugging


sharwell:
## Description

It is not currently possible to "clone and go" with this repository.

## Steps to reproduce

1. Clone the repository
1. Open **TestPlatform.sln**
1. Right click on a project and select **Set as StartUp Project**
1. Press <kbd>F5</kbd> (Debug &rarr; Start Debugging)

## Expected behavior

The solution builds, and an experimental instance of Visual Studio 2017 opens with the new local build of the test platform installed and enabled.

## Actual behavior

There is no such project in the solution.

## Environment

Windows 10, Visual Studio 2017 15.1

codito:
The vsix project today doesn't include project references to other product projects (they are all multi targeted, and don't play nice with project references in a vsix). This definitely makes getting started harder.

It is possible to generate a vsix in command line:
```
> build
> ls artifacts\Debug\VSIX\TestPlatform.vsix
```

It can be installed into any VS instance.

pvlakshm:
Closing this; reactive is still necessary.

sharwell:
@pvlakshm I don't understand?

PR #802, opened 2017-05-11
Build VSIX as part of the solution

sharwell:
This pull request implements the foundations of #798. It addresses all of the blocking items which previously prevented this from moving forward, but leaves several simpler items related to an incomplete localization implementation in [TestPlatform.Localization.targets](https://github.com/Microsoft/vstest/blob/master/scripts/build/TestPlatform.Localization.targets) and the VSIX project not referencing the correct locations for individual project build outputs.

* [x] Work around CPS multi-targeting limitations by explicitly including all source items
  * [x] Use solution references to ensure correct build order, since project references don't work
* [x] Add VSIX project to the solution
  * [x] Fix NuGet restore for the VSSDK
  * [x] Fix debugging properties for the experimental instance
* [ ] Fix solution build producing incorrect output
  * [ ] Fix solution build not localizing files
  * [ ] Fix solution build not placing files in the correct output location (or fix VSIX project to use the actual output location)

Please let me know if any of the remaining work items are not clear so we can unblock other contributors and efficiently as possible. :smile:


codito:
@sharwell thanks for starting this :+1: We may need to add a task to disable building VSIX from .sln in CLI, the VSSDK build tasks assembly seems to be desktop only. I will play around with this branch and post an update.

singhsarab:
@sharwell 
This seems like a good change to me. What is remaining to complete this one?

singhsarab:
Closing this PR. Please let me know if someone wants to take this up.

sharwell:
@singhsarab As far as I can tell, the remaining items are still the items which are not checked off above

Issue #685, opened 2017-04-04
Enables Fakes support in vstest


pvlakshm:
## Description
Currently test projects taking a dependency on Fakes cannot be migrated to **vstest**. This issue tracks adding support for adding Fakes, that will then enable such projects to be migrated.

PR #810, opened 2017-05-15
Fakes support in TestPlatform

mayankbansal018:
Fixes https://github.com/Microsoft/vstest/issues/685

### DIFF
--- a/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
+++ b/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
@@ -60,7 +60,6 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.DataCollector
         /// </summary>
         private DataCollectorExtensionManager dataCollectorExtensionManager;
 
-
         /// <summary>
         /// Initializes a new instance of the <see cref="DataCollectionManager"/> class.
         /// </summary>
@@ -378,7 +377,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.DataCollector
         /// <param name="dataCollectorSettings">
         /// The data collector settings.
         /// </param>
-        /// <param name="settingsXml"></param>
+        /// <param name="settingsXml"> runsettings Xml</param>
         private void LoadAndInitialize(DataCollectorSettings dataCollectorSettings, string settingsXml)
         {
             DataCollectorInformation dataCollectorInfo;
@@ -518,20 +517,8 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.DataCollector
                 return;
             }
 
-
-            //do not send events multiple times
-
-            try
-            {
-                this.events.RaiseEvent(args);
-            }
-            catch (Exception ex)
-            {
-                if (EqtTrace.IsErrorEnabled)
-                {
-                    EqtTrace.Error("DataCollectionManger:SendEvent: Error while RaiseEvent {0} to datacollector : {1}.", args.GetType(), ex);
-                }
-            }
+            // do not send events multiple times
+            this.events.RaiseEvent(args);
         }
 
         /// <summary>
--- a/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
+++ b/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
@@ -518,23 +518,18 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.DataCollector
                 return;
             }
 
-            foreach (var dataCollectorInfo in this.RunDataCollectors.Values)
-            {
-                if (EqtTrace.IsVerboseEnabled)
-                {
-                    EqtTrace.Verbose("DataCollectionManger:SendEvent: Raising event {0} to collector {1}", args.GetType(), dataCollectorInfo.DataCollectorConfig.FriendlyName);
-                }
 
-                try
+            //do not send events multiple times
+
+            try
+            {
+                this.events.RaiseEvent(args);
+            }
+            catch (Exception ex)
+            {
+                if (EqtTrace.IsErrorEnabled)
                 {
-                    this.events.RaiseEvent(args);
-                }
-                catch (Exception ex)
-                {
-                    if (EqtTrace.IsErrorEnabled)
-                    {
-                        EqtTrace.Error("DataCollectionManger:SendEvent: Error while RaiseEvent {0} to datacollector {1} : {2}.", args.GetType(), dataCollectorInfo.DataCollectorConfig.FriendlyName, ex);
-                    }
+                    EqtTrace.Error("DataCollectionManger:SendEvent: Error while RaiseEvent {0} to datacollector : {1}.", args.GetType(), ex);
                 }
             }
         }
@@ -555,6 +550,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.DataCollector
             var dataCollectorEnvironmentVariable = new Dictionary<string, DataCollectionEnvironmentVariable>(StringComparer.OrdinalIgnoreCase);
             foreach (var dataCollectorInfo in this.RunDataCollectors.Values)
             {
+                dataCollectorInfo.SetTestExecutionEnvironmentVariables();
                 try
                 {
                     this.AddCollectorEnvironmentVariables(dataCollectorInfo, dataCollectorEnvironmentVariable);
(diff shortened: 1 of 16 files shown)


src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
```diff
@@ -518,23 +518,18 @@ private void SendEvent(DataCollectionEventArgs args)
...
-                {
-                    EqtTrace.Verbose("DataCollectionManger:SendEvent: Raising event {0} to collector {1}", args.GetType(), dataCollectorInfo.DataCollectorConfig.FriendlyName);
-                }
 
-                try
-                {
-                    this.events.RaiseEvent(args);
-                }
-                catch (Exception ex)
+            //do not send events multiple times
```
codito: :) Do we have test coverage for this?

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
```diff
@@ -518,23 +518,18 @@ private void SendEvent(DataCollectionEventArgs args)
...
+
+            try
+            {
+                this.events.RaiseEvent(args);
+            }
+            catch (Exception ex)
```
codito: Is `catch` required? It seems we are already using `SafeInvoke` to raise the individual events, it captures and logs the exception?

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectorInformation.cs
```diff
@@ -176,7 +175,7 @@ internal void DisposeDataCollector()
         /// <summary>
         /// The get test execution environment variables sync.
```
codito: nit: out of date comment

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectorInformation.cs
```diff
@@ -176,7 +175,7 @@ internal void DisposeDataCollector()
...
         /// </summary>
-        private void SetTestExecutionEnvironmentVariables()
+        public void SetTestExecutionEnvironmentVariables()
```
codito: Why is this change required?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            try
+            {
+                var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), FakesConfiguratorAssembly + ".dll");
+
+                Assembly assembly = null;
+#if NET46
+                assembly = Assembly.LoadFrom(assemblyPath);
+#else
+                // I think we should not do this, since FakeConfigurator is built for net46?
```
codito: Yes, we shouldn't do this.

codito: Possibly we can remove the System.Reflection reference in csproj?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -0,0 +1,214 @@
...
+
+        private static bool TryGetFakesDataCollectorConfigurator(out Func<IEnumerable<string>, string> configurator)
+        {
+            try
+            {
+                var assemblyPath = Path.Combine(Path.GetDirectoryName(Assembly.GetEntryAssembly().Location), FakesConfiguratorAssembly + ".dll");
+
+                Assembly assembly = null;
+#if NET46
+                assembly = Assembly.LoadFrom(assemblyPath);
```
codito: Is it okay to use `Assembly.Load` in this case since TP.Fakes.dll is always in the app base?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -0,0 +1,214 @@
...
+            Func<IEnumerable<string>, string> configurator;
+
+            // fakes supported?
+            if (!TryGetFakesDataCollectorConfigurator(out configurator))
+            {
+                return false;
+            }
+
+            // if no fakes, return settings unchanged
+            var fakesConfiguration = configurator(sources);
```
codito: Can this configuration be done within the datacollector itself?

It will be difficult to perform such magic for external data collectors. May be our datacollector extensibility APIs require an addition to support this?

src/vstest.console/Processors/DisableAutoFakesArgumentProcessor.cs
```diff
@@ -0,0 +1,110 @@
...
+    using System;
+    using System.Globalization;
+
+    using CommandLineResources = Microsoft.VisualStudio.TestPlatform.CommandLine.Resources.Resources;
+
+    /// <summary>
+    /// An argument processor that allows the user to disable fakes
+    /// from the command line using the --DisableAutoFakes|/DisableAutoFakes command line switch.
+    /// </summary>
+    internal class DisableAutoFakesArgumentProcessor : IArgumentProcessor
```
codito: This is interesting. For `code coverage` we require users to specify `/collect`, for fakes we're automatically trying to figure if fakes should be enabled from the provided test sources. I believe this is for compatibility with TPv1.

cc: @Faizan2304 @pvlakshm fyi

codito: I think this will break cloud build if we don't auto enable fakes. Is it a possibility that this can be done in the caller of `vstest.console.exe`?

test/Microsoft.TestPlatform.Common.UnitTests/Utilities/FakesUtilitiesTests.cs
```diff
@@ -0,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace TestPlatform.Common.UnitTests.Utilities
```
codito: We need positive tests?

codito: nit: should the namespace be fully qualified?

src/vstest.console/Processors/DisableAutoFakesArgumentProcessor.cs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    internal class DisableAutoFakesArgumentProcessorCapabilities : BaseArgumentProcessorCapabilities
+    {
+        public override bool AllowMultiple => false;
+        public override string CommandName => DisableAutoFakesArgumentProcessor.CommandName;
+
+        public override ArgumentProcessorPriority Priority => ArgumentProcessorPriority.Normal;
+
+        public override HelpContentPriority HelpPriority => HelpContentPriority.DisableAutoFakesArgumentProcessorHelpPriority;
```
codito: Does this require help text?

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectorInformation.cs
mayankbansal018: We are delaying getting Environment variables for datacollectors. If we do it early(that is before initializing all DataCollectors) then it creates a partial collection plan, which contains info about only one(current) datacollector.

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
mayankbansal018: If we remove this then we can remove reference to System,Reflection, I did it for now so that if we do work for fakes in NetCore, we won't need to make change here.
Either way is fine with me.

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
mayankbansal018: The whole logic to check whether or not to enable fakes lies inside this function (configurator), what you are suggesting is correct, but that would need for us to refactor code so that atleast we figure out whether to enable fakes or not, & then generate appropriate settings for it inside it's data collector.

src/vstest.console/Processors/DisableAutoFakesArgumentProcessor.cs
mayankbansal018: It was present before, so I let it be

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
mayankbansal018: yes I'll remove it

src/vstest.console/Processors/DisableAutoFakesArgumentProcessor.cs
mayankbansal018: The text is there, but is not displayed since the property HelpContentResourceName is not assigned

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
mayankbansal018: Resolved

PR #818, opened 2017-05-17
Pass version parameter to dotnet restore

smadala:
- Pass version parameter to `dotnet restore`, otherwise default "15.0.3-dev" has been consider as version for restore, that show up in deps.json file.


codito:
Please update build.sh

Faizan2304 (approved):

Issue #832, opened 2017-05-25
.NET 4.5 XUnit project using Newtonsoft.Json with incorrect binding redirect doesn't work with TPv2


KevinH-MS:
## Description
> I have a project (``` \\scratch2\scratch\kevinh\XUnitTestProject.Desktop```) that doesn't work with Live Unit Testing (tests time out during execution with no errors).

## Steps to reproduce
> Open above project
> Test > Live Unit Testing > Start

I'm not seeing this with an analogous MSTest project, so I'm guessing it has something to do with xunit.runner.visualstudio, but I can't figure out what it could be.  AFAICT, nothing in xunit references Newtonsoft.Json, so I'm guessing it must have something to do with AppDomain creation or policy.

## Expected behavior
> Tests run and "failed" coverage glyphs appear

## Actual behavior
> Nothing happens for ~60 seconds and then "uncovered" coverage glyphs appear (because tests never execute)

## Environment
> Dev15.3 preview 1

KevinH-MS:
When running with TPv1 (Test Explorer), I get:
```
Test Name:	XUnitTestProject.TestClass1.Test1
Test FullName:	XUnitTestProject.TestClass1.Test1
Test Source:	C:\Users\kevinh\Documents\Visual Studio 2017\Projects\XUnitTestProject.Desktop\XUnitTestProject.Desktop\Class1.cs : line 11
Test Outcome:	Failed
Test Duration:	0:00:00.039

Result StackTrace:	
at XUnitTestProject.TestClass1.Test1()
----- Inner Stack Trace -----
Result Message:	
System.IO.FileLoadException : Could not load file or assembly 'Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
---- System.IO.FileLoadException : Could not load file or assembly 'Newtonsoft.Json, Version=10.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

Faizan2304:
@KevinH-MS : What is the target framework version of test project. 

[...]

PR #835, opened 2017-05-29
Testhost crash info is not coming to Testwindow.

Faizan2304:
Issue: When testhost is crashing before setting up channel with vstest.console, vstest.console should log error message in TestWindow. https://github.com/Microsoft/vstest/issues/832 

Fix: Send error message to Test hub of output window.

Testing: Added UT and done manual testing


src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyDiscoveryManager.cs
```diff
@@ -99,6 +103,13 @@ public void DiscoverTests(DiscoveryCriteria discoveryCriteria, ITestDiscoveryEve
             catch (Exception exception)
             {
                 EqtTrace.Error("ProxyDiscoveryManager.DiscoverTests: Failed to discover tests: {0}", exception);
+
```
smadala: Handling same case [here](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs#L140) too.

Faizan2304: Done!!

PR #752, opened 2017-04-25
Use netcoreapp 1.0.5 shared runtime.

codito:
Fixes an issue in case-sensitive OS where packages are not located
with netcoreapp2.0 sdk and 1.0.4 runtime. See dotnet/cli#6390.

PR #839, opened 2017-06-03
Fix flaky test in datacollector attachments manager.

codito:


codito:
Taking the fix in https://github.com/Microsoft/vstest/pull/752.

PR #869, opened 2017-06-15
Harden parallel execution

codito:
* Update termination condition to use sources count
* Handle a scenario where the concurrent execution manager thread can save leading
to parallel execution hang
* Send abort in case of exception during Discovery or Execution in the runner. This will ensure a new execution manager is spawned to replace in case of abort.
* Do not count tests in case a discovery aborts for a parallel discovery scenario.


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -91,15 +95,22 @@ public int StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsHandler e
...
+                var testCaseLists = testCasesBySource.Values.ToArray();
+                this.testCaseListEnumerator = testCaseLists.GetEnumerator();
+                this.availableTestSources = testCaseLists.Length;
             }
             else
             {
                 // Set the enumerator for parallel yielding of sources
                 // Whenever a concurrent executor becomes free, it picks up the next source using this enumerator
                 this.sourceEnumerator = testRunCriteria.Sources.GetEnumerator();
+                this.availableTestSources = testRunCriteria.Sources.Count();
```
singhsarab: >availableTestSources [](start = 21, length = 20)

nit: should we rename it to TestSourceCount ?

singhsarab:
:shipit:

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
Faizan2304: I am fine with name but if you are changing here then please change in ParallelProxyDiscoveryManager.cs also

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -22,6 +24,8 @@ internal class ParallelProxyExecutionManager : ParallelOperationManager<IProxyEx
         #region TestRunSpecificData
 
         private int runCompletedClients = 0;
+        private int runStartedClients = 0;
```
Faizan2304: Can you please explain why we need this variable? Why runCompletedClients is not fulfilling our requirement?

Faizan2304 (approved):
I have a very small comment, rest look very fine to me 👍 :shipit:

PR #877, opened 2017-06-19
async/await additions to TranslationLayer

JakenVeina:
In the process of writing a custom Test Explorer UI, I found myself writing essentially an async wrapper layer around the synchronous methods of the TranslationLayer. Eventually, I realized the TranslationLayer methods are mostly synchronous wrappers of async methods on ICommunicationManager, so I thought I'd take a stab at bubbling those up to the top as async all the way.

I tried to model all the existing methods as closely as possible. Definitely open to inputs and criticism.

I also wanted to add a new sample project to the solution to demo the use of the new methods with a simple WPF app, but my first attempt at adding a new project to the solution, and setting up dependencies, failed miserably. There's a lot of custom stuff going on in the solution that I don't understand at all, so some help on that, if it can be spared, would be appreciated.


codito:
@JakenVeina thanks for the enhancement! We've been busy with few critical fixes for 15.3 release, will review this PR in the next couple of days.

samples/Microsoft.TestPlatform.TranslationLayer.E2ETest/Program.cs
```diff
@@ -51,6 +51,8 @@ public static int Main(string[] args)
                 }
             }
 
+            Console.WriteLine(Directory.GetCurrentDirectory());
```
codito: Should we prefix this output with `Current directory: {0}` to give context when the tool is run?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITranslationLayerRequestSender.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer.Interfaces
 {
     using System;
     using System.Collections.Generic;
+    using System.Threading.Tasks;
```
codito: Since this interface is internal to vstest, we can probably make this async only.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
```diff
@@ -4,6 +4,7 @@
 namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer.Interfaces
 {
     using System.Collections.Generic;
+    using System.Threading.Tasks;
```
codito: Should we create a `IVsTestConsoleWrapper2` for async methods? Async is probably the way forward for the API set. @Microsoft/vstest what does everybody think?

From implementation of the interface, probably the `VsTestConsoleWrapper` can implement both interfaces. We can possibly reduce the business logic duplication across async and non-async.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -337,6 +470,60 @@ private void SendMessageAndListenAndReportTestCases(IEnumerable<string> sources,
...
+                this.communicationManager.SendMessage(
+                            MessageType.StartDiscovery,
+                            new DiscoveryRequestPayload() { Sources = sources, RunSettings = runSettings },
+                            this.protocolVersion);
+                var isDiscoveryComplete = false;
+
+                // Cycle through the messages that the vstest.console sends.
+                // Currently each of the operations are not separate tasks since they should not each take much time.
+                // This is just a notification.
+                while (!isDiscoveryComplete)
```
codito: This while loop based implementation will eventually get replaced with a event based one. For example, we can compare: [TestRequestSender](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs#L141) with the newer [TestRequestSender2](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender2.cs#L180).

codito: Since the calls will not be blocking with the event based model, we need to consider how they will play with the async interface.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITranslationLayerRequestSender.cs
JakenVeina: If we do that, then the public non-async methods on VsTestConsoleWrapper would just end up calling the async methods internally and waiting them. Same end result, but with extra pointless overhead in the internal use of Tasks.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
JakenVeina: I think it's appropriate to have both async and non-async methods available to consumers, particularly since console applications are one of the more common uses of the platform. However, splitting the methods into async and non-async interfaces, and having the concrete class implement both, sounds like an excellent idea. I kind of hate the idea of anything named MyThing2, but this isn't my codebase. :)

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
JakenVeina: I imagine something like this...

```csharp
public Task DiscoverTestsAsync()
{
    var discoveryCompletionSource = new TaskCompletionSource<int>();

    this.communicationManager.MessageReceived += (sender, e) => 
    {
        switch(e.Type)
        {
            case MessageType.DiscoveryComplete:
                discoveryCompletionSource.SetResult(0);
                break;
        }
    };

    this.communicationManager.SendMessage(startDiscoveryMessage);

    await discoveryCompletionSource.Task;
}
```

And equivalently for DiscoverTests().

Unless you're planning on exposing an event-based interface publicly?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
singhsarab: Agree. Separate interface would look clean.

JakenVeina: I ended up not having VsTestConsoleWrapper implement both interfaces, since it has a constructor that takes ITranslationLayerRequestSender as an injected dependency, and all the async methods would require ITranslationLayerRequestSenderAsync instead. Instead, I just forked a new class for the async interface.

codito: @JakenVeina on a quick glance, the new changes looks great to me.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -268,16 +369,48 @@ private bool HandShakeWithVsTestConsole()
...
+        private async Task<bool> HandShakeWithVsTestConsoleAsync()
+        {
+            var success = false;
+            var message = await this.communicationManager.ReceiveMessageAsync(this.processExitCancellationTokenSource.Token);
+            if (message.MessageType == MessageType.SessionConnected)
+            {
+                this.communicationManager.SendMessage(MessageType.VersionCheck, this.protocolVersion);
+                message = await this.communicationManager.ReceiveMessageAsync(this.processExitCancellationTokenSource.Token);
+
+                if (message.MessageType == MessageType.VersionCheck)
```
codito: The switch logic remains same across the sync and async variants of `HandShake`, `Discovery` and `Execution`. Could we extract it to a method to reduce duplication?

JakenVeina: Internal static method on the non-async class?

codito: I meant in the current class `VsTestConsoleRequestSender`, for example, only change between the business logic in `SendMessageAndListenAndReportTestCases` and `SendMessageAndListenAndReportTestCasesAsync` is the `TryReceiveMessage` call.

codito (approved):

codito:
Merging this, will take up the changes for DRY violation in a separate PR. Thanks for contributing this @JakenVeina!

PR #906, opened 2017-06-30
Building platform abstraction for UWP10.0

mayankbansal018:
-Adding Abstraction for Stream


codito:
Can we add the usages for these API please? Few use cases like `PlatformStream.Instance` are not very clear.

codito (changes requested):
We should add the usage of these APIs in the current PR. It will help understand the design better.

Faizan2304 (approved):
I am assuming this is just the base to build uap code. In later PR implementation for uwp will come.

PR #935, opened 2017-07-14
Copied only datacollector dll and pdb to extensions folder.

harshjain2:
Copied only datacollector dll and pdb to extensions folder.


singhsarab (approved):
Please validate before pushing in.

harshjain2:
Manually validated, also there is an e2e test that is running as part of pr build.

codito (approved):

Issue #938, opened 2017-07-17
Reliability: test host clean up issues


codito:
(This issue shows up inconsistently)

## Callstack 1:
Unhandled Exception: System.InvalidOperationException: No process is associated with this object.
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.get_HasExited()
   at Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.ProcessHelper.TryGetExitCode(Object process, Int32& exitCode)
   at Microsoft.TestPlatform.TestHostProvider.Hosting.TestHostManagerCallbacks.ExitCallBack(IProcessHelper processHelper, Object process, StringBuilder testHostProcessStdError, Action`1 onHostExited)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting.DefaultTestHostManager.<get_ExitCallBack>b__32_0(Object process)
   at Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.ProcessHelper.<>c__DisplayClass0_0.<LaunchProcess>b__1(Object sender, EventArgs args)
   at System.Diagnostics.Process.OnExited()
   at System.Diagnostics.Process.RaiseOnExited()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading._ThreadPoolWaitOrTimerCallback.PerformWaitOrTimerCallback(Object state, Boolean timedOut)

## Callstack 2
Unhandled Exception: System.InvalidOperationException: No process is associated with this object.
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.GetProcessHandle(Int32 access, Boolean throwIfExited)
   at System.Diagnostics.Process.WaitForExit(Int32 milliseconds)
   at Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.ProcessHelper.<>c__DisplayClass0_0.<LaunchProcess>b__1(Object sender, EventArgs args)
   at System.Diagnostics.Process.OnExited()
   at System.Diagnostics.Process.RaiseOnExited()
[...]

PR #941, opened 2017-07-17
Fix for InvalidOperationException thrown by Process.Exited

harshjain2:
Fix for InvalidOperationException thrown by Process.Exited.
Fixes https://github.com/Microsoft/vstest/issues/938


src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -121,18 +133,24 @@ public void SetExitCallback(int processId, Action callbackAction)
...
         /// <inheritdoc/>
         public void TerminateProcess(object process)
         {
             var proc = process as Process;
-            if (proc != null && !proc.HasExited)
+            try
+            {
+                if (proc != null && !proc.HasExited)
+                {
+                    proc.Kill();
```
Faizan2304: This can return `Win32Exception` as well if the process is terminating?
Shouldn't we catch that too?

Faizan2304 (approved):
Gave a comment, please take a look. Rest look good to me 👍

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -50,8 +50,14 @@ public object LaunchProcess(string processPath, string arguments, string working
                     {
                         // Call WaitForExit again to ensure all streams are flushed
                         var p = sender as Process;
-                        p.WaitForExit();
-                        exitCallBack(p);
+                        try
+                        {
+                            p.WaitForExit();
+                            exitCallBack(p);
```
codito: Is it possible that `exitCallBack` may not be called?

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
harshjain2: IMHO, we should bubble up `Win32Exception` because this exception can also be thrown if process cannot be terminated.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
harshjain2: In my opinion, we should not call exitCallBack if InvalidOperationException is thrown while accessing Process.HasExited API.

However, since we are also logging the testHostProcessStdError, I am moving this method call outside try catch block.

codito (approved):

mayankbansal018 (commented):
This still doesn't solve the issue, when in cleanup testhost we are disposing the test host. Agreed that no more exceptions will come, but we will lose streams, & exit codes, if disposed is called, before we complete those operations

One option is that we only dispose if host exited event is raised.

PR #943, opened 2017-07-18
Lazy initialized TraceSource

harshjain2:
Lazy initialized TraceSource, added `DoNotInitialize` boolean to prevent initialization TraceSource when EqtTrace is not enabled.

This is bound to give a perf benefit of ~ 5% (400 ms).


src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/EqtTrace.cs
```diff
@@ -86,7 +102,7 @@ public partial class PlatformEqtTrace : IPlatformEqtTrace
         /// <summary>
         /// Specifies whether the trace is initialized or not
         /// </summary>
-        private static bool isInitialized = false;
+        private static bool isInitialized = false;
```
Faizan2304: nit: Remove extra white space

Faizan2304 (approved):
👍

codito:
Will take a look at this in conjunction with the TraceListener.Add related performance investigation results for completeness.

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
```diff
@@ -131,6 +130,35 @@ public static string ErrorOnInitialization
             set;
         }
 
+        // This is added to ensure that traceSource should not be instantiated in when creating appdomains if EqtTrace is not enabled.
+        internal static bool DoNotInitialize
```
AbhitejJohn: wouldn't this permeate to non-remote trace flow as well?

src/Microsoft.TestPlatform.PlatformAbstractions/net451/Tracing/PlatformEqtTrace.cs
```diff
@@ -29,25 +29,32 @@ public partial class PlatformEqtTrace : IPlatformEqtTrace
         public void SetupRemoteEqtTraceListeners(AppDomain childDomain)
         {
             Debug.Assert(childDomain != null, "domain");
-            if (childDomain != null && !Enum.Equals(TraceLevel, TraceLevel.Off))
+            if (childDomain != null)
             {
                 RemoteEqtTrace remoteEqtTrace = (RemoteEqtTrace)childDomain.CreateInstanceFromAndUnwrap(
                     typeof(RemoteEqtTrace).Assembly.Location,
                     typeof(RemoteEqtTrace).FullName);
```
AbhitejJohn: so we are creating a remote trace object in a new appdomain even the trace is off. Seems odd.

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
```diff
@@ -147,6 +175,11 @@ public bool InitializeVerboseTrace(string customLogFile)
...
         public bool ShouldTrace(PlatformTraceLevel traceLevel)
         {
+            if (DoNotInitialize)
+            {
+                return false;
+            }
+
             switch (traceLevel)
             {
                 case PlatformTraceLevel.Off:
```
AbhitejJohn: <GitHub does not let me add comments to the line I want :-/ >
If I understand the reason for this change - Calling `ShouldTrace` even if `PlatformTraceLevel.Off` is the target, we end up initializing the Source instance since it is static. Can we just pull that out to a separate method instead please? You would not need a DoNotInitialize then.
That apart, should we be caching Source.Switch.ShouldTrace value and reuse?

AbhitejJohn (changes requested):

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
harshjain2: No...
Default value of DoNotInitialize is `false`. This variable is only set while creating AppDomains. 
Since `PlatformEqtTrace` is not inheriting from `marshalbyrefobject` or Serializable, this value will not interfere with parent appdomain traces.

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
harshjain2: The intent of this PR is to prevent initialization of EqtTrace in child app domains when EqtTrace is not enable in parent app domain. 
In the parent app domain, we need to initialize `TraceSource` object to find out whether the EqtTrace is enabled using app.config file. And therefore, we can't prevent instantiating `TraceSource` at least once.

harshjain2: Also, once the `TraceSource` is initialized, the values are cached, `Source.Switch.ShouldTrace` is read from cache.

src/Microsoft.TestPlatform.PlatformAbstractions/net451/Tracing/PlatformEqtTrace.cs
harshjain2: We need to create the `RemoteEqtTrace` object because we want to set the value of DoNotInitialize variable. 
If EqtTrace is not initialized in parent app domain, then we set `DoNotInitialize` in child app domain, will further stop instantiation of `TraceSource` which is a costly operation as it involves reading app.config file.

Faizan2304:
@dotnet-bot test Ubuntu16.04 / Release Build please

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
AbhitejJohn: <Spoke to harsh offline>
We plan to cache the current PlatformTraceSource value and use that to bail out faster checking for `PlatformTraceSource.Off`.

src/Microsoft.TestPlatform.PlatformAbstractions/net451/Tracing/PlatformEqtTrace.cs
AbhitejJohn: <Spoke to harsh offline>
We could default the current `PlatformTraceSource` discussed above to Off.

AbhitejJohn (approved):

codito (approved):

PR #948, opened 2017-07-20
Datacollector env variables should be given more priority.

smadala:
- When Datacollector want to set env variables and if same env variable already set in system level. Datacollector value should be given more priority.

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/net451/System/ProcessStartInfoExtensions.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/net451/System/ProcessStartInfoExtensions.cs
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
         /// <param name="value">Environment Variable value.</param>
         public static void AddEnvironmentVariable(this ProcessStartInfo startInfo, string name, string value)
         {
-            startInfo.EnvironmentVariables.Add(name, value);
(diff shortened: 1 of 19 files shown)


mayankbansal018 (commented):
Changes look good, but any particular reason why we are not adding test case now?

src/Microsoft.TestPlatform.PlatformAbstractions/net451/System/ProcessStartInfoExtensions.cs
```diff
@@ -15,7 +15,7 @@ public static class ProcessStartInfoExtensions
         /// <param name="value">Environment Variable value.</param>
         public static void AddEnvironmentVariable(this ProcessStartInfo startInfo, string name, string value)
         {
-            startInfo.EnvironmentVariables.Add(name, value);
+            startInfo.EnvironmentVariables[name] = value;
```
mayankbansal018: Have we tested that this doesn't throw exception on overriding?
Earlier even add was not throwing exception.

mayankbansal018 (approved):

codito (changes requested):
Require tests

src/Microsoft.TestPlatform.PlatformAbstractions/net451/System/ProcessStartInfoExtensions.cs
smadala: Tested it. Worked as expected.

codito (approved):

Issue #919, opened 2017-07-10
Test Run level timeout


AbhitejJohn:
The test platform should have the ability for users to specify a test run level timeout which allows users to terminate a test session when it exceeds a given timeout. This ensures that resources are well consumed and test sessions are constrained to a set time. A proposal for having this setting could be in the runsettings:
```
<Runsettings>
  <RunConfiguration>
     <!-- Specify timeout in ms -->
     <TestSessionTimeout>10000</TestSessionTimeout>
  </RunConfiguration>
</Runsettings>
```
Users could also pass this in as a command line argument as follows:
```
vstest.console <testContainersList> -- RunConfiguration.TestSessionTimeout=10000
```

Faizan2304:
Completed. Will come in next release of Visual Studio/Dotnet cli.

Faizan2304:
For detailed info about this feature https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0011-Test-Session-Timeout.md

PR #949, opened 2017-07-20
Implement TestSessionTimeout

Faizan2304:
Issue: https://github.com/Microsoft/vstest/issues/919


src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
```diff
@@ -96,7 +98,10 @@ public virtual int StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsH
                     testSources = testRunCriteria.Tests.GroupBy(tc => tc.Source).Select(g => g.Key);
                 }
 
-                this.isCommunicationEstablished = this.SetupChannel(testSources, this.cancellationTokenSource.Token);
+                lock (locObject)
```
mayankbansal018: why do we need a lock while setting up channel? Is it possible that we try to setup communication even though test run times out?

Faizan2304: Consider a scenario where Channel is getting setup (testhost has been launched but setup channel didn't return back ( First thread is at this point [code](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs#L130)) ) and in between Cancelation request come. Since at this point channel is not setup completely cancel function will not call cancel of test host ( Second thread is at this point [code](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs#L166))

To deal this situation we need lock here

mayankbansal018: In case when the test request timed out, before the set up channel could complete, you are allowing the testhost launch to complete, & wait for the communication to establish, & then you will send the cancellation request. Would it not be helpful if you just set cancellation token & hope that if the test host was not launched, then it might get cancel, & lock() the Cancel request?

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -155,6 +161,22 @@ public long BatchSize
             }
         }
 
+        /// <summary>
+        /// Gets or sets the frequency of the runStats/discoveredTests event. Should be non-negative integer.
```
codito: nit: fix the doc

test/Microsoft.TestPlatform.Client.UnitTests/Execution/TestRunRequestTests.cs
```diff
@@ -140,6 +142,66 @@ public void CancelAsyncIfTestRunStateInProgressCallsExecutionManagerCancel()
...
+                     </RunConfiguration>
+                </RunSettings>";
+
+            var testRunCriteria = new TestRunCriteria(new List<string> { "foo" }, 1, true, settingsXml);
+            var executionManager = new Mock<IProxyExecutionManager>();
+            var testRunRequest = new TestRunRequest(testRunCriteria, executionManager.Object);
+
+            ManualResetEvent onTestSessionTimeoutCalled = new ManualResetEvent(true);
+
+            executionManager.Setup(o => o.StartTestRun(testRunCriteria, testRunRequest)).Callback(() => { System.Threading.Thread.Sleep(3 * 1000); });
```
codito: Can we not depend on Thread.Sleep?

test/Microsoft.TestPlatform.Client.UnitTests/Execution/TestRunRequestTests.cs
```diff
@@ -140,6 +142,66 @@ public void CancelAsyncIfTestRunStateInProgressCallsExecutionManagerCancel()
...
+
+            executionManager.Setup(o => o.Cancel()).Callback(() => onTestSessionTimeoutCalled.Set());
+
+            testRunRequest.ExecuteAsync();
+
+            onTestSessionTimeoutCalled.WaitOne(20 * 1000);
+
+            executionManager.Verify(o => o.Cancel(), Times.Once);
+
```
codito: nit: please remove extra lines. Divide every test into three blocks of code: arrange, act and assert.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -84,6 +101,17 @@ public int ExecuteAsync()
             }
         }
 
+        internal void OnTestSessionTimeout(object obj)
+        {
+            string message = String.Format(ClientResources.TestSessionTimeoutMessage, this.testSessionTimeout);
```
codito: Knowledge of json serializer and that we need to log to two instances (regular and rawmessage) has now percolated to TestRunRequest. Is there a better way to hide this complexity?

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -84,6 +101,17 @@ public int ExecuteAsync()
...
+            var testMessagePayload = new CommunicationObjectModel.TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = message };
+            var rawMessage = this.dataSerializer.SerializePayload(CommunicationObjectModel.MessageType.TestMessage, testMessagePayload);
+
+            this.HandleLogMessage(TestMessageLevel.Error, message);
```
codito: Suggest adding an integration test for this change.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -464,5 +492,11 @@ protected virtual void Dispose(bool disposing)
         /// Tracks the time taken by each run request
         /// </summary>
         private Stopwatch runRequestTimeTracker;
+
+        private IDataSerializer dataSerializer;
```
codito: nit: suggest to move variable declarations up.

src/Microsoft.TestPlatform.Client/Resources/Resources.resx
```diff
@@ -120,6 +120,9 @@
   <data name="InvalidStateForExecution" xml:space="preserve">
     <value>The test run could not be executed because the initial state was invalid.</value>
   </data>
+  <data name="TestSessionTimeoutMessage" xml:space="preserve">
+    <value>Canceling test run as it crossed testSessionTimeout: {0} milliseconds.</value>
```
codito: Can we please add few screenshots and request @pvlakshm to review?

codito: `testSessionTimeout` looks like an internal implementation name, it may not be exposed in user message?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
codito: Does `POM.Close` wait until testhost cleans up?

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -84,6 +101,17 @@ public int ExecuteAsync()
...
 
+        internal void OnTestSessionTimeout(object obj)
+        {
+            string message = String.Format(ClientResources.TestSessionTimeoutMessage, this.testSessionTimeout);
+            var testMessagePayload = new CommunicationObjectModel.TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = message };
+            var rawMessage = this.dataSerializer.SerializePayload(CommunicationObjectModel.MessageType.TestMessage, testMessagePayload);
+
+            this.HandleLogMessage(TestMessageLevel.Error, message);
+            this.HandleRawMessage(rawMessage);
+            this.CancelAsync();
```
codito: Do we need any verbose logs in the changes in current PR? E.g. cancel is called by test host cleanup takes time, how can user know?

codito (changes requested):

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -84,6 +101,17 @@ public int ExecuteAsync()
             }
         }
 
+        internal void OnTestSessionTimeout(object obj)
+        {
+            string message = String.Format(ClientResources.TestSessionTimeoutMessage, this.testSessionTimeout);
+            var testMessagePayload = new CommunicationObjectModel.TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = message };
+            var rawMessage = this.dataSerializer.SerializePayload(CommunicationObjectModel.MessageType.TestMessage, testMessagePayload);
```
AbhitejJohn: This looks odd... We are serializing a message even though it is used in the same process again. Would be great to not do that. 
One more question: Why do we need both a `HandleLogMessage` and a `HandleRawMessage` ? Does not seem very evident since they both seem to be doing the same thing.

src/vstest.console/CommandLine/TestRunResultAggregator.cs
```diff
@@ -107,13 +107,11 @@ public void Reset()
         /// </summary>
         private void TestRunCompletionHandler(object sender, TestRunCompleteEventArgs e)
         {
-            if (e.TestRunStatistics == null)
+            if (e.TestRunStatistics == null || e.IsCanceled || e.IsAborted)
             {
```
AbhitejJohn: are we not returning different exit codes?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
Faizan2304: > Does POM.Close wait until testhost cleans up?

Yes it wait for 100 ms only after that it forcefully kill testhost.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: Good point, need to think through it. Currently in all the places we are doing this way only.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: `HandleLogMessage` is used to log message to console and `HandleRawMessage` is for Designmode Client.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: Fixed!

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
Faizan2304: We are not using Lock now, instead we are checking cancelation token after initialization to cancel run.

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
Faizan2304: Fixed

test/Microsoft.TestPlatform.Client.UnitTests/Execution/TestRunRequestTests.cs
Faizan2304: Fixed

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: Fixed

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
AbhitejJohn: oh man... hmm.. are we planning to fix this? seems like a hack.

test/Microsoft.TestPlatform.Client.UnitTests/Execution/TestRunRequestTests.cs
Faizan2304: Yes we can achieve without Thread.Sleep()

src/vstest.console/CommandLine/TestRunResultAggregator.cs
Faizan2304: For now we are returning exit code `1` for cancel , fail and abort tests. In future we are planning to have different exit code for each  cancel , fail and abort

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: Currently at every place we are doing this only :(

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
Faizan2304: Added

codito (approved):

PR #956, opened 2017-07-25
Dump Creation In Blame Mode added

vagisha-nidhi:
Added option to create dump in blame mode by --blame:dump|/blame:dump


singhsarab:
https://github.com/Microsoft/vstest/pull/1458

PR #963, opened 2017-07-28
Platform Trace refactor stab.

AbhitejJohn:
Took a stab at refactoring the trace class to optimize the flow. This is something I stumbled upon and thought could be improved. Let me know if I'm missing anything obvious that requires all the fluff. 

This requires some comments cleanup.


src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
```diff
@@ -36,45 +36,31 @@ public partial class PlatformEqtTrace : IPlatformEqtTrace
         private static readonly TraceSource Source = new TraceSource("TpTrace", SourceLevels.Off);
 
         /// <summary>
-        /// Create static maps for TraceLevel to SourceLevels. The APIs need to provide TraceLevel
-        /// for backward compatibility with older versions of Object Model.
-        /// </summary>
```
AbhitejJohn: I'm guessing this refers to TP V1's object model. If yes, that does not know of something called PlatformEqtTrace - It is well abstracted from the EqtTrace class in CoreUtilities.

PR #964, opened 2017-07-31
Allow assemblies to specify default resource

mayankbansal018:
UWP release does not pick resources if this attribute is not specified.


src/testhost.x86/DefaultEngineInvoker.cs
```diff
@@ -23,8 +23,9 @@ internal class DefaultEngineInvoker :
     {
         /// <summary>
         /// The timeout for the client to connect to the server.
+        /// Increasing Timeout to allow client to connect, not always the client can connect within 5 seconds
         /// </summary>
-        private const int ClientListenTimeOut = 5 * 1000;
+        private const int ClientListenTimeOut = 30 * 1000;
```
codito: Is this related to UWP?

codito (approved):

PR #974, opened 2017-08-03
This will make vstest.console run as 32bit process on all machines

mayankbansal018:
while if loaded inside a 64 bit process, it will load normally.

### DIFF
--- a/src/vstest.console/vstest.console.csproj
+++ b/src/vstest.console/vstest.console.csproj
@@ -10,6 +10,7 @@
     <WarningsAsErrors>true</WarningsAsErrors>
     <OutputType>Exe</OutputType>
     <PlatformTarget Condition="'$(TargetFramework)' == 'net451'">AnyCPU</PlatformTarget>
+    <Prefer32Bit Condition="'$(TargetFramework)' == 'net451'">true</Prefer32Bit>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


src/vstest.console/vstest.console.csproj
```diff
@@ -10,6 +10,7 @@
     <WarningsAsErrors>true</WarningsAsErrors>
     <OutputType>Exe</OutputType>
     <PlatformTarget Condition="'$(TargetFramework)' == 'net451'">AnyCPU</PlatformTarget>
+    <Prefer32Bit Condition="'$(TargetFramework)' == 'net451'">true</Prefer32Bit>
```
codito: This will break universal applications support in `dotnet test`. I think it is a deal breaker.

mayankbansal018: Irrespective of this UWP support won't work in dotnet test, but with this we can atleast bring it up in vstest.console
I agree that we don't want to be specific to vstest.console, but right now I see no option
I'll update UWP team, to take consider that scenario as well, & try to port themselves to 64bit as well

codito: We will not support dotnet test immediately in the first iteration of UWP support.

codito (approved):

PR #976, opened 2017-08-04
Use test container capability in test projects.

codito:
Update build scripts to not download runtimes if they are already installed.

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -52,6 +52,11 @@
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
       </PropertyGroup>
 
+      <!-- Test projects are not discovered in test window without test container capability -->
+      <ItemGroup>
+        <ProjectCapability Include="TestContainer" />
+      </ItemGroup>
+
       <!-- Test project references -->
       <ItemGroup>
         <!-- Test projects take a direct dependency on TestHost. Indirect dependencies must be included.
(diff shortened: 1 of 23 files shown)


scripts/build/TestPlatform.Settings.targets
```diff
@@ -52,6 +52,11 @@
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
       </PropertyGroup>
 
+      <!-- Test projects are not discovered in test window without test container capability -->
+      <ItemGroup>
```
Faizan2304: This file will get included in product code as well. Shouldn't we add this only for test project?

Faizan2304 (approved):

scripts/build/TestPlatform.Settings.targets
```diff
@@ -52,6 +52,11 @@
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
       </PropertyGroup>
 
+      <!-- Test projects are not discovered in test window without test container capability -->
```
smadala: Please remove `<Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />`. I think tests should get discovered.

scripts/build/TestPlatform.Settings.targets
codito: It is under a TestProject condition, see Line 50.

smadala (approved):

scripts/build/TestPlatform.Settings.targets
Faizan2304: Ohh yes.. My bad 👍

PR #978, opened 2017-08-04
Update xunit assets to latest beta. Use net452 for perf test project.

codito:


scripts/build/TestPlatform.Dependencies.props
```diff
@@ -10,8 +10,8 @@
     <MSTestAdapterVersion>1.2.0-beta</MSTestAdapterVersion>
     <MSTestAssertExtensionVersion>1.0.3-preview</MSTestAssertExtensionVersion>
     
-    <XUnitFrameworkVersion>2.2.0</XUnitFrameworkVersion>
-    <XUnitAdapterVersion>2.2.0</XUnitAdapterVersion>
+    <XUnitFrameworkVersion>2.3.0-beta3-build3705</XUnitFrameworkVersion>
+    <XUnitAdapterVersion>2.3.0-beta3-build3705</XUnitAdapterVersion>
 
     <NUnit3FrameworkVersion>3.6.1</NUnit3FrameworkVersion>
     <NUnitFrameworkVersion>2.6.4</NUnitFrameworkVersion>
```
smadala: Due to `XUPerfTestProject.csproj` file encoded as Unicode, Change the encoding to UTF-8.  It didn't get merge properly, service still guid exists. Change to net452 at DebugType also.

smadala (approved):

scripts/build/TestPlatform.Dependencies.props
smadala: Please import `Microsoft.NET.Test.Sdk.props` & `Microsoft.NET.Test.Sdk.targets` directly in TestPlatform.Settings.targets for Test project. This can avoid thinks like DebugType, Entry point for .netcore and ProjectCapabilities miss match for test platform test projects.

smadala: > Service guids are not changed for test assets in last PR since we don't care if those tests display in IDE/Editor.

I guess either way service guid is not need relevant anymore.
> Good point, will do it outside this PR.

Makes sense.

Issue #543, opened 2017-02-25
Errors related to incorrect adapter for a unit test project 


sbaid:
## Description
If in a VS instance mstest project is opened and later xunit project is opened, mstest adapter related errors are seen in test output pane and vice-versa.

## Steps to reproduce
1. Open VS. Create new project -> .NET core mstest
2.  run all tests, close solution.
3. Create new project -> .NET core xunit 
4. run all tests.

## Expected behavior
No Errors in test output pane.

## Actual behavior
Errors in test output pane.

`````````````
------ Discover test started ------
[xUnit.net 00:00:00.8335609]   Discovering: XUnitTestProject1
[xUnit.net 00:00:00.9900386]   Discovered:  XUnitTestProject1
File does not exist: C:\Users\sbaid\Documents\Visual Studio 2017\Projects\XUnitTestProject1\XUnitTestProject1\bin\Debug\netcoreapp1.1\XUnitTestProject1.dll
========== Discover test finished: 1 found (0:00:01.7870696) ==========
------ Run test started ------
[xUnit.net 00:00:00.7583411]   Discovering: XUnitTestProject1
[xUnit.net 00:00:00.8918778]   Discovered:  XUnitTestProject1
[xUnit.net 00:00:00.9480573]   Starting:    XUnitTestProject1
[xUnit.net 00:00:01.1502194]   Finished:    XUnitTestProject1
File does not exist: C:\Users\sbaid\Documents\Visual Studio 2017\Projects\XUnitTestProject1\XUnitTestProject1\bin\Debug\netcoreapp1.1\XUnitTestProject1.dll
An exception occurred while invoking executor 'executor://mstestadapter/v2': Could not load file or assembly 'System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.
========== Run test finished: 1 run (0:00:02.0190726) ==========
`````````````

## Environment
latest d15rel build. Projects were targeting `netcoreapp1.1` . Latest build template defaults to `netcoreapp1.1`

PR #1006, opened 2017-08-18
Updating Extensions Correctly

singhsarab:
Changes to ensure UpdateAdditionalExtensions does not append to previous list of extensions. It should rather replace PathToExtensions with new list of extensions.

Fixes : https://github.com/Microsoft/vstest/issues/543


src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -304,6 +304,14 @@ public void UpdateExtensions(IEnumerable<string> additionalExtensionsPath, bool
             }
         }
 
+        /// <summary>
+        /// Clear the previously cached extensions
+        /// </summary>
+        public void ClearExtensions()
+        {
+            this.pathToExtensions?.Clear();
```
Faizan2304: Shouldn't we clear TestExtensions cache?

Faizan2304 (changes requested):
LGTM 👍 

After this change design mode client will be able to initialize extension only once. Subsequent request to update extension will clear the previous extensions.

smadala (approved):

PR #1059, opened 2017-09-08
Support for UWP

mayankbansal018:
Moving System.Xml.ReaderWriter to lower version to allow UWP support
Adding Executor layer to support Old UWP projects
Adding fallback in dia session, to read pdb from current directory if path is not rooted
Adding exception for UWP C++ adapters


mayankbansal018:
@dotnet-bot test Windows_NT / Release Build

scripts/build.ps1
```diff
@@ -161,9 +161,9 @@ function Restore-Package
     $dotnetExe = Get-DotNetPath
 
     Write-Log ".. .. Restore-Package: Source: $TPB_Solution"
-    & $dotnetExe restore $TPB_Solution --packages $env:TP_PACKAGES_DIR -v:minimal -warnaserror -p:Version=$TPB_Version
+    & $dotnetExe restore $TPB_Solution --packages $env:TP_PACKAGES_DIR -v:minimal -warnaserror:false -p:Version=$TPB_Version
```
codito: Reason for this change

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
```diff
@@ -63,6 +63,17 @@ internal class TestPluginDiscoverer
 
 #if !WINDOWS_UAP
 
+            if(!extensionPaths.Any())
+            {
+                // C++ UWP adaters do not follow TestAdapater naming convention, so making this exception)
```
codito: nit: spelling

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
```diff
@@ -63,6 +63,17 @@ internal class TestPluginDiscoverer
 
 #if !WINDOWS_UAP
 
+            if(!extensionPaths.Any())
```
codito: Is this code supposed to be in `!WINDOWS_UAP`?

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
```diff
@@ -63,6 +63,17 @@ internal class TestPluginDiscoverer
...
+            {
+                // C++ UWP adaters do not follow TestAdapater naming convention, so making this exception)
+                extensionPaths = extensionPaths.Concat<string>(new List<string> { "Microsoft.VisualStudio.TestTools.CppUnitTestFramework.ComInterfaces.dll",
```
codito: Shouldn't this be an const array? Should there be a check if these files exist before adding them for discovery?

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
```diff
@@ -26,6 +26,11 @@
       <FromP2P>true</FromP2P>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
```
codito: Please add a comment to maintain context of this change. Same comment everywhere.

Which API in XmlReaderWriter is not available for UAP?

scripts/build.ps1
harshjain2: Why are we setting `-warnaserror` as false?

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
```diff
@@ -47,7 +47,7 @@ public DiaSession(string binaryPath)
         /// search path.
         /// </param>
         public DiaSession(string binaryPath, string searchPath)
-            : this(binaryPath, searchPath, GetSymbolReader(binaryPath))
+            : this(binaryPath, searchPath, GetSymbolReader(binaryPath, !Path.IsPathRooted(binaryPath) ? Directory.GetCurrentDirectory() : null))
```
codito: Added code looks like magic. What are we trying to do? Is there a platform specific assumption? When would the path be not rooted?

May be we should add a comment.

codito: Why not move the `PathRooted` logic to `GetSymbolReader`?

scripts/build.ps1
mayankbansal018: since we are downgrading Xml.ReaderWriter, the dotnet compiler throws warning, as the one that comes as part of NetStandard meta package is of higher version. Although this does not fail the build, but these warnings show up in red, so disabling marking warning as errors

src/testhost.x86/UnitTestClient.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Initialization point for Old UWP application
+    /// </summary>
+    public static class UnitTestClient
+    {
+        /// <summary>
+        /// Create default UI for UWP app
+        /// </summary>
+        public static void CreateDefaultUI()
+        {
+
```
codito: nit: remove extra lines

src/testhost.x86/UnitTestClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+
+        /// <summary>
+        /// Entry point for testhost, in App Model(UWP) scenario
+        /// </summary>
+        /// <param name="arguments">testhost initialization arguments</param>
+        public static void Run(string arguments)
```
codito: Require an E2E test so that this logic is validated.

src/testhost.x86/UnitTestClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            Task.Run(() =>
+            {
+                Program.Run(arguments.Split(' ').ToArray());
```
codito: Assumption that an argument cannot contain a string like: `/someKey="some value"`. Is it correct?

codito (changes requested):

scripts/build.ps1
mayankbansal018: since we are downgrading Xml.ReaderWriter, the dotnet compiler throws warning, as the one that comes as part of NetStandard meta package is of higher version. Although this does not fail the build, but these warnings show up in red, so disabling marking warning as errors

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
mayankbansal018: the other code has no significance, since we are no longer compiling for UAP, we can remove all other code.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
mayankbansal018: adding check is not necessary, since it will fail at worst during assembly load, with file not file exception. I'm okay with it, since it doesn't break anything, and assembly load will check for file existence anyway

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
mayankbansal018: it's not the API, but since we have taken dependency on version 4.3.0, it files while trying to load xmlreader, as the one present in UWP is 4.0.11 by default.

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
mayankbansal018: for all remote scenario, the path is always not rooted, the check for it could be moved to GetSymbolReader

src/testhost.x86/UnitTestClient.cs
mayankbansal018: Yes, we are passing these arguments, & this is how we are creating dictionary args.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
harshjain2: Where are we shipping these dlls? Extensions folder?

scripts/build.ps1
harshjain2: Why are we setting -warnaserror as false?

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
harshjain2: Where are we shipping these dlls? Extensions folder?
If in Extensions folder, then we need not to do this explicitly.

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
harshjain2: Can we use IFileHelper API here.

harshjain2: Directory.GetCurrentDirectory() might not give the right path. 
https://github.com/dotnet/project-system/issues/2239

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
mayankbansal018: they are not in extension folder, they come appx folder when UWP app is deployed

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
```diff
@@ -47,7 +47,7 @@ public DiaSession(string binaryPath)
         /// search path.
         /// </param>
         public DiaSession(string binaryPath, string searchPath)
```
harshjain2: Add/update UTs for the changes made in UT.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
codito: Please remove it.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
codito: not fixed.

mayankbansal018:
muddled

PR #1074, opened 2017-09-12
Exposing filter support on test discovery

abhishekkumawat23:
**Contains following changes**:
1. Setting filter expression in discovery context
2. Exposing GetTestCaseFilter from discovery context

**Workaround**:
Instead of exposing GettestcaseFilter from IDiscoveryContext, GetTestCaseFilter is a method of DiscoveryContext. Thus, all adapters need to use reflection to access GetTestCaseFilter. 

Addition of GetTestCaseFilter in IDiscoveryContext can be done when callout to all adapters is made that adapter version having these changes can no longer support old vs versions.

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
@@ -242,6 +242,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.Parallel
 
                 // Kick off another discovery task for the next source
                 var discoveryCriteria = new DiscoveryCriteria(new[] { nextSource }, this.actualDiscoveryCriteria.FrequencyOfDiscoveredTestsEvent, this.actualDiscoveryCriteria.DiscoveredTestEventTimeout, this.actualDiscoveryCriteria.RunSettings);
+                discoveryCriteria.TestCaseFilter = this.actualDiscoveryCriteria.TestCaseFilter;
                 Task.Run(() =>
                     {
                         if (EqtTrace.IsVerboseEnabled)
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
@@ -3,6 +3,7 @@
 
 namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Discovery
 {
+    using Microsoft.VisualStudio.TestPlatform.Common.Filtering;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
 
     /// <summary>
@@ -14,5 +15,10 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Discovery
         /// Gets the run settings specified for this request.
         /// </summary>
         public IRunSettings RunSettings { get; internal set; }
+
+        /// <summary>
+        /// Gets or sets the FilterExpressionWrapper instance as created from filter string.
+        /// </summary>
+        internal FilterExpressionWrapper FilterExpressionWrapper {get; set; }
     }
 }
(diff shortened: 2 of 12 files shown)


codito (approved):
Looks good to me.

codito:
@smadala @singhsarab can you take a look please?

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
```diff
@@ -14,5 +23,44 @@ public class DiscoveryContext : IDiscoveryContext
...
+                }
+            }
+
+            return adapterSpecificTestCaseFilter;
+        }
+
+        /// <summary>
+        /// Gets or sets the FilterExpressionWrapper instance as created from filter string.
+        /// </summary>
+        internal FilterExpressionWrapper FilterExpressionWrapper {get; set; }
```
acesiddhu: private?

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
```diff
@@ -14,5 +23,44 @@ public class DiscoveryContext : IDiscoveryContext
...
         public IRunSettings RunSettings { get; internal set; }
+
+        /// <summary>
+        /// Returns TestCaseFilterExpression validated for supportedProperties.
+        /// If there is a parsing error in filter expression, TestPlatformFormatException() is thrown.
+        /// </summary>
+        /// <param name="supportedProperties"> The supported Properties. </param>
+        /// <param name="propertyProvider"> The property Provider. </param>
+        /// <returns> The <see cref="ITestCaseFilterExpression"/>. </returns>
+        public virtual ITestCaseFilterExpression GetTestCaseFilter(IEnumerable<string> supportedProperties, Func<string, TestProperty> propertyProvider)
```
acesiddhu: remove virtual

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -242,6 +242,7 @@ private void DiscoverTestsOnConcurrentManager(IProxyDiscoveryManager proxyDiscov
 
                 // Kick off another discovery task for the next source
                 var discoveryCriteria = new DiscoveryCriteria(new[] { nextSource }, this.actualDiscoveryCriteria.FrequencyOfDiscoveredTestsEvent, this.actualDiscoveryCriteria.DiscoveredTestEventTimeout, this.actualDiscoveryCriteria.RunSettings);
```
acesiddhu: >DiscoveryCriteria [](start = 44, length = 17)

check if overriden constructor for criteria

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
abhishekkumawat23: FilterExpressionWrapper is set by DiscovererEnumerator in case of discovery and by RunTestsWithSources in case of test run. Thus cant be private.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
abhishekkumawat23: Letting behavior consistent with TestRunCriteria and thus not adding overriden constructor accepting testCaseFilter arg

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryContext.cs
abhishekkumawat23: done

singhsarab (approved):
Looks good. Please add an acceptance test for this.

smadala (approved):
Add acceptance test [here](https://github.com/Microsoft/vstest/blob/master/test/Microsoft.TestPlatform.AcceptanceTests/TestCaseFilterTests.cs#L124)

abhishekkumawat23:
Acceptance tests are not required as these changes are exposing filtering support at discovery and not actually implementing them. Implementation will happen at adapter side.

nohwnd:
For anyone looking at this in the future, this only applies to MSTest test framework. Current the filtering during Discovery only works when using MSTest and XUnit, but not NUnit.

PR #1082, opened 2017-09-13
Exposing discovery events to logger extensibility

abhishekkumawat23:
Changes done:
1. Enabled logger extensibility at discovery
2. Exposed discoveryMessage, discoveredTests, discoveryComplete events to loggers


src/Microsoft.TestPlatform.Common/Logging/TestLoggerManager.cs
```diff
@@ -439,12 +443,14 @@ protected virtual void Dispose(bool disposing)
                         this.runRequest.TestRunMessage -= this.TestRunMessageHandler;
                         this.runRequest.OnRunStatsChange -= this.TestRunStatsChangedHandler;
                         this.runRequest.OnRunCompletion -= this.TestRunCompleteHandler;
-                        this.runRequest.DataCollectionMessage -= this.DiscoveryMessageHandler;
```
singhsarab: >DiscoveryMessageHandler [](start = 70, length = 23)

Good catch.

src/Microsoft.TestPlatform.ObjectModel/Logger/TestLoggerEvents.cs
```diff
@@ -39,6 +39,21 @@ protected TestLoggerEvents()
         /// </summary>
         public abstract event EventHandler<TestRunCompleteEventArgs> TestRunComplete;
 
+        /// <summary>
+        /// Raised when a discovery message is received.
+        /// </summary>
+        public abstract event EventHandler<TestRunMessageEventArgs> DiscoveryMessage;
```
singhsarab: >TestRunMessageEventArgs [](start = 43, length = 23)

Why aren't we following the convention for this one?

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -92,6 +92,16 @@ public IDiscoveryRequest CreateDiscoveryRequest(DiscoveryCriteria discoveryCrite
...
             this.AddExtensionAssemblies(discoveryCriteria.RunSettings);
 
+            // Update and initialize loggers only when DesignMode is false
+            var runConfiguration = XmlRunSettingsUtilities.GetRunConfigurationNode(discoveryCriteria.RunSettings);
+            if (runConfiguration.DesignMode == false)
+            {
+                this.AddExtensionAssembliesFromSource(discoveryCriteria.Sources);
+
+                // Initialize loggers
+                TestLoggerManager.Instance.InitializeLoggers();
```
singhsarab: Now since this happens for both discovery and execution, we should think if there is a better way to do this.

src/Microsoft.TestPlatform.ObjectModel/Logger/TestLoggerEvents.cs
abhishekkumawat23: OnDiscoveryMessage sent by test platform sent TestRunMessageEventArgs object. Even though the naming is incorrect, TestLoggerEvents is just listening to what platform is exposing.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
abhishekkumawat23: Functioning is not exactly same. In case of execution and discovery, we are calling different overriden  methods.

singhsarab (approved):
:shipit:

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -92,6 +92,16 @@ public IDiscoveryRequest CreateDiscoveryRequest(DiscoveryCriteria discoveryCrite
             // Update cache with Extension Folder's files
             this.AddExtensionAssemblies(discoveryCriteria.RunSettings);
 
+            // Update and initialize loggers only when DesignMode is false
+            var runConfiguration = XmlRunSettingsUtilities.GetRunConfigurationNode(discoveryCriteria.RunSettings);
```
harshjain2: RunConfiguration is heavily used. Should we consider making `RunConfiguration` as property in `RunSettingsManager` instead of creating an object of RunConfiguration every time we need to read run config?

harshjain2: cc : @codito

harshjain2 (approved):
LGTM

PR #1086, opened 2017-09-14
Support for list extensions

codito:
Add following command line switches to `vstest.console.exe`:
* /listDiscoverers
* /listExecutors
* /listLoggers
* /listSettingsProviders

These will list the corresponding test platform extensions. They are added for compatibility with earlier version of `vstest.console.exe`.


src/vstest.console/Processors/ListExtensionsArgumentProcessor.cs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                ConsoleOutput.Instance.WriteLine(extension.Value.GetType().FullName, OutputLevel.Information);
+                ConsoleOutput.Instance.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.AvailableExtensionsMetadataFormat, "SettingName", extension.Metadata.SettingsName), OutputLevel.Information);
+            }
+
+            return ArgumentProcessorResult.Success;
+        }
+
+        private class NullMessageLogger : IMessageLogger
+        {
```
chakradhar2: Do we require NullMessageLogger here?

codito: It is required by this argument processor.

codito: @chakradhar2 thanks, I've removed the extraneous code.

src/vstest.console/Processors/ListExtensionsArgumentProcessor.cs
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
singhsarab: Are these getting deprecated ? If yes, should we let the user know that these options are getting deprecated via message.

singhsarab (approved):
With a suggestion.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginManager.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
```
Faizan2304: nit: Removed unused using

test/Microsoft.TestPlatform.AcceptanceTests/ListExtensionsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    public class ListExtensionsTests : AcceptanceTestBase
+    {
+        [CustomDataTestMethod]
+        [NETFullTargetFramework(inIsolation: false, inProcess: true)]
+        public void ListDiscoverersShouldShowInboxDiscoverers(RunnnerInfo runnerInfo)
+        {
+            AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
```
smadala: Is it okay to create one test for all arguments?

src/vstest.console/Processors/ListExtensionsArgumentProcessor.cs
```diff
@@ -0,0 +1,236 @@
...
+
+    internal class ListDiscoverersArgumentExecutor : IArgumentExecutor
+    {
+        public void Initialize(string argument)
+        {
+        }
+
+        public ArgumentProcessorResult Execute()
+        {
+            ConsoleOutput.Instance.WriteLine(CommandLineResources.AvailableDiscoverersHeaderMessage, OutputLevel.Information);
```
smadala: How come `AvailableDiscoverersHeaderMessage` resources strings already exists? Are we having any more unused resources strings?

smadala (approved):

smadala:
What is the behavior when running using `dotnet vstest`?

Faizan2304 (approved):

Faizan2304:
@codito : I am assuming it will discover extensions from --TeatAdapterPath as well.

smadala:
`vstest.console.exe /TestAdapterPath:"Path\to\core-adapters" /listDiscoverers`, will show the discoverer?

src/vstest.console/Processors/ListExtensionsArgumentProcessor.cs
codito: I think these were copied as-is from older version of test platform.

codito:
@smadala @Faizan2304 no, this scenario is not supported. In tpv2, the runner process doesn't load the external test adapters; they are always known in the test host (which is specific to user provided runtime). This is different than tpv1 which only supported test adapters authored in .NET full framework.

PR #1103, opened 2017-09-17
Fix for Object Model due to change in TestRunCompleteEventsArgs

khilangoel:
Fix where while running tests from IDE , testruncomplete eventsargs default constructor was not found. Added one more constructor.

Permanent fix would be to update object model in VS repo.


src/Microsoft.TestPlatform.ObjectModel/Client/Events/TestRunCompleteEventArgs.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Client
 
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
+    using Newtonsoft.Json;
```
codito: ObjectModel cannot depend on NewtonSoft.Json

khilan2goel:
Changing fix in VS repo. Not required.

PR #1052, opened 2017-09-07
[Port] TL support for filtering

AbhitejJohn:
Porting this change from 15.4. Not porting specific commits to change th vsix for now since similar changes are already in master.

PR #1107, opened 2017-09-19
Provide a fast path for filter consists of only "FQN=" and "|"

genlu:
**Scenario:** 
The matching algorithm used for testcase filter is based on a tree structure and is recursive. It is too inefficient (O(n^2)) when used to run a large amount of tests by name. 

For example, running 10k mstest tests using TestCase takes ~4s whereas using filter takes ~150s. As you can see from the screenshot below, in the 10k tests scenario, more than 90% of time is spent in filtering testcases.

image

This change replaces the FilterExpression with a hash table of fully qualified name of test methods, if we detect that only "FullyQualifiedName=" and "|" is used in the filter. As a result, running 10k tests by filter now takes ~15s.

This PR is based on #1052, which should be merged first. And all my change is in bc84847


genlu:
FYI @AbhitejJohn @peterwald

AbhitejJohn:
@genlu : Had a quick chat on this fix internally. Could we make fast filter generic, so other properties could benefit from this as well? The logic we have here does seem quite generic in that we optimize the case where we only have "|" and one property(could be FQN, could be category etc).
/cc @codito , @smadala

AbhitejJohn:
An orthogonal query: Is the delta of 11s in execution acceptable? How does this compare to discovery + execution of 10k tests using the traditional approach?

codito:
@AbhitejJohn @genlu @peterwald for theory tests can we form a filter expression like `FullyQualifiedName=NS.Class.A | FullyQualifiedName~NS.Class.TheoryB`? In this case, all the child tests for the theory test `NS.Class.TheoryB` should match.

It will be costlier than a regular `=` check, I think filter based execution is anyways a onetime operation. We shouldn't add test framework specific constructs in the platform, it stays forever and impacts other frameworks (learning from TPv1 and mstest days).

peterwald:
@AbhitejJohn @genlu @codito I think the `FullyQualifiedName~NS.Class.TheoryB` could work well. When doing execute by name, we'd need to basically run every test this way (theory or not) in order for it to work on all frameworks. 

Basically, it would still be an `=` comparison but in the case of a filter defined with `~`, we'd just need to extract the method name from FullyQualifiedName (remove extra trailing data) before doing the comparison.

I like the idea of using a different comparison operator since it remains backwards compatible with the current functionality.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -213,6 +223,13 @@ internal static FilterExpression Parse(string filterString)
             var operatorStack = new Stack<Operator>();
             var filterStack = new Stack<FilterExpression>();
 
+            // We try to build a hash table for an filter expression consists of only a single kind of property with 'equal' operation and '|' operator in parallel.
+            // If by the time we finished parsing the entire expression and `isRunByFullyQualifiedName` is still true, then this hash table will be used instead
+            // for "run by fully qualified name".
```
navin22: Need to change this comment, to use `canUseFastFilter` ?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -296,19 +334,66 @@ internal static FilterExpression Parse(string filterString)
...
+        internal bool Evaluate(Func<string, Object> propertyValueProvider, Func<string, string> propertyValueRegexMatchOpt)
         {
             ValidateArg.NotNull(propertyValueProvider, "propertyValueProvider");
 
+            if (this.UseFastFilter)
+            {
+                Debug.Assert(this.condition == null && this.left == null && this.right == null);
+
+                if (fastFilterPropertyName.Equals(FilterExpression.NormalizedFullyQualifiedNameFilterKeyword, StringComparison.OrdinalIgnoreCase))
+                {
```
AbhitejJohn: Don't think we would be needing a normalized FQN anymore.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,40 @@ public class FilterExpressionWrapper
...
+            this.FilterOptions = options;            
+
             try
             {
                 this.filterExpression = FilterExpression.Parse(filterString);
+
+                var regexString = options?.FilterRegEx;
+                if (!string.IsNullOrEmpty(regexString))
+                {
+                    filterRegex = new Regex(regexString);
```
AbhitejJohn: Should we be using `RegexOptions.Compiled`?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,40 @@ public class FilterExpressionWrapper
...
-        
+        private readonly FilterExpression filterExpression;
+
+        private readonly Regex filterRegex;
+
         /// <summary>
         /// Initializes FilterExpressionWrapper with given filterString.  
         /// </summary>
-        public FilterExpressionWrapper(string filterString)
+        public FilterExpressionWrapper(string filterString, FilterOptions options = null)
```
AbhitejJohn: Might want to add another constructor since this is a public class :(

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITranslationLayerRequestSenderAsync.cs
```diff
@@ -50,7 +50,7 @@ internal interface ITranslationLayerRequestSenderAsync : IDisposable
         /// <summary>
         /// Asynchronous equivalent of <see cref="ITranslationLayerRequestSender.StartTestRunWithCustomHost(IEnumerable{string}, string, TestPlatformOptions, ITestRunEventsHandler, ITestHostLauncher)"/>.
         /// </summary>
-        Task StartTestRunWithCustomHostAsync(IEnumerable<string> sources, string runSettings, ITestRunEventsHandler runEventsHandler, ITestHostLauncher customTestHostLauncher);
+        Task StartTestRunWithCustomHostAsync(IEnumerable<string> sources, string runSettings, TestPlatformOptions options, ITestRunEventsHandler runEventsHandler, ITestHostLauncher customTestHostLauncher);
```
AbhitejJohn: Thanks for adding it here too :)

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -142,17 +142,12 @@ public void CancelDiscovery()
         /// <inheritdoc/>
         public void RunTests(IEnumerable<string> sources, string runSettings, ITestRunEventsHandler testRunEventsHandler)
         {
-            this.RunTests(sources, runSettings, new TestPlatformOptions(), testRunEventsHandler);
+            this.RunTests(sources, runSettings, null, testRunEventsHandler);
```
AbhitejJohn: Do we want users to use the below API when `TestPlatformOptions` is null? Would expect users to use this one in that case.

AbhitejJohn:
@Genlu: Looks fairly good to me. Were you able to get the numbers with this flow?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
genlu: We check for null down stream, so I figured there's no harm.

genlu:
@shyamnamboodiripad @drognanar @navin22 @codito @smadala Could you guys please review this change?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -296,19 +294,23 @@ internal static FilterExpression Parse(string filterString)
...
+            }
             return filterStack.Pop();
         }
 
         /// <summary>
         /// Evaluate filterExpression with given propertyValueProvider.
         /// </summary>
-        /// <param name="propertyValueProvider"> The property Value Provider. </param>
+        /// <param name="propertyValueProvider"> The property Value Provider.</param>
+        /// <param name="propertyValueRegexMatchOpt">Applies RegEx pattern on the property value</param>
```
jimmy9988: Should this be removed?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,68 @@ public class FilterExpressionWrapper
...
             ValidateArg.NotNullOrEmpty(filterString, "filterString");
 
             this.FilterString = filterString;
+            this.FilterOptions = options;
+
             try
             {
-                this.filterExpression = FilterExpression.Parse(filterString);
+                // We prefer fast filter when it's available.
+                this.filterExpression = FilterExpression.Parse(filterString, out this.fastFilter);
```
jimmy9988: Consider: Make both FilterExpression and FastFilter out parameters or use value tuple.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,68 @@ public class FilterExpressionWrapper
...
+                // TODO: surface an error message to user.
+                // We don't match any tests in case of a regex parsing error.
+                this.fastFilter = null;
+            }
+        }
+
+        /// <summary>
+        /// Initializes FilterExpressionWrapper with given filterString.  
+        /// </summary>
+        public FilterExpressionWrapper(string filterString)
```
jimmy9988: After adding optional parameter `FilterOptions options = null` for FilterExpressionWrapper, this should be duplicated?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -60,25 +113,28 @@ public string ParseError
...
         {
             ValidateArg.NotNull(propertyValueProvider, "propertyValueProvider");
-            if (null == this.filterExpression)
-            {
-                return false;
-            }
-            return this.filterExpression.Evaluate(propertyValueProvider);
-        }
 
+            return (UseFilterExpression && this.filterExpression.Evaluate(propertyValueProvider)) ||
```
jimmy9988: Why the order of UseFilterExpression and UseFastFilter is opposite here?

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
```diff
@@ -42,5 +42,9 @@
   <PropertyGroup Label="Configuration">
     <RootNamespace>Microsoft.VisualStudio.TestPlatform.Common</RootNamespace>
   </PropertyGroup>
+
```
jimmy9988: Nit: Empty line

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapperAsync.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.Interfaces;
+    using System;
```
jimmy9988: NIT: unnecessary using

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
jimmy9988: Consider: Use named parameters for all these method call with null option.

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                IsFilteredOutWhenMatched = false;
+            }
+            else if (filterOperation == Operation.NotEqual && (filterOperator == Operator.And || filterOperator == Operator.None))
+            {
+                IsFilteredOutWhenMatched = true;
+            }
+            else
+            {
+                throw new ArgumentException();
```
AbhitejJohn: Please add in text defining the support of fast filter in here.

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+        internal bool Evaluate(Func<string, Object> propertyValueProvider)
+        {
+            ValidateArg.NotNull(propertyValueProvider, "propertyValueProvider");
+
+            if (!TryGetSinglePropertyValue(this.FilterPropertyName, propertyValueProvider, out var value))
+            {
+                return false;
+            }
+
+            if (PropertyValueRegex != null)
```
AbhitejJohn: nit: Since `Evaluate` would be called for every test case, this if and the `IsFilteredOutWhenMatched` condition could be avoided if the implementation changes to have an inheritance with a Factory to return back the right FastFilter instance. Might just make it faster. Thoughts?

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+                {
+                    return false;
+                }
+            }
+
+            var matched = this.FilterPropertyValues.Contains(value);
+            return IsFilteredOutWhenMatched ? !matched : matched;
+        }
+
+        private static bool TryGetSinglePropertyValue(string name, Func<string, Object> propertyValueProvider, out string singleValue)
```
AbhitejJohn: nit: Looks like there is only one caller. Can we inline this?

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+            private ImmutableHashSet<string>.Builder filterHashSetBuilder = ImmutableHashSet.CreateBuilder<string>();
+
+            private bool containsValidFilter = true;
+
+            internal bool ContainsValidFilter => containsValidFilter && conditionEncountered;
+
+            internal void AddOperator(Operator @operator)
+            {
+                if (!containsValidFilter || !(@operator == Operator.And || @operator == Operator.Or))
+                {
```
AbhitejJohn: should `containsValidFilter` be set to false if `!(@operator == Operator.And || @operator == Operator.Or)) ` is satisfied?

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -0,0 +1,161 @@
...
+                else
+                {
+                    conditionEncountered = true;
+                    fastFilterOperation = condition.Operation;
+                    filterPropertyName = condition.Name;
+
+                    filterHashSetBuilder.Add(condition.Value);
+
+                    // Don't support `Contains`.
+                    if (fastFilterOperation != Operation.Equal && fastFilterOperation != Operation.NotEqual)
```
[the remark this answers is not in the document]
AbhitejJohn: I guess its because we want to keep the logic tight in case a new operation is added in later.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -50,7 +50,7 @@ internal class FilterExpression
 
         #region Constructors
 
-        internal FilterExpression()
+        private FilterExpression()
```
AbhitejJohn: I think we can get rid of this.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -296,19 +294,23 @@ internal static FilterExpression Parse(string filterString)
                 throw new FormatException(string.Format(CultureInfo.CurrentCulture, CommonResources.TestCaseFilterFormatException, CommonResources.MissingOperator));
             }
 
+            if (fastFilterBuilder.ContainsValidFilter)
```
AbhitejJohn: ToFastFilter already seems to have this check.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,68 @@ public class FilterExpressionWrapper
...
+        private bool UseFilterExpression => this.filterExpression != null;
+
+        private bool UseFastFilter => this.fastFilter != null;
+
         /// <summary>
-        /// Initializes FilterExpressionWrapper with given filterString.  
+        /// Initializes FilterExpressionWrapper with given filterString and options.  
         /// </summary>
-        public FilterExpressionWrapper(string filterString)
+        public FilterExpressionWrapper(string filterString, FilterOptions options = null)
```
AbhitejJohn: We can make this a required parameter now?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,68 @@ public class FilterExpressionWrapper
...
             }
             catch (FormatException ex)
             {
                 this.ParseError = ex.Message;
             }
+            catch (ArgumentException)
+            { 
+                // TODO: surface an error message to user.
+                // We don't match any tests in case of a regex parsing error.
+                this.fastFilter = null;
```
AbhitejJohn: this.ParseError does not surface the message?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -60,25 +113,28 @@ public string ParseError
         public string[] ValidForProperties(IEnumerable<String> supportedProperties, Func<string, TestProperty> propertyProvider)
         {
             string[] invalidProperties = null;
-            if (null != this.filterExpression)
+
+            if (UseFastFilter)
+            {
+                invalidProperties = this.fastFilter.ValidForProperties(supportedProperties);
+            }
+            else if (UseFilterExpression)
```
AbhitejJohn: Do we need a `UseFilterExpression`? Slightly confusing since it seems it is just a mutual exclusion to `UseFastFilter` although not all code paths ensure a null check to `filterexpression`. Maybe we can just make it a null check as before instead?

src/Microsoft.TestPlatform.Common/Interfaces/Engine/ClientProtocol/TestExecutionContext.cs
```diff
@@ -44,7 +45,8 @@ public TestExecutionContext()
             bool areTestCaseLevelEventsRequired,
             bool hasTestRun,
             bool isDebug,
-            string testCaseFilter)
+            string testCaseFilter,
+            FilterOptions filterOptions)
```
AbhitejJohn: I'm hoping this would not break older versions of testhost. I do not think so. @navin22 ?

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -0,0 +1,198 @@
...
+
+namespace Microsoft.TestPlatform.Common.UnitTests.Filtering
+{
+    using System.Collections.Generic;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+    using Microsoft.VisualStudio.TestPlatform.Common.Filtering;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
+
+    [TestClass]
+    public class FastFilterTests
```
AbhitejJohn: Can we have a test for something like this too please: `FullyQualifiedName=a|(FullyQualifiedName=b|FullyQualifiedName=c)`?

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -0,0 +1,198 @@
...
+            filterExpressionWrapper.ValidForProperties(new List<string>() { "FullyQualifiedName" }, null);
+
+            Assert.IsFalse(fastFilter.Evaluate((s) => "Test1"));
+            Assert.IsFalse(fastFilter.Evaluate((s) => "Test2"));
+            Assert.IsFalse(fastFilter.Evaluate((s) => "Test3"));
+            Assert.IsTrue(fastFilter.Evaluate((s) => "Test4"));
+        }
+
+        [TestMethod]
+        public void FastFilterWithMultipleNotEqualsClauseAndRegex()
```
AbhitejJohn: nit: Please include the expected test outcome in the name.

AbhitejJohn (approved):

AbhitejJohn:
@dotnet-bot : test Windows_NT / Debug Build please.
@dotnet-bot : test Windows_NT / Release Build please.

AbhitejJohn:
@genlu : Just wanted to bring this up here again(We had an offline discussion on this earlier): This would end up executing tests with the same FQN across multiple assemblies. This is a limitation with the filtering logic unfortunately. I do not think this is a prevalent scenario however. 
You might want to validate how this works with MTFM and the new project system though. I believe the source we send over there would only be of one framework - so that should *not* ideally cause problems.
/cc @ManishJayaswal

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -16,26 +18,68 @@ public class FilterExpressionWrapper
...
+                this.filterExpression = FilterExpression.Parse(filterString, out this.fastFilter);
+
+                if (UseFastFilter)
+                {
+                    this.filterExpression = null;
+
+                    // Property value regex is only supported for fast filter, 
+                    // so we ignore it if no fast filter is constructed.
+
+                    // TODO: surface an error message to suer.
```
navin22: What's the error message to surface it here ?

src/Microsoft.TestPlatform.Common/Interfaces/Engine/ClientProtocol/TestExecutionContext.cs
navin22: I don't think so. 
@genlu Is these scenarios tested ? And also the Object Model changes in VS IDE TestExplorer scenarios need to be verified.

navin22:
@genlu Changes LGTM. Can we add one E2E Acceptance tests in `Microsoft.TestPlatform.AcceptanceTests`

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
genlu: Sure.

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
genlu: You are right.

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
genlu: well, I don't think the logic and overhead here warrants an abstraction like that yet.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
genlu: This is for compatibility, since `FilterExpressionWrapper` is public.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
genlu: Yup.

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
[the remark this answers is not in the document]
genlu: Yup.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
genlu: Replaced with `FilterExpression(FilterExpression left, FilterExpression right, bool areJoinedByAnd)`

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
genlu: You are right.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
genlu: Fixed.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
genlu: I didn't set `ParseError` when there's regex parsing error, in which case, both `filterExpression` and `fastFilter` would be `null`. Seems `TestCaseFilterExpression.validForMatch` would short-circuit validation and evaluation if `ParseError` is set. Will fix.

genlu:
@navin22 I think existing acceptance tests already cover this change. They helped me caught a bug for matching property with multiple values.

navin22 (approved):
:shipit:

AbhitejJohn:
@dotnet-bot : test Windows_NT / Debug Build please.

PR #1108, opened 2017-09-19
Put quotes around TestHost path so in case of spaces in name it starts correctly

DavidKarlas:
This is issue for Visual Studio for Mac, because application is installed in "/Applications/Visual Studio.app/..." which means it has space in path..


DavidKarlas:
I made existing unit test more strict, I'm having problems building and running unit tests on my machine, but CI should be able to verify...

codito (approved):

Faizan2304 (approved):

PR #1111, opened 2017-09-19
Moving discovery output to console logger

abhishekkumawat23:
Moving logic of writing discovery output from ListTestsArgumentProcessor to console logger.


Faizan2304:
After your change the output will change for run selected scenario.
Consider a scenario where test project has one test method with name TestMethod1
Run the following command `vstest.console.exe TestProject1.dll --Tests:Test --logger:Console;Verbosity=normal`

### Old before your change
```
Starting test discovery, please wait...
Passed   Project1.UnitTest1.TestMethod1

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
```

### Output after your change
```
Starting test discovery, please wait...
    TestMethod1
Passed   Project1.UnitTest1.TestMethod1

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
```


### What is happening here
It will list all the test which will meet the filter

### Why it is happening 
Run selected first call discovery and collect matching TestCases and then call Run on that selected test. Since we are now enabling ConsoleLogger for discovery as well and consoleLogger id listening for discovery events, its printing discovered tests on console.

@codito  @singhsarab

navin22:
It will actually print all the discovery result and then the filtered tests list that ran. It's a special case needs to be handled, where both discovery and execution requests are made.

smadala:
Please print number of tests discovered also.

singhsarab:
@abhishkk Can we close this PR?

PR #1112, opened 2017-09-20
Used DisplayName from TestCase instead of TestResult.

harshjain2:
Used DisplayName from TestCase instead of TestResult.


src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -353,7 +353,7 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
             this.testsTotal++;
 
             string name = null;
-            name = !string.IsNullOrEmpty(e.Result.DisplayName) ? e.Result.DisplayName : e.Result.TestCase.FullyQualifiedName;
+            name = !string.IsNullOrEmpty(e.Result.TestCase.DisplayName) ? e.Result.TestCase.DisplayName : e.Result.TestCase.FullyQualifiedName;
```
Faizan2304: Is there any specific reason for doing this?
What is the difference b/w Result.DisplayName and TestCase.DisplayName?

codito (approved):

PR #1114, opened 2017-09-20
Added Event Collector and Video for Legacy flows as part of nuget package and Vsix package

acesiddhu:
testing done
1) Checked Vsix and Nuget package has these dlls inside Extension/v1
2) validated xcopy scnerios with testsettings


codito (approved):

Faizan2304 (approved):
Rest looks good to me 👍

Issue #1118, opened 2017-09-20
Platform throws a null key for data collectors when used in class hierarchy 


prawalagarwal:
## Description
Data Collector not working as expected

## Steps to reproduce
> What steps can reproduce the defect?
Create a class say A() with collector attributes. Now create another class B which is extended by A and extends Platform.DataCollector() but does not have the attributes.
[uri]
[friendleName]
A() : B()

//no attributes
B(): DataCollector

On parsing the test extensions, the platform throws a null key exception while accessing a dictionary.

## Expected behavior
This class structure should work.

## Actual behavior
On parsing the test extensions, the platform throws a null key exception while accessing a dictionary.

PR #1121, opened 2017-09-21
Fix for failure in loading extensions without Identifier Data

harshjain2:
Loading extensions that don't have identifier data was resulting in failures in DataCollection Framework.
If a DataCollector class doesn't declare URI and FriendlyName, then it throws KeyNotFound Exception and remaining types are not probed for Extensions.  

Proposed Solution : If Identifier Data is not present for any extension, log it in EqtTrace and continue probing remaining types.

Fixes issue : https://github.com/Microsoft/vstest/issues/1118


singhsarab (approved):

Faizan2304 (approved):

harshjain2:
@dotnet_bot test Windows_NT / Debug Build please.

harshjain2:
@dotnet_bot test Windows_NT / Debug Build please.

PR #1130, opened 2017-09-26
Refactored console runner.

codito:
Moved the console runner logic to a library. The actual application
will only invoke the business logic from the library.

The console runner app config provides a private probing path for the
library. This will allow us to keep the runner app in a different location
than the console runner library.


src/Microsoft.TestPlatform.CommandLine/ConsoleRunner.cs
```diff
@@ -0,0 +1,34 @@
...
+    using Microsoft.VisualStudio.TestPlatform.Utilities;
+
+    public static class ConsoleRunner
+    {
+        public static int Start(string[] args)
+        {
+            var debugEnabled = Environment.GetEnvironmentVariable("VSTEST_RUNNER_DEBUG");
+            if (!string.IsNullOrEmpty(debugEnabled) && debugEnabled.Equals("1", StringComparison.Ordinal))
+            {
+                ConsoleOutput.Instance.WriteLine("Waiting for debugger attach...", OutputLevel.Information);
```
chakradhar2: Can we read this from resource, So that it can be displayed in all locales?

codito:
Abandoning this PR, we are taking a different approach.

PR #1131, opened 2017-09-26
Making publisher constructor call async and refactoring code to call dispose only once.

khilangoel:
Added ETW event for dispose method also.

### DIFF
--- a/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
+++ b/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
@@ -61,6 +61,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.TestPlatformHelpers
 
         private Task<IMetricsPublisher> metricsPublisher;
 
+        private bool isDisposed;
+
         #region Constructor
 
         public TestRequestManager()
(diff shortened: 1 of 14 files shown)


src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -66,17 +69,24 @@ public TestRequestManager()
...
                 TestRunResultAggregator.Instance,
-                TestPlatformEventSource.Instance)
+                TestPlatformEventSource.Instance,
+                MetricsPublisherFactory.GetMetricsPublisher(IsTelemetryOptedIn(), CommandLineOptions.Instance.IsDesignMode))
+        {
+        }
+
+        public void Dispose()
         {
+            this.metricsPublisher.Result.Dispose();
```
navin22: Add the regular pattern of handling the Dispose with one more method `Dispose(bool disposing)`

navin22: null check required ?

khilan2goel: Should I add destructor as well in that case?

navin22 (commented):
LGTM

khilan2goel:
test Windows_NT / Release Build

harshjain2 (approved):

PR #1136, opened 2017-09-27
Help should display when invalid argument is provided

jayaranigarg:


codito:
I think the command line should show an usage only message for incorrect argument. We can guide the customer to use `--help`. For example, here's how `git` and `ls` behave:

```shell
C:\Users\armahapa\Source>git --somearg
Unknown option: --somearg
usage: git [--version] [--help] [-C <path>] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]

C:\Users\armahapa\Source>"C:\Program Files\Git\usr\bin\ls.exe" --somearg
ls: unknown option -- somearg
Try '/usr/bin/ls --help' for more information.

C:\Users\armahapa\Source>wget --somearg
C:\ProgramData\chocolatey\lib\Wget\tools\wget.exe: unrecognized option '--somearg'
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
```

pvlakshm:
While this is good to have, we can live with the current behaviour (i.e. doing what vstest.console.exe currently does).

test/Microsoft.TestPlatform.AcceptanceTests/ArgumentProcessorTests.cs
```diff
@@ -28,5 +28,29 @@ public void PassingNoArgumentsToVsTestConsoleShouldPrintHelpMessage(RunnerInfo r
             //Check for help examples text
             this.StdOutputContains("To run tests: >vstest.console.exe tests.dll");
         }
+
+        [CustomDataTestMethod]
+        [NETFullTargetFramework]
+        [NETCORETargetFramework]
```
Faizan2304: Since this code doesn't depends on Target framework, can we use only one attribute `NETFullTargetFramework` ?

Faizan2304 (approved):
Looks good to me 👍 
Just a small comment. Resolve that and merge the PR

PR #1138, opened 2017-09-27
Send discovery/execution complete only when all Test-Host's are closed

mayankbansal018:


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -189,10 +187,10 @@ public void Close()
 
                 // Dispose concurrent executors
                 // Do not do the cleanuptask in the current thread as we will unncessarily add to execution time
-                this.lastParallelRunCleanUpTask = Task.Run(() =>
+                Task.Run(() =>
```
codito: There doesn't seem any benefit of creating a Task and Waiting on it synchronously. Same comment for discovery logic.

codito: How does an exception during cleanup affect the parallel logic?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -122,7 +120,7 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
 
                 // Dispose concurrent executors
                 // Do not do the cleanuptask in the current thread as we will unncessarily add to discovery time
-                this.lastParallelDiscoveryCleanUpTask = Task.Run(() => this.UpdateParallelLevel(0));
```
codito: How does this change affect Discovery/Execution benchmarks? Should we validate with few scale test projects? cc: @navin22

src/Microsoft.TestPlatform.Client/Discovery/DiscoveryRequest.cs
```diff
@@ -355,7 +374,16 @@ public void HandleLogMessage(TestMessageLevel level, string message)
...
-            this.OnRawMessageReceived?.Invoke(this, rawMessage);
+            var message = this.dataSerializer.DeserializeMessage(rawMessage);
+
+            if (!string.Equals(MessageType.DiscoveryComplete, message.MessageType))
+            {
+                this.OnRawMessageReceived?.Invoke(this, rawMessage);
+            }
+            else
+            {
+                this.discoveryCompleteTask = Task.Run(() => SendRawMessageAsync(rawMessage));
```
codito: We can avoid the additional Task if we just store the `rawMessage` content and send it just before the `runCompletionEvent` is set.

src/Microsoft.TestPlatform.Client/Discovery/DiscoveryRequest.cs
```diff
@@ -249,6 +264,10 @@ public void HandleDiscoveryComplete(DiscoveryCompleteEventArgs discoveryComplete
                     if (this.discoveryCompleted != null)
                     {
                         this.discoveryCompleted.Set();
+
+                        //Send the Discovery Complete Raw Message
+                        this.discoveryCompleteTask.Wait();
```
codito: Should the order be `rawMessage` is send first, `discoveryCompleted` is set?

codito (changes requested):

src/Microsoft.TestPlatform.Client/Discovery/DiscoveryRequest.cs
mayankbansal018: I do want to DiscoveryRequest or TestRunRequest to cache messages so they can send them later.  Their responsibility is to send the messages, but not to store them.

src/Microsoft.TestPlatform.Client/Discovery/DiscoveryRequest.cs
mayankbansal018: It is waiting for this event to be set before it can send message, that's why the order

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
mayankbansal018: The only place exception can be thrown is while Closing testhost(we also handle this under try catch)), so this means that testhost did not close(possibly), in such cases we should not be sending discovery/execution complete raw message anyways.

mayankbansal018:
@dotnet-bot Test Windows_NT / Release Build

PR #1150, opened 2017-09-29
Aspdotnet test is not working

Faizan2304:
Inserting new bits of nuget package Microsoft.VisualStudio.QualityTools which has fix for

https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems/edit/496477


smadala (approved):
Add new dlls to signproj, and verify_sign if required.

codito (approved):

PR #1162, opened 2017-10-04
Compatibility for verbosity and test execution thread.

codito:
Modified console logger verbosity to be normal by default (for vstest.console.exe) scenarios.
Today the verbosity is also used as a progress marker, this will provide updates to CI users.

Updated test execution thread to be STA by default (for tests targeting .NET desktop framework).


src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -69,7 +69,7 @@ internal enum Verbosity
         /// <summary>
         /// Level of verbosity
         /// </summary>
-        private Verbosity verbosityLevel = Verbosity.Minimal;
+        private Verbosity verbosityLevel = Verbosity.Normal;
```
codito: TODO: update verbosity only for `vstest.console.exe`, waiting on spec closure for this change. Will do it in next commit in this PR.

singhsarab (approved):

src/Microsoft.TestPlatform.ObjectModel/Constants.cs
```diff
@@ -101,7 +101,13 @@ public static class Constants
         /// The default execution thread apartment state.
         /// </summary>
         [CLSCompliant(false)]
+#if NET451
```
Faizan2304: Do we have any tests for this default behavior?

Faizan2304 (approved):

src/Microsoft.TestPlatform.ObjectModel/Constants.cs
codito: Yes, STA thread acceptance tests cover this

PR #1177, opened 2017-10-09
Fix generate runsettings for fakes scenarios

codito:
A check for empty runsettings is not necessary to enables fakes.

### DIFF
--- a/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
+++ b/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
@@ -247,8 +247,10 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.TestPlatformHelpers
 
             this.CollectMetrics(requestData, runConfiguration);
 
-            if (!commandLineOptions.IsDesignMode && string.IsNullOrWhiteSpace(runsettings))
+            if (!commandLineOptions.IsDesignMode)
             {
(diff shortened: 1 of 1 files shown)


src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -247,8 +247,10 @@ public bool RunTests(TestRunRequestPayload testRunRequestPayload, ITestHostLaunc
 
             this.CollectMetrics(requestData, runConfiguration);
 
-            if (!commandLineOptions.IsDesignMode && string.IsNullOrWhiteSpace(runsettings))
+            if (!commandLineOptions.IsDesignMode)
```
Faizan2304: Do we have any test covering this scenario?

Faizan2304 (approved):

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
codito: No we can't test fakes in PR/CI.

smadala:
There will be ArgumentNullException [here]( https://github.com/Microsoft/vstest/blob/ea7c65941a74dff54fa3f25e82e443efc2d1c8df/src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs#L42), if runsettings is null. Should we add null check here?

PR #1182, opened 2017-10-09
Add timeout for process WaitForExit

smadala:
- Vstest.console waiting indefinitely if testhost launch a process and won't kill it.

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
@@ -49,7 +49,7 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
                     process.Exited += (sender, args) =>
                     {
                         // Call WaitForExit without again to ensure all streams are flushed,
-                        // Add timeout to avoid indefinate waiting on child process exists.
+                        // Add timeout to avoid indefinite waiting on child process exist.
                         var p = sender as Process;
                         try
                         {
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
@@ -48,11 +48,12 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
                 {
                     process.Exited += (sender, args) =>
                     {
-                        // Call WaitForExit again to ensure all streams are flushed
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -48,11 +48,12 @@ public object LaunchProcess(string processPath, string arguments, string working
...
                     process.Exited += (sender, args) =>
                     {
-                        // Call WaitForExit again to ensure all streams are flushed
+                        // Call WaitForExit without again to ensure all streams are flushed,
+                        // Add timeout to avoid indefinate waiting on child process exists.
                         var p = sender as Process;
                         try
                         {
-                            p.WaitForExit();
+                            p.WaitForExit(500);
```
codito: Timeouts are approximate. What is the worst case? Is it only limited to loss of error stream data?

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -48,11 +48,12 @@ public object LaunchProcess(string processPath, string arguments, string working
                 {
                     process.Exited += (sender, args) =>
                     {
-                        // Call WaitForExit again to ensure all streams are flushed
+                        // Call WaitForExit without again to ensure all streams are flushed,
+                        // Add timeout to avoid indefinate waiting on child process exists.
```
codito: nit: s/indefinate/indefinite

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
navin22: Is it good to take this timeout as part of argument while launching the process ?

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
smadala: Done.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
smadala: @codito Verified writing 1.3MB of data continuously to testhost standard error, which got captured in vstest.console within 500 ms.  Yes, Theoretically error stream can be lost. At cost of losing error stream we can avoid hang.
@navin22 As scenarios is very rare, I think adding this variable part of arguments won't required.

codito (approved):

PR #1186, opened 2017-10-11
Updating data collectors package

prawalagarwal:
These changes are to support the missing dll for legacy scenarios in case of media recorder collector

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -356,11 +356,11 @@ function Create-VsixPackage
     $testImpactComComponentsDir = Join-Path $extensionsPackageDir "TestImpact"
 
     # Copy legacy dependencies
-    $legacyDir = Join-Path $env:TP_PACKAGES_DIR "Microsoft.Internal.TestPlatform.Extensions\15.5.0-preview-1046721\contentFiles\any\any"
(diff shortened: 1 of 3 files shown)


Faizan2304 (approved):

scripts/build.ps1
```diff
@@ -364,7 +364,7 @@ function Create-VsixPackage
     Copy-Item -Recurse $legacyDir\* $packageDir -Force
 
     # Copy Legacy data collectors Related depedencies
-    $legacyDir = Join-Path $env:TP_PACKAGES_DIR "Microsoft.VisualStudio.QualityTools.DataCollectors\15.5.0-preview-1046721\contentFiles\any\any"
+    $legacyDir = Join-Path $env:TP_PACKAGES_DIR "Microsoft.VisualStudio.QualityTools.DataCollectors\15.5.0-preview-1050541\contentFiles\any\any"
```
codito: Should we keep versions of both `Microsoft.VisualStudio.QualityTools` and `Microsoft.VisualStudio.QualityTools.DataCollectors` packages in sync?

prawalagarwal: I was not sure about that so did not change. What is the guidance here?

PR #1192, opened 2017-10-12
Skip filtering for VSIX extensions

codito:
* Add API to skip filtering of extensions by name.
* For UseVsixExtensions option, the extensions are specified by name in a <Asset Type=UnitTestExtension Path=xyz />. Test platform must use the path directly (even if it doesn't adhere to the test adapter naming guidelines (compat with VS 2017.3 and previous).
* Removed the `loadWellKnownExtensions` API in Test Plugin component. It was not used anywhere.

### DIFF
--- a/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
+++ b/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
@@ -149,8 +149,9 @@ namespace Microsoft.VisualStudio.TestPlatform.Client.DesignMode
 
                         case MessageType.ExtensionsInitialize:
                             {
+                                // Do not filter the Editor/IDE provided extensions by name
                                 var extensionPaths = this.communicationManager.DeserializePayload<IEnumerable<string>>(message);
-                                testRequestManager.InitializeExtensions(extensionPaths, skipFiltering: false);
+                                testRequestManager.InitializeExtensions(extensionPaths, skipExtensionFilters: true);
                                 break;
                             }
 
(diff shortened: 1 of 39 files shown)


sudiptadmsft (approved):
As there si some refactoring, re-confirm test coverage

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -149,8 +149,9 @@ private void ProcessRequests(ITestRequestManager testRequestManager)
 
                         case MessageType.ExtensionsInitialize:
                             {
+                                // Do not filter the Editor/IDE provided extensions by name
```
codito: This is required because VSIX and other inbox extensions don't conform to the file naming logic (compat with TPv1).

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -163,11 +163,11 @@ public void Dispose()
...
         /// </summary>
         /// <param name="pathToAdditionalExtensions"> The path to additional extensions. </param>
-        /// <param name="loadOnlyWellKnownExtensions"> The load only well known extensions. </param>
-        public void UpdateExtensions(IEnumerable<string> pathToAdditionalExtensions, bool loadOnlyWellKnownExtensions)
+        /// <param name="skipExtensionFilters">Skips filtering by name (if true).</param>
+        public void UpdateExtensions(IEnumerable<string> pathToAdditionalExtensions, bool skipExtensionFilters)
         {
             this.TestEngine.GetExtensionManager()
-                   .UseAdditionalExtensions(pathToAdditionalExtensions, loadOnlyWellKnownExtensions);
+                   .UseAdditionalExtensions(pathToAdditionalExtensions, skipExtensionFilters);
```
Faizan2304: we should sent false for skipExtensionFilter in https://github.com/Microsoft/vstest/pull/1192/files#diff-0dd53ffcac47f4d639b0baccbea77462R209

codito: Fixed

PR #1197, opened 2017-10-13
Fix return code for framework and platform warnings.

codito:
Do not set exit code as non-zero for warnings. Add checks for acceptance tests
for exit code.

### DIFF
--- a/test/Microsoft.TestPlatform.AcceptanceTests/ExecutionTests.cs
+++ b/test/Microsoft.TestPlatform.AcceptanceTests/ExecutionTests.cs
@@ -21,10 +21,12 @@ namespace Microsoft.TestPlatform.AcceptanceTests
         {
             AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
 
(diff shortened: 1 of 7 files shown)


test/Microsoft.TestPlatform.AcceptanceTests/ExecutionTests.cs
```diff
@@ -216,6 +223,7 @@ public void IncompatiableSourcesWarningShouldBeDisplayedInTheConsole(RunnerInfo
             this.InvokeVsTest(arguments);
 
             this.ValidateSummaryStatus(1, 1, 1);
+            this.ExitCodeEquals(1); // there are failing tests
```
navin22: When one dll is picked for execution and other is ignored, exit code will be 1 ?

smadala: Exit code is 1 because there is failing test.

smadala (approved):

navin22 (approved):

Issue #1185, opened 2017-10-10
NullReferenceException


Ed-Hill:
## Description
vstest\src\Microsoft.TestPlatform.Client\TestPlatform.cs

 var testHostManager = this.testHostProviderManager.GetTestHostManagerByRunConfiguration(testRunCriteria.TestRunSettings);
 testHostManager.Initialize(TestSessionMessageLogger.Instance, testRunCriteria.TestRunSettings);

vstest\src\microsoft.testplatform.common\hosting\testruntimeprovidermanager.cs
GetTestHostManagerByRunConfiguration can return null which will cause a NullReferenceException on the attempt to call Initialize.

Ed-Hill:
Looks like this will be addressed in : 
https://github.com/Microsoft/vstest/pull/1198

smadala:
@Ed-Hill Thanks for reporting issue. You are right #1198 will address it. Did you run into trouble because this?

codito:
Closing since #1198 is merged.

PR #1198, opened 2017-10-13
Exit gracefully if no testHostProvider found

Faizan2304:
Issue: Try to run tests from a dll built for framework= `.NETPortable,Version=v4.5,Profile=Profile259`. There is no testHostProvider for target framework `portable and` hence `testHostProviderManager.GetTestHostManagerByRunConfiguration` return null.

Fix: Check for null and exit gracefully.


src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -101,6 +103,15 @@ public IDiscoveryRequest CreateDiscoveryRequest(IRequestData requestData, Discov
             }
 
             var testHostManager = this.testHostProviderManager.GetTestHostManagerByRunConfiguration(discoveryCriteria.RunSettings);
+            if (testHostManager == null)
```
harshjain2: DRY: Abstract out the common code in a method.

harshjain2 (approved):
Approved with suggestions.

Faizan2304:
issue: https://github.com/Microsoft/vstest/issues/1185

PR #1206, opened 2017-10-17
Results directory when running through vstest.console.dll

Faizan2304:
issue: Results directory is coming wrong to datacollector.dll if running test from vstest.console.dll


Faizan2304:
Abandoning this PR as we got better fix here https://github.com/Microsoft/vstest/pull/1208

PR #1218, opened 2017-10-24
Assembly Resolver perf improvement

Shyam-Gupta:
Adding null in resolvedAssemblies dictionary for assemblies which couldn't be resolved.
It is giving ~3.1 seconds perf improvement for 10k xunit tests from vstest.console during execution.


navin22 (approved):

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -175,9 +175,10 @@ private Assembly OnResolve(object sender, AssemblyResolveEventArgs args)
                         }
                     }
                 }
-            }
 
-            return null;
+                this.resolvedAssemblies[args.Name] = null;
```
smadala: Add this at line:112, because there is exit path at 126 and we should operate on resolvedAssemblies inside `lock (this.resolvedAssemblies)` block.

smadala: Add EqtTrace.IsInfoEnabled check for main flows.

Shyam-Gupta: Done

Shyam-Gupta: Its already under lock block. As per discussion, added this statement in one catch block.

PR #1221, opened 2017-10-24
BuildConfiguration is now a static property.

harshjain2:
BuildConfiguration is now a static property of IntegrationTestEnvironment.


codito:
The changes are already in the master branch.

PR #1226, opened 2017-10-25
Fix for changes in XPathNavigator for netcoreapp2.0

harshjain2:
Fix for changes in XPathNavigator for netcoreapp2.0.
Issue: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems/edit/491724

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
@@ -209,15 +209,18 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
             var dataCollectorsNode = doc.CreateElement(Constants.DataCollectorsSettingName);
             dataCollectionRunSettingsNode.AppendChild(dataCollectorsNode);
 
+#if NET451
+            return doc;
+#else
+            // Xmldocument doesn't inherit from XmlDocument for netcoreapp2.0
             var ret = doc as IXPathNavigable;
             if (ret == null)
             {
-#if !NET451
                 return doc.ToXPathNavigable();
-#endif
             }
 
             return ret;
+#endif
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/Microsoft.TestPlatform.Utilities/Microsoft.TestPlatform.Utilities.csproj
```diff
@@ -20,6 +20,11 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Xml" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
```
codito: Not ideal :-/ This will require all the nuget packages to declare an extra dependency.

codito: Please remove it.

src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
```diff
@@ -209,11 +209,15 @@ public static IXPathNavigable CreateDefaultRunSettings()
             var dataCollectorsNode = doc.CreateElement(Constants.DataCollectorsSettingName);
             dataCollectionRunSettingsNode.AppendChild(dataCollectorsNode);
 
-#if NET451
-            return doc;
-#else
-            return doc.ToXPathNavigable();
+            var ret = doc as IXPathNavigable;
```
codito: Since this fix applies only to `!NET451`, could we move the fix accordingly. E.g.
```
#if NET451
    return doc;
#else
   var ret = doc as IXPathNavigable;
   if ...
#endif
```

Please add a comment on why line 212 works for netcoreapp2.0. Can we link to the bug also?

codito (approved):

test/Microsoft.TestPlatform.Utilities.UnitTests/InferRunSettingsHelperTests.cs
```diff
@@ -236,38 +236,37 @@ public void UpdateRunSettingsShouldThrowIfArchitectureSetIsIncompatibleWithCurre
         [DataTestMethod]
         [DataRow("DesignMode")]
         [DataRow("CollectSourceInformation")]
-        public void UpdateRunSettingsShouldNotModifyXmlIfNavigatorIsNotAtRootNode(string settingName)
+        [Ignore]
+        public void UpdateRunSettingsShouldNotModifyXmlIfxmlDocumentIsNotAtRootNode(string settingName)
```
smadala: Why this is ignored?

harshjain2: It is no more required. 
We are using xmldoc and therefore, there is no concept of positioning as it used to be in xpathnavigatoor

smadala: Got it, Please remove this test.

smadala (approved):

PR #1229, opened 2017-10-26
Adding null in resolvedAssemblies dictionary for assemblies which cou…

Shyam-Gupta:
…ldnt be resolved.

It is giving ~3.1 seconds perf improvement for 10k xunit tests from vstest.console during execution.


src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -175,9 +176,15 @@ private Assembly OnResolve(object sender, AssemblyResolveEventArgs args)
...
                 }
-            }
 
-            return null;
+                if (EqtTrace.IsInfoEnabled)
+                {
+                    EqtTrace.Info("AssemblyResolver: {0}: Failed to load assembly.", args.Name);
+                }
+
+                this.resolvedAssemblies[args.Name] = null;
```
codito: Please validate if this cache is cleaned up when extension paths are updated (IDE scenario).

codito:
Abandoning this since the change requires baking time, it will flow into 15.6 release.

PR #1233, opened 2017-10-26
Set anycpu arch for non dotnet test sources

smadala:
- Settings DefaultPlatform for js, appx or other test source causing warning message showing up if /platform specify not not default platform.

### DIFF
--- a/src/vstest.console/CommandLine/InferHelper.cs
+++ b/src/vstest.console/CommandLine/InferHelper.cs
@@ -44,8 +44,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLineUtilities
                         }
                         else
                         {
-                            //TODO what to do for js, appx and others? Using default for now.
-                            arch = Constants.DefaultPlatform;
+                            //TODO what to do for js, appx and others? Using AnyCPU for now.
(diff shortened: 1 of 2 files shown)


src/vstest.console/CommandLine/InferHelper.cs
```diff
@@ -44,8 +44,8 @@ public Architecture AutoDetectArchitecture(List<string> sources, IDictionary<str
                         }
                         else
                         {
-                            //TODO what to do for js, appx and others? Using default for now.
-                            arch = Constants.DefaultPlatform;
+                            //TODO what to do for js, appx and others? Using AnyCPU for now.
```
codito: We should remove `TODO` please.

codito: Should we make the chutzpah tests in our acceptance suite fail with issues like this? Seems like warnings are a P0 scenario blocker.

codito: Validate nodejs, python, chutzpah, uwp with this change.

codito (approved):

harshjain2 (approved):

PR #1243, opened 2017-10-30
Add fail fast on build failures of TestPlatform.sln

smadala:
- Build should fail if there is error in build or publish steps.
- Build wasn't failing with latest dotnet-cli due to this, But it was failing in "test" step.


scripts/build.ps1
```diff
@@ -190,6 +190,10 @@ function Invoke-Build
     & $dotnetExe build $TPB_Solution --configuration $TPB_Configuration -v:minimal -p:Version=$TPB_Version -p:CIBuild=$TPB_CIBuild -p:LocalizedBuild=$TPB_LocalizedBuild
     Write-Log ".. .. Build: Complete."
 
+    if ($lastExitCode -ne 0) {
```
harshjain2: DRY violation. A single function can be used.

harshjain2: Can we move the logic to check to $lastExitCode to Set-ScriptFailed function?

harshjain2 (approved):

scripts/build.ps1
smadala: Done.

scripts/build.ps1
```diff
@@ -604,8 +602,12 @@ function Get-ElapsedTime([System.Diagnostics.Stopwatch] $timer)
     return $timer.Elapsed
 }
 
-function Set-ScriptFailed
+function Set-ScriptFailed-OnError
```
codito: nit: function name doesn't match the Powershell naming convention.

smadala: Renamed.

codito (approved):

PR #1244, opened 2017-10-30
Added null check for parent node.

harshjain2:
Added null check for parent node.


src/Microsoft.TestPlatform.Utilities/XmlUtilities.cs
```diff
@@ -75,11 +75,11 @@ internal static bool IsValidNodeXmlValue(string xmlNodeValue, Func<string, bool>
                 }
 
                 var parentNode = xmlDocument.SelectSingleNode(nodeXPath.Substring(0, nodeXPath.LastIndexOf('/')));
-                parentNode.AppendChild(childElement);
+                parentNode?.AppendChild(childElement);
```
smadala: Add test for this if already not exist.

smadala (approved):

codito (approved):

PR #1241, opened 2017-10-30
Source build support

codito:
Allow vstest repository to build with `dotnet/source-build`.

- [x] Fix existing mono based builds (#1222, #1153)
- [x] Use Ubuntu 14.04 for linux builds
- [x] Remove dependency on mono, build TestHost by default (https://github.com/dotnet/source-build/issues/96)
- [x] Port patches to vstest (https://github.com/dotnet/source-build/issues/178)
- [x] Support [RepoApi](https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/RepoApi.md) for build

Validated these changes with symlink in `source-build` repo and built it successfully.

Localization requires xlftool (which requires mono to be available). Will fix this outside current PR.

PR #1245, opened 2017-10-30
Reverted to stable version of dotnet cli

harshjain2:
Reverted to stable version of dotnet cli as latest version is having issues.


codito:
Already part of PR #1241.

Issue #1240, opened 2017-10-30
TestCaseFilter with & operator fails


AbhitejJohn:
## Description
A TestCaseFilter with a single property and an & operator throws the following and does not execute the test.
```
Value does not fall within the expected range.
```

## Steps to reproduce
1. With the following code
```
    [TestClass]
    public class Tests
    {
        [TestMethod]
        [TestCategory("smoke")]
        [TestCategory("p0")]
        public void Test1()
        {
        }

        [TestMethod]
        [TestCategory("smoke")]
        [TestCategory("p1")]
        public void Test2()
        {
            
        }
    }
```
2. Run 
```
vstest.console.exe <Test.dll>/testcasefilter:"TestCategory=smoke&TestCategory=p1"
```

## Expected behavior
It should run test2.

## Actual behavior
No test is run and the following is printed out:
```
Value does not fall within the expected range.
```

## Environment
> Build version of vstest.console

15.5.0-preview-20171025-02

AbhitejJohn:
Could be related to the Fast Filter changes we made.

PR #1247, opened 2017-10-30
FastFilter Fix

singhsarab:
Adding the conditions for the negative scenario in the AddOperator for FastFilter.

Fixes : https://github.com/Microsoft/vstest/issues/1240


src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
```diff
@@ -150,6 +150,11 @@ internal void AddOperator(Operator @operator)
                     {
                         operatorEncountered = true;
```
AbhitejJohn: Might be unrelated to this change but should we be putting out a meaningful message when we throw ArgumentException at Ln44 above?

AbhitejJohn (approved):

src/Microsoft.TestPlatform.Common/Filtering/FastFilter.cs
genlu: +1. Also might want to make `ToFastFilter` more robust by catching the `ArgumentException`.

genlu (approved):

codito (approved):

codito:
Suggest to make acceptance tests for stringent for filters.

Issue #1214, opened 2017-10-23
Microsoft.TestPlatform nuget package throws PathTooLongException 


harshjain2:
## Description
We have recently introduced Microsoft.TestPlatform nuget package [here](https://github.com/Microsoft/vstest/pull/1211). 
Because of long path structure of this package, users of this nuget package might face PathTooLongException.

It is required to break `Common7\IDE\...` dependency and have a relatively flat structure.

## Workaround
Keep the Microsoft.Testplatform nuget package at the root folder of directory (D:\Microsoft.TestPlatfrom, etc.)

pvlakshm:
I don't think end users can extract to the ```...\Common7\IDE\...``` folder. And no matter how short our package name, if the user were to extract it to a deep folder structure (s)he could still face the PathTooLongException, right?

I think we can live with this.

smadala:
@pvlakshm Following is path if I install the package by default `C:\Users\samadala\.nuget\packages\Microsoft.TestPlatform\15.5.0-preview-20170810-02\tools\net451\Common7\IDE\Extensions\TestPlatform\Extensions\CppUnitFramework\Microsoft.VisualStudio.TestTools.CppUnitTestFramework.Executor.x64.dll.manifest` which is `240` character.
If I change the nuget package folder to %temp% it is 260 character. `C:\Users\samadala\AppData\Local\Temp\.nuget\packages\Microsoft.TestPlatform\15.5.0-preview-20170810-02\tools\net451\Common7\IDE\Extensions\TestPlatform\Extensions\CppUnitFramework\Microsoft.VisualStudio.TestTools.CppUnitTestFramework.Executor.x64.dll.manifest
`
There was live site issue with cloudbuild due to this very long layout last month. We should consider to break this long layout.

smadala:
https://github.com/Microsoft/vstest/pull/1259 unblocks cloudbuild team. 
/cc @acesiddhu  JFYI.

PR #1259, opened 2017-11-02
Get Latest extensions from vsualm

smadala:
- fixes for cpp folder name
- Latest TMI adapter for NAR results fix


mayankbansal018 (approved):
Since we are picking from vsualm, & binaries are not signed there, please do make sure that in our external package, what all binaries we are bringing, are at least added in verify signing steps, & that we do verify we check if native dll's are signed as well

harshjain2:
@dotnet-bot test this please.

codito:
@smadala does this fix https://github.com/Microsoft/vstest/issues/1214?

codito (approved):

smadala:
@codito CPP directory path reduce 13 character, this folder was blocking cloudbuild tpv2 packages to restore on dev machine and try out tpv2 with qtest on local dev machine.

PR #1270, opened 2017-11-06
Fix test run for x64 c++ tests.

codito:
Add x64/dbghelp to the vsix package.

An existing integration already exists, we couldn't catch it since it is always running in xcopy mode.


src/package/VSIXProject/TestPlatform.csproj
```diff
@@ -143,6 +143,9 @@
     <VsixSourceItem Include="$(VsixInputFileLocation)\Extensions\Cpp\*.*" Exclude="$(VsixInputFileLocation)\Extensions\Cpp\*.pdb">
       <VSIXSubPath>Extensions\Cpp</VSIXSubPath>
     </VsixSourceItem>
+    <VsixSourceItem Include="$(VsixInputFileLocation)\Extensions\CppUnitFramework\x64\*.*" Exclude="$(VsixInputFileLocation)\Extensions\CppUnitFramework\x64\*.pdb">
```
smadala: `/s/CppUnitFramework/Cpp`

PR #1274, opened 2017-11-07
Load C++, & Old Mstest adapter for UWP

mayankbansal018:
## Description
In UWP .Net Native Compilation mode managed dll's are packaged differently, & File.Exists() fails.
Include these two dll's if so far no adapters(extensions) were found, & let Assembly.Load() fail if they are not present.

## Related issue
Bug 520306: C++ UWP Unit tests are not discovered in release mode if targeting Tpv2
https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems?id=520306&_a=edit&triage=true


codito (approved):
Please take @sudiptadmsft approval before merge

PR #1275, opened 2017-11-07
Load C++, & Old Mstest adapter for UWP

mayankbansal018:
## Description
In UWP .Net Native Compilation mode managed dll's are packaged differently, & File.Exists() fails.
Include these two dll's if so far no adapters(extensions) were found, & let Assembly.Load() fail if they are not present.

## Related issue
Bug 520306: C++ UWP Unit tests are not discovered in release mode if targeting Tpv2
https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems?id=520306&_a=edit&triage=true

Removing tests for loading c++/c# UWP adapter.
One way to fix the test would be to add platform abstraction for Assembly Load

### DIFF
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
@@ -95,17 +95,10 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework
 
         private void AddKnownExtensions(ref IEnumerable<string> extensionPaths)
         {
-            // For C++ UWP adatper
-            if(this.fileHelper.Exists("Microsoft.VisualStudio.TestTools.CppUnitTestFramework.CppUnitTestExtension.dll"))
-            {
-                extensionPaths = extensionPaths.Concat(new[] { "Microsoft.VisualStudio.TestTools.CppUnitTestFramework.CppUnitTestExtension.dll" });
-            }
-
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
```diff
@@ -95,17 +95,10 @@ internal TestPluginDiscoverer(IFileHelper fileHelper)
...
-
-            // For OLD C# UWP(MSTest V1) adatper
-            if (this.fileHelper.Exists("Microsoft.VisualStudio.TestPlatform.Extensions.MSAppContainerAdapter.dll"))
-            {
-                extensionPaths = extensionPaths.Concat(new[]{ "Microsoft.VisualStudio.TestPlatform.Extensions.MSAppContainerAdapter.dll" });
-            }
+            // For C++ UWP adatper, & OLD C# UWP(MSTest V1) adatper
+            // In UWP .Net Native Compilation mode managed dll's are packaged differently, & File.Exists() fails.
+            // Include these two dll's if so far no adapters(extensions) were found, & let Assembly.Load() fail if they are not present.
+            extensionPaths = extensionPaths.Concat(new[] { "Microsoft.VisualStudio.TestTools.CppUnitTestFramework.CppUnitTestExtension.dll", "Microsoft.VisualStudio.TestPlatform.Extensions.MSAppContainerAdapter.dll" });
```
codito: Is there a way to test these?

PR #1286, opened 2017-11-10
Updated Nuget package description.

harshjain2:
Updated Nuget package description.


PBoraMSFT (commented):
Looks good. I realized that we haven't added support for Coded UI yet, so we need to remove that portion. Rest is good.

smadala (approved):

Issue #373, opened 2017-01-24
Source info not available for desktop projects if it created from netcore template


smadala:
## Description
> Source info not available for tests.

## Steps to reproduce
> Create netcore unit test project from VS templates
> Change `TargetFramework` from netcoreapp1.0 to net46 in csproj file.

## Expected behavior
> Source info should be available.

## Actual behavior
> Source info not available.

## Workaround 
> Add following PropertyGroup to test csproj file.
```
<PropertyGroup>
    <DebugType Condition="'$(TargetFramework)' != '' AND !$(TargetFramework.StartsWith('netcoreapp'))">Full</DebugType>
</PropertyGroup>
```

codito:
This is happening because there's an [assumption](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs#L50) that projects that target desktop can only build with Full PDBs.

smadala:
Add following PropertyGroup to test csproj file.
```
<PropertyGroup>
    <DebugType Condition="'$(TargetFramework)' != '' AND !$(TargetFramework.StartsWith('netcoreapp'))">Full</DebugType>
</PropertyGroup>
```

Faizan2304:
This has been fixed with https://github.com/Microsoft/vstest/pull/528

tmat:
Reopening this issue. The problem has not been fixed fully. 

The SDK still has the following lines, which break debugging tests built with embedded PDBs:

```xml
  <!--
     In IDE scenario for full CLR projects default test platform is v1, which expects full pdbs for source information.
     This can be removed once TPv2 is default for full CLR. Related issue https://github.com/Microsoft/vstest/issues/373.
   -->
  <PropertyGroup>
    <DebugType Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">Full</DebugType>
  </PropertyGroup>
```

Note that this actually completely breaks these tests -- the targets file sets ```DebugType``` after it has been used by the SDK to determine whether symbols need to be copied to the output directory. If the project sets ```DebugType``` to ```embedded``` the SDK determines that no PDBs need to be copied but then the test SDK overwrites ```DebugType``` value which makes the compiler to produce Windows PDB. This PDB is not copied to the output directory. XUnit will not find it there when displaying stack traces so the stack traces will miss line and file information.

johnwc:
[...]

Issue #800, opened 2017-05-11
Code coverage requires DebugType full


codito:
For projects using the roslyn based project system, coverage requires `DebugType = Full` as shown below:

```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp1.1</TargetFramework> 
    <DebugType>Full</DebugType>
  </PropertyGroup>
```

This issue tracks the requirement to support portable symbols in coverage infrastructure.

smadala:
Making DebugType to Full for netcore breaks source navigation and context "Run Tests" option. related bug:https://github.com/Microsoft/vstest/issues/373

cvs79:
When is this expected to be fixed out of the box. We cant advise clients to manually add <DebugType>Full</DebugType> to their projects to get this working. Also adding that setting shows a very strange code coverage list.  See screenshots

<img width="843" alt="code coverage" src="https://user-images.githubusercontent.com/8415874/32727001-4990a328-c87b-11e7-80d4-a32887f9d277.png">

singhsarab:
There is a lot of feedback coming in around this.

tmat:
Is this being worked on? What component needs to be updated to support Portable PDBs?

smadala:
For timeline https://github.com/Microsoft/vstest/pull/1301#discussion_r151942331, Code coverage profiler need to support portable PDBs.

Mohammed-ElSayed:
is there any update regarding this issue?
or is it still the fix that we need to in order to enable code coverage reports on TFS?

max-buecker:
I am also genuinely interested in the status of this issue. For my netcore2.0 projects, I've added the debugtype full setting without any side-effects as far as I am aware. Is this still the way to go? Will there be a solution in the future, where this won't be required anymore?

smadala:
@max-buecker Sorry for delay in fixing this issue.

CodeCoverage profiler has dependency on [msdia.dll](https://docs.microsoft.com/en-us/visualstudio/debugger/debug-interface-access/debug-interface-access-sdk) library, Currently msdia won't support for portable PDB. 
 
 Luckily @VCZYK Implemented to support portable with msdia. I have tested private bits on vstest repo, Code coverage works without any workaround. We are planning to ship it in VS 15.8.

stephenegriffin:
The above workaround is correct for .net. For C++ and v141, full symbols is handled differently:
<GenerateDebugInformation>DebugFull</GenerateDebugInformation> 
[...]

PR #1301, opened 2017-11-17
Remove incorrect DebugType setting

tmat:
## Description
Remove incorrect DebugType setting.

Currently the targets file sets ```DebugType``` after it has been used by the SDK to determine whether symbols need to be copied to the output directory. If the project sets ```DebugType``` to ```embedded``` the SDK determines that no PDBs need to be copied but then the test SDK overwrites ```DebugType``` value which makes the compiler to produce Windows PDB. This PDB is not copied to the output directory. XUnit will not find it there when displaying stack traces so the stack traces will miss line and file information.

## Related issue
Fixes https://github.com/Microsoft/vstest/issues/373

### DIFF
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.targets
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.targets
@@ -33,14 +33,6 @@
   <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
     <OutputType>Exe</OutputType>
   </PropertyGroup>
-
-  <!--
-     In IDE scenario for full CLR projects default test platform is v1, which expects full pdbs for source information.
-     This can be removed once TPv2 is default for full CLR. Related issue https://github.com/Microsoft/vstest/issues/373.
-   -->
-  <PropertyGroup>
-    <DebugType Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">Full</DebugType>
-  </PropertyGroup>
  
   <!--
      Generate config file for test project targeting .NET Framework. This config file has binding redirect which is needed at time of running tests.


livarcocc (approved):

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
```diff
@@ -33,14 +33,6 @@
   <PropertyGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
     <OutputType>Exe</OutputType>
   </PropertyGroup>
-
-  <!--
-     In IDE scenario for full CLR projects default test platform is v1, which expects full pdbs for source information.
-     This can be removed once TPv2 is default for full CLR. Related issue https://github.com/Microsoft/vstest/issues/373.
-   -->
-  <PropertyGroup>
-    <DebugType Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">Full</DebugType>
```
smadala: Removing this before fixing https://github.com/Microsoft/vstest/issues/800 breaks code coverage for .NET Full framework. We can change the condition to add Full debug type if user not specified.

tmat: @smadala This is not the right place to change DebugType value. As I pointed out by the time this is evaluated its value is already used to calculate other variables.

smadala (approved):

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
pvlakshm: Starting next sprint, we will begin the work to add code coverage support for .NET Core in the presence of portable and embedded PDBs.

@tmat, in the meantime we will merge in this PR.

tmat:
@smadala Thanks!

PR #1336, opened 2017-12-18
C++ UWP Reflection based discovery

mayankbansal018:


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -203,8 +204,21 @@ public IEnumerable<string> GetTestPlatformExtensions(IEnumerable<string> sources
         /// <inheritdoc/>
         public IEnumerable<string> GetTestSources(IEnumerable<string> sources)
         {
-            // We do not have scenario where full CLR tests are deployed to remote machine, so no need to udpate sources
-            return sources;
+            List<string> actualSources = new List<string>();
+
+            // We are doing this specifically for UWP, should we extract it out to some other utility?
```
codito: Absolutely. This code takes us back couple of years, albeit the `if` used to be `#if ` in those days.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/System/PlatformFile.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Checks if give file exists on disk
+        /// </summary>
+        /// <param name="filePath">input filePath</param>
+        /// <returns>True if file Exists</returns>
+        public static bool Exists(string filePath)
+        {
+            // Appxrecipe gets renamed vs.appxrecipe, which test platform is unaware of, so if queried for appxrecipe, return true
+            if (filePath.EndsWith(".appxrecipe", System.StringComparison.OrdinalIgnoreCase))
```
codito: This is business logic getting in platform layer :) Would the consumer of platform APIs expect such filtering?

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -221,7 +223,7 @@ private Guid GetTestId()
             // For example in the database adapter case this is not a file path.
             string source = this.Source;
 
-            if (File.Exists(source))
+            if (PlatformFile.Exists(source))
```
codito: IFileHelper.Exists: https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CoreUtilities/Helpers/Interfaces/IFileHelper.cs#L34

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/System/PlatformFile.cs
mayankbansal018: This is the most unfortunate hack I'm introducing since we have to go back to doing UWP discovery in fullCLR process, just like I predicted it would come back to haunt me.
I hated doing this, but business requirement trumps.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
mayankbansal018: I know, but I do not want to create a new Instance of FileHelper to check for file Exists. Correct thing to do would be to move it out of coreutilities, & make it a static class in platform abstraction, but that would require a lot of tests changes.

I'm still thinking on how to do it better.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
mayankbansal018: I'm thinking of moving this to UWP provider, & invoke it via reflection so they own always own the ownership of giving us the correct exe, but that would still mean Desktop Provider would need to understand what all packages types it can support, however it need to understand the logic to extract sources from that package. Thoughts?

mayankbansal018:
I'm not taking the fix in C++ adapters, as it required us to introduce package concept over there as well. Also we had to add the logic of creating TestCase ID in both C++ Desktop, & UWP adapters.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/PlatformPath.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
+{
+    using System.IO;
+
+    /// <summary>
+    /// File Abstraction
+    /// </summary>
+    public static class PlatformPath
```
singhsarab: Have the code there itself. remove this.

singhsarab (approved):

mayankbansal018:
@dotnet-bot test Windows_NT / Debug Build

PR #1351, opened 2017-12-29
Add UWP testhost deps in meta package

smadala:
## Related issue
[unit test app (universal windows) don't compile with Newtonsoft.Json](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/534799)

PR #1352, opened 2017-12-29
Fix for coded UI test debug broken

smadala:
- Porting to master. 
- TMIHelper from d15.6stg.
- Keep the default VSIX version in Int32 range.
#### Related issue
[Can't debug coded UI test in latest VS2017](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/534824)

### DIFF
--- a/samples/Microsoft.TestPlatform.E2ETest/Microsoft.TestPlatform.E2ETest.csproj
+++ b/samples/Microsoft.TestPlatform.E2ETest/Microsoft.TestPlatform.E2ETest.csproj
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+      <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">../../</TestPlatformRoot>
+  </PropertyGroup>
+  <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Dependencies.props" />
+  <PropertyGroup>
     <VersionPrefix>15.0.0</VersionPrefix>
     <TargetFramework>net451</TargetFramework>
     <AssemblyName>Microsoft.TestPlatform.E2ETest</AssemblyName>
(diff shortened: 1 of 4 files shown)


samples/Microsoft.TestPlatform.E2ETest/Microsoft.TestPlatform.E2ETest.csproj
```diff
@@ -22,7 +26,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Internal.TestPlatform.Extensions">
-      <Version>15.6.0-preview-1202328</Version>
+      <Version>$(TestPlatformExternalsVersion)</Version>
```
codito: Thank you!

scripts/build.ps1
```diff
@@ -433,16 +433,17 @@ function Create-VsixPackage
     $testImpactComComponentsDir = Join-Path $extensionsPackageDir "TestImpact"
     $legacyTestImpactComComponentsDir = Join-Path $extensionsPackageDir "V1\TestImpact"
 
+    $testPlatformExternalsVersion = "15.6.0-preview-1251113"
```
codito:
```powershell
[xml](Get-Content $TP_ROOT_DIR\scripts\build\TestPlatform.Dependencies.props).Project.PropertyGroup.TestPlatformExternalsVersion
```

smadala: Thanks, I was searching for this 😄

codito:
I was about to declare https://github.com/Microsoft/vstest/pull/1351 as the last PR to vstest for 2017, and then this one... @smadala on a roll today ;) :+1:

mayankbansal018 (commented):
Do we want to investigate that, rather than acquiring dependencies from nuget, & inserting it back, we create a vsix in devdiv, which directly drop these dependencies into Testplatform\Externals, just like we didf for UWP recently?

scripts/build.ps1
smadala: Done.

smadala:
@mayankbansal018 I had discussed same with @codito. We thought that will cause  different versions of external assemblies(MSTestv1 adapter, etc) ships in Microsoft.Testplatform nuget package and VS.

mayankbansal018:
@smadala, I agree there could be different bits in VS preview, & TP nuget preview bits, but we can make sure that same version ends up in non-preview bits of Testplatform nugets, & in official VS release.

Which should be acceptable. It also avoids us making insertions in VS, just to update some dependencies, & we also need to go through the check-points, which we can avoid. I think it's worth it.

codito:
In my opinion, we shouldn't fragment the distribution with more packages. While we get to do less VS insertions, but the cost of validation (`i` TPv2 vsix with `j` external vsix) and risk of human errors will increase. For instance, right now the acceptance tests give good confidence, with `i` and `j` variable, how should we ensure quality? And how do we ensure that VSTS task == VS insertion for every commit?

I'd rather fix the issues with VS insertion process if there are any, or plan the insertions better.

singhsarab (approved):

Issue #1183, opened 2017-10-09
Tests that call Environment.FailFast do not fail immediately...


KevinH-MS:
## Description
> Tests that call Environment.FailFast do not fail immediately...

## Steps to reproduce
> Create a new Unit Test Project
> Add a simple test method
> Turn on Live Unit Testing
> See the test run and pass
> Add ```Environment.FailFast("test")``` to the test method

## Expected behavior
> Test should fail immediately and report a "not run" or "failed" result.

## Actual behavior
> Test times out after ~30 seconds with: 
```
System.IO.IOException: Unable to communicate with test host process.
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestSender.TryReceiveRawMessage()
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestSender.ListenAndReportTestResults(ITestRunEventsHandler testRunEventsHandler)
```

## Diagnostic logs
> Please share test platform diagnostics logs. Instructions to collect logs are [here](https://github.com/Microsoft/vstest-docs/blob/master/docs/diagnose.md#test-platform-diagnostics).  
> The logs may contain test assembly paths, kindly review and mask those before sharing.

## Environment
> Please share additional details about the test environment.
> Operating system, Build version of vstest.console

smadala:
Fixed in https://github.com/Microsoft/vstest/commit/8d8d744c73cdf26aed0c9f64846c941204845590

PR #1358, opened 2018-01-04
Add exit call back to custom test host process

smadala:
## Description
- `TcpClient.Poll()` not throwing the exception on testhost process exit if testhost process have child process like geckodriver.exe(Which do some socket operations), but normal processes ( notepad.exe, etc..).
- Don't just relay on `TcpClient.Poll()`, Have process exit call back for custom testhost.

## Related issue
- [[Feedback] Cannot Stop/Cancel Debugged Unit Tests when using Selenium WebDriver](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/537068?src=WorkItemMention&src-action=artifact_link&fullScreen=true)
- https://github.com/Microsoft/vstest/issues/1183
- https://developercommunity.visualstudio.com/content/problem/161037/cannot-stopcancel-debugged-unit-tests-when-using-s.html

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
@@ -97,5 +97,7 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces
         /// </summary>
         /// <param name="process">Reference of process to terminate.</param>
         void TerminateProcess(object process);
+
+        void SetExitCallback(int processId, Action<object> exitCallBack);
     }
 }
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
@@ -90,14 +90,12 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces
         /// <param name="callbackAction">
         /// Callback on process exit.
         /// </param>
-        void SetExitCallback(int processId, Action callbackAction);
+        void SetExitCallback(int processId, Action<object> callbackAction);
 
         /// <summary>
         /// Terminates a process.
         /// </summary>
         /// <param name="process">Reference of process to terminate.</param>
         void TerminateProcess(object process);
-
-        void SetExitCallback(int processId, Action<object> exitCallBack);
     }
 }
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
@@ -51,10 +51,10 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
                     process.Exited += (sender, args) =>
                     {
                         // Call WaitForExit without again to ensure all streams are flushed,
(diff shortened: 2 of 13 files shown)


codito:
I think the issue also shows when debugging a simple unittest and then
terminate the process from VS (debug - terminate). Can we confirm? Is it a
dependent on child process or the debugger attached?

Why the additional arg in exit callback?

smadala:
In normal debug stop scenario, TcpClient.Poll() throwing IOException(remote host closed the connection...). Because there was no exitcall back we wait for 10 sec(Which causes #1183) to avoid race condition between communication break and exit callback. 

>dependent on child process or the debugger attached?
 
Indefinite time on child process case, 10s on debugger attached.

>Why the additional arg in exit callback?

Sometimes call back depends on exit code of process. I choose one method rather than adding overload method.

src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
```diff
@@ -90,7 +90,7 @@ public interface IProcessHelper
         /// <param name="callbackAction">
         /// Callback on process exit.
         /// </param>
-        void SetExitCallback(int processId, Action callbackAction);
+        void SetExitCallback(int processId, Action<object> callbackAction);
```
cltshivash: >object [](start = 51, length = 6)

explicitly as Process or have an eventArgs..

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -131,15 +131,18 @@ public bool TryGetExitCode(object process, out int exitCode)
...
         {
-            var process = Process.GetProcessById(processId);
-
-            process.EnableRaisingEvents = true;
-            process.Exited += (sender, args) =>
+            try
+            {
+                var process = Process.GetProcessById(processId);
+                process.EnableRaisingEvents = true;
+                process.Exited += (sender, args) => callbackAction.Invoke(process);
```
cltshivash: >callbackAction.Invoke(process); [](start = 52, length = 31)

if there's a process crash before the hookup.. call the handler..

cltshivash (approved):
:shipit:

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
codito: In line 140, try `sender as Process` within Invoke (reference: https://github.com/smadala/vstest/blob/0848eafca059e3e49790a8a91ea2b9919970de83/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs#L55)

codito:
Looks good. Please cover LUT, UWP scenarios explicitly. These may listen to testhost process exit as well. So depending on which event callback is called, good to validate there are no race/premature cleanup conditions possible.

Also good to cover cases where the debugged testhost produces bunch of stdout/stderr messages.

mayankbansal018 (approved):

src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/IProcessHelper.cs
mayankbansal018: Since Process API's are not available in UWP, using it explicitly here would mean we cannot built uwp platform abstraction.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
smadala: Done.

PR #1362, opened 2018-01-05
VideoRecorder Fix

singhsarab:
Pulling in the latest pkgs with d15.6stg bits, with the video recorder related fix

## Description
This ensures users do not hit BadImageFormatException related to VideoRecorderEngine.


mayankbansal018:
Would this would require insertion in VS as well?

smadala (approved):

PR #1363, opened 2018-01-07
RunSettings compat fix with embedded testsettings

cltshivash:
## Description
Updated the runsettings validation to permit codecoverage and fakes collectors usage along with an embedded testsettings


singhsarab (approved):
Lets validate what happens when static code coverage is enabled in testsettings.

PR #1364, opened 2018-01-08
Runsettings compat fix with testsettings

cltshivash:
## Description
Updated the runsettings validation to permit codecoverage and fakes collectors usage along with an embedded testsetting

## Related issue
Performing code coverage analysis from VS IDE for a project with testsettings fails today.


singhsarab (approved):
:shipit:

PR #1371, opened 2018-01-14
DataCollector Environment Variables propogation during debugging

cltshivash:
## Description
Fixed the execution manager to pass on the environment variables configured by the data collectors to the execution process when under debugger

## Related issue
Debugging fails for a MSTest V1 test using fakes. [https://developercommunity.visualstudio.com/content/problem/176422/new-error-in-155-data-collectors-configured-via-ru.html]


src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
```diff
@@ -134,6 +134,31 @@ public override void Cancel()
...
+
+                foreach(var envVariable in this.dataCollectionEnvironmentVariables)
+                {
+                    if (testProcessStartInfo.EnvironmentVariables.ContainsKey(envVariable.Key))
+                    {
+                        testProcessStartInfo.EnvironmentVariables[envVariable.Key] = envVariable.Value;
+                    }
+                    else
+                    {
+                        testProcessStartInfo.EnvironmentVariables.Add(envVariable.Key, envVariable.Value);
```
codito: We can just keep line 150, it will not throw if key doesn't exist and will replace the value if it exists.

codito (approved):
Looks good to me

acesiddhu (approved):

PR #1387, opened 2018-01-24
Creating log file for UWP via GUID

mayankbansal018:
Issue: https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems?id=555336&_a=edit

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
@@ -64,7 +64,8 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
         /// <inheritdoc/>
         public bool InitializeVerboseTrace(string customLogFile)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
```diff
@@ -64,7 +64,8 @@ public void WriteLine(PlatformTraceLevel level, string message)
         /// <inheritdoc/>
         public bool InitializeVerboseTrace(string customLogFile)
         {
-            LogFile = Path.GetTempPath() + Path.GetFileNameWithoutExtension(customLogFile).Replace(" ", "_") + ".TpTrace.log";
+            // Generate log file name using a GUID, as Path.GetFileNameWithoutExtension is throwing exception
```
singhsarab: Should we use string operation to get the name ?
This can be a fallback if the above fails, thoughts ?

mayankbansal018: I'm waiting for UWP team to come back & tell us why the API GetFileNameWithoutExtension is failing in local machine scenario.
I don't want to write any extra logic till we have to.

Let me know if there are any concerns.

smadala (approved):

PR #1391, opened 2018-01-25
Poll for Connection in case vstest.console is acting as client.

mayankbansal018:
## Description

## Related issue


src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
```diff
@@ -56,8 +62,28 @@ public string Start(string endPoint)
...
+                {
+                    connectionTask = this.tcpClient.ConnectAsync(ipEndPoint.Address, ipEndPoint.Port);
+                    connectionTask.Wait();
+                }
+                catch (Exception ex)
+                {
+                    EqtTrace.Verbose("Connection Failed with error {0}, retrying", ex.Message);
+                }
+            }
+            while ((this.tcpClient != null) && !this.tcpClient.Connected && watch.ElapsedMilliseconds < CONNECTIONRETRYTIMEOUT);
```
singhsarab: Do we need this null check? How can tcpClient become null?

src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
```diff
@@ -56,8 +62,28 @@ public string Start(string endPoint)
...
+            Stopwatch watch = new Stopwatch();
+            watch.Start();
+
+            // It is possible that first ConnectAsync call could not connect to server, so do a repoll to connect to server for 50secs
+            do
+            {
+                try
+                {
+                    connectionTask = this.tcpClient.ConnectAsync(ipEndPoint.Address, ipEndPoint.Port);
+                    connectionTask.Wait();
```
singhsarab: Should we sleep for few ms before retry ?

mayankbansal018: We don't need to, ConnectAsync is an async call, & it does not time out immediately.

src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
```diff
@@ -19,6 +20,11 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities
     /// </summary>
     public class SocketClient : ICommunicationEndPoint
     {
+        /// <summary>
+        /// Time for which the client wait for executor/runner process to start, and host server
+        /// </summary>
+        private const int CONNECTIONRETRYTIMEOUT = 50 * 1000;
```
cltshivash: >CONNECTIONRETRYTIMEOUT [](start = 26, length = 22)

nitpick: fix the casing

cltshivash: >50 * 1000 [](start = 51, length = 9)

50 seconds is too long.. any reason for this high timeout./.?

src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
cltshivash: >(this.tcpClient != null) && !this.tcpClient.Connected [](start = 19, length = 53)

suggestion: can this move into the try block.. after the wait.. and can break based on it..

src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
```diff
@@ -56,8 +62,28 @@ public string Start(string endPoint)
...
+                    connectionTask.Wait();
+                }
+                catch (Exception ex)
+                {
+                    EqtTrace.Verbose("Connection Failed with error {0}, retrying", ex.Message);
+                }
+            }
+            while ((this.tcpClient != null) && !this.tcpClient.Connected && watch.ElapsedMilliseconds < CONNECTIONRETRYTIMEOUT);
+
+            this.OnServerConnected(connectionTask);
```
cltshivash: >OnServerConnected [](start = 17, length = 17)

what was the problem earlier..? (any link to the issue..?)

src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
mayankbansal018: In case exception is thrown in ConnectAsync call then wait won't execute, If the connection does happen it will immediately go to while block after wait statement.
I'm not sure if we need to put the check's up, & break on basis of that.

PR #1395, opened 2018-01-29
Generate proper default settings for EnableCodeCoverage

### LINKED ISSUE #813: Code coverage should exclude auto generated entrypoint
## Description
Code coverage for .net core projects is including `AutoGeneratedProgram` in coverage metrics. It should be excluded.

## Steps to reproduce
1. Use these instructions to run code coverage for .net core in test explorer: https://github.com/Microsoft/vstest-docs/blob/master/docs/analyze.md#working-with-code-coverage

## Expected behavior
Coverage includes user code only.

## Actual behavior
image

## Environment
VS 2017 15.3 Preview 1

mayankbansal018:
Description
Generate proper default settings for EnableCodeCoverage
Add appropriate CodeGenerated attribute for netcore projects

Related issue
Fixes #813.


src/vstest.console/Processors/EnableCodeCoverageArgumentProcessor.cs
```diff
@@ -105,6 +105,79 @@ internal class EnableCodeCoverageArgumentExecutor : IArgumentExecutor
 
         private static string codeCoverageCollectorSettingsTemplate =
```
cltshivash: >codeCoverageCollectorSettingsTemplate [](start = 30, length = 37)

please have this content as an embedded resource instead of using Environment.NewLines all over the place

cltshivash (changes requested):
:clock1:

PR #1396, opened 2018-01-29
Fixed the normal verbosity level to not log the full information for non-failed tests

cltshivash:
## Description
Updated the normal verbosity behavior to not output the detailed information for a test other than failed test. Added a detailed verbosity option to do the same


test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -334,7 +394,7 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
         }
 
         [TestMethod]
-        public void TestResultHandlerShouldShowStdOutMsgOfPassedTestIfVerbosityIsNormal()
+        public void TestResultHandlerShouldShowNotStdOutMsgOfPassedTestIfVerbosityIsNormal()
```
singhsarab: >TestResultHandlerShouldShowNotStdOutMsgOfPassedTestIfVerbosityIsNormal [](start = 20, length = 70)

Nit: Should*Not*Show
please change the name for test following this one as well.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -337,7 +337,6 @@ private void TestMessageHandler(object sender, TestRunMessageEventArgs e)
                     Output.Warning(ConsoleLogger.AppendPrefix, e.Message);
                     break;
                 case TestMessageLevel.Error:
-                    this.testOutcome = TestOutcome.Failed;
```
smadala: Why this is removed? This impact behavior. If the adapter/platform send any error message after this change we not failing test run.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -142,9 +142,9 @@ public void Initialize(TestLoggerEvents events, string testRunDirectory)
             }
 
             // Register for the events.
-            events.TestRunMessage += this.TestMessageHandler;
```
singhsarab: We have been following this convention to use 'this' with members, makes it readable, any specific reason why we are removing it ?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -355,53 +354,82 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
...
+
+                case TestOutcome.Passed:
+                    {
+                        testsPassed++;
+                        if (verbosityLevel == Verbosity.Quiet)
+                        {
+                            break;
+                        }
+
+                        var output = string.Format(CultureInfo.CurrentCulture, CommandLineResources.PassedTestIndicator, testDisplayName);
```
smadala: On Minimal level the passed test name should not get displayed. Update the test to catch this scenario.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -453,16 +481,16 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
             {
                 Output.Error(false, CommandLineResources.TestRunAborted);
             }
-            else if (this.testOutcome == TestOutcome.Failed && this.testsTotal > 0)
+            else if (testsFailed > 0)
```
singhsarab: >testsFailed [](start = 21, length = 11)

We should revert this one.
We could have a passing tests and then we some failure occurred.
TestOutcome is set to failed in case some error occurs during the run, it is not necessary that we would have failing test.

cltshivash: Ok. Will fix this. Didnt see any  UT failing though.. will add one.

src/vstest.console/Internal/ConsoleLogger.cs
cltshivash: Will fix this and add an UT. Currently didn't see any UT fail.. Also why is the failure of a run decided at a logger level ?

singhsarab (approved):

src/vstest.console/Internal/ConsoleLogger.cs
cltshivash: Fixed

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -61,7 +61,8 @@ internal enum Verbosity
         {
             Quiet,
             Minimal,
-            Normal
+            Normal,
+            Detailed
```
smadala: Add detailed to [help](https://github.com/Microsoft/vstest/blob/01ad93c84bd5afcc5a521b80049527d52fa25324/src/vstest.console/Resources/Resources.resx#L241) for customer to aware this option.

PR #1397, opened 2018-01-31
Fix for a Cancellation timing issue 

AbhitejJohn:
Fix for a timing issue where a cancel request before test run request has been started is being ignored.

We have the right plumbing in place but just seem to have missed initializing a var. Put that in place.
All interface changes are internal only.

## Related issue
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/559064 

From testhost diag logs, this is the stack trace:
TpTrace Error: 0 : 17096, 8, 2018/01/30, 15:50:44.560, 1284210642450, testhost.x86.exe, LengthPrefixCommunicationChannel: MessageReceived: Exception occurred while calling handler of type Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestHandler for MessageReceivedEventArgs: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.ExecutionManager.Cancel()
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestHandler.OnMessageReceived(Object sender, MessageReceivedEventArgs messageReceivedArgs)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Delegate.DynamicInvokeImpl(Object[] args)
   at Microsoft.VisualStudio.TestPlatform.Utilities.MulticastDelegateUtilities.SafeInvoke(Delegate delegates, Object sender, EventArgs args, String traceDisplayName)

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
@@ -168,7 +168,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client
         /// <summary>
         /// Cancels the test run.
         /// </summary>
-        public virtual void Cancel()
+        /// <param name="eventHandler"> EventHandler for handling execution events from Engine. </param>
+        public virtual void Cancel(ITestRunEventsHandler eventHandler)
         {
             // Cancel fast, try to stop testhost deployment/launch
             this.cancellationTokenSource.Cancel();
@@ -187,7 +188,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client
         /// <summary>
         /// Aborts the test run.
         /// </summary>
-        public void Abort()
+        /// <param name="eventHandler"> EventHandler for handling execution events from Engine. </param>
+        public void Abort(ITestRunEventsHandler eventHandler)
         {
             this.RequestSender.SendTestRunAbort();
         }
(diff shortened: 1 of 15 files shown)


AbhitejJohn:
This might be targeted to the wrong branch. Do let me know which branch I should target to for 15.6.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
```diff
@@ -168,7 +168,8 @@ public virtual int StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsH
         /// <summary>
         /// Cancels the test run.
         /// </summary>
-        public virtual void Cancel()
+        /// <param name="eventHandler"> EventHandler for handling execution events from Engine. </param>
+        public virtual void Cancel(ITestRunEventsHandler eventHandler)
         {
             // Cancel fast, try to stop testhost deployment/launch
```
singhsarab: Based on our assumption everywhere else, that is Cancel comes before StartTestRun.
If Cancel comes, it will result in a HandleTestRunComplete, where we have a baseTestRunEventsHandler which only gets set if Start was called, else it could also potentially cause null reference.

singhsarab:
@AbhitejJohn Does this have all the changes? Please update and push this one in.

singhsarab (approved):

singhsarab:
Created another for port https://github.com/Microsoft/vstest/pull/1436

PR #1402, opened 2018-01-31
Fixing few flaky tests port.

AbhitejJohn:


singhsarab (approved):
This is based on assumption that sw.ElapsedMilliseconds is not that reliable.
Abhitej, let's push this only if this blocks the build.

mayankbansal018:
@singhsarab has fixed these test, abandoning PR

PR #1406, opened 2018-02-01
Increase timeouts for datacollector to connect to Testhost

### LINKED ISSUE #1410: Communication b/w testhost, & datacollector fails causing tests processes to hang
## Description
On slow machines there is a delay in testhost process launch. In addition CLR profiler is causing additional delay as testhost is launched under profiler.
Due to this delay in testhost process launch, testhost is sometimes unable to communicate with datacollector process, as datacollector process times out after 15 seconds.

## Expected behavior
Test should not hang, & should fail with appropriate message

## Actual behavior
Test process hangs.

## Environment
vsts Hosted agents, with VS2017.5 build

mayankbansal018:
## Description
On Slower Machines(hosted agents) with Profiling enabled 15secs is not enough for testhost to get started,
hence with data collection enabled, the connection b/w testhost, & dc.exe fails.

TestHost process currently waits infinitely to connect to dc.exe, if the above connection does not happen, it causes a hang in th.exe

Fixes #1410

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
@@ -31,7 +31,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.DataCollect
     {
         // On Slower Machines(hosted agents) with Profiling enabled 15secs is not enough for testhost to get started(weird right!!),
         // hence increasing this timeout
-        private const int DATACOLLECTIONCONNTIMEOUT = 60 * 1000;
+        private const int DataCollectionCommTimeOut = 60 * 1000;
 
         private static readonly object SyncObject = new object();
 
@@ -200,7 +200,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.DataCollect
                                         {
                                             if (
                                                 this.dataCollectionTestCaseEventHandler.WaitForRequestHandlerConnection(
-                                                    DATACOLLECTIONCONNTIMEOUT))
+                                                    DataCollectionCommTimeOut))
                                             {
                                                 this.dataCollectionTestCaseEventHandler.ProcessRequests();
                                             }
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
@@ -29,7 +29,9 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.DataCollect
     /// </summary>
     internal class DataCollectionRequestHandler : IDataCollectionRequestHandler, IDisposable
     {
-        private const int DATACOLLECTIONCONNTIMEOUT = 15 * 1000;
+        // On Slower Machines(hosted agents) with Profiling enabled 15secs is not enough for testhost to get started(weird right!!),
+        // hence increasing this timeout
+        private const int DATACOLLECTIONCONNTIMEOUT = 60 * 1000;
 
         private static readonly object SyncObject = new object();
 
--- a/src/testhost.x86/DefaultEngineInvoker.cs
+++ b/src/testhost.x86/DefaultEngineInvoker.cs
@@ -32,6 +32,8 @@ namespace Microsoft.VisualStudio.TestPlatform.TestHost
         /// </summary>
         private const int ClientListenTimeOut = Timeout.Infinite;
 
+        private const int DataConnectionClientListenTimeOut = 60 * 1000;
+
         private const string EndpointArgument = "--endpoint";
 
         private const string RoleArgument = "--role";
@@ -107,7 +109,13 @@ namespace Microsoft.VisualStudio.TestPlatform.TestHost
                 {
                     var dataCollectionTestCaseEventSender = DataCollectionTestCaseEventSender.Create();
                     dataCollectionTestCaseEventSender.InitializeCommunication(dcPort);
-                    dataCollectionTestCaseEventSender.WaitForRequestSenderConnection(ClientListenTimeOut);
+
+                    // It's possible that connection to vstest.console happens, but to datacollector fails, why?
+                    // DataCollector keeps the server alive for testhost only for 15secs(increased to 60 now), 
+                    // if somehow(on slower machines, with Profiler Enabled) testhost can take considerable time to launch,
+                    // in such scenario dc.exe would have killed the server, but testhost will wait infinitely to connect to it,
+                    // hence do not wait to connect to datacollector process infinitely, as it will cause process hang.
+                    dataCollectionTestCaseEventSender.WaitForRequestSenderConnection(DataConnectionClientListenTimeOut);
                 }
 
                 // Checks for Telemetry Opted in or not from Command line Arguments.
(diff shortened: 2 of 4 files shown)


abhishkk (approved):

test/vstest.console.PlatformTests/AssemblyMetadataProviderTests.cs
```diff
@@ -109,7 +111,9 @@ public void GetFrameWorkForDotNetAssembly(string framework)
             }
 
             Console.WriteLine("Framework:{0}, {1}", framework, string.Format(PerfAssertMessageFormat, expectedElapsedTime, stopWatch.ElapsedMilliseconds));
-            Assert.IsTrue(stopWatch.ElapsedMilliseconds < expectedElapsedTime, string.Format(PerfAssertMessageFormat, expectedElapsedTime, stopWatch.ElapsedMilliseconds));
+
```
nigurr: if we remove asserts what are we testing?

nigurr (changes requested):
Please do change tests as per new logic or remove old code. no comment out

src/testhost.x86/DefaultEngineInvoker.cs
```diff
@@ -32,6 +32,8 @@ internal class DefaultEngineInvoker :
         /// </summary>
         private const int ClientListenTimeOut = Timeout.Infinite;
```
cltshivash: where is ClientListenTimeout used ? And can it be removed ?

cltshivash:
>                 if (requestHandler.WaitForRequestSenderConnection(ClientListenTimeOut))

do we need infinite timeout here ..?

---
Refers to: src/testhost.x86/DefaultEngineInvoker.cs:163 in 1251029. [](commit_id = 12510294f90ed4c01e591d2379cbece1f335c7b2, deletion_comment = False)

src/testhost.x86/DefaultEngineInvoker.cs
```diff
@@ -107,7 +109,13 @@ public void Invoke(IDictionary<string, string> argsDictionary)
...
                     var dataCollectionTestCaseEventSender = DataCollectionTestCaseEventSender.Create();
                     dataCollectionTestCaseEventSender.InitializeCommunication(dcPort);
-                    dataCollectionTestCaseEventSender.WaitForRequestSenderConnection(ClientListenTimeOut);
+
+                    // It's possible that connection to vstest.console happens, but to datacollector fails, why?
+                    // DataCollector keeps the server alive for testhost only for 15secs(increased to 60 now), 
+                    // if somehow(on slower machines, with Profiler Enabled) testhost can take considerable time to launch,
+                    // in such scenario dc.exe would have killed the server, but testhost will wait infinitely to connect to it,
+                    // hence do not wait to connect to datacollector process infinitely, as it will cause process hang.
+                    dataCollectionTestCaseEventSender.WaitForRequestSenderConnection(DataConnectionClientListenTimeOut);
```
cltshivash: >WaitForRequestSenderConnection [](start = 54, length = 30)

if we fail to connect we aren't logging anything .. minimal we should do that.. also what will happen when timeout kicks in,,,

test/vstest.console.PlatformTests/AssemblyMetadataProviderTests.cs
```diff
@@ -75,9 +75,11 @@ public void GetArchitectureForNativeDll(string platform)
             var arch = this.assemblyMetadataProvider.GetArchitecture(assemblyPath);
             stopWatch.Stop();
 
-            Assert.AreEqual(Enum.Parse(typeof(Architecture), platform, ignoreCase: true), arch);
             Console.WriteLine("Platform:{0}, {1}", platform, string.Format(PerfAssertMessageFormat, expectedElapsedTime, stopWatch.ElapsedMilliseconds));
-            Assert.IsTrue(stopWatch.ElapsedMilliseconds < expectedElapsedTime, string.Format(PerfAssertMessageFormat, expectedElapsedTime, stopWatch.ElapsedMilliseconds));
+            Assert.AreEqual(Enum.Parse(typeof(Architecture), platform, ignoreCase: true), arch);
+
+            // We should not assert on time elapsed, it will vary depending on machine, & their state, commenting below assert
+            // Assert.IsTrue(stopWatch.ElapsedMilliseconds < expectedElapsedTime, string.Format(PerfAssertMessageFormat, expectedElapsedTime, stopWatch.ElapsedMilliseconds));
```
cltshivash: can we remove it then ?

test/vstest.console.PlatformTests/AssemblyMetadataProviderTests.cs
cltshivash: Mayank please remove the commented out code. 

---
In reply to: [165563074](https://github.com/Microsoft/vstest/pull/1406#discussion_r165563074) [](ancestors = 165563074)

cltshivash (changes requested):
:clock1:

src/testhost.x86/DefaultEngineInvoker.cs
mayankbansal018: ClientListenTimeout is used by testhost to connect to vstest.console. We are waiting infinitely here so as to allow testhost to keep on polling vstest for comm. However vstest.console only waits for 60secs as default(this is configurable) to connect to testhost. If this does not connect to testhost in this time, it will itself kill testhost process. So this timeout will not cause deadlock

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -29,7 +29,9 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.DataCollect
     /// </summary>
     internal class DataCollectionRequestHandler : IDataCollectionRequestHandler, IDisposable
     {
-        private const int DATACOLLECTIONCONNTIMEOUT = 15 * 1000;
+        // On Slower Machines(hosted agents) with Profiling enabled 15secs is not enough for testhost to get started(weird right!!),
+        // hence increasing this timeout
+        private const int DATACOLLECTIONCONNTIMEOUT = 60 * 1000;
```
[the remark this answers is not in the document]
mayankbansal018: resolved

src/testhost.x86/DefaultEngineInvoker.cs
mayankbansal018: ClientListenTimeout is used by testhost to connect to vstest.console. We are waiting infinitely here so as to allow testhost to keep on polling vstest for comm. However vstest.console only waits for 60secs as default(this is configurable) to connect to testhost. If this does not connect to testhost in this time, it will itself kill testhost process. So this timeout will not cause deadlock

src/testhost.x86/DefaultEngineInvoker.cs
mayankbansal018: Logging done. If timeout kicks in vstest will only run test, & data collection will fail

test/vstest.console.PlatformTests/AssemblyMetadataProviderTests.cs
mayankbansal018: I don't want to remove this commented out as:

1. It is in test code so I can live with it.
2. If I had removed it to being with, I would have been able to explain why I did so in first place.
3. It's kind of a reminder that do not add perf test in Unit Tests

src/testhost.x86/DefaultEngineInvoker.cs
cltshivash: Let us not have infinite timeouts anywhere in the code.. keep the timeout then the twice of the vstest.console.exe timeout.. but not infinite.. 

---
In reply to: [165572248](https://github.com/Microsoft/vstest/pull/1406#discussion_r165572248) [](ancestors = 165572248)

mayankbansal018: Like I said vstest.console will kill testhost process if cannot connect to it. So keeping an infinite timeout here is not harmful. Also the timeout at vstest.console is configurable, & not here. What I can do however is make this also configurable using same environment variable as we use for vstest.console, but then again allowing testhost to wait indefinitely to connect to vstest.console is not harfmful.

Let me know if I still need to make this change

mayankbansal018:
@dotnet-bot Windows_NT / Debug Build

mayankbansal018:
@dotnet-bot test Windows_NT / Debug Build

acesiddhu (approved):
:shipit:

cltshivash (approved):
:shipit:

PR #1440, opened 2018-02-22
Fixing a hang in clients on request init issues in vstest.console

AbhitejJohn:
Fixing  a hang in the design mode clients when the start up code fails. We were not actually sending a test run complete in the line above. That was just notifying loggers that test run is complete.

However we are blanket handling this scenario in the DesignModeClient API. So I just went ahead and made sure that the exception flows through.


singhsarab:
This has been addressed with https://github.com/Microsoft/vstest/pull/1444

AbhitejJohn:
Abandoning this change in favor of a more comprehensive fix - #1444

PR #1444, opened 2018-02-22
Fix: Design mode clients hang for errors

abhishkk:
## Description
There are few error scenarios in which we were not sending discovery/run completion event and thus causing Design mode to hang.

In this fix, sending discovery complete for those scenarios.

### DIFF
--- a/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
+++ b/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
@@ -287,6 +287,8 @@ namespace Microsoft.VisualStudio.TestPlatform.Client.DesignMode
             Task.Run(
             delegate
             {
+                var success = false;
+                var exception = default(Exception);
                 try
                 {
                     testRequestManager.ResetOptions();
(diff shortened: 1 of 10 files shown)


src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -287,27 +287,34 @@ private void StartTestRun(TestRunRequestPayload testRunPayload, ITestRequestMana
...
+                    exception = ex;
+                }
 
+                if (!success)
+                {
                     // If there is an exception during test run request creation or some time during the process
                     // In such cases, TestPlatform will never send a TestRunComplete event and IDE need to be sent a run complete message
                     // We need recoverability in translationlayer-designmode scenarios
-                    var testMessagePayload = new TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = ex.ToString() };
+                    var testMessagePayload = new TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = exception?.ToString() ?? "Exception in StartTestRun." };
```
abhishkk: Better error message? Also error  message need to be present in resource file. Should we take that much changes for 15.6?

singhsarab: let's handle the exception case differently.
If we get a success false, we would have already sent the TestMessagePayload.

smadala (approved):

singhsarab (approved):
:shipit:

mayankbansal018 (approved):
:shipit:

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
```diff
@@ -152,13 +152,19 @@ public virtual int StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsH
             catch (Exception exception)
             {
                 EqtTrace.Error("ProxyExecutionManager.StartTestRun: Failed to start test run: {0}", exception);
+
+                // Log error message to design mode and CLI.
+                var testMessagePayload = new TestMessagePayload { MessageLevel = TestMessageLevel.Error, Message = exception.ToString() };
+                this.HandleRawMessage(this.dataSerializer.SerializePayload(MessageType.TestMessage, testMessagePayload));
```
AbhitejJohn: nit: I feel that the commandline does not need to pay this serialization cost since HandleRawMessage in that case would be a no-op. However since this is a one time thing I think it should be ok.

PR #1452, opened 2018-02-27
Add NgenApplication so that default (vsn.exe.config) is not used...

KevinH-MS:
## Description
...when attempting to ngen testhost(.x86).exe.  Some dependent
assemblies not normally loaded in VS will fail to load during ngen
install when vsn.exe.config is used.

## Related issue
https://devdiv.visualstudio.com/DevDiv/_workitems?id=571813&_a=edit


KevinH-MS:
FYI @cltshivash @abhishekkumawat23

abhishkk (approved):

PR #1460, opened 2018-03-05
String comparison test fix

singhsarab:
Trying out the fix with one test.


singhsarab:
Fixing this via https://github.com/Microsoft/vstest/pull/1462

PR #1498, opened 2018-03-23
New Perf test changes

msk2208:
- Included a ExecutionPerfTests class to run the Mstest,Xunit and Nunit tests.
- Categorized Mstest,Xunit and Nunit tests respectively


test/Microsoft.TestPlatform.PerformanceTests/TranslationLayer/DiscoveryPerfTests.cs
```diff
@@ -21,6 +21,7 @@ public DiscoveryPerfTests()
         }
 
         [TestMethod]
+        [TestCategory("MsTest")]
```
singhsarab: TestCategory should be the same across these tests.
Use "TelemetryPerf"

abhishkk (approved):

PR #1508, opened 2018-03-27
Trim Quotes from start, & end of log file

mayankbansal018:
Add try catch as Path.GetFileNameWithoutExtension() can throw exception. If exception is thrown application hangs, so avoiding that.

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
@@ -64,7 +64,17 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
         /// <inheritdoc/>
         public bool InitializeVerboseTrace(string customLogFile)
         {
-            LogFile = Path.GetTempPath() + Path.GetFileNameWithoutExtension(customLogFile).Replace(" ", "_") + ".TpTrace.log";
+            string logFileName = string.Empty;
+            try
+            {
+                logFileName = Path.GetFileNameWithoutExtension(customLogFile.TrimStart('"').TrimEnd('"')).Replace(" ", "_");
+            }
+            catch
+            {
+                logFileName = Guid.NewGuid().ToString();
+            }
+
+            LogFile = Path.GetTempPath() + logFileName + ".TpTrace.log";
             TraceLevel = PlatformTraceLevel.Verbose;
 
             return this.TraceInitialized();
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
@@ -74,7 +74,7 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
                 logFileName = Guid.NewGuid().ToString();
             }
 
-            LogFile = Path.GetTempPath() + logFileName + ".TpTrace.log";
+            LogFile = Path.Combine(Path.GetTempPath(), logFileName, ".TpTrace.log");
             TraceLevel = PlatformTraceLevel.Verbose;
 
             return this.TraceInitialized();


src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
```diff
@@ -64,7 +64,17 @@ public void WriteLine(PlatformTraceLevel level, string message)
...
+            try
+            {
+                logFileName = Path.GetFileNameWithoutExtension(customLogFile.TrimStart('"').TrimEnd('"')).Replace(" ", "_");
+            }
+            catch
+            {
+                logFileName = Guid.NewGuid().ToString();
+            }
+
+            LogFile = Path.GetTempPath() + logFileName + ".TpTrace.log";
```
smadala: Should we use `Path.Conbine` here?

smadala (commented):
Log a bug for this or explain what is user will see if this fix is not available. So that it will show up in search results.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
mayankbansal018: it's all same internally

smadala: `Path.GetTempPath()` value depends on [Env variable](https://msdn.microsoft.com/en-us/library/system.io.path.gettemppath(v=vs.110).aspx). User may set it to C:\Temp or C:\Temp.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
```diff
@@ -64,7 +64,17 @@ public void WriteLine(PlatformTraceLevel level, string message)
...
         {
-            LogFile = Path.GetTempPath() + Path.GetFileNameWithoutExtension(customLogFile).Replace(" ", "_") + ".TpTrace.log";
+            string logFileName = string.Empty;
+            try
+            {
+                logFileName = Path.GetFileNameWithoutExtension(customLogFile.TrimStart('"').TrimEnd('"')).Replace(" ", "_");
+            }
+            catch
+            {
+                logFileName = Guid.NewGuid().ToString();
```
abhishkk: Should we not throw error here as this is possible only if customLogFile path has invalid chars. Should we write logs in new file name?

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
mayankbansal018: the behavior for UWP is different, all BCL have implementation according to CLR, UWP apps cannot write in C:\Temp, they are allocated their own temp folder inside app local data.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
mayankbansal018: We want to disable logging because file name was incorrect, which we constructed?

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
smadala: C:\Temp was just an example I gave. In UWP how the Path.GetTempPath() get determined? User can modify that? If so can he ignore `\` at end?

mayankbansal018: no idea how it is overwritten.

smadala: Then we better use `Path.Combine()`.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
abhishkk: Does this log file path not coming as diag arg value?

mayankbansal018: yes, but we also add bunch of data to it to make it unique.

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
acesiddhu: lets use Path.Combine. There is a reason why Path.Combine is there in BCL and we shouldn't re implement it

src/Microsoft.TestPlatform.PlatformAbstractions/uap10.0/Tracing/PlatformEqtTrace.cs
abhishkk: ok. got it. 
So in case user passed some invalid chars in logFilePath, and in result we are creating a log file with different name, is that ok with us?

mayankbansal018: What if user passed valid file name, & we corrupted it?

abhishkk: Yes, got that case. Just pointing that there is a scenario in which user given file path will not be respected and user will not be aware of the same (as we will change the log file name and not inform the user)

mayankbansal018: We cannot inform user of that change, this is the first step that we are doing while launching App.
Also your logging should not be based on file name, but unique extension.

abhishkk (approved):

PR #1518, opened 2018-03-28
Extend FastFilter to support multiple properties...

KevinH-MS:
## Description
Previously, we would only use the "fast" filter if the filter expression used a single property name in all sub-expressions.  To improve Live Unit Testing execution performance, we need to be able to support an expression like this:
```
    FQN!=Test1&FQN!=Test2&TestCategory!=SkipWhenLiveUnitTesting
```
This change allows for use of multiple property names with the ```FastFilter```.  In cases where a single property is used, the time complexity remains O(n).  For cases using multiple properties, the worst case is O(m*n), but this is still better than the "slow" case, and in general, I expect "m" (number of property names used in the filter) to be small.

## Related issue
https://github.com/dotnet/testimpact/pull/2557

### DIFF
--- a/test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
+++ b/test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
@@ -5,6 +5,7 @@ namespace Microsoft.TestPlatform.Common.UnitTests.Filtering
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using Microsoft.VisualStudio.TestTools.UnitTesting;
     using Microsoft.VisualStudio.TestPlatform.Common.Filtering;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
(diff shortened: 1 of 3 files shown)


KevinH-MS:
FYI @genlu @mayankbansal018 @abhishkk 

/CC @Microsoft/testimpact

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -321,13 +313,74 @@ public void FastFilterShouldThrowExceptionForUnsupportedOperatorOperationCombina
...
             }
             catch (Exception ex)
             {
                 Assert.IsTrue(ex is ArgumentException);
                 Assert.AreEqual("An error occured while creating Fast filter.", ex.Message);
             }
         }
+
+        [TestMethod]
+        public void MultiplePropertyNames()
```
genlu: Could you also add a test for the other pattern? e.g. "FQN!=Test1&TestCategory!=SkipWhenLiveUnitTesting"

genlu (approved):
:shipit:

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
KevinH-MS: Will do?

KevinH-MS:
@cltshivash @mayankbansal018 @abhishkk would one of you be able to review/merge this if you approve?  We'd like to get this in for 15.7, as it allows for significant improvement in the Live Unit Testing case (it's required as part of the work to avoid sending the extraneous Discovery request in dotnet/testimpact#2557)

/CC @ManishJayaswal

mayankbansal018 (approved):
:shipit:

PR #1521, opened 2018-03-29
Dev/shgu/crlf fix

Shyam-Gupta:
**Repro Steps:**

1. Using vstest.console.exe, user runs a test which has Trace.Writeline() statement with some line breaks, and generates a trx file 
2. Open the trx file in Visual Studio
3. Select the test result, right click --> View Test Results Details

**Result**: In the Results window, check the ‘Standard Console Output section’. Notice that the text is present in a single line and not in multiple lines.

**Issue**: Carriage Return character is not getting retained while generating TRX file

**Fix**: In TrxLogger, PopulateTrxFile(), while saving the file use XmlWriter and in its settings, specify NewLineHandling.Entitize option


cltshivash (approved):

Shyam-Gupta:
@dotnet-bot retest Windows_NT / Release Build please

Issue #1540, opened 2018-04-10
Cancel/Abort is not honored when triggered before a custom host is launched.


AbhitejJohn:
## Description
When a Cancel/Abort is sent to the DesignModeClient before the custom host is launched(for debugging let's say) we tend to ignore this and continue on with sending out requests to the client to launch the custom host and continue with the workflow.

## Steps to reproduce
1. Open a solution with some tests
2. Open Test Explorer
3. Run all tests, quickly cancel before tests are run.
4. From Test menu, Debug All, again quickly cancel in Test Explorer

## Expected behavior
The test run should cancel even before the launch custom host is triggered.

## Actual behavior
The workflow ignores the cancel/abort operations and continues on.

## Environment
latest 15.7 dogfood build of VS.

Repro project: [ManualTest_CSharp1.zip](https://github.com/Microsoft/vstest/files/1896311/ManualTest_CSharp1.zip)

Tracking bug in VS: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/595438

PR #1543, opened 2018-04-16
Cancel/abort not honored when sent before custom host launch

abhishkk:
**Issue**:
https://github.com/Microsoft/vstest/issues/1540

**Changes done**:
1. If cancellation/abort happens before custom host launch, TestPlatform will cancel waiting for IDE. In this case, IDE will get an error level message that Could not connect to test host process.
2. Cancel and abort request will not keep on waiting on same lock as of execution request, so as to respect cancel/abort quickly.
3. On cancellation, data collector will not send any attachments (that happens today also) and quickly cancel.
4. Methods are updated in ITestRequestSender API so that we can cancel waiting for test host to get connected to vstest.console.
5. With every request sent by console to test host, now has a pre-check if cancellation requested or not.
6. AfterTestRunEnd request to data collectors will be called only once when test run completes. (Currently in case of cancellation, we were sending AfterTestRunEnd 2 times, once just as soon as cancellation comes and second when test run completes.)
7. ITestHostLauncher API now has one more overriden LaunchTestHost method which takes CancellationToken. Any custom runtime provider needs to call this overriden method to respect cancellation request.
8. TestRequestManager is no longer holding cancel and abort request. Till now they were holding cancel/abort request for five seconds in case execution request is active.

@AbhitejJohn Please verify IDE side changes considering above changes. I have done validations from my end. Let me know if there are any issues.

@mayankbansal018 Regarding # 7, do we need make any changes in UWP runtime provider to respect cancellation?


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -383,7 +383,7 @@ private bool LaunchHost(TestProcessStartInfo testHostStartInfo, CancellationToke
                 }
                 else
                 {
-                    int processId = this.customTestHostLauncher.LaunchTestHost(testHostStartInfo);
+                    int processId = this.customTestHostLauncher.LaunchTestHost(testHostStartInfo, cancellationToken);
```
abhishkk: We are breaking the API here. Is customTestHostLauncher used by others also? Interface ITesHostLauncher is in ObjectModel.
If its used by others, we need to introduce something called **IsCancellable** in ITestHostLauncher which will be used here to switch between overriden calls of customTestHostLauncher.LaunchTestHost.

cltshivash: will it be used in UWP sceanrio ..?

---
In reply to: [181616862](https://github.com/Microsoft/vstest/pull/1543#discussion_r181616862) [](ancestors = 181616862)

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeTestHostLauncher.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Client.DesignMode
 {
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.Interfaces;
+    using System.Threading;
```
cltshivash: >Threading [](start = 17, length = 9)

nitpick: sort using..

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -265,7 +275,7 @@ public void Abort()
         {
             EqtTrace.Verbose("TestRunRequest.Abort: Aborting.");
 
-            lock (this.syncObject)
+            lock (this.abortSyncObject)
```
cltshivash: >abortSyncObject [](start = 23, length = 15)

is the usage of the sync object to synchronize the access to the executionmanager ?

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestSender.cs
```diff
@@ -160,7 +160,7 @@ public Collection<AttachmentSet> SendAfterTestRunStartAndGetResult(ITestMessageE
 
             // Cycle through the messages that the datacollector sends.
             // Currently each of the operations are not separate tasks since they should not each take much time. This is just a notification.
-            while (!isDataCollectionComplete)
+            while (!isDataCollectionComplete && !isCancelled)
```
cltshivash: >isCancelled [](start = 49, length = 11)

won't we fail to receive data (testruncomplete )

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -260,13 +270,19 @@ public void InitializeExecution(IEnumerable<string> pathToAdditionalExtensions)
 
         /// <inheritdoc />
         public void StartTestRun(TestRunCriteriaWithSources runCriteria, ITestRunEventsHandler eventHandler)
```
cltshivash: >StartTestRun [](start = 20, length = 12)

why do we need two signatures..? TP is an internal interface..?

src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/ITestRequestSender.cs
```diff
@@ -63,13 +72,29 @@ public interface ITestRequestSender : IDisposable
...
         /// <param name="eventHandler">EventHandler for test run events</param>
         void StartTestRun(TestRunCriteriaWithTests runCriteria, ITestRunEventsHandler eventHandler);
 
+        /// <summary>
+        /// Starts the TestRun with given test cases and criteria
+        /// </summary>
+        /// <param name="runCriteria">RunCriteria for test run</param>
+        /// <param name="eventHandler">EventHandler for test run events</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        void StartTestRun(TestRunCriteriaWithTests runCriteria, ITestRunEventsHandler eventHandler, CancellationToken cancellationToken);
```
cltshivash: >StartTestRun [](start = 13, length = 12)

same question.. is this a public interface ?

cltshivash (changes requested):
:clock1:

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
abhishkk: All execution related requests, like ExecuteAsync, HandleTestRunComplete, HandleTestRunStatsChange, HandleLogMessage are under same sync object.

But we want to respect cancel, abort and don't want to wait if some of the above request are going on. Thus separate lock objects for them

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
abhishkk: I confirmed from Mayank. It will not be breaking change.

mayankbansal018: Add this is dotnet runtime

mayankbansal018 (approved):
:shipit:

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -250,6 +253,9 @@ public int LaunchCustomHost(TestProcessStartInfo testProcessStartInfo)
                     waitHandle.Set();
                 };
 
+                // Registering cancellationToken to set waitHandle (whenever request is cancelled).
+                var cancellationTokenRegistration = cancellationToken.Register(() => waitHandle.Set());
```
AbhitejJohn: nit: would something like [this](https://stackoverflow.com/questions/12129749/cancellation-with-waithandle) be more direct? Msdn usage [here](https://docs.microsoft.com/en-us/dotnet/standard/threading/how-to-listen-for-cancellation-requests-that-have-wait-handles)

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
AbhitejJohn: Potentially reuse the same object for cancel and abort then?

src/Microsoft.TestPlatform.ObjectModel/Client/Interfaces/ITestHostLauncher.cs
```diff
@@ -19,5 +21,13 @@ public interface ITestHostLauncher
...
         /// <returns>Process id of the launched test host</returns>
         int LaunchTestHost(TestProcessStartInfo defaultTestHostStartInfo);
+
+        /// <summary>
+        /// Launches custom test host using the default test process start info
+        /// </summary>
+        /// <param name="defaultTestHostStartInfo">Default TestHost Process Info</param>
+        /// <param name="cancellationToken">The cancellation Token.</param>
+        /// <returns>Process id of the launched test host</returns>
+        int LaunchTestHost(TestProcessStartInfo defaultTestHostStartInfo, CancellationToken cancellationToken);
```
AbhitejJohn: Will require a corresponding change in TPV1 OM. TE implements this for debugging which would need a change as well.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
abhishkk: done.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
abhishkk: done.

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -554,15 +559,18 @@ private string GetAbortErrorMessage(Exception exception, bool getClientError)
 
                 // Set a default message and wait for test host to exit for a moment
                 reason = CommonResources.UnableToCommunicateToTestHost;
-                if (this.clientExited.Wait(this.clientExitedWaitTime))
+                if (this.clientExited.Wait(this.clientExitedWaitTime) && !cancellationToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
abhishkk: No. If cancellation flag is checked first and it returns false, and then while waiting cancellation occurs, then waitHandle will be set because of cancellation and thus enter in if condition (which is wrong as wait is set by cancel)

src/Microsoft.TestPlatform.ObjectModel/Client/Interfaces/ITestHostLauncher.cs
abhishkk: This interface is used by run time provider only for launching the custom host via custom host launcher.
Why changes are required in TPV1 OM?

Even though if TE uses it as an interface for their internal purpose, there will no breaking change here (as TPv1 OM is used both while build and run time. TPV2 OM is not loaded.) Please correct me if i am missing something here.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
abhishkk: done

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
abhishkk: done

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
abhishkk: done

mayankbansal018:
Do validate that UWP does work with these changes

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs
```diff
@@ -184,6 +189,24 @@ public virtual bool SetupChannel(IEnumerable<string> sources, CancellationToken
             return true;
         }
 
+        private bool LaunchTestHostAsync(TestProcessStartInfo testHostStartInfo, CancellationToken token)
+        {
+            this.CancellationTokenSource.Token.ThrowIfCancellationRequested();
```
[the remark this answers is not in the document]
AbhitejJohn: yup, got that but most clients would not expect a TestRunComplete when negotiating protocol unless told otherwise.... Was a separate message for cancellation considered for scenarios like these? Might be a bigger change but worth a fluid protocol.

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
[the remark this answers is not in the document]
AbhitejJohn: I was thinking the other way round where we actually end up waiting for a while, figure out its cancelled and head to the else part which seems like a wasted wait...

src/Microsoft.TestPlatform.ObjectModel/Client/Interfaces/ITestHostLauncher.cs
abhishkk: @cltshivash @singhsarab @AbhitejJohn 
Am I missing something here? AFIAK ITestHostLauncher is meant only for runtime provider and it is loaded in vstest console only. Neither translation layer, nor test host or any other process from TP side uses this except vstest.console.

If TE is using this interface for their internal purpose, then it is using TPV1 OM and loading TPV1 OM. So how the behavior is inconsistent? I don't feel comfortable changing TPV1 OM unless it's a breaking change.

cltshivash (approved):

abhishkk:
@dotnet-bot test this please.

abhishkk:
@dotnet-bot test Windows_NT / Release Build

PR #1547, opened 2018-04-18
Fix no stacktrace information when net472 is installed

smadala:
## Description
- To get stacktrace information for portable and embedded pdbs when net472 installed on machine, Add `AppContextSwitchOverrides` element in appconfigs for all the applications(vstest.console, testhost and datacollector). More details: [here](https://github.com/dotnet/designs/blob/master/accepted/diagnostics/debugging-with-symbols-and-sources.md#stack-traces)
- Add `system.diagnostics` element to vstest.console appconfig. To make easy to user to enable logs.
## Related issue
 [DevDiv Issue](https://devdiv.visualstudio.com/DevDiv/VS.in%20Agile%20Testing%20IDE/_workitems/edit/586862).

**Testing**
Verified by installing net472.
net472_sourceinfo

### DIFF
--- a/src/vstest.console/app.config
+++ b/src/vstest.console/app.config
@@ -26,4 +26,38 @@
     <!-- This flag is added to support test execution for net35 tests through TMI adapter. -->
     <add key="TestProjectRetargetTo35Allowed" value="true" />
   </appSettings>
+
+  <!--
+        To collect logs for vstest.console and it's child processes, follow below steps.
+          1. Uncomment below system.diagnostics element content.
+          2. Replace "c:\tmp\log.txt" with valid filename. Make sure "c:\tmp" folder exists. log.txt file will be created by vstest.console if does not exists.
+          3. Restart vstest.console or it's client(VS, any other IDE).
+        More details: https://github.com/Microsoft/vstest-docs/blob/master/docs/diagnose.md#collect-trace-using-config-file
+  -->
+
+  <!--
+  <system.diagnostics>
+    <sources>
+      <source name="TpTrace"
+        switchName="sourceSwitch"
+        switchType="System.Diagnostics.SourceSwitch">
+        <listeners>
+          <add name="logfile"/>
+          <remove name="Default"/>
+        </listeners>
+      </source>
+    </sources>
+    <switches>
+      <add name="sourceSwitch" value="All"/>
+    </switches>
+    <sharedListeners>
+      <add name="logfile"
+        type="System.Diagnostics.TextWriterTraceListener"
+        initializeData="c:\tmp\log.txt">
+        <filter type="System.Diagnostics.EventTypeFilter"
+          initializeData="Verbose"/>
+      </add>
+    </sharedListeners>
+  </system.diagnostics> -->
+
 </configuration>
--- a/src/vstest.console/app.config
+++ b/src/vstest.console/app.config
@@ -5,6 +5,11 @@
   </startup>
   <runtime>
     <legacyUnhandledExceptionPolicy enabled="1" />
+
+    <!-- To get stacktrace information for portable and embedded pdbs when net472 installed on machine.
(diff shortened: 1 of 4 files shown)


src/vstest.console/app.config
```diff
@@ -21,4 +26,38 @@
...
+    </switches>
+    <sharedListeners>
+      <add name="logfile"
+        type="System.Diagnostics.TextWriterTraceListener"
+        initializeData="c:\tmp\log.txt">
+        <filter type="System.Diagnostics.EventTypeFilter"
+          initializeData="Verbose"/>
+      </add>
+    </sharedListeners>
+  </system.diagnostics> -->
```
cltshivash: why isn't diag option sufficient ..? Don't want a config file based approach here..

cltshivash (changes requested):
:clock1:

src/vstest.console/app.config
smadala: This is for IDE(VS , VS Mac, VS code test explorer extensions, Raider, etc.. ) scenarios.

mayankbansal018: Logging by any client should be enabled by starting vstest.console in diag mode. I'm not sure what we are saving here apart from copy paste from internet

smadala: @mayankbansal018 
> Logging by any client should be enabled by starting vstest.console in diag mode.

I think currently no client do this. Which need more infra like giving option to use to enable diag. 

> I'm not sure what we are saving here apart from copy paste from internet.

Yes, We are saving copy paste from internet. Copy paste from internet is not great experience if user have to do more than once. Below is the detailed steps.

1. Find out the link which contains 20+ lines of xml to enable logs(Search engine, github issue, developer community, email from MSFT, browser history, etc..).
2. Goto the link and copy the 20+ lines of xml.
3. Keep it in under right node in vstest.console config file.

- If you upgrade your client(E.g: VS), Boom follow the above steps again.
- If you have multiple SKUs/Release channels, Copy it from previously enabled config file(lucky there is no upgrade for it) or follow the above steps again.
- If you are on different machine follow the above steps or copy it from previous machine.

**I follow these steps or ask the customer to follow them could be more than 100 time.**

mayankbansal018: None of the above is solved by this, expect copy paste, on upgrade, on different machine, boom you still have to go and manually edit the file, restart VS.

To make it any easy of customer, we could actually implement some default logging in platform, & user can maybe modify it via setting these diagnostics, in case more detailed logging is needed

smadala: I clearly mentioned it solves only copy past 3 steps out of 5/6 steps.

mayankbansal018: At best what we could be solving it copy-paste alone, so only step 2. 
Step 1, & 3 still have to be performed, user still doesn't know how to enable logging, user still needs to edit the config file

smadala: Please read "Copy paste from internet is not great experience if user have to do more than once."

cltshivash (approved):

singhsarab (approved):

src/vstest.console/app.config
```diff
@@ -21,4 +26,38 @@
...
+          1. Uncomment below system.diagnostics element content.
+          2. Replace "c:\tmp\log.txt" with valid filename. Make sure "c:\tmp" folder exists. log.txt file will be created by vstest.console if does not exists.
+          3. Restart vstest.console or it's client(VS, any other IDE).
+        More details: https://github.com/Microsoft/vstest-docs/blob/master/docs/diagnose.md#collect-trace-using-config-file
+  -->
+
+  <!--
+  <system.diagnostics>
+    <sources>
+      <source name="TpTrace"
```
singhsarab: nit: We can make this these tags inline, make it look less complicated.

smadala: Done.

Issue #1539, opened 2018-04-10
`FIlterOptions` is not serialized correctly when running .NET Core tests


genlu:
We have noticed that for .NET Core tests, when we run them using filter and a valid `FilterOptions` (through `IVsTestConsoleWrapper`, see the first screenshot below), the `FilterOptions` received on the testhost side will be `null` (see 2nd screenshot below). This doesn't happen for desktop framework tests.

Here's the internal bug tracking this
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/597060

1. When TestExplorer sending run request through console wrapper:

![image](https://user-images.githubusercontent.com/788783/38578091-bd942e7e-3cb7-11e8-987e-12b4a6b67de6.png)

2. When message received and deserialized on the other side:

![image](https://user-images.githubusercontent.com/788783/38578228-29c78d3e-3cb8-11e8-8b33-1162ae4ad1d2.png)

genlu:
FYI @ManishJayaswal

mayankbansal018:
Fixed in VS 15.7 RTM

PR #1551, opened 2018-04-19
FilterOptions is not serialized correctly when running .NET Core tests

abhishkk:
**Issue**:
#1539 

**Fix**:
ParallelProxyExecutionManager send execution request to each execution manager passing a copy of testRunCriteria with different sources.

While copying from TestRunCriteria, FilterOptions was not passed to each execution manager's testRunCriteria.

Fix contains changes to pass FilterOptions to each execution manager in case of Parallal run.


smadala (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -287,32 +287,15 @@ private void StartTestRunOnConcurrentManager(IProxyExecutionManager proxyExecuti
...
-                                          new[] { nextSource },
-                                          this.actualTestRunCriteria.FrequencyOfRunStatsChangeEvent,
-                                          this.actualTestRunCriteria.KeepAlive,
-                                          this.actualTestRunCriteria.TestRunSettings,
-                                          this.actualTestRunCriteria.RunStatsChangeEventTimeout,
-                                          this.actualTestRunCriteria.TestHostLauncher)
-                                          {
-                                              TestCaseFilter = this.actualTestRunCriteria.TestCaseFilter
-                                          };
+                    testRunCriteria = new TestRunCriteria(new[] { nextSource }, this.actualTestRunCriteria);
```
mayankbansal018: >actualTestRunCriteria [](start = 85, length = 21)

how was parallel ever working?

abhishkk: When FilterOptions are used, it never works. otherwise with test case filter it does.

genlu (approved):

singhsarab (approved):
:shipit:

PR #1563, opened 2018-04-25
Adding node in runsettings

kaadhina:
## Description
Adding a node in runsettings to point to the location, that can be used by adapters down the line.


singhsarab (approved):

smadala (approved):
Add documentation [here](https://github.com/Microsoft/vstest-docs/blob/master/docs/configure.md) and link user story.

PR #1576, opened 2018-05-04
Pass TargetPlatform value to DataCollectors

smadala:
## Description
- Add TargetPlatform element to DataCollector configuration node, This can be used by DataCollector. E.g Code Coverage datacollector to determine which platform profiler(covrun32.dll/covrun64.dll) to use.

## Related issue
https://github.com/Microsoft/vstest/issues/981


mayankbansal018:
We can use "COR_PROFILER_PATH_64", & "COR_PROFILER_PATH_32" to set profiler paths, we won't need this change after it

PR #1594, opened 2018-05-15
Enable CodeCoverage x64 .NET Full framework test

smadala:
## Description
- CodeCoverage x64 .NET Full framework test failing on CI server only.  Enabling it to find root cause.


smadala:
tracking [here](https://mseng.visualstudio.com/VSOnline/_workitems/edit/1278924)

PR #1595, opened 2018-05-15
Fix running .NETCore unit tests

smadala:
## Description
Fix TestPlatform.CI.Real failure https://devdiv.visualstudio.com/DevDiv/_build?buildId=1688685&_a=summary


singhsarab (changes requested):
Let's fix the root cause as well. we should have a check so that it fails the PR itself.

mayankbansal018 (approved):
Please do file an Eng bug to publish trx file from jenkins PR

smadala:
@mayankbansal018  created https://github.com/Microsoft/vstest/issues/1597

singhsarab (approved):

PR #1598, opened 2018-05-16
Fix Exception thrown while creating framework based on default enums

mayankbansal018:
## Related issue
Fixes https://developercommunity.visualstudio.com/content/problem/224359/systemargumentexception-frameworkname-cannot-have.html.


smadala (approved):
Good change 👍

PR #1625, opened 2018-05-31
Fix attachment uri in trx if same attachment filename is same

smadala:
- Update the attachment filename with targetFilename while moving all the attachments to one directory for trx report.

Fixes: https://github.com/Microsoft/vstest/issues/1626

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs
@@ -149,15 +149,16 @@ namespace Microsoft.TestPlatform.Extensions.TrxLogger.Utility
         public static List<CollectorDataEntry> ToCollectionEntries(IEnumerable<ObjectModel.AttachmentSet> attachmentSets, TestRun testRun, string trxFileDirectory)
         {
 
-            if (EqtTrace.IsInfoEnabled)
-            {
-                EqtTrace.Info($"Converter.ToCollectionEntries: Converting attachmentSets {string.Join(",", attachmentSets ?? Enumerable.Empty<AttachmentSet>())} to collection entries.");
-            }
-
             List<CollectorDataEntry> collectorEntries = new List<CollectorDataEntry>();
             if (attachmentSets == null)
             {
+                EqtTrace.Info($"Converter.ToCollectionEntries: Received {nameof(attachmentSets)} as null returning empty collection entries.");
                 return collectorEntries;
+            }
+
+            if (EqtTrace.IsInfoEnabled)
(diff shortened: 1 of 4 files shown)


src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs
```diff
@@ -148,6 +148,11 @@ public static TrxObjectModel.TestOutcome ToOutcome(ObjectModel.TestOutcome rockS
 
         public static List<CollectorDataEntry> ToCollectionEntries(IEnumerable<ObjectModel.AttachmentSet> attachmentSets, TestRun testRun, string trxFileDirectory)
         {
+            if (EqtTrace.IsInfoEnabled)
+            {
+                EqtTrace.Info($"Converter.ToCollectionEntries: Converting attachmentSets {string.Join(",", attachmentSets)} to collection entries.");
```
nigurr: attachmentSets null check is below and here we are printing before null check?

src/Microsoft.TestPlatform.ObjectModel/AttachmentSet.cs
```diff
@@ -40,6 +40,11 @@ public AttachmentSet(Uri uri, string displayName)
             DisplayName = displayName;
             Attachments = new List<UriDataAttachment>();
         }
+
+        public override string ToString()
+        {
+            return $"{nameof(Uri)}: {Uri.AbsoluteUri}, {nameof(DisplayName)}: {DisplayName}, {nameof(Attachments)}: [{ string.Join(",", Attachments)}]";
```
nigurr: can Uri be null?

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs
smadala: Thanks. added null check.

src/Microsoft.TestPlatform.ObjectModel/AttachmentSet.cs
smadala: Not sure about that. However we are using AbsoluteUri in [other places](https://github.com/smadala/vstest/blob/2cfc8a2abc01adb740bd9f89c7cbcffd69e4eaac/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/Converter.cs#L164) without any check. Not adding null check here.

nigurr (approved):

smadala:
@dotnet-bot test this please.

smadala:
@dotnet-bot test this please

PR #1627, opened 2018-05-31
Add support to escape/unescape testcase filter strings

genlu:
This change adds support of special character escape/unescape to testcase filter, so they can be used as part of filter condition value. Those special characters and their escape sequence are:

- `%` -> `%0`  This is used as the prefix for escape sequence.
- `(`   -> `%1`
- `)`  -> `%2`
- `&`  -> `%3`
- `|` -> `%4`
- `=`  -> `%5`
- `!`  -> `%6`
- `~`  -> `%7`

The escape sequences I chose here might not be the most intuitive ones for manually constructing filter strings. But it minimizes the change to existing paring code. My assumption is, there's very limited usage of filter that contains those special characters from command-line, otherwise we'd run into this problem earlier.

Here's an example that shows a currently broken scenario, the fully qualified name for the NUnit test below is `Class1.HelloMethod(1.5)`, which contains special characters `'('` and `')'`. When this name is used to construct a testcase filter, parsing will fail.

```cs
    [TestFixture]
    public class Class1
    {
        [Test]
        [TestCase(1.5)]
        public void HelloMethod(decimal randomAmount)
        {  
            Assert.AreEqual(1, 1);
        }
    }
```

I have done some quick test to ensure that this doesn't regress perf for regular scenario: the filter parser was invoked 10 times on a filter string in the form of `FullyQualifiedName=...|FullyQualifiedName=...|...`, which consists of 6987 conditions with total length of 792243 (this is the filter Test Explorer generated from running all tests in Roslyn C# semantic tests project). As you can see from the results below, no impact on perf was observed.

Before:
```
Avg: 91.1 ms Max: 223 ms Min: 72 ms
```
After:
```
Avg: 92.9 ms Max: 219 ms Min: 75 ms
```

## **UPDATE:**

I made additional change to the parser so now we can use typical escape syntax instead, which will make it easier in CLI scenario. Another benefit of this is improved perf, since we get rid of Regex matching. Now the same perf test above takes about 50 ms in average.

New syntax:
- `\` -> `\\`  This is used as the prefix for escape sequence.
- `(`   -> `\(`
- `)`  -> `\)`
- `&`  -> `\&`
- `|` -> `\|`
- `=`  -> `\=`
- `!`  -> `\!`
- `~`  -> `\~`

Fixes: https://developercommunity.visualstudio.com/content/problem/257779/live-unit-testing-does-not-work-with-nunit-testcas.html


genlu:
FYI @ManishJayaswal

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace Microsoft.VisualStudio.TestPlatform.Common.Filtering
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Text;
+    using Microsoft.VisualStudio.TestPlatform.Common.Resources;
+
+    public static class FilterHelpers
```
smadala: Can this be internal? LUT has dependency on `Microsoft.TestPlatform.Common`?

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+    {
+        private const char EscapePrefix = '%';
```
smadala: Can we  use `\`(backslash) as escape character convert ( `(` -> `\(`, `&` -> `\&`,etc..) and write the our own parsing logic rather than depending upon regex? To make easy to construct the filter from CLI?

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            var builder = new StringBuilder();
+            for (int i = 0; i < str.Length; ++i)
+            {
+                var currentChar = str[i];
+                if (currentChar == EscapePrefix)
+                {
+                    if (++i == str.Length || !ReverseLookUpMap.TryGetValue(currentChar = str[i], out var specialChar))
+                    {
+                        // "\" should be followed by a valid escape seq. i.e. '0' - '7'.
```
smadala: Nit: `%` should be followed.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FilterStringEscapeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace Microsoft.TestPlatform.Common.UnitTests.Filtering
+{
+    using System;
+    using System.Globalization;
+    using Microsoft.VisualStudio.TestPlatform.Common.Filtering;
+    using Microsoft.VisualStudio.TestPlatform.Common.Resources;
+    using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+    [TestClass]
+    public class FilterStringEscapeTests
```
smadala: Nit: `FilterHelpersTests`

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FilterStringEscapeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            Assert.AreEqual("printf%1\"%0s %4 %0d\", text, 0%2", escapedValue);
+            Assert.AreEqual(value, FilterHelpers.Unescape(escapedValue));
+        }
+
+        [TestMethod]
+        public void UnescapeInvalid1()
+        {
+            var invalidString = @"TestClass%8""a %4 b""%2.TestMethod";     
+            Assert.ThrowsException<ArgumentException>(() => FilterHelpers.Unescape(invalidString), string.Format(CultureInfo.CurrentCulture, Resources.TestCaseFilterEscapeException, invalidString));
```
smadala: Can we hardcode the expected string? So that if anyone accidentally changes it then test fails. Any change in  product code behavior should fail a test.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+        private const char EscapePrefix = '%';  
+        private static readonly Dictionary<char, char> LookUpMap;  
+        private static readonly Dictionary<char, char> ReverseLookUpMap;    
+        private static readonly char[] SpecialCharacters;
+
+        static FilterHelpers()
+        {
+            LookUpMap = new Dictionary<char, char>()
+            {
+                { '%', '0'},
```
singhsarab: Is this exhaustive list of special characters we expect ? If not, this won't scale as we can add just 2 more with current design as %9 will be the last supported value.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            SpecialCharacters = LookUpMap.Keys.ToArray();
+        }
+
+        /// <summary>
+        /// Escapes a set of special characters for filter (%, (, ), &, |, =, !, ~) by replacing them with their escape sequences. 
+        /// </summary>
+        /// <param name="str">The input string that contains the text to convert.</param>
+        /// <returns>A string of characters with special characters converted to their escaped form.</returns>
+        public static string Escape(string str)
```
singhsarab: I am guessing. we are just going to replicate this code on the IDE side. If yes, it is problematic, maintenance issues and what not.
Also, we would need to document this for other consumers of TP to use this escaping logic

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+                builder.Append(currentChar);
+            }
+
+            return builder.ToString();
+        }
+
+        /// <summary>
+        /// Converts any escaped characters in the input filter string.
+        /// </summary>
+        /// <param name="str"></param>
```
singhsarab: Nit: doc for str

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FilterStringEscapeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var value = "TestClass(1).TestMethod(2)";
+            var escapedValue = FilterHelpers.Escape(value);
+
+            Assert.AreEqual(@"TestClass%11%2.TestMethod%12%2", escapedValue);
+            Assert.AreEqual(value, FilterHelpers.Unescape(escapedValue));
+        }
+
+        [TestMethod]
+        public void EscapeStringWithSpecialCharacters()
```
singhsarab: Add test to cover '&' and '~' as well.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: Yes, made this public for LUT. LUT has dependency on TranslationLayer, which pulls in Common as well.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: I agree with you that this is hard to use in CLI scenario. As you can see from the out-of-sync comment you pointed out below, I actually started with exactly what you suggested, but quickly realized that it would involve a bigger change, which I was reluctant to do this late and I really want this fix for preview 3, as it's a serious regression for LUT. Let me give this a try today and see how it goes.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FilterStringEscapeTests.cs
genlu: Good point.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: I don't know if there's plan to expand operator/operation support for filter, but we can always use "%a", "%b", etc

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: I made the helper class public, so it can be used by other components. But I'm not sure if this is the right place for public helpers.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: Done.

genlu:
@shyamnamboodiripad @AbhitejJohn @drognanar Could you folks please take a look at this as well?

smadala (approved):
Good job 👍 Approving with suggestions.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,231 @@
...
+                    if (current == FilterHelpers.EscapePrefix)
+                    {
+                        // We just encountered "\\" (escaped '\'), this will set last to '\0' 
+                        // so the next char will not be treated as a suffix of escape sequence.
+                        current = '\0';
+                    }
+                }
+                else
+                {
+                    switch (current)
```
singhsarab: Just put a comment telling the allowed special tokens, eg "=", "!=", "!", "~"

singhsarab: When do we expect just "!" ? Can you share an example, also add to tests.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,231 @@
...
+                last = current;
+            }
+
+            if (tokenBuilder.Length > 0)
+            {
+                yield return tokenBuilder.ToString();
+            }
+        }
+
+        internal static IEnumerable<string> TokenizeFilterConditionString(string str)
```
singhsarab: Nit: Please add documentation for these as well.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,231 @@
...
+            var last = '\0';
+            for (int i = 0; i < str.Length; ++i)
+            {
+                var current = str[i];
+
+                if (last == FilterHelpers.EscapePrefix)
+                {
+                    // Don't check if `current` is one of the special characters here.
+                    // Instead, we blindly let any character follows '\' pass though and 
+                    // relies on `FiltetrHelpers.Unescape` to report such errors.
```
singhsarab: Nit: pass through*/ FilterHelpers* (Same in the other method)
I guess you'd change the name though.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/ConditionTests.cs
```diff
@@ -50,5 +50,67 @@ public void ParseShouldCreateProperConditionOnValidConditionString()
...
+            Condition condition = Condition.Parse(conditionString);
+            Assert.AreEqual("FullyQualifiedName", condition.Name);
+            Assert.AreEqual(Operation.Contains, condition.Operation);
+            Assert.AreEqual(@"TestClass1(""~"").TestMethod(1.5)", condition.Value);
+        }
+
+        [TestMethod]
+        public void ParseStringWithSingleUnescapedBangShouldFail1()
+        {
+
```
singhsarab: Nit: extra lines

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: I guess the tokenizer code should remain private to FilterExpression/Condition in this case, don't see a point of exposing them.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
genlu: We don't expect '!', but it's not handled in tokenizer. Instead, an exception is thrown in `Condition.GetOperator`.

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,231 @@
...
+                            break;
+
+                        case '!':
+                            if (tokenBuilder.Length > 0)
+                            {
+                                yield return tokenBuilder.ToString();
+                                tokenBuilder.Clear();
+                            }
+                            // Determine if this is a "!=" or just a single "!".
+                            var next = str[i + 1];
```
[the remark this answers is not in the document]
genlu: Good catch! Fixed

src/Microsoft.TestPlatform.Common/Filtering/FilterHelpers.cs
```diff
@@ -0,0 +1,231 @@
...
+
+                builder.Append(currentChar);
+            }
+
+            return builder.ToString();
+        }
+
+        internal static IEnumerable<string> TokenizeFilterExpressionString(string str)
+        {
+            if (str == null)
```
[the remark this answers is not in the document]
genlu: This is to ensure the exact same behavior as before (with regex). we might be able to do further adjustment in `Parse` methods but I think it can be done in a separate PR.

ManishJayaswal (approved):

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -168,16 +164,16 @@ internal string[] ValidForProperties(IEnumerable<string> properties, Func<string
         /// </summary>
         internal static FilterExpression Parse(string filterString, out FastFilter fastFilter)
         {
-            ValidateArg.NotNull(filterString, "filterString");
+            ValidateArg.NotNull(filterString, nameof(filterString));
 
-            // below parsing doesn't error out on pattern (), so explicitly search for that (empty parethesis).
+            //// below parsing doesn't error out on pattern (), so explicitly search for that (empty parethesis).
```
AbhitejJohn: nit nit nit: might want to undo the changes to this comment (remove the extra `//`).

src/Microsoft.TestPlatform.Common/Filtering/Condition.cs
```diff
@@ -281,7 +273,95 @@ private string[] GetPropertyValue(Func<string, Object> propertyValueProvider)
...
+                    {
+                        // Don't check if `current` is one of the special characters here.
+                        // Instead, we blindly let any character follows '\' pass though and 
+                        // relies on `FilterHelpers.Unescape` to report such errors.
+                        tokenBuilder.Append(current);
 
+                        if (current == FilterHelper.EscapeCharacter)
+                        {
+                            // We just encountered "\\" (escaped '\'), this will set last to '\0' 
+                            // so the next char will not be treated as a suffix of escape sequence.
```
AbhitejJohn: nit: treated as a **prefix** of escape sequence

src/Microsoft.TestPlatform.Common/Resources/Resources.Designer.cs
```diff
@@ -107,7 +107,7 @@ internal class Resources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The diagnostic data adapter &apos;{0}&apos; requested environment variable &apos;{1}&apos; with value &apos;{2}&apos; to be set in test execution environment, but another diagnostic data adapter &apos;{3}&apos; has already requested same environment variable with different value &apos;{4}&apos;..
+        ///   Looks up a localized string similar to The data collector &apos;{0}&apos; requested environment variable &apos;{1}&apos; with value &apos;{2}&apos; to be set in test execution environment, but another data collector &apos;{3}&apos; has already requested same environment variable with different value &apos;{4}&apos;..
```
AbhitejJohn: question: Are all these resource changes required too?

src/Microsoft.TestPlatform.Common/Resources/Resources.Designer.cs
```diff
@@ -159,20 +159,18 @@ internal class Resources {
...
+        internal static string FastFilterException {
+            get {
                 return ResourceManager.GetString("FastFilterException", resourceCulture);
             }
         }
-
+        
         /// <summary>
-        ///   Looks up a localized string similar to There are multiple configurations that have diagnostic data adapter  type &apos;{0}&apos; or Uri &apos;{1}&apos;. Duplicate configurations will be ignored in the test run..
+        ///   Looks up a localized string similar to There are multiple configurations that have data collector FriendlyName as &apos;{0}&apos;. Duplicate configurations will be ignored in the test run..
```
AbhitejJohn: this removed a `{1}` too - Might want to edit the Format message?

AbhitejJohn (approved):
LGTM

src/Microsoft.TestPlatform.Common/Resources/Resources.Designer.cs
genlu: Hmm, those aren't from my change, not sure what happened here...

src/Microsoft.TestPlatform.Common/Filtering/Condition.cs
genlu: I think this should be "suffix". But there's definitely issue with my wording so I rewrote it :)

PR #1635, opened 2018-06-07
[For testing not for merge] Remove mscoreei.dll from default exclude module list

smadala:
## Description
- Will get package from this changes to make https://github.com/dotnet/cli/pull/9355 green if we fix the crash issue.


smadala:
Addressed in https://github.com/Microsoft/vstest/pull/1636

Issue #1617, opened 2018-05-25
Running single Desktop(.NET Fullframework) test taking more than one minute


smadala:
## Description
- Running one simple desktop(net46) test takes more than one minute.
## Steps to reproduce
> Create simple test project with below csproj.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>netcoreapp2.1;net46</TargetFrameworks>
    <RootNamespace>cc_mstest</RootNamespace>

    <IsPackable>false</IsPackable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0-preview-20180525-01" />
    <PackageReference Include="MSTest.TestAdapter" Version="1.2.1" />
    <PackageReference Include="MSTest.TestFramework" Version="1.2.1" />
  </ItemGroup>

</Project>


```

- dotnet test

## Expected behavior
> Desktop tests should takes approximately same time as .NET Core tests.

## Actual behavior
> See below output "Test execution time: 1.1655 Minutes" for desktop tests where as .NET Core tests takes "Test execution time: 1.2214 Seconds".

```ps
PS C:\Users\samadala\src\samples-projects\cc-mstest> dotnet test
Build started, please wait...
Build completed.

Test run for C:\Users\samadala\src\samples-projects\cc-mstest\bin\Debug\netcoreapp2.1\cc-mstest.dll(.NETCoreApp,Version=v2.1)

Microsoft (R) Test Execution Command Line Tool Version 15.8.0-preview-20180524-01
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 1.2214 Seconds
Test run for C:\Users\samadala\src\samples-projects\cc-mstest\bin\Debug\net46\cc-mstest.dll(.NETFramework,Version=v4.6)
Microsoft (R) Test Execution Command Line Tool Version 15.8.0-preview-20180524-01
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 1.1655 Minutes
PS C:\Users\samadala\src\samples-projects\cc-mstest>
```

singhsarab:
[...]

PR #1637, opened 2018-06-08
Show time taken for acceptance test

smadala:
## Description
- Print time taken for each acceptance test, So that we can verify the issues like https://github.com/Microsoft/vstest/issues/1617 . And verify the regressions easily .


test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
```diff
@@ -455,6 +455,9 @@ private void Execute(string args, out string stdOut, out string stdError, out in
                 Console.WriteLine("IntegrationTestBase.Execute: Path = {0}", vstestconsole.StartInfo.FileName);
                 Console.WriteLine("IntegrationTestBase.Execute: Arguments = {0}", vstestconsole.StartInfo.Arguments);
 
+                Stopwatch stopwatch = new Stopwatch();
```
acesiddhu: wrap Stopwatch inside a class which implements IDisposable pattern. in Dispose stop and print the time. for reference please check with me

smadala: As discussed offline, IDisposable pattern very interesting. We have few places where we are using stopwatch like this. I'm scoping this PR to just add stopwatch in acceptance tests.

singhsarab (approved):

Issue #1628, opened 2018-06-01
Test platform crashes when there's null value in Payload messages


nigurr:
When tests are being discovered/executed if the adapter sends the value as "null" in the payload, platform fails to deserialize the payload causing it crash.

For example:

       {
          "Key": {
            "Id": "VCTestDiscoverer.TestCategory",
            "Label": "TestCategory",
            "Category": "",
            "Description": "",
            "Attributes": 1,
            "ValueType": "System.String[]"
          },
          "Value": null
        },

```
TpTrace Verbose: 0 : 4936, 7, 2018/05/31, 17:21:44.662, 12516221489, vstest.console.exe, TestRequestSender: GetAbortErrorMessage: Exception: Newtonsoft.Json.JsonSerializationException: Error setting value to 'StoreKeyValuePairs' on 'Microsoft.VisualStudio.TestPlatform.ObjectModel.TestCase'. ---> System.NotSupportedException: CustomStringArrayConverter cannot convert from (null).
   at System.ComponentModel.TypeConverter.GetConvertFromException(Object value)
   at System.ComponentModel.TypeConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.CustomStringArrayConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.TestObject.ConvertPropertyFrom[T](TestProperty property, CultureInfo culture, Object value)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.TestObject.SetPropertyValue[T](TestProperty property, T value, CultureInfo culture)
   at Microsoft.VisualStudio.TestPlatform.ObjectModel.TestObject.set_StoreKeyValuePairs(List`1 value)
   at SetStoreKeyValuePairs(Object , Object )
   at Newtonsoft.Json.Serialization.DynamicValueProvider.SetValue(Object target, Object value)
   --- End of inner exception stack trace ---
   at Newtonsoft.Json.Serialization.DynamicValueProvider.SetValue(Object target, Object value)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.SetPropertyValue(JsonProperty property, JsonConverter propertyConverter, JsonContainerContract containerContract, JsonProperty containerProperty, JsonReader reader, Object target)
[...]

PR #1640, opened 2018-06-11
Fix issue while deserializing null values

nigurr:
## Description
Handle null values in TestCategory. 

C# throws exception if null value passed to ConvertFrom and with our custom handlers, we shouldn't call base when the value is null.
ref: https://referencesource.microsoft.com/#System/compmod/system/componentmodel/TypeConverter.cs,106

https://github.com/Microsoft/vstest/issues/1628


singhsarab (approved):

nigurr:
@dotnet-bot test Windows_NT / Debug Build please

Issue #1642, opened 2018-06-12
Few UWP VC++ unit tests are not executing


smadala:
## Description
-  Every 10th tests case of UWP VC++ unit test projects are not executing.
## Steps to reproduce
[UnitTestApp3.zip](https://github.com/Microsoft/vstest/files/2094590/UnitTestApp3.zip)
- Download attached UnitTestApp3.zip
- Click Run all in VS 15.7.3

## Expected behavior
> All the 12 test cases should execute.

## Actual behavior
> Only 11 test cases are executing.

## Diagnostic logs

[UWP-cpp-testhost-logs.txt](https://github.com/Microsoft/vstest/files/2094597/UWP-cpp-testhost-logs.txt)

smadala:
Fix will be available in VS 15.8.

PR #1649, opened 2018-06-15
Fix Few UWP VC++ unit tests are not executing

smadala:
## Description
**RCA**  
Changing the testcase's object source for inprogress tests for clients(VS Test explorer) causing the issue in UWP C++ scenario as the adapter and testplatform share same testcase object by changing the testcase source the testcase becoming invalid in UWP C++ adapter.
## Related issue
https://github.com/Microsoft/vstest/issues/1642

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
@@ -152,6 +152,13 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities
             return JsonConvert.SerializeObject(message);
         }
 
+        /// <inheritdoc/>
+        public T Clone<T>(T obj)
+        {
+            var stringObj = this.Serialize<T>(obj);
+            return this.Deserialize<T>(stringObj);
+        }
+
         /// <summary>
         /// Serialize an object to JSON using default serialization settings.
         /// </summary>
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
@@ -155,8 +155,13 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities
         /// <inheritdoc/>
         public T Clone<T>(T obj)
         {
-            var stringObj = this.Serialize<T>(obj);
-            return this.Deserialize<T>(stringObj);
+            if (obj == null)
+            {
+                return default(T);
+            }
+
+            var stringObj = this.Serialize<T>(obj, 2);
+            return this.Deserialize<T>(stringObj, 2);
         }
 
         /// <summary>
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
@@ -13,7 +13,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution
     using System.Linq;
     using System.Reflection;
     using System.Threading.Tasks;
-
+    using CommunicationUtilities.Interfaces;
     using Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework;
     using Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.Utilities;
     using Microsoft.VisualStudio.TestPlatform.Common.Interfaces;
@@ -78,6 +78,11 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution
         /// </summary>
         private RunConfiguration runConfiguration;
 
+        /// <summary>
+        /// The Serializer to clone testcase object incase of user input test source is package. E.g UWP scenario(appx/build.appxrecipe).
+        /// </summary>
+        private IDataSerializer dataSerializer;
+
         #endregion
 
         #region Constructor
@@ -107,7 +112,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution
                 testRunEventsHandler, 
                 testPlatformEventSource, 
                 testCaseEventsHandler as ITestEventsPublisher, 
-                new PlatformThread())
+                new PlatformThread(),
+                JsonDataSerializer.Instance)
         {
         }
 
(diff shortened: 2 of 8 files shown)


src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -152,6 +152,13 @@ public string SerializePayload(string messageType, object payload, int version)
             return JsonConvert.SerializeObject(message);
         }
 
+        /// <inheritdoc/>
```
singhsarab: Please write a comment on why we chose this approach, just for reference.
Also consider making it an extension method.

singhsarab (approved):
Tests please :)

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -152,6 +152,13 @@ public string SerializePayload(string messageType, object payload, int version)
...
+        public T Clone<T>(T obj)
+        {
+            var stringObj = this.Serialize<T>(obj);
```
cltshivash: Suggestion: Check for ICloneable implemention and if so use it. else fallback

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
```diff
@@ -624,22 +624,44 @@ private bool TryToRunInSTAThread(Action action, bool waitForCompletion)
...
+
             // Before sending the testresults back, update the test case objects with source provided by IDE/User.
-            if (!string.IsNullOrEmpty(package))
+            foreach (var tr in testResults)
             {
-                foreach (var tr in testResults)
-                {
-                    tr.TestCase.Source = package;
-                }
+                tr.TestCase.Source = package;
```
cltshivash: >tr.TestCase.Source = package; [](start = 16, length = 29)

Is this safe ?

cltshivash (changes requested):
:clock1:

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
smadala: There are lot of discussion over stackoverflow, how to do deep copy.
- https://stackoverflow.com/questions/129389/how-do-you-do-a-deep-copy-of-an-object-in-net-c-specifically  
- https://stackoverflow.com/questions/78536/deep-cloning-objects

1. BinaryFormatter
This very generic and easy to implement, But it is not available in NS14.  
2. ICloneable
This required changes in all the class which are involved and difficult to maintain(complicated code).

As we are already using `JsonDataSeializer` to transfer the object over network. We can use the same functionality to do clone Which is well tested.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
smadala: Please see https://github.com/Microsoft/vstest/pull/1649#discussion_r196123434

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
smadala: No, this not safe. As we discussed offline,  currently there is no known scenario to which break  because of not copying(better performance). However it is better to copy the test results(which adapter creates) to make sure correctness over performance.

I have made changes to copy the test results too.

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
```diff
@@ -557,6 +566,12 @@ private void SetAdapterLoggingSettings()
                 // Collecting Number of Adapters Used to run tests. 
                 this.requestData.MetricsCollection.Add(TelemetryDataConstants.NumberOfAdapterUsedToRunTests, this.ExecutorUrisThatRanTests.Count());
 
+                if (lastChunk.Any() && string.IsNullOrEmpty(package) == false)
```
cltshivash: nitpick: inaddition to package being null should we also compare with source.. also depending on package being null for a business logic seems a little odd.. ideally there shold be a flag setup upstream (to indicate this behavior)

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
```diff
@@ -557,6 +566,12 @@ private void SetAdapterLoggingSettings()
...
+                {
+                    Tuple<ICollection<TestResult>, ICollection<TestCase>> updatedTestResultsAndInProgressTestCases = this.UpdateTestCaseSourceToPackage(lastChunk, null);
```
cltshivash: nitpick: can use var ? (also usage of Tuple indicates a missing struct)

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
```diff
@@ -586,13 +597,19 @@ private void SetAdapterLoggingSettings()
             }
         }
 
-        private void OnCacheHit(TestRunStatistics testRunStats, ICollection<TestResult> results, ICollection<TestCase> inProgressTests)
+        private void OnCacheHit(TestRunStatistics testRunStats, ICollection<TestResult> results, ICollection<TestCase> inProgressTestCases)
         {
             if (this.testRunEventsHandler != null)
             {
-                UpdateTestResults(results, inProgressTests, this.package);
+                if (string.IsNullOrEmpty(package) == false)
```
cltshivash: i see this being repeated.. opportunity to have it in a meaningful method..

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/BaseRunTests.cs
```diff
@@ -624,22 +641,50 @@ private bool TryToRunInSTAThread(Action action, bool waitForCompletion)
...
-        private static void UpdateTestResults(IEnumerable<TestResult> testResults, IEnumerable<TestCase> testCases, string package)
+        private Tuple<ICollection<TestResult>, ICollection<TestCase>> UpdateTestCaseSourceToPackage(
+            ICollection<TestResult> testResults,
+            ICollection<TestCase> inProgressTestCases)
         {
-            // Before sending the testresults back, update the test case objects with source provided by IDE/User.
-            if (!string.IsNullOrEmpty(package))
+
+            EqtTrace.Verbose("BaseRunTests.UpdateTestCaseSourceToPackage: Update source details for testResults and testCases.");
+
```
cltshivash: remove unnecessary empty lines

cltshivash (approved):
Please address the comments

smadala:
@cltshivash I have to wait for one hour for PR checks to pass if I address the PR comments. As there no behavior change comments I'm going ahead with merge. Will address the comments in separate PR.

smadala:
Created https://github.com/Microsoft/vstest/pull/1651 to fix nits.

PR #1664, opened 2018-06-29
Auto-generate F# program file

### LINKED ISSUE #1171: Generate the F# file who contains the main (`EntryPoint`) based on runtime
F# bundle the `Program.fs` (who contains the main, `[<EntryPoint>]` in f#), because a test project targeting `netcoreapp2.0` is built as exe (so need a main)
But if i switch to `net461`, the test project is a library, by design, and should include it.

Atm template is going to be changed to conditionally include the `Program.fs` (ref https://github.com/dotnet/templating/issues/1271 ) to workaround this.

But that's annoying, because all test framework templates need to do so. otherwise bug is subtle (ref https://github.com/dotnet/netcorecli-fsc/issues/79 ).

C# autogenerate the `Program.cs` needed in the `CoreGenerateProgramFile` target.

F# should do the same.

Minor notes:

- need to add the file at end (for f# file order matter, and `EntryPoint` MUST be the last), so just `Compile` is not enough (=> require support by f# sdk, like adding `CompileLast` or `CompileMain` item)
- need a property to disable this behaviour. For example the [expecto](https://github.com/haf/expecto) test library, used a lot in F#, configure itself in main, so cannot be autogenerated
- something to generate the main empty file, `obj/{projectName}.Program.fs` (in F# sdk or msbuild sdk or here)
- a feature flag (a property?), so project based on previous template (who add a `Program.fs` explicit) doesnt break with new sdk version

/cc @dsyme @cartermp @KevinRansom

## Environment

dotnet sdk >= 2.0

saul:
## Description

Allows F# Program.fs to be autogenerated.

**Todo**:

- [x] Do not auto-generate a 'Program.fs' is already in Compile

## Related issue
Fixes #1171. cc @enricosada


src/package/nuspec/Microsoft.NET.Test.Sdk.targets
```diff
@@ -76,15 +76,26 @@
       <Line Include="End Sub"/>
       <Line Include="End Module"/>
     </ItemGroup>
+
+    <ItemGroup Condition="'$(Language)'=='F#'">
+      <Line Include="module Program"/>
```
Alxandr: This will fail if another module is named `Program`. It would probably be better to do something like 

```fsharp
module ``VSTest-AutoGenerated-Program``
```

It can still have collisions, but way less likely. Also, both the C# and the VB version is decorated with `TestSDKAutoGeneratedCodeAttribute` which is probably a good idea.

Alxandr:
My "program" files are never called `Program.fs`, so having that as the way to turn off auto-generation of would be a bit too opinionated IMHO.

saul:
Yeah, but you still have the option to specify `<GenerateProgramFile>false</GenerateProgramFile>` as you do with C#/VB.

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
saul: Good catch - I'll fix that.

mayankbansal018 (approved):

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
```diff
@@ -76,15 +76,28 @@
...
 
     <!--
     Compile Include the generated Program File
     -->
-    <ItemGroup>
+    <ItemGroup Condition="'$(Language)'!='F#'">
       <Compile Include="$(GeneratedProgramFile)"/>
     </ItemGroup>
+
+    <ItemGroup Condition="'$(Language)'=='F#'">
```
smadala: I'm not sure `runtimeconfig.*.json` will gets generates because `CompileAfter`?
FYI https://github.com/dotnet/sdk/issues/201 https://github.com/Microsoft/vstest/issues/287.

@saul did you test simple unit test project of F# with your changes?

smadala (changes requested):
@saul Thanks a lot 👍 for PR.

saul:
Hi @smadala - I texted the changes with a unit test project I had and the Program.fs file was generated as expected and the tests passed.

smadala:
@saul Can you please test E2E?

 Add  `Microsoft.NET.Test.Sdk.15.8.0-dev.nupkg` (  
 Example path: `C:\Users\samadala\src\vstest\artifacts\Debug\packages\Microsoft.NET.Test.Sdk.15.8.0-dev.nupkg`, obtained by `build.cmd`) package to your test project and update nuget feed.

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
smadala: I think I understood what is [CompileAfter](https://github.com/Microsoft/visualfsharp/blob/61907dcd859702924021bc0aad74f64ce459c9f0/src/fsharp/FSharp.Build/Microsoft.FSharp.Targets#L296). Now it makes sense after reading issue one more time. 

I have checked out your branch and verified E2E scenario. Below are the issue we need to address.
-  Newly created F# unit test project not getting compiled because of below compilation error.
```ps
PS C:\Users\samadala\tmp\fs-test-project> dotnet test
Build started, please wait...
C:\Users\samadala\tmp\fs-test-project\Program.fs(1,1): error FS0222: Files in libraries or multiple-file applications must begin with a namespace or module declaration. When using a module declaration at the start of a file the '=' sign is not allowed. If this is a top-level module, consider removing the = to resolve this error. [C:\Users\samadala\tmp\fs-test-project\fs-test-project.fsproj]
```
This can be fixed by removing `<Compile Include="Program.fs" />` from template fsproj [here](https://github.com/dotnet/templating/blob/4cc28db23041bfbeae2797fbfbc7356b50c09e84/template_feed/Microsoft.DotNet.Test.ProjectTemplates.2.0/content/MSTest-FSharp/Company.TestProject1.fsproj#L14) and removing [Program.fs](https://github.com/dotnet/templating/blob/4cc28db23041bfbeae2797fbfbc7356b50c09e84/template_feed/Microsoft.DotNet.Test.ProjectTemplates.2.0/content/MSTest-FSharp/Program.fs), We can address this with separate https://github.com/dotnet/templating PR 

- Test project created with old templates fail to compile with above error on updating to latest `Microsoft.NET.Test.Sdk` package.

This can be fixed by adding check [here](https://github.com/Microsoft/vstest/pull/1664/files#diff-251b10661dfb4f6d4d69fb78fe9127b6R99) that don't include  `GeneratedProgramFile` in `CompileAfter` if `Compile` item group contains `Program.fs`.

@saul Can you please fix the seconds issue in this PR?

saul: Thanks @smadala - I've fixed the second problem.

I've added a warning: `A 'Program.fs' file can be automatically generated for F# .NET Core test projects. To fix this warning, either delete the file from the project, or set the <GenerateProgramFile> property to 'false'.`

This lets the user either remove their Program.fs, or disables `<GenerateProgramFile>` if they need a complex Program.fs for whatever reason.

smadala: @saul Adding warning is very good idea.

src/package/nuspec/Microsoft.NET.Test.Sdk.targets
```diff
@@ -76,15 +76,31 @@
...
+      <ProgramCompiles Include="@(Compile)" Condition=" '%(Identity)' == 'Program.fs' "/>
+      <CompileAfter Include="$(GeneratedProgramFile)" Condition="@(ProgramCompiles-&gt;Count()) == 0"/>
+    </ItemGroup>
+
+    <Warning Condition=" @(ProgramCompiles-&gt;Count()) != 0 " Text="A 'Program.fs' file can be automatically generated for F# .NET Core test projects. To fix this warning, either delete the file from the project, or set the &lt;GenerateProgramFile&gt; property to 'false'." />
```
smadala: This might need localization. We will take it in separate PR if required.

smadala (approved):

smadala:
@dotnet-bot test this

smadala:
@dotnet-bot test this

ErikSchierboom:
Merged, great news! Will this be in the next VSTest release?

Issue #1661, opened 2018-06-27
Unexpected exception during cancel operation.


singhsarab:
## Description
> This is a translation layer scenario. 
When cancellation request is raised by client, some times unexpected exceptions is reported back instead of clean cancel.
This is an inconsistent issue.

## Steps to reproduce
* Create a managed UT project in VS
* Add some tests that take some time to run
* Run All using Test Explorer
* Cancel the run via Test Explorer

## Expected behavior
> The cancel operation cancels the current run and no unexpected exceptions get logged.

## Actual behavior
> The following exception gets logged.

```
 Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: vstest.console process failed to connect to testhost process after 90 seconds. This may occur due to machine slowness, please set environment variable VSTEST_CONNECTION_TIMEOUT to increase timeout.

at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.ThrowExceptionOnConnectionFailure(Int32 connTimeout)

at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.SetupChannel(IEnumerable`1 sources)

at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyExecutionManager.StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsHandler eventHandler)
```

AbhitejJohn:
@singhsarab : Sure here you are: #1692

singhsarab:
Thanks. Closing this one as it has been addressed via #1667

PR #1667, opened 2018-07-02
Cancellation message fix

singhsarab:
## Description
Fixing error messaging for the cancel during WaitForRequestHandlerConnection and LaunchHost

## Related issue
https://github.com/Microsoft/vstest/issues/1661

### DIFF
--- a/src/Microsoft.TestPlatform.Common/Resources/Resources.resx
+++ b/src/Microsoft.TestPlatform.Common/Resources/Resources.resx
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CancellationRequested" xml:space="preserve">
+    <value>Cancelling the operation as requested.</value>
+  </data>
   <data name="DataCollectionMessageFormat" xml:space="preserve">
     <value>Data collection : {0}</value>
   </data>
--- a/src/Microsoft.TestPlatform.Common/Utilities/CancellationTokenExtensions.cs
+++ b/src/Microsoft.TestPlatform.Common/Utilities/CancellationTokenExtensions.cs
@@ -3,8 +3,8 @@
 
 namespace Microsoft.VisualStudio.TestPlatform.Common.Utilities
 {
-    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using System.Threading;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
     public static class CancellationTokenExtensions
     {
--- a/src/Microsoft.TestPlatform.Common/Utilities/CancellationTokenExtensions.cs
+++ b/src/Microsoft.TestPlatform.Common/Utilities/CancellationTokenExtensions.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.Common.Utilities
+{
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+    using System.Threading;
+
+    public static class CancellationTokenExtensions
+    {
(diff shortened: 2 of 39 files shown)


src/Microsoft.TestPlatform.Common/Resources/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CancellationRequested" xml:space="preserve">
+    <value>Cancelling the operation as requested.</value>
```
abhishkk: Something like "The operation was cancelled." seems better. We can check the existing IDE cancellation message to get more input.

src/Microsoft.TestPlatform.Common/Utilities/CancellationTokenExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.Common.Utilities
+{
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+    using System.Threading;
```
abhishkk: System using at top.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs
```diff
@@ -138,6 +140,11 @@ public virtual bool SetupChannel(IEnumerable<string> sources)
                 catch (Exception ex)
                 {
                     EqtTrace.Error("ProxyOperationManager: Failed to launch testhost :{0}", ex);
+
+                    if (ex is AggregateException && ex.InnerException is TaskCanceledException)
+                    {
+                        throw new TestPlatformException(Common.Resources.Resources.CancellationRequested);
```
abhishkk: TaskCanceledException might happen because of something other than cancel also. In that case, we might not want exception to part of TestPlatformException.

Here in case of TaskCancelledException, we can check `this.CancellationTokenSource.Token.ThrowTestPlatformExceptionIfCancellationRequested();` which will throw TestPlatformException only in case of cancellation.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/ProxyExecutionManagerTests.cs
```diff
@@ -402,6 +402,17 @@ public void StartTestRunShouldCatchExceptionAndCallHandleRawMessageAndHandleLogM
             mockTestRunEventsHandler.Verify(s => s.HandleLogMessage(TestMessageLevel.Error, It.IsAny<string>()));
         }
 
+        [TestMethod]
+        public void StartTestRunForCancelRequestShouldHandleLogMessageWithProperErrorMessage()
```
abhishkk: we can have a unit test for case when exception is not testplatformexception

src/Microsoft.TestPlatform.Common/Resources/Resources.resx
singhsarab: Okay. I just took this as reference. https://github.com/Microsoft/vstest/blob/09057425a192fd329c32bbee2fd9bdd401751450/src/Microsoft.TestPlatform.CrossPlatEngine/Resources/Resources.resx#L180 
And I guess we should remove this string.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs
singhsarab: Makes sense. Done

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/ProxyExecutionManagerTests.cs
singhsarab: Added.

abhishkk (approved):

PR #1678, opened 2018-07-06
fix for settingsprovider.dll not being loaded in teshost

karanjitsingh:
load paths for SettingsProviderEndsWithPattern along with TestAapterEndsWithPattern in ProxyExecutionManager when initializing extensions
added tests for the same

## Description
SettingsProvider.dll paths are not sent to testhost.exe from vstest.console which causes settingsprovider.dll to not load in testhost

This is affected in case of --isolation and --platform:x64


karanjitsingh:
@dotnet-bot test

karanjitsingh:
@dotnet-bot test this

PR #1697, opened 2018-07-20
Remove temporary cc package

smadala:
## Description
- Remove temporary Code coverage package.
- update vstest version to 15.9.0


singhsarab (approved):
Good change

scripts/vstest-codecoverage.runsettings
```diff
@@ -20,6 +20,7 @@
               <ModulePath>.*mscorlib.ni.dll$</ModulePath>
               <ModulePath>.*cryptbase.dll$</ModulePath>
               <ModulePath>.*bcryptPrimitives.dll$</ModulePath>
+              <ModulePath>.*CodeCoverage.exe$</ModulePath>
```
mayankbansal018: Why are we adding CC.exe in exclusion list?

smadala: This we have discussed earlier.
- We launch codecoverage.exe in unit tests like [this](https://github.com/Microsoft/vstest/blob/master/test/DataCollectors/TraceDataCollector.UnitTests/VanguardTests.cs#L97), Enable code coverage for those tests makes tests fails. As per code coverage basic principal the behavior of tests should not change under codecoverage.
- We are not interested in coverage info of codecovergae.exe here

Issue #1643, opened 2018-06-13
Confusing log message when all tests are filtered out


genlu:
## Description
I noticed a confusing warning in the LUT log after restart LUT (see below). Essentially, after filtering out tests, none left to execute, which is expected because there's no code change since LUT was stopped. But this message makes it sounds like an error.  @Satya Madala Could you guys change this message at least for filter scenario (if there’s tests discovered but all filtered out).

```
[15:02:09.546 Info] [TestRunner 1] No test is available in C:\Users\gel\source\repos\XUnitTestProject1\.vs\NUnitCoreTests\lut\0\t\NUnitCoreTests\Debug\netcoreapp2.0\NUnitCoreTests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.
```

## Steps to reproduce
With latest dogfood build:
Open a test project -> Start LUT -> Stop LUT -> Start LUT again

smadala:
@genlu  Is below message looks good?

```ps
PS C:\Users\samadala\src\vstest> .\artifacts\Debug\net451\win7-x64\vstest.console.exe .\test\TestAssets\SimpleTestProject\bin\Debug\net451\SimpleTestProject.dll /testcasefilter:"Name=abc"
Microsoft (R) Test Execution Command Line Tool Version 15.8.0-dev
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
No test available for testcase filter `Name=abc` in C:\Users\samadala\src\vstest\.\test\TestAssets\SimpleTestProject\bin\Debug\net451\SimpleTestProject.dll
```

smadala:
What is the timeline your looking for this issue to get fixed?

genlu:
The message above is fine. But for LUT, the filter can be very long (e.g. all tests in project). So can we not include the filter string itself in the message?

In terms of timeline, 15.8 preview 4 would be great. @ManishJayaswal

ManishJayaswal:
"No test matches the given testcase filter" may be better.

For LUT, in most cases, this would not be an error condition. It would just point to the fact that no tests are covering the line which was edited. Is there any way we can make this message "informational" so that it only shows up with verbose logging? Else this will keep showing as error when technically there is no error- the program is working as expected. If it only shows up in verbose logging then spending the filter text to the message may help with diagnostics so it may be ok.

For timeline - preview 4 of 15.8 is fine.

smadala:
[...]

PR #1723, opened 2018-08-07
Show error message on Framework35

smadala:
## Description
- Show error message when user pass `/Framework:Framework35` as we don't have testhost built for CLR 2.0

## Related issue
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/493289

### DIFF
--- a/src/vstest.console/Resources/Resources.resx
+++ b/src/vstest.console/Resources/Resources.resx
@@ -276,6 +276,9 @@
   <data name="FileNotFound" xml:space="preserve">
     <value>'{0}' not found.</value>
   </data>
+  <data name="Framework35NotSupported" xml:space="preserve">
+    <value>Framework35 not supported. Use Framework40 or above to run tests in CLR 4.0 "compatibly mode".</value>
+  </data>
   <data name="FrameworkArgumentHelp" xml:space="preserve">
     <value>--Framework|/Framework:&lt;Framework Version&gt;
       Target .Net Framework version to be used for test execution. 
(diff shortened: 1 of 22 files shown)


src/vstest.console/Resources/Resources.resx
```diff
@@ -276,11 +276,14 @@
   <data name="FileNotFound" xml:space="preserve">
     <value>'{0}' not found.</value>
   </data>
+  <data name="Framework35NotSupported" xml:space="preserve">
+    <value>Framework35 not supported. Use Framework40 or above to run tests in CLR 4.0 "compatibly mode".</value>
```
singhsarab: Suggestion:
Framework35 is not supported. For projects targeting .Net Framework 3.5, please use Framework40 to run tests in CLR 4.0 "compatibility mode".

singhsarab (approved):
Approving with suggestions.

@cltshivash @pvlakshm please review the change in behaviour

pvlakshm:
Are we erroring out if Framework35 is specified? We should not. Today, we run such tests in CLR 4.0's compat mode. Just call that out as a warning so the user is more aware.

smadala:
@pvlakshm In Tpv1 if user passes Framework35, tests runs in CLR 2.0 process. From VS 15.5 we have changed that behavior to run tests CLR 4.0 process in `compact mode`. If we don't error out user might think his tests running in CLR2.0 process.

As we don't support running tests in CLR 2.0 process, we should remove the `/Framework:Framework35`.

smadala:
>Just call that out as a warning so the user is more aware.

Error makes more awareness. More details refer https://github.com/Microsoft/vstest/issues/1643#issuecomment-399493125

PR #1725, opened 2018-08-10
Add overwrite parameter to trx logger

### LINKED ISSUE #1603: dotnet test on a multi-target projects logs only the last target
## Description
`dotnet test "--logger:trx;LogFileName=results.trx"` on a multi-target project (e.g. `<TargetFrameworks>netcoreapp2.0;net45</TargetFrameworks>` in the `.csproj`) causes the logger to only log the last target. Possibly because the previous targets' logger output file is overwritten by subsequent targets. FYI, this also happens with the [nunit logger extension](https://github.com/Faizan2304/LoggerExtensions).

## Steps to reproduce

`dotnet test "--logger:trx;LogFileName=results.trx"`

## Expected behavior
Either generate as many .trx files as there are targets (e.g. `results.net45.trx`, `results.netcoreapp2.0.trx`) OR assemble the results into the single .trx (if file format allows for it). Perhaps the user can pick a behavior.

## Actual behavior
Only a single `results.trx` is created and it's contents only hold the results of the last target framework in the `.csproj`.

## Environment
In case it matters, our multi-target project uses nunit 3.9 with the the following inside it's `.csproj`
```
<PackageReference Include="NUnit" Version="3.9.0" />
<PackageReference Include="NUnit3TestAdapter" Version="3.9.0" />
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
<PackageReference Include="NunitXml.TestLogger" Version="1.1.0" />
``` 
that last bit is so we can also do `dotnet test --logger:"nunit;LogFilePath=results.xml"` since our CI can use either trx or nunit xml - but that also fails similarly.

OS: Windows 10 v1803
.NET SDK: `dotnet --version` shows `2.1.0`
Not sure if/how dotnet test wraps over vstest but this is `vstest.console.exe`
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow>vstest.console.exe
Microsoft (R) Test Execution Command Line Tool Version 15.7.1
Copyright (c) Microsoft Corporation.  All rights reserved.
```

smadala:
## Description
- **Before fix**
```PS
PS C:\Users\samadala\src\vstest> C:\Users\samadala\src\vstest\artifacts\Debug\net451\win7-x64\vstest.console.exe "D:\g\samples\Net35TestProject\Net35TestProject\bin\Debug\Net35TestProject.dll" /logger:"trx;LogfileName=abc.trx"
Microsoft (R) Test Execution Command Line Tool Version 15.9.0-dev
Copyright (c) Microsoft Corporation.  All rights reserved.

Passed   TestMethod1
WARNING: Overwriting results file: C:\Users\samadala\src\vstest\TestResults\abc.trx
Results File: C:\Users\samadala\src\vstest\TestResults\abc.trx

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 0.5303 Seconds
```

- **After fix**

```PS
PS C:\Users\samadala\src\vstest> C:\Users\samadala\src\vstest\artifacts\Debug\net451\win7-x64\vstest.console.exe "D:\g\samples\Net35TestProject\Net35TestProject\bin\Debug\Net35TestProject.dll" /logger:"trx;LogfileName=abc.trx;overwrite=false"
Microsoft (R) Test Execution Command Line Tool Version 15.9.0-dev
Copyright (c) Microsoft Corporation.  All rights reserved.

Passed   TestMethod1
Results File: C:\Users\samadala\src\vstest\TestResults\abc[1].trx

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 0.5841 Seconds
```

## Related issue
Fixes https://github.com/Microsoft/vstest/issues/1603


abhishkk:
@smadala
It seems like this parameter can be avoided. Why can't we add `abc-{frameworkname}.trx` for each framework whenever user gives `abc` as trx file name. We do similar for diag file path also for logging test host and data collector logs also. Adding a parameter for just this purpose doesn't seem reasonable.

singhsarab:
Need a small RFC for this.
Based on the feedback we got #1603, if we create proper name appended with framework, it is much easier and intuitive for user as well.

We should be careful while adding options, because adding is easy, removing them is difficult. Reason I want us evaluate the value we are providing with this new option.
Default, I think we should not overwrite, onus of deleting the file if it already exists should be with the user. If we see folks asking for overwrite we can then give it as an option.

/cc: @cltshivash @pvlakshm for reviewing the spec here

mayankbansal018:
see comment https://github.com/Microsoft/vstest/issues/243#issuecomment-263226391 , Pratap has already defined, that append target framework to trx file.

PR #1718, opened 2018-08-07
Updating TestWindow.Interfaces assembly redirection 

AbhitejJohn:
Pointing this to 16.0.0.0 for the next version of VS. This fixes discovery/execution for Google Test Adapter and other 3rd party adapters. 
Please let me know if master is not the right branch for this.

PR #1730, opened 2018-08-13
Revert "Updating TestWindow.Interfaces assembly redirection"

singhsarab:
Reverts Microsoft/vstest#1718

### DIFF
--- a/src/testhost.x86/app.config
+++ b/src/testhost.x86/app.config
@@ -19,7 +19,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.VisualStudio.TestWindow.Interfaces" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-        <bindingRedirect oldVersion="11.0.0.0-15.0.0.0"  newVersion="16.0.0.0"/>
+        <bindingRedirect oldVersion="11.0.0.0-14.0.0.0"  newVersion="15.0.0.0"/>
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.VisualStudio.QualityTools.UnitTestFramework" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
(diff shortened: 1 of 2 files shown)


src/testhost.x86/app.config
```diff
@@ -19,7 +19,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.VisualStudio.TestWindow.Interfaces" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-        <bindingRedirect oldVersion="11.0.0.0-15.0.0.0"  newVersion="16.0.0.0"/>
+        <bindingRedirect oldVersion="11.0.0.0-14.0.0.0"  newVersion="15.0.0.0"/>
```
mayankbansal018: do we need this redirection at all? Can't we remove this entirely?

singhsarab:
@AbhitejJohn FYI

singhsarab:
This already happened with https://github.com/Microsoft/vstest/pull/1731

src/testhost.x86/app.config
AbhitejJohn: So certain extensions seem to be extending both TestWindow related extensibility points(for which they need TW.Interfaces) and TP in the same assembly which sometimes ends up triggering a load of TW.Interfaces in the testhost process. That starts bombing test discovery/execution. Now devenv.exe has a similar redirection so the TW extension points as such work perfectly fine but discovery/execution through the platform would bomb out without this.

singhsarab: How do these extensions work in portable scenarios, when VS isn't present?

These extensions should either have this interface shipped along with them or not have this dependency in the adapter least.

In Dev16, I wish to stop supporting these extensions and would like them to react to this.

AbhitejJohn: Well they do not support portable scenarios. I've seen C++ extensions - Google Test, Cmake etc having such dependencies. That should be ok as long as we have documentation on this.

AbhitejJohn: If we do want to stop doing this, we might want to communicate with the extension authors sooner because these scenarios are blocked in the latest Dev16 bits.

singhsarab: Can you point to an existing bug for this ?

AbhitejJohn: @singhsarab : This one: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/664972

Issue #1335, opened 2017-12-18
JetBrains dotCover can't get coverage when testhost is killed


AM5800:
## Description
> **IMPORTANT**: if the defect is reproduced only in a workflow from within the Visual Studio IDE then do not report the issue here - instead, please report it using Visual Studio's "Send Feedback" option that can be accessed from the Help menu OR using this link https://developercommunity.visualstudio.com.
>
> For a defect reproducable from the vstest command line, describe the issue you've observed.

Hi, I am from dotCover team. Recently we have detected an issue with vstest that prevents our users from covering unit tests in console mode. 
dotCover is a classic coverage profiler. We store intermediate coverage data in the profiled process. And when this process is about to exit - we save this data on disk. But if the process is killed - coverage data is lost. This is what I am observing now. 

Say we have a command `dotnet vstest testAssembly.dll` and we would like to get unit test coverage with dotCover console tool. We feed this command into dotCover and the following happens(a little simplified):
1) dotCover spawns dotnet.exe process
2) dotnet.exe launches testhost.dll 
3) tests are run in testhost, they are instrumented by dotCover and coverage is collected in memory.
4) When tests are finished dotCover starts to save data
5) And suddenly testhost process is killed and coverage is lost.

I believe this happens in this line: https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs#L217 - Test host is killed after 100ms timeout. 
Most of the time 100ms is enough to save coverage. But sometimes not.

So do you have any ideas on how we can resolve this issue?

## Steps to reproduce
> What steps can reproduce the defect?
> Please share the setup, commandline for vstest.console, sample project, target
> framework etc.

I don't have a stable repro for you. It happens not always and you need really big project to exceed those 100 ms. I am able to reproduce this issue if I insert sleeps into dotCover executable. But I can't share it

So far I tried to reproduce this only with netcoreapp2.0 projects.

## Expected behavior
> Share the expected output

## Actual behavior
> What is the behavior observed?

## Diagnostic logs
> Please share test platform diagnostics logs. Instructions to collect logs are [here](https://github.com/Microsoft/vstest-docs/blob/master/docs/diagnose.md#test-platform-diagnostics).  
> The logs may contain test assembly paths, kindly review and mask those before sharing.

## Environment
> Please share additional details about the test environment.
> Operating system, Build version of vstest.console

AM5800:
[...]

PR #1741, opened 2018-08-20
Add a way to override test host close timeout by the environment variable VSTEST_CLOSE_TIMEOUT for #1335

NikolayPianikov:


mayankbansal018:
@NikolayPianikov , thanks for raising this PR, can you also provide some details, as to why this change is needed? Maybe a reference link to some bug

ManishJayaswal:
tagging @mayankbansal018 and I think I was accidently tagged by @NikolayPianikov

singhsarab:
@NikolayPianikov This timeout based on solution does not seem deterministic. On slower machine for an instance, it might not suffice.

I think inproc datacollector is actually the right approach here.

/cc: @pvlakshm @PBoraMSFT

NikolayPianikov:
@singhsarab It is hard coded so far. The suggested PR allows to specify a required timeout. The timeout will be large for our case (TeamCity CI). We are controlling a process tree and can finish processes if it is needed, for example when tests are hanging.

gighen76 (approved):

NikolayPianikov:
This PR does not change the current behavior by default, it just allows to override hard-coded constant by the environment variable `VSTEST_CLOSE_TIMEOUT `, for instance like as `VSTEST_CONNECTION_TIMEOUT`.

@mayankbansal018, @singhsarab what can I do to make this PR accepted?

NikolayPianikov:
@cltshivash what is the difference between `VSTEST CONNECTION_TIMEOUT` and `VSTEST_CLOSE_TIMEOUT`? Why one timeout should be hard-coded and another does not?

Why the current hard-coded timeout 100 ms is enough from your point fo view and
what if computer is too slow to have enough time to finish all operations (what is actually happening now) in normal way? May be 10 ms or 1000 ms would be better?

`The fix to increase the exit time is not deterministic and will need the customer to change the exit timeout.` What do you mean? By default the timeout is 100 ms as earlier and and here nothing has changed. But we would like to have an ability to set a value for instance like 1 minute, because we are controling the process tree and timeouts as I mentioned in a previous post. 

The inproc data collector scenario supposes that we should change user test settings file to specify our custom collector. It looks like workaround but not as basic usage scenario.

cltshivash:
@NikolayPianikov VSTEST_CONNECTION_TIMEOUT is used in the context of establishing the connection between the different processes involved. (vstest.console.exe, testhost.x86.exe, datacollector.exe). It is kept configurable as we have seen in cases where the profilers are involved, the startup time of the processes can be high. This has been observed on lower config machines used in pipeline pools.

The 100 ms timeout (default) is to provide the testhost process to gracefully exit instead of a termination from the runner. The testhost would have completed all the necessary work and is waiting for the final message to end and we don't expect any other processing to happen at this stage.

singhsarab:
@mayankbansal018 @cltshivash @NikolayPianikov 
I am closing the stale PRs, Should I go ahead close this one?

singhsarab:
Closing this. Please let me know if anyone has concerns.

Issue #411, opened 2017-01-31
add support for `dotnet test path/to/my.sln`


TheRealPiotrP:
As folks are spending more time with the new command line experience, the desire for a sln-level test command is growing.

Some desirable characteristics:
- parallelism! allow MSBuild's `/m` to control the number of projects to test in parallel
- Unified output. Provide a summary of how the overall test run did, not just individual projects

ErikSchierboom:
I would love to see this implemented! It will makes build scripts just a little bit shorter and easier.

bernardbr:
In my case, I solved it with a python script. I've tried to do it with a PowerShell script, but I got restriction error.

### Python solution:
```python
import subprocess
from argparse import ArgumentParser

parser = ArgumentParser()
parser.add_argument("-f", "--files", nargs="+", dest="files", help="Define a .csproj file list.")

args = parser.parse_args()

for f in args.files:
	subprocess.call(r"C:\Program Files\dotnet\dotnet.exe" +  " test " + f)
```

### PowerShell solution:
```powershell
param(
    [Parameter(Mandatory = $true, Position = 0)]
    [string[]]$csProjFiles
)    

foreach ($csProjFile in $csProjFiles)
{
    dotnet test $csProjFile
}
```

### My task.json
```javascript
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "0.1.0",
    "isShellCommand": true,
    "args": [],
    "tasks": [
        {
            "taskName": "build",
            "command": "dotnet",
            "args": [
                "build",
                "solution.sln"
            ],
            "isBuildCommand": true,
            "showOutput": "silent",
            "problemMatcher": "$msCompile"
        },
        {
            "taskName": "test",
            "command": "py",
            "args": [
                "${cwd}/run_tests.py",
                "-a",
                "${cwd}/test/unit/model/model.csproj",
                "${cwd}/test/unit/repository/repository.csproj"
            ],
            "isTestCommand": true,
            "showOutput": "silent"
        }
    ]
}
```

tlycken:
Unified output would be really nice, but could prove quite complex given the wide range of testing frameworks and how they present their output. However, there's another killer reason for me to want this: reduced test running time. It's quite easy to find the project files and pass each of them to `dotnet test`, but doing that triggers a new build for each project, meaning that in a larger solution with many test projects, I have to wait for many builds just to run the tests.

tverboon:
[...]

PR #1745, opened 2018-08-22
Run tests only for test projects

smadala:
## Description
- Avoid obvious failure for non test projects
 
## Related issue
https://github.com/Microsoft/vstest/issues/1129
https://github.com/Microsoft/vstest/issues/705
https://github.com/Microsoft/vstest/issues/411

### DIFF
--- a/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
+++ b/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
@@ -25,7 +25,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     Main entry point for running tests through vstest.console.exe
     ============================================================
   -->
-  <Target Name="VSTest" >
+  <Target Name="VSTest" Condition="'$(IsTestProject)' == 'true'">
     <CallTarget Condition="'$(VSTestNoBuild)' != 'true'" Targets="BuildProject" />
     
     <CallTarget Targets="ShowCallOfVSTestTaskWithParameter" />


dasMulli:
Btw I was protoyping a different idea where:

(note: currently broken for multi-targeting projects!!)
each project exhibits a target to collect test assemblies.
```xml
<Project>
  <Target Name="CollectBuildTestOutput"
    Condition="'$(TargetFramework)' != '' or '$(TargetPath)' != ''"
    DependsOnTargets="Build"
    Returns="@(TestAssembly)">
    <ItemGroup>
      <TestAssembly Include="$([System.IO.Path]::GetFullPath($(TargetPath)))" TargetFrameworkMoniker="$(TargetFrameworkMoniker)" />
    </ItemGroup>
  </Target>
  
  <Target Name="CollectInnerBuildTestOutput"
    Condition="'$(TargetFramework)' == '' and '$(TargetFrameworks)' != ''">
    <ItemGroup>
      <_InnerTestFramework Include="$(TargetFrameworks)" />
    </ItemGroup>
    <MSBuild Projects="$(MSBuildProject)" Targets="CollectBuildTestOutput" Properties="TargetFramework=%(_InnerTestFramework.Identity)">
      <Output TaskParameter="TargetOutputs" ItemName="TestAssembly" />
    </MSBuild>
  </Target>

  <Target Name="CollectTestBuildOutput"
     Condition="'$(IsTestProject)' == 'true'"
     DependsOnTargets="CollectBuildTestOutput;CollectInnerBuildTestOutput"
     Returns="@(TestAssembly)" />
</Project>
```

then the solution collects the assemblies and calls `dotnet vstest` (since the `VSTest` task only accepts a single assembly but this could be done using solution extensions targets that make an updated task available)
```xml
  <Target Name="VSTest">
    <MSBuild Projects="@(ProjectReference)" Targets="CollectTestBuildOutput">
      <Output TaskParameter="TargetOutputs" ItemName="TestAssembly" />
    </MSBuild>
    <Exec Command="dotnet vstest --Framework:%(TestAssembly.TargetFrameworkMoniker) @(TestAssembly,' ')"
    StandardOutputImportance="high" />
  </Target>
```

dasMulli:
(updated to fix it for multi-targeting projects)

I have no idea how this behaves with multiple test runners or different runners in the same solution (or different versions of xunit etc.).
I guess there is some logic to handle this in the VS "run all" experience already?
But currently this works quite well when run using `dotnet msbuild -restore -t:VSTest`

dasMulli:
(sry for spamming)

I'm fine with however this is resolved to make typing `dotnet test` work.
I see it as a dev experience and my tests are usually factored well enough to not have weird dependency issues.
Some users however prefer having a single trx output. (but I don't mind since VSTS can merge multiple results anyway.. need this for JS + .NET tests anyway)

smadala:
>I have no idea how this behaves with multiple test runners or different runners in the same solution (or different versions of xunit etc.)

Different test frameworks(NUnit, xUnit, MSTest V2,etc..) and version will be handled by vstest.console.

smadala:
I'm planning to address below problems in next few days. I will create an issue for each task very soon.



image

smadala:
```xml
<Target Name="CollectBuildTestOutput"
    Condition="'$(TargetFramework)' != '' or '$(TargetPath)' != ''"
    DependsOnTargets="Build"
    Returns="@(TestAssembly)">
    <ItemGroup>
      <TestAssembly Include="$([System.IO.Path]::GetFullPath($(TargetPath)))" TargetFrameworkMoniker="$(TargetFrameworkMoniker)" />
    </ItemGroup>
  </Target>
```

I'm just wondering can  we avoid build (`DependsOnTargets="Build"`) every time?

dasMulli:
I think there's a `NoBuild` property already, so `DependsOnTargets="Build"` => `DependsOnTargets="BuildForTestIfNecessary"`.
plus a `BuildForTestIfNecessary` target that conditions a depends on or a `CallTarget` on `Build` based on the property passed in.

Alternatively: Build the solution form the outer `VSTest` target and then remove the `DependsOnTargets` completely. The `TargetPath` property should be available regardless.
This will build non-test projects as well but is probably what users would expect.

singhsarab (approved):

smadala:
@dasMulli Sure 😄

gitfool:
@smadala when and how will this be available? Ideally it’s in the next drop of the 2.1.x SDK. Failing that, the next 2.2.x SDK preview. Finally, more out of curiosity, is there a way to drop in a nightly build of just MS Build Core?

Issue #1791, opened 2018-10-09
Make TRXLogger public


dotMorten:
## Description

In order for other test runners or frameworks to generate TRX reports, it would be useful if the TrxLogger class, it's default constructor and Initialize method were made public and published in a NuGet Package (with .NET Standard support).
I successfully referenced the TRX Logger source code myself and called these two methods, and was able to generate TRX Reports with just these two members.

smadala:
@dotMorten Are you planning to create PR for creating NuGet package?

dotMorten:
@smadala I could, if you can tell me where to start

smadala:
@dotMorten Nuspec file location is https://github.com/Microsoft/vstest/tree/master/src/package/nuspec .
After build.cmd `.\artifacts\Debug\packages` location contains nuget packages.

PR #1792, opened 2018-10-09
Allow external use of the TRX Logger

dotMorten:
## Description
Make the TRX Logger public, as described in #1791
Only constructor and initialize methods ara being exposed with this change.

## Related issue
#1791


smadala (approved):
Thanks for contribution 💯

PR #1806, opened 2018-10-19
[Revert] Allow external use of the TRX Logger

singhsarab:
Postponing the change to make TRX public as per discussion.

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -28,7 +28,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Extensions.TrxLogger
     /// </summary>
     [FriendlyName(TrxLoggerConstants.FriendlyName)]
     [ExtensionUri(TrxLoggerConstants.ExtensionUri)]
-    public class TrxLogger : ITestLoggerWithParameters
+    internal class TrxLogger : ITestLoggerWithParameters
     {
         #region Fields
 
@@ -185,7 +185,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Extensions.TrxLogger
         /// <param name="e">
         /// Event args
         /// </param>
-        internal void TestMessageHandler(object sender, TestRunMessageEventArgs e)
+        public void TestMessageHandler(object sender, TestRunMessageEventArgs e)
         {
             ValidateArg.NotNull<object>(sender, "sender");
             ValidateArg.NotNull<TestRunMessageEventArgs>(e, "e");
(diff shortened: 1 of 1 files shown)


singhsarab:
@mayankbansal018 Should we go ahead ?

smadala (dismissed):
> Postponing the change to make TRX public as per discussion.

@singhsarab Can you please elaborate this?

dotMorten:
I'd like to understand the reasoning too (not questioning it, but just want to understand the thinking so I can operate under the same understanding)

mayankbansal018:
@dotMorten , we were internally debating the cost of making TRX logger public, the key thing for us to consider was the fact, that we have to maintain backward compatibility so that the trx do not fail to open in older versions of VS.

Unfortunately the initial change that we took to make it public was not discussed, which caused this issue & we apologize for that.

We are in-fact glad that with your changes folks can generate trx from other test runners as well, two key things I would like to point out.

- TRX is a proprietary format and as a Microsoft policy, we generally maintain back compat, hence schema changes will not be accommodated, which by extension means that we cannot take any change to trx logger, which can cause this to happen.  
- If that constraint is alright, we will go ahead and accept it. If that is not acceptable, we should revert the change.

Given this constraint we were wondering how significant it would be to make trx logger public, w/o allowing changes in schema, which caused the delay. I hope you understand our concerns, & we can proceed safely to make TRX logger as public.

dotMorten:
I don't understand the schema argument. No changes where done that would change the schema, and as minimal as possible amount of members were exposed to just allow other test runners to reuse it to generate a report based on already published objects (ie testresult). No public APIs allow extending the output beyond what it already supports.

Trx might be a MSFT proprietary format, but it's also a defacto standard that so many 3rd party tools already support.

mayankbansal018:
@dotMorten , I agree so far no changes have been made that require changes in schema. 

My argument was not w.r.t to current changes, but from future looking perspective, where if we make TRX logger as public, then a valid ask can come which requires us to alter the schema, but unfortunately we won't be able to take such changes, as it will break the back compat, which we follow as a rule. Again I repeat, I'm talking this from current changes, I hope this clarifies things. 😃

dotMorten:
Not sure I get the argument. You should of course deny any PR/Feature Request that breaks existing customers., including schema changes. That's a given, but IMHO rather irrelevant to this issue.

dotMorten:
I'm assuming there's no problem with adding a v.next schema that customers could opt in to if you one day do want to expand the schema, and if that happens it would just be another overload for this constructor that enables that (or it would be an entirely new logger binary)

mayankbansal018:
@dotMorten I think we can agree to this. I'll close this PR, & we'll wait for you fix the other PR's you raised w.r.t to logger.

smadala:
@mayankbansal018 Thanks for detail explanation. It will be great if the future discussions also happens on github as this is **open source** project.

/cc @singhsarab @PBoraMSFT @cltshivash @codito @smalpani-msft

PR #1812, opened 2018-10-23
[CC]Adding Coverage.Interop.dll dependency for x64 in TPv2

jayaranigarg:
## Description
Adding Microsoft.VisualStudio.Coverage.Interop.dll dependency in TestPlatform Vsix to work for x64 scenarios.


mayankbansal018 (approved):

mayankbansal018 (approved):
As discussed, do modify the files count in nupkg,

PR #1839, opened 2018-11-14
Fix incorrect XML in error text.

vcsjones:
This addresses a small typo in the error text to contain an XML snippet that is easier to copy-and-paste.

Fixes #1838.

PR #1821, opened 2018-10-25
LOC CHECKIN | Microsoft/vstest master | 20181025

v-zbsail:

PR #1844, opened 2018-11-19
Fix invalid XML in resource string

martincostello:
When the `NoIsTestProjectProperty` resource string is displayed, the XML shown is not valid XML for copying and pasting into a project file. This fixes that by adding the missing `/` from the second tag.

Looks like this was partially fixed in #1839 but missed from #1821.


singhsarab (approved):

singhsarab:
Good catch  @martincostello

PR #1877, opened 2018-12-24
Fix - dotnet test on a multi-target projects logs only the last target

abhishkk:
## Related issue
https://github.com/Microsoft/vstest/issues/1603

## Description
Timestamp is appended to trx filename with millisecond precision. Thus for multi-target projects, trx for both targets will be stored.

Example: `dotnet test testproject.csproj --logger:"trx;logFileName=E:\results.trx"` will save trx file `E:\results_2018-12-24_14-01-07-176.trx` where `2018-12-24_14-01-07-176` is the timestamp.

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestRun.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestRun.cs
@@ -191,7 +191,7 @@ namespace Microsoft.TestPlatform.Extensions.TrxLogger.ObjectModel
         {
             // We use custom format string to make sure that runs are sorted in the same way on all intl machines.
             // This is both for directory names and for Data Warehouse.
-            return timeStamp.ToString("yyyy-MM-dd HH:mm:ss", DateTimeFormatInfo.InvariantInfo);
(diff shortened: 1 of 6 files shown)


cltshivash (approved):
:shipit:

singhsarab (approved):

stefanforsberg:
Hello! I'm using version "16.0.0-preview-20190203-03" which, if I interpret the roadmap doc correctly, should include this PR.

However, when I execute dotnet test --logger "trx;logFileName=C:\Path\File.trx" against a solution containing multiple test projects the result file is not appending the timestamp mentioned above and thus the result file is overwritten by the next test project. I also tried executing the command against a single csproj file with the same result (trx file without timestamp).

Any idea of what I could be doing wrong?

JeffLeFevre:
What if the timestamp is unwanted? I don't see an option to omit it for single-target projects. This breaks our CI workflow.

masterpoi:
Thanks for breaking all our ci workflows.

jnm2:
This change is has some problems.

- VSTest inventing directories that don't exist because the string.Replace is hitting all containing directory names, not just the file name.
- The filename is unsightly and there's no way to turn this off. 
- Our CI scripts have to guess at the resulting filename based on the directory listing or else direct VSTest to a temp folder and copy TRX files back out with proper filenames.

PR #1867, opened 2018-12-11
Do not show skip message always

singhsarab:
Show skip message only when the IsTestProject is not defined

Fixes #1866

PR #1882, opened 2018-12-30
use log in VSTestLogsTask

adamralph:
## Description

See https://github.com/Microsoft/vstest/pull/1867/files#r240614971

Before this change, messages for build start/completion, and a missing `IsTestProject` property, are printed to the console only, and regardless of the current verbosity level.

After this change, these messages are printed to all configured loggers, with the build start/completion messages only at normal verbosity or higher, and the `IsTestProject` property message at detailed verbosity or higher.

### DIFF
--- a/src/Microsoft.TestPlatform.Build/Tasks/VSTestLogsTask.cs
+++ b/src/Microsoft.TestPlatform.Build/Tasks/VSTestLogsTask.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.TestPlatform.Build.Tasks
 {
     using System;
+    using Microsoft.Build.Framework;
     using Microsoft.Build.Utilities;
     using Microsoft.TestPlatform.Build.Resources;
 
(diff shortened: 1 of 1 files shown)


singhsarab (approved):

adamralph:
@singhsarab done.

adamralph:
🎉 thanks for the merge!

Do you know which version of the .NET SDK this will be released in? There's no urgency, I'd just like to know when to expect it.

adamralph:
Which version will this be released in?

singhsarab:
@mayankbansal018 is tracking that. Mayank, let's expedite the release.

adamralph:
Any update? I don't necessarily want to expedite the release. I'd just like to know _which version_ it will be released in, so I can set my expectations accordingly.

mayankbansal018:
@adamralph we have already made a release of Testpaltform, please check the [releases](https://github.com/Microsoft/vstest/releases), this is available in latest pre-release.

adamralph:
Thanks @mayankbansal018.

FYI I can't find any mention of this PR in neither https://github.com/Microsoft/vstest/releases nor https://github.com/Microsoft/vstest-docs/blob/master/docs/releases.md

singhsarab:
@adamralph We might have missed added this to the release notes.
It was inserted into the dotnet cli via https://github.com/dotnet/cli/pull/10654

adamralph:
@singhsarab OK, but in any case, this will be released in 16.0.0, correct? Would it be reasonable to put it on the `16.0.0` milestone, so that anyone browsing the PR knows when to expect the change?

PR #1927, opened 2019-02-21
Adding ARM64 support to the test console

dotMorten:
## Description
I've found that I'm not able to run UWP ARM64 unit test on ARM64 devices.
Looking through the code, there's no indication ARM64 was ever supported, and the OS Architecture detection was getting the wrong values.
This is the beginnings of adding support for executing ARM64 unit tests.
I'm submitting as a draft for now, for feedback, and will also require some more testing at my end.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
@@ -34,6 +34,8 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
                         return ObjectModel.Architecture.X64;
                     case PlatformArchitecture.X86:
                         return ObjectModel.Architecture.X86;
+                    case PlatformArchitecture.ARM64:
+                        return ObjectModel.Architecture.ARM64;
                     default:
                         return ObjectModel.Architecture.ARM;
                 }
--- a/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
@@ -26,6 +26,29 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
         {
             get
             {
+#if !NETSTANDARD1_4
+                try
+                {
+                    // use the new IsWow64Process2 to detect ARM64 as well.
+                    var processHandle = System.Diagnostics.Process.GetCurrentProcess().Handle;
+                    if (IsWow64Process2(processHandle, out ImageFileMachine pProcessMachinem, out ImageFileMachine pNativeMachine))
+                    {
+                        switch (pNativeMachine)
+                        {
+                            case ImageFileMachine.AMD64:
+                            case ImageFileMachine.IA64:
+                                return Architecture.X64;
+                            case ImageFileMachine.X86:
+                                return Architecture.X86;
+                            case ImageFileMachine.ARM:
+                                return Architecture.ARM;
+                            case ImageFileMachine.ARM64:
+                                return Architecture.ARM64;
+                        }
+                    }
+                }
+                catch { }
+#endif
                 var arch = new PlatformEnvironment().Architecture;
 
                 switch (arch)
@@ -41,6 +64,18 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
                 }
             }
         }
+#if !NETSTANDARD1_4
+        private enum ImageFileMachine : ushort
+        {
+            X86 = 0x014c,
+            ARM = 0x01c0,
+            IA64 = 0x0200,
+            AMD64 = 0x8664,
+            ARM64 = 0xAA64
+        }
+        [System.Runtime.InteropServices.DllImport("kernel32")]
+        private static extern bool IsWow64Process2([System.Runtime.InteropServices.In] IntPtr processHandle, [System.Runtime.InteropServices.Out] out ImageFileMachine pProcessMachinem, [System.Runtime.InteropServices.Out] out ImageFileMachine pNativeMachine);
+#endif
 
         /// <summary>
         /// Gets the settings to be used while creating XmlReader for runsettings.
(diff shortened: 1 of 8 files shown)


src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
```diff
@@ -26,6 +26,29 @@ public static ObjectModel.Architecture OSArchitecture
         {
             get
             {
+#if !NETSTANDARD1_4
```
singhsarab: Can't this come from the PlatformEnvironment ?

dotMorten: No that enum doesn't have ARM63 in it

dotMorten: No due to backwards compat reasons this API normally lies about the OS architecture, so this is the way to do it.

ViktorHofer:
@dotMorten just because I'm curious, will you finish your work here?

src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
```diff
@@ -26,6 +26,29 @@ public static ObjectModel.Architecture OSArchitecture
...
+                try
+                {
+                    // use the new IsWow64Process2 to detect ARM64 as well.
+                    var processHandle = System.Diagnostics.Process.GetCurrentProcess().Handle;
+                    if (IsWow64Process2(processHandle, out ImageFileMachine pProcessMachinem, out ImageFileMachine pNativeMachine))
```
AntonLapounov, line 34: Is the `m` at the end of `pProcessMachinem` a typo?

src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
```diff
@@ -26,6 +26,29 @@ public static ObjectModel.Architecture OSArchitecture
...
+                    {
+                        switch (pNativeMachine)
+                        {
+                            case ImageFileMachine.AMD64:
+                            case ImageFileMachine.IA64:
+                                return Architecture.X64;
```
AntonLapounov, line 40: `IA64` (Itanium) should not be mapped to `X64`.

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -516,7 +516,11 @@ private static void VerifyCompatibilityWithOSArchitecture(Architecture architect
         {
             var osArchitecture = XmlRunSettingsUtilities.OSArchitecture;
 
-            if (architecture == Architecture.X86 && osArchitecture == Architecture.X64)
+            if (architecture == Architecture.X86 && (osArchitecture == Architecture.X64 || osArchitecture == Architecture.ARM64))
+            {
+                return;
+            }
+            if ((architecture == Architecture.ARM || architecture == Architecture.X86) && osArchitecture == Architecture.ARM64)
```
AntonLapounov, line 523: The combination `architecture == Architecture.X86 && osArchitecture == Architecture.ARM64` is checked twice. Using `IsWow64GuestMachineSupported` may be more appropriate here.

src/Microsoft.TestPlatform.ObjectModel/Utilities/XmlRunSettingsUtilities.cs
dotMorten: I submitted this PR almost a year and a half ago, so honestly I have no clue. To be honest, I don't even remember submitting this PR.
Feel free to take what I have and work with it, but with such a delay in response and in the mean time I found other means for testing on ARM64, I don't see myself picking this PR up.

nohwnd: I will most likely pick this up, when we get ARM64 hardware.

src/vstest.console/Resources/Resources.resx
```diff
@@ -336,7 +336,7 @@
     <value>The --ParentProcessId|/ParentProcessId argument requires the process id which is an integer. Specify the process id of the parent process that launched this process.</value>
   </data>
   <data name="InvalidPlatformType" xml:space="preserve">
-    <value>Invalid platform type:{0}. Valid platform types are x86, x64 and Arm.</value>
+    <value>Invalid platform type:{0}. Valid platform types are x86, x64, Arm and Arm64.</value>
```
danmosemsft, line 339: capitalize ARM and ARM64 in messages?

ViktorHofer:
@nohwnd I assume you will continue with this work? I guess we still want/need the additional supported architectures?

nohwnd:
@ViktorHofer yes, but plain tests seem to work just fine on ARM64 already with the custom path to dotnet or the PR that was merged this morning #2479 .

ViktorHofer:
Do you know the reason why VSTest isn't AnyCPU by default? Why do we need to pass in the architecture on Core at all?

nohwnd:
the runner (vstest.console) is anyCPU, but we have architecture specific hosts (testhost), to run the processes at the correct bittness that you select via the settings, to run your tests.

ViktorHofer:
I see, that makes sense. Still wondering why the testhost isn't part of the (platform specific) SDK but delivered via the nuget package.

nohwnd:
@ViktorHofer Because then it can be updated independently, so you don't need to fix your tests every time you install new dotnet sdk. Also you can get updates faster in case you are stuck with a given dotnet sdk.

Evangelink:
@dotMorten Thank you for the hard work and sorry it took so long! I will close this PR as we have been merging a couple of PRs providing support for ARM64 in the past months.

Feel free to test it out and let us know if we missed any case.

src/vstest.console/Resources/Resources.resx
dotMorten: Nope! This is the correct capitalization

AntonLapounov: I would use lower case only as in C# compiler usage help. In any case the two lists below should be consistent, and they are not (@nohwnd): https://github.com/microsoft/vstest/blob/b99950648b6cd0a34fcf83137e09492021814ff2/src/Microsoft.TestPlatform.ObjectModel/Architecture.cs#L11-L16 https://github.com/microsoft/vstest/blob/b99950648b6cd0a34fcf83137e09492021814ff2/src/vstest.console/Resources/Resources.resx#L452

PR #1946, opened 2019-03-07
Flow warnings to IDE

singhsarab:
## Description
Warnings should flow to IDE. Make sure all the warnings logged before the request is created are logged and sent to the IDE.


src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -391,13 +391,13 @@ private bool UpdateRunSettingsIfRequired(string runsettingsXml, List<string> sou
                     if (!string.IsNullOrEmpty(incompatibleSettingWarning))
                     {
                         EqtTrace.Info(incompatibleSettingWarning);
-                        ConsoleLogger.RaiseTestRunWarning(incompatibleSettingWarning);
+                        logger.LogWarning(incompatibleSettingWarning);
```
mayankbansal018: >logger [](start = 24, length = 6)

how will these flow to Console Logger?

src/Microsoft.TestPlatform.Client/DesignMode/IDesignModeClient.cs
```diff
@@ -37,5 +32,18 @@ public interface IDesignModeClient : IDisposable
...
+        /// </summary>
+        /// <param name="rawMessage"></param>
+        void SendRawMessage(string rawMessage);
+
+        /// <summary>
+        /// Send the test session messages to IDE
+        /// </summary>
+        /// /// <param name="level"></param>
+        /// <param name="message"></param>
+        void SendTestMessage(TestMessageLevel level, string message);
```
mayankbansal018: We would also need a change in Translation Layer, so that it knows a new message is coming in, & it needs to handle it?

singhsarab: It can already handle these messages. It is aware of TestMessages.
TestMessages flow as Raw messages since they are already serialized.

test/vstest.console.UnitTests/TestPlatformHelpers/TestRequestManagerTests.cs
```diff
@@ -876,6 +876,7 @@ public void RunTestsShouldNotThrowForFramework35()
 
             this.testRequestManager.RunTests(payload, mockCustomlauncher.Object, mockRunEventsRegistrar.Object, this.protocolConfig);
 
+            mockRunEventsRegistrar.Verify(lw => lw.LogWarning("Framework35 is not supported. For projects targeting .Net Framework 3.5, test will run in CLR 4.0 \"compatibility mode\"."), Times.Once);
```
abhishkk: we can add test for incompatibleSettingWarning as well.

abhishkk (approved):

Issue #1985, opened 2019-04-10
testhost.x86 is not large address aware


sharwell:
Currently testhost.x86 is compiled with platform set to x86:

https://github.com/Microsoft/vstest/blob/53c0a341b07c1f224a31b2378e696f535a1f6080/src/testhost.x86/testhost.x86.csproj#L12

This compilation setting causes the application to omit the large address aware setting, which causes problems on larger multi-core machines (I'm observing frequent OOM errors on 2990WX).

The correct setting for an x86 build which is large address aware is a combination of the AnyCPU platform and the Prefer32Bit setting. See Microsoft/perfview@1582d04aefd7869a1ea6e1530c49ed2e6a1c5939 for an example of this configuration.

AbhitejJohn:
@sharwell : Just trying to understand the affect of moving to LAA, how would this affect low config systems?

ivonin:
From what I read, setting the performance hit from setting the /LAA flag exists although it's not too big.

Can't find anything more thourough than [this StackOverflow post](https://stackoverflow.com/a/26001862/501518) at the moment. @sharwell what's your take on this?

sharwell:
I've never heard of there being a difference. I can't benchmark the difference in a meaninful way because the current release just crashes.

PR #1987, opened 2019-04-10
Enabled gcServer and disabled gcConcurrent for test runners

### LINKED ISSUE #1967: Setting ServerGC for the test runner
## Description
VSTest.console.exe is currently configured to run in workstation GC (also concurrent). There is currently no way to change this behavior other than patching the `vstest.console.exe.config` file in the VS installation folder. This solution is obviously not pleasant.

We have a big test assembly (11.5k tests) and it makes a *big* difference: the tests take 9 minutes to run with `ServerGC=off` and 4.5 minutes with `ServerGC=on`.

## Expected behavior
1. It's probably not a bad thing to make `ServerGC=on` by default (MSTest v1 was using server GC, see `QTAgent32.exe.Config`)
2. If not, it would be nice to set the behavior via the `.runsettings` file.

ivonin:
## Description
Changed the GC settings to match the settings for MS Test v1 runners (QTAgent.exe)

## Related issue
 Fixes #1967

### DIFF
--- a/src/testhost.x86/app.config
+++ b/src/testhost.x86/app.config
@@ -4,6 +4,8 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
   </startup>
   <runtime>
+    <gcServer enabled="true" />
+    <gcConcurrent enabled="false"/>
     <legacyUnhandledExceptionPolicy enabled="1"/>
     <legacyCorruptedStateExceptionsPolicy enabled="true"/>
 
--- a/src/testhost/app.config
+++ b/src/testhost/app.config
@@ -4,6 +4,8 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
   </startup>
   <runtime>
+    <gcServer enabled="true" />
+    <gcConcurrent enabled="false"/>
     <legacyUnhandledExceptionPolicy enabled="1" />
     <legacyCorruptedStateExceptionsPolicy enabled="true" />
 
--- a/src/vstest.console/app.config
+++ b/src/vstest.console/app.config
@@ -4,6 +4,8 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
   </startup>
   <runtime>
+    <gcServer enabled="true" />
+    <gcConcurrent enabled="false"/>
     <legacyUnhandledExceptionPolicy enabled="1" />
 
     <!-- To get stacktrace information for portable and embedded pdbs when net472 installed on machine.


ivonin:
The check is failed with some acceptance tests broken, but I'm not sure if my changes could affect any of it... The tests have bad history and are failing randomly it seems?

sharwell:
Locally I observed `gcServer` for testhost.x86 reducing test run time significantly. Here's a message I sent on the subject internally:

> The only difference between these two test runs is I turned on server GC for testhost.x86.exe:
>
> ```
> [4/2/2019 7:18:43 AM Informational] ------ Run test started ------
> [4/2/2019 7:18:46 AM Informational] [xUnit.net 00:00:01.8254068]   Discovering: StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:18:48 AM Informational] [xUnit.net 00:00:03.9438646]   Discovered:  StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:18:48 AM Informational] [xUnit.net 00:00:03.9871306]   Starting:    StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:21:12 AM Informational] [xUnit.net 00:02:27.5838838]   Finished:    StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:21:13 AM Informational] ========== Run test finished: 6446 run (0:02:29.701225) ==========
> [4/2/2019 7:21:52 AM Informational] ------ Run test started ------
> [4/2/2019 7:21:55 AM Informational] [xUnit.net 00:00:01.6501179]   Discovering: StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:21:57 AM Informational] [xUnit.net 00:00:03.8203892]   Discovered:  StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:21:57 AM Informational] [xUnit.net 00:00:03.8693271]   Starting:    StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:22:33 AM Informational] [xUnit.net 00:00:40.0539226]   Finished:    StyleCop.Analyzers.Test.CSharp7
> [4/2/2019 7:22:33 AM Informational] ========== Run test finished: 6446 run (0:00:41.4295855) ==========
> ```
> 
> Sam

I have not reviewed the scenario with gcConcurrent set.

sharwell:
⚠️ After [locally setting this]() I started to see frequent OutOfMemoryException failures in x86 test runs, which appear to be caused by the substantial increase in memory required to use Server GC on many cores (2990WX). It will be important to address #1985 at the same time.

ivonin:
GCconcurrent=true makes performance _slightly_ worse on our local test stand: 4 minutes 34 seconds median duration vs 4 minutes 28 seconds with GCconcurrent=false (this is the median from 10 runs).

`QTAgent.exe` has GCconcurrent=false, so combined with my benchmark it seems to be the right call.

ivonin:
Some tests are indeed failing with my changes... I'll look into it shortly, but help will be appreciated

singhsarab:
Closing stale PRs please re-open if required.

sharwell:
I have not seen OOM errors since #1985 was addressed, even with gcServer enabled.

yar-shukan:
@sharwell , @ivonin , can this be reopened?

I think gcServer should be either default or there should be the ability to configure that through .runsettings file (I guess this is more complicated).

Considering the build agent case where it's not trivial (or even impossible due to permissions) to modify the shared vsconsole configuration I think enabling or giving the ability to configure `gcServer` mode has to be reconsidered again.

PR #1996, opened 2019-04-26
[Revert] Fix for dotnet test on a multi-target projects logs only the last target 

singhsarab:
Revert "Fix - dotnet test on a multi-target projects logs only the last target (#1877)"

This reverts commit 37696768786ef1018f0b4499eab9ca27af141057.

## Description
Reverting the above fix because it came out as a breaking change for a lot of customers.
We will be revisiting this soon with a proper fix and RFCs

This will be a take back for https://github.com/Microsoft/vstest/issues/1603
(Timestamp is appended to trx filename with millisecond precision. Thus for multi-target projects, trx for both targets will be stored.)


cltshivash (approved):

mayankbansal018 (approved):

singhsarab:
cc: @SidShetye

PR #2007, opened 2019-05-05
Do not show TPv2 vsix in Extensions and Updates.

singhsarab:
## Description
Making the TPv2 vsix a system component so that it cannot be disabled by the user.

## Issue
https://developercommunity.visualstudio.com/content/problem/531135/test-explorer-fails-with-exception.html


mayankbansal018 (commented):
why is this needed?

singhsarab:
@mayankbansal018 Updated the description.

cltshivash (approved):

mayankbansal018 (approved):

Issue #1992, opened 2019-04-20
Trx TestDefinition classname


cheng93:
## Description
> **IMPORTANT**: if the defect is reproduced only in a workflow from within the Visual Studio IDE then do not report the issue here - instead, please report it using Visual Studio's "Send Feedback" option that can be accessed from the Help menu OR using this link https://developercommunity.visualstudio.com.
>
> For a defect reproducable from the vstest command line, describe the issue you've observed.

The classname under test definition is wrong.

For example

```

    <UnitTest name="XunitTrx.FactTests.Test" storage="/home/xunittrx/xunittrx/bin/debug/netcoreapp2.2/xunittrx.dll" id="add8c0e4-fdbc-6929-f132-2c65282e85bb">
      <Execution id="cb57a616-fd2e-49c7-9da1-00f6109a1849" />
      <TestMethod codeBase="/home/XunitTrx/XunitTrx/bin/Debug/netcoreapp2.2/XunitTrx.dll" adapterTypeName="executor://xunit/VsTestRunner2/netcoreapp" className="XunitTrx" name="XunitTrx.FactTests.Test" />
    </UnitTest>
```

the classname is `XunitTrx` which is **wrong**

## Steps to reproduce
> What steps can reproduce the defect?

Instructions are here https://github.com/cheng93/XunitTrx

Example output is in the `TestResults` folder

> Please share the setup, commandline for vstest.console, sample project, target
> framework etc.

|docker tag|vstest|
|-|-|
|2.2|16.0.1|
|2.2.105|15.9.0|

## Expected behavior
> Share the expected output

```
    <UnitTest name="XunitTrx.FactTests.Test" storage="/home/xunittrx/xunittrx/bin/debug/netcoreapp2.2/xunittrx.dll" id="add8c0e4-fdbc-6929-f132-2c65282e85bb">
      <Execution id="cb57a616-fd2e-49c7-9da1-00f6109a1849" />
      <TestMethod codeBase="/home/XunitTrx/XunitTrx/bin/Debug/netcoreapp2.2/XunitTrx.dll" adapterTypeName="executor://xunit/VsTestRunner2/netcoreapp" className="XunitTrx.FactTests" name="XunitTrx.FactTests.Test" />
    </UnitTest>
```

the classname should be `XunitTrx.FactTests`

## Actual behavior
> What is the behavior observed?

See in **Description**

## Diagnostic logs
[...]

PR #2014, opened 2019-05-10
Trx ClassName fix

singhsarab:
## Description
Fix for the trx classname being wrongly stamped when testname and fullyqualifiedname are same.

## Related issue
https://github.com/microsoft/vstest/issues/1992


mayankbansal018 (approved):
:shipit:

PR #2023, opened 2019-05-23
[Draft] [Design mode] Invoking vstest.console with environment variables

singhsarab:
## Description
Changes to allow clients to provide environment variable while initializing VsTestConsoleWrapper

### DIFF
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
@@ -4,7 +4,7 @@
 namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer
 {
     using System;
-    using System.Collections.Specialized;
+    using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
 
@@ -44,7 +44,7 @@ namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer
         /// TODO: Remove the #if when project is targeted to netstandard2.0
         /// Environment variables to be set for the process
         /// </summary>
-        public StringDictionary EnvironmentVariables { get; set; } = new StringDictionary();
+        public Dictionary<string, string> EnvironmentVariables { get; set; }
 
 #endif
 
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer
 {
     using System;
+    using System.Collections.Specialized;
     using System.Diagnostics;
     using System.IO;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -106,6 +107,13 @@ public void StartProcess(ConsoleParameters consoleParameters)
 
             EqtTrace.Verbose("VsTestCommandLineWrapper: Process Start Info {0} {1}", info.FileName, info.Arguments);
 
+#if NET451
+            info.EnvironmentVariables.Clear();
```
AbhitejJohn: Should we only clear if environment variables are supplied? Might affect other clients otherwise.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
```diff
@@ -37,6 +38,16 @@ public ConsoleParameters(IFileHelper fileHelper)
             this.fileHelper = fileHelper;
         }
 
+#if NET451
+
+        /// <summary>
+        /// TODO: Remove the #if when project is targeted to netstandard2.0
+        /// Environment variables to be set for the process
+        /// </summary>
+        public StringDictionary EnvironmentVariables { get; set; } = new StringDictionary();
```
AbhitejJohn: Should this just be IDictionary<string, string> instead? Looks like StringDictionary is pre generics from [here](https://stackoverflow.com/questions/627716/stringdictionary-vs-dictionarystring-string).

AbhitejJohn (approved):

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
```diff
@@ -37,6 +38,16 @@ public ConsoleParameters(IFileHelper fileHelper)
             this.fileHelper = fileHelper;
         }
 
+#if NET451
```
acesiddhu: why do we need this check

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
acesiddhu: agree

---
In reply to: [287176793](https://github.com/microsoft/vstest/pull/2023#discussion_r287176793) [](ancestors = 287176793)

acesiddhu: >StringDictionary [](start = 15, length = 16)

new clients who take dependency on this property when they try to run with older version of TP will fail. is that a scenario?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -106,6 +107,13 @@ public void StartProcess(ConsoleParameters consoleParameters)
...
+            foreach (DictionaryEntry envVariable in consoleParameters.EnvironmentVariables)
+            {
+                info.EnvironmentVariables.Add(envVariable.Key.ToString(), envVariable.Value.ToString());
```
acesiddhu: >envVariable.Value [](start = 74, length = 17)

u r assuming client will give non null values. handle that here

acesiddhu (dismissed):

mayankbansal018:
@singhsarab can you please share the issues. Also I'm not sure what is the requirement here, to start vstest.console under certain Env vars, or the goal is to move them to testhost.exe, & we are using vstest.console as intermediate to do so. Can you please confirm

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
cltshivash: Yes. That is a possibility if the translation layer isn't bundled.

cltshivash (approved):

PR #2028, opened 2019-05-27
Add netstandard20 dependencies to the TestHost pkg

### LINKED ISSUE #2027: Microsoft.TestPlatform.ObjectModel add netstandard2.0 configuration
The Microsoft.TestPlatform.Object package should add the `netstandard2.0` configuration when multi-targeting to minimize the dependency graph.

ViktorHofer:
## Description
To reduce the dependency graph when restoring with .NET Core >= 2.0, add a .NET Standard 2.0 configuration to `Microsoft.TestPlatform.ObjectModel` and reference a newer package (prerelease but considered as stable) of `Microsoft.Extensions.DependencyModel` with a .NET Standard 2.0 configuration.

CoreFx currently can't use the Microsoft.NET.Test.Sdk because of dependencies that are copied to the output directory. All framework references that come from netstandard < 2.0 are creating conflicts with the live built framework assets in corefx. We really want to enable Test Explorer integration in CoreFx + simple F5 debugging.

## Related issue
Fixes https://github.com/microsoft/vstest/issues/2027

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -28,7 +28,7 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <!-- .NetStandardLibrary1.6.0 brings in dependencies which are supported in UWP,
          any higher version will bring in dependencies which are not supported by current UWP version -->
-    <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
+    <NetStandardImplicitPackageVersion Condition="'$(OverrideImplicitNSVersion)' != 'true'">1.6.0</NetStandardImplicitPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
--- a/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
+++ b/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
@@ -1,13 +1,14 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <OverrideImplicitNSVersion Condition=" '$(TargetFramework)' == 'netstandard2.0' ">true</OverrideImplicitNSVersion>
     <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
   </PropertyGroup>
   <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
   <PropertyGroup>
     <AssemblyName>Microsoft.VisualStudio.TestPlatform.ObjectModel</AssemblyName>
-    <TargetFrameworks>net451;netstandard1.4</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(DotNetBuildFromSource)' == 'true' ">netstandard1.4</TargetFrameworks>
+    <TargetFrameworks>net451;netstandard1.4;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(DotNetBuildFromSource)' == 'true' ">netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.TestPlatform.ObjectModel</PackageId>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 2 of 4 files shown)


ViktorHofer:
ping @singhsarab

singhsarab:
@ViktorHofer There are a couple of failures because this change. Can you please take a look.

scripts/build/TestPlatform.Settings.targets
```diff
@@ -28,7 +28,7 @@
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <!-- .NetStandardLibrary1.6.0 brings in dependencies which are supported in UWP,
          any higher version will bring in dependencies which are not supported by current UWP version -->
-    <NetStandardImplicitPackageVersion>1.6.0</NetStandardImplicitPackageVersion>
+    <NetStandardImplicitPackageVersion Condition="'$(OverrideImplicitNSVersion)' != 'true'">1.6.0</NetStandardImplicitPackageVersion>
```
mayankbansal018: how is OverrideImplicitNSVersion set?

src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
```diff
@@ -1,13 +1,14 @@
...
+    <OverrideImplicitNSVersion Condition=" '$(TargetFramework)' == 'netstandard2.0' ">true</OverrideImplicitNSVersion>
     <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
   </PropertyGroup>
   <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
   <PropertyGroup>
     <AssemblyName>Microsoft.VisualStudio.TestPlatform.ObjectModel</AssemblyName>
-    <TargetFrameworks>net451;netstandard1.4</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(DotNetBuildFromSource)' == 'true' ">netstandard1.4</TargetFrameworks>
+    <TargetFrameworks>net451;netstandard1.4;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(DotNetBuildFromSource)' == 'true' ">netstandard2.0</TargetFrameworks>
```
mayankbansal018: why is it changes only for one project? Also I don't get the need to upgrade it to 2.0 when NS1.4 will work? Is it because of below dependency of "System.Runtime.InteropServices.RuntimeInformation"

src/package/nuspec/TestPlatform.TestHost.nuspec
```diff
@@ -17,7 +17,7 @@
       <group targetFramework="netcoreapp1.0">
         <dependency id="Microsoft.TestPlatform.ObjectModel" version="$Version$"/>
         <dependency id="Newtonsoft.Json" version="$JsonNetVersion$"/>
-        <dependency id="Microsoft.Extensions.DependencyModel" version="1.0.3"/>
+        <dependency id="Microsoft.Extensions.DependencyModel" version="3.0.0-preview5-27626-15"/>
```
mayankbansal018: why is this needed?

src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
ViktorHofer: The reason for that is that netstandard < 2.0 brings framework assemblies with it which results in them being copied to the output directory. With >= netstandard2.0 we changed that so that nothing will be brought and copied over. We always offer both the minimum possible NS version + a NS2.0 version for our packages in .NET Core. Upgrading both the ObjectModel package and using a newer version of the DependencyModel package results in a clean output graph and we can finally use it in CoreFx.

As we produce the framework assemblies ourselves we would hit conflicts if Microsoft.NET.Test.Sdk brings their own with them. The remaining issue is Newtonsoft.Json of course but I think there is a workaround on our side for that.

ViktorHofer:
cc @tmat I'm working around the clashing framework assets with this. For Newtonsoft.Json I will packagereference a newer version with NS2.0 support in CoreFx.

src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
tmat: I don't think we need this `DotNetBuildFromSource` conditional override. Build from source can build any TFM.

mayankbansal018:
@ViktorHofer lets have a quick discussion over it, I think these changes alone won't help resolve the bug, but we might also need to repackage the correct dependencies in our nugets ?
Can you please set up a call b/w us for next week?

ViktorHofer:
I agree, the correct fix would be to add a netstandard2.0 configuration to all shipped assemblies, repackage the content with correct tfms and only reference netstandard2.0 dependencies. I understand that this isn't possible for Newtonsoft.Json as testhost dependencies aren't isolated currently from the test app. But that is fine for corefx at least. Long term there should be an isolation.

ViktorHofer:
Sounds good, will send you an invite.

ViktorHofer:
Closing as discussed offline, there might be more work necessary.

PR #2060, opened 2019-06-21
Exclude tests from source-build

adaggarwal:
Exclude tests from source-build as they are note required. 

@ViktorHofer

### DIFF
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -1,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
+  </PropertyGroup>
+</Project>


mayankbansal018:
@adaggarwal we are already not building test projects when `ExcludeFromSourceBuild` is set to true, why is this additionally needed? When sourcebuild is set to true our build scripts uses `TestPlatform_BuildFromSource.sln`, which does not have reference to test projects.

ViktorHofer:
what @mayankbansal018 said makes sense. Also I don't have write permissions here and probably not enough context to help.

adaggarwal:
@mayankbansal018 can you point me to the file where ```ExcludeFromSourceBuild``` is set to true?

mayankbansal018:
[TPB_Build_From_Source_Solution ](https://github.com/microsoft/vstest/blob/fb3545cf67530577c7bc6d3340c7f92057e07f11/scripts/build.sh#L121) is used when [DOTNET_BUILD_FROM_SOURCE](https://github.com/microsoft/vstest/blob/fb3545cf67530577c7bc6d3340c7f92057e07f11/scripts/build.sh#L70) is set when building via testplatform's script.

adaggarwal:
@mayankbansal018 I get the point. Thanks for the details. I see that ```JsonNetVersion``` is being used to set newtonsoft.json version. Might I suggest using ```NewtonsoftJsonVersion``` so that vstest can utilize the sourcebuild version. And other thing could be, having a way to pass this newtonsoft version in the build script:

https://github.com/microsoft/vstest/blob/fb3545cf67530577c7bc6d3340c7f92057e07f11/scripts/build.ps1#L375-L379

I have updated [896](https://github.com/dotnet/source-build/issues/896) to include newtonsoftJson prebuilt version.

PR #2066, opened 2019-06-25
Change the error message to display name instead of version

Cosifne:
The current incompatible message uses the version number as an error message. So it could give a confusing message like:
ProjectA is built for framework 2.1 and platform Any CPU. But the project could be built for .NET Core 2.1.

This message is quite confusing since it uses the word 'framework' so the user may think it is a strange version of .NET Framework


AbhitejJohn (approved):

singhsarab:
This is already fixed via https://github.com/microsoft/vstest/pull/2059/files

Cosifne:
> 
> 
> This is already fixed via https://github.com/microsoft/vstest/pull/2059/files

Oh, it is nice to know that.

PR #2067, opened 2019-06-26
VSTest targeting netstandard 2.0

vagisha-nidhi:
Related issue: [https://github.com/microsoft/vstest/issues/2073](https://github.com/microsoft/vstest/issues/2073)

PR #2074, opened 2019-06-26
Add netstandard2 target to object model and lower

### LINKED ISSUE #2073: TestPlatform ObjectModel has bad dependencies
The TestPlatform ObjectModel has old/bad dependencies. It causes lots of package downgrade warnings.

image

It should reference .NET Standard Library 1.6 and not the other packages. There should also be a netstandard2 target that reduces those all-together.

clairernovotny:
## Description
This adds a .NET Standard 2.0 build target to ObjectModel, CoreUtilities and PlatformAbstractions. This is necessary to prevent package downgrade warnings when using them on newer platforms.

## Related issue
Fixes #2073


singhsarab:
This will be take care of via https://github.com/microsoft/vstest/pull/2067.

@onovotny Can you please review the above change.

Issue #149, opened 2016-10-20
`VsTestConsoleWrapper.CancelDiscovery` is not implemented


genlu:
It'd be nice to be able to cancel test discovery when LUT is paused or discovery hangs due to bugs in test platform/test adapter/test framework.

Also, behavioral wise, I think it'd make sense to let `CancelDiscovery`  to immediately abort current discovery, kill testhost process and cause the call to DiscoverTests to return (with i`sAborted`set to true). @saikrishnav Does this sound reasonable?

pvlakshm:
Closing this; reactivate if still required.

AbhitejJohn:
I've gone ahead and reopened this issue. It would be really great to have a form of discovery cancellation where other high priority operations need to supersede a currently running potentially long discovery operation. For instance a build/rebuild/clean operation or a solution close. 

From an implementation standpoint if I understand this correctly it could probably just be killing the testhost process.

ManishJayaswal:
@cltshivash - any update on this issue?

cltshivash:
@ManishJayaswal The changes are in and we should be inserting the TP version with the fix for 16.3 this week.

ManishJayaswal:
@cltshivash great to hear this!

AbhitejJohn:
@cltshivash : I just tried this out on a solution with 100,000 MSTest tests -> Closed solution while discovery is still in progress but that did not stop the operation. It looks like we bail out from discovery between adapter calls but if an adapter is still servicing a request, there isn't a way of hard-stopping it. That's probably a change in the adapter interface but until that is fixed, we could probably use a testhost process kill similar to AbortTestRun. That should fix the scenario above for us.

singhsarab:
This should be resolved by https://github.com/microsoft/vstest/pull/2076 and https://github.com/microsoft/vstest/pull/2134

PR #2076, opened 2019-06-27
Implemented the cancellation of discovery request

cltshivash:
## Description
Implemented the changes to cancel a discovery request through the design mode client

### DIFF
--- a/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
+++ b/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
@@ -196,6 +196,12 @@ namespace Microsoft.VisualStudio.TestPlatform.Client.DesignMode
                                 break;
                             }
 
+                        case MessageType.CancelDiscovery:
+                            {
+                                testRequestManager.CancelDiscovery();
+                                break;
+                            }
+
                         case MessageType.CancelTestRun:
                             {
                                 testRequestManager.CancelTestRun();
--- a/test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
+++ b/test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
@@ -3,15 +3,17 @@
 
 namespace Microsoft.TestPlatform.AcceptanceTests.TranslationLayerTests
 {
-    using Microsoft.TestPlatform.TestUtilities;
-    using Microsoft.TestPlatform.VsTestConsole.TranslationLayer.Interfaces;
-    using Microsoft.VisualStudio.TestPlatform.Common.Telemetry;
-    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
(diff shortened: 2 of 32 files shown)


src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -196,6 +196,12 @@ private void ProcessRequests(ITestRequestManager testRequestManager)
                                 break;
                             }
 
+                        case MessageType.CancelDiscovery:
```
singhsarab: We'll need to add acceptance tests for this.

cltshivash: You mean targeting at the vstestconsole wrapper layer ?

singhsarab: Yes.

AbhitejJohn:
@cltshivash : Would be great to have this in whenever you think its ready.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
cltshivash: Added the acceptance test

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
```diff
@@ -169,6 +171,42 @@ public void DiscoverTestsUsingSourceNavigation(RunnerInfo runnerInfo)
...
+            discoveryEvents.Setup((events) => events.HandleDiscoveredTests(It.IsAny<IEnumerable<TestCase>>())).Callback
+                ((IEnumerable<TestCase> testcases) => { discoveredTests.AddRange(testcases); });
+
+            // Act
+            var discoveryTask = Task.Run(() =>
+            {
+                this.vstestConsoleWrapper.DiscoverTests(testAssemblies, this.GetDefaultRunSettings(), discoveryEvents.Object);
+            });
+
+            Task.Delay(2000).Wait();
```
singhsarab: We should not have this delay. This is possibly one of the scenarios. When the user immediately cancels.

In case of run, I recall a scenario where there was a race between the actual run request and cancel request.

singhsarab (approved):
Approving with suggestion

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
cltshivash: The delay is for the discovertest to actually start. It isn't an async operation. It would be good to cover the race condition through UTs and not through acceptance test as it is never going to validate the scenario in a deterministic

AbhitejJohn:
@singhsarab , @cltshivash : Wondering if we could update the test platform insertion on the VS side to include this change.

cltshivash:
Addresses issue #149

Issue #2126, opened 2019-08-09
Test Console fails if path contains forward slash


tannergooding:
## Description

When using the latest SDK, you get an error if your `VSTestResultsDirectory` contains a forward slash (on Windows).

```
Microsoft (R) Test Execution Command Line Tool Version 16.3.0-preview-20190808-04
Copyright (c) Microsoft Corporation.  All rights reserved.

The path 'C:\Users\tagoo\repos\clangsharp\artifacts/tst/Debug/' specified in the 'ResultsDirectory' is invalid. Error: Illegal characters in path.
```

## Steps to reproduce

1. Clone https://github.com/microsoft/clangsharp
2. Run .\scripts\cibuild.cmd

## Expected behavior

Everything succeeds (as with previous versions)

## Actual behavior

The Test Execution Command Line Tool fails with the above diagnostics.

tannergooding:
This does not repro on preview 7 (3.0.100-preview7-012821)

ShreyasRmsft:
@tannergooding i took a look a have an idea what is causing the issue.

https://github.com/microsoft/vstest/blob/03cb97c1e7a57dfed561a872e5797110524a1d70/src/vstest.console/Processors/ResultsDirectoryArgumentProcessor.cs#L178

this methods "IsValidPath"

is first splitting the given path using Path.DirectorySeparatorChar which returns '\' for windows.

Then for each folder (results of the split) we are checking if it contains any of Path.GetInvalidFileNameChars one of which is a '/'

The fix would be to normalize (replace '/' and '\' with ) the filepath before calling isvalid on it

Thank you for reporting this. We would be delighted if you could contribute this fix. And i will provide any help needed. Otherwise i will pick this up sometime soon. Thanks again.

tannergooding:
@ShreyasRmsft, why does the path need to be validated at all? Why aren't we just passing the path (as given) down to the right APIs and allowing the file operation to fail if it determines it should?

There are a range of characters that may or may not be valid based on where the destination exists and the file system it falls under. For example, it could be a network path in which case normalizing `/` or `\` would be incorrect. You could be targeting a file system (say FAT16, FAT32, or a Unix file system) which has a different set of allowed characters than NTFS. You could be targeting a virtual file system, etc

@JeremyKuhne, @carlossanlop. Could you possibly comment on "best practice" here, with regards to taking in user specified file paths and having them "just work" for cross-platform code?

ShreyasRmsft:
[...]

PR #2133, opened 2019-08-19
Test Console fails if path contains forward slash

vagisha-nidhi:
## Description
Removing pre validation checks for results directory.
Instead, creating the results directory if possible.

## Related issue
#2126

### DIFF
--- a/src/vstest.console/Processors/ResultsDirectoryArgumentProcessor.cs
+++ b/src/vstest.console/Processors/ResultsDirectoryArgumentProcessor.cs
@@ -139,20 +139,14 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Processors
 
             try
             {
-                // to check valid directory path
-                if (!string.IsNullOrEmpty(argument) && !IsValidPath(argument) )
-                {
-                    throw new ArgumentException("Illegal characters in path.");
-                }
-
-                var di = new DirectoryInfo(argument);
-
                 if (!Path.IsPathRooted(argument))
                 {
                     argument = Path.GetFullPath(argument);
                 }
+
+                var di = Directory.CreateDirectory(argument);
             }
-            catch (Exception ex) when( ex is NotSupportedException || ex is SecurityException || ex is ArgumentException || ex is PathTooLongException)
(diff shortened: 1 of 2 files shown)


tannergooding (approved):

tannergooding:
CC. @JeremyKuhne

src/vstest.console/Processors/ResultsDirectoryArgumentProcessor.cs
```diff
@@ -139,20 +139,14 @@ public void Initialize(string argument)
...
-                }
-
-                var di = new DirectoryInfo(argument);
-
                 if (!Path.IsPathRooted(argument))
                 {
                     argument = Path.GetFullPath(argument);
                 }
+
+                var di = Directory.CreateDirectory(argument);
```
tannergooding: Just as an FYI.

This will only validate that you have access to this path "at this moment" in time. It could fail here and still be valid later (some step that runs after this check, but before results are actually written could make the path available).

It could also succeed here and fail later (a network share or path could exist when checked and go down before the results are actually written).

src/vstest.console/Processors/ResultsDirectoryArgumentProcessor.cs
```diff
@@ -139,20 +139,14 @@ public void Initialize(string argument)
...
             {
-                // to check valid directory path
-                if (!string.IsNullOrEmpty(argument) && !IsValidPath(argument) )
-                {
-                    throw new ArgumentException("Illegal characters in path.");
-                }
-
-                var di = new DirectoryInfo(argument);
-
                 if (!Path.IsPathRooted(argument))
```
JeremyKuhne: `Path.IsPathRooted` will tell you if the path is fixed to a drive, not if it is relative or not. `D:foo.txt` is relative to wherever the working directory is. `Path.IsPathFullyQualified()` was added in core to tell you if the location would change based on the current working directory. Fully qualified, however, doesn't mean normalized, so if you're doing any manipulation on the path you should always call `Path.GetFullPath()`.

In this case it is pointless to call `Path.GetFullPath()` as `Directory.CreateDirectory()` will do that anyway.

JeremyKuhne (changes requested):
The `IsPathRooted()` block should be removed.

JeremyKuhne:
cc: @carlossanlop

singhsarab (approved):

Issue #2138, opened 2019-08-21
Allow sources to be parallelized across processes and controlled by adapters


NGloreous:
## Description
Today vstest doesn't split tests within a dll across multiple processes (e.g. if I only have 1 dll with thousands of tests they will all run sequentially in one process, assuming the adapter doesn't support parallelization or the tests can't be parallelized in the same process).

Ideally control of how tests within a dll are split across processes is determined by the adapter to allow consumers to provide custom logic. 

## Context
My team is trying to migrate off a custom built test harness to vstest/mstest but we have built up logic within our harness to effectively stripe/parallelize our tests across multiple machines and processes. This allows up to load balance our tests based on runtime to minimize the overall test pass time. Losing the ability to load balance tests within a dll will greatly increase our overall test pass time.

nohwnd:
This is a great idea but we won't be implementing it into TestPlatform any time soon, sorry.

PR #2139, opened 2019-08-21
Allow sources to be parallelized across processes and controlled by adapters

NGloreous:
Wanted to get feedback on the general idea. I can clean up and add unit tests if this is something vstest is open to.

## Description
Add Bucketize command to allow adapters to bucketize tests within a source and parallelize across buckets.

## Related issue
Add support for #2138

### DIFF
--- a/src/vstest.console/Processors/RunSpecificTestsArgumentProcessor.cs
+++ b/src/vstest.console/Processors/RunSpecificTestsArgumentProcessor.cs
@@ -88,6 +88,11 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Processors
         #region Fields
 
         /// <summary>
+        /// Indicates if bucketization is enabled.
+        /// </summary>
+        private bool enableBucketization;
+
+        /// <summary>
         /// Used for getting sources.
         /// </summary>
         private CommandLineOptions commandLineOptions;
@@ -154,7 +159,19 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Processors
             IRunSettingsProvider runSettingsProvider,
             ITestRequestManager testRequestManager,
             IOutput output)
+            : this(options, runSettingsProvider, testRequestManager, output, false)
         {
+        }
+
+        public RunSpecificTestsArgumentExecutor(
+            CommandLineOptions options,
+            IRunSettingsProvider runSettingsProvider,
+            ITestRequestManager testRequestManager,
+            IOutput output,
+            bool enableBucketization)
+        {
+            this.enableBucketization = enableBucketization;
+
             Contract.Requires(options != null);
             Contract.Requires(testRequestManager != null);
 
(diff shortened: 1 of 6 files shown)


singhsarab:
Need to understand this better

-	Where is the execution time coming from ?
-	Why is bucket decided by adapter ? Does not in-assembly parallelization suffice ?
-	Can buckets be created  by platform ?

Note:
Changing the OM (TestCase) in the Testplatform is not encouraged, There are compat issues that need to be handled. We'd rather go with a well known property that goes in the store property bag in the test case object.

NGloreous:
@singhsarab 

- The execution time comes form multiple factors. The test suites in question are long running functional tests which make many HTTP calls to a service. We have around 6K tests in just a handful of dlls. With our custom test harness we run 70 processes at a time and stripe the tests across these processes regardless of which dll the test comes from. 
- My understanding is in-assembly parallelization is handled by the adapter but will always be parallelized within the same process. We need in-assembly parallelization across process (for multiple reasons).
- The bucket is decided in the adapter because the adapter is already reasoning about what tests need to be run and is an existing extension point. I guess a whole new extension point could be added. 
- No the platform can't do the bucketing since we want to provide our own custom code to decide how to bucketize tests. 

Regarding the change in the TestPlatform, I'm not actually changing the TestClass class, just adding a new well known property to TestCaseProperties which will be stored in the property bag. Is there a concern with adding new properties to TestCaseProperties?

singhsarab:
/cc: @mayankbansal018 Let's review this tomorrow.

@NGloreous Apologies for the delay.

src/vstest.console/Processors/RunSpecificTestsArgumentProcessor.cs
```diff
@@ -315,6 +340,15 @@ private void discoveryRequest_OnDiscoveredTests(object sender, DiscoveredTestsEv
             }
         }
 
+        private void discoveryRequest_OnDiscoveredBucketizedTests(object sender, DiscoveredTestsEventArgs args)
```
vagisha-nidhi: Why is this needed? Why are we not filtering discovered tests?

NGloreous: So the logic for bucketizing was almost identical to test filtering (minus the filtering). So I reused this class. For clarity probably worth creating different classes that can share common logic. I'm happy to clean things up but wanted to make sure you all are fine with the general idea before I spend any more time on this.

NGloreous: @vagisha-nidhi does this make sense?

NGloreous:
@singhsarab @vagisha-nidhi is this change something you are open to taking? If so I spend a bit more time making the change nicer.

HeroMaxPower:
bump, need this :)

MarcoRossignoli:
We did some improvement for the parallelization and for the moment we don't have a lot of request to provide new feature on this side.
I'm going to close for now to reopen if the interest will raise.

PR #2163, opened 2019-09-11
MIT license for Nuget pkgs

singhsarab:
## Description
Update the license for nuget packages to point to MIT license


mayankbansal018 (approved):
Just confirm with Ariel, if this is right license that we can use.

PR #2181, opened 2019-09-19
Redirect procdump process output to diag files

vagisha-nidhi:
The procdump process std output is written to console in vstesttask. 
Redirecting the std output to diag files.


singhsarab (approved):
:shipit:

PR #2190, opened 2019-09-24
Enabling native code debugging of test host

vagisha-nidhi:
## Description
Added environment variable VSTEST_HOST_NATIVE_DEBUG to directly attach native debugger to testhost.exe process.

### DIFF
--- a/src/testhost.x86/Program.cs
+++ b/src/testhost.x86/Program.cs
@@ -101,7 +101,6 @@ namespace Microsoft.VisualStudio.TestPlatform.TestHost
             // else check for host debugging enabled
             else
             {
-                IProcessHelper processHelper = new ProcessHelper();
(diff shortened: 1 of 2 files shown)


src/testhost.x86/Program.cs
```diff
@@ -83,25 +84,35 @@ private static IEngineInvoker GetEngineInvoker(IDictionary<string, string> argsD
...
                     System.Threading.Tasks.Task.Delay(1000).Wait();
                 }
 
-                Debugger.Break();
+                DebugBreak();
+            }
+            // else check for host debugging enabled
+            else
+            {
+                IProcessHelper processHelper = new ProcessHelper();
```
mayankbansal018: processHelper is not used

mayankbansal018 (approved):

src/testhost.x86/Program.cs
vagisha-nidhi: Removed

PR #2191, opened 2019-09-24
Console output for minimal and quiet

hvinett:
## Description
modified printing level in verbosity level is quiet

console_suggested

##issue
1836


src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -155,6 +155,17 @@ protected static IOutput Output
         /// </summary>
         public Verbosity VerbosityLevel => verbosityLevel;
 
+        public class Summary
+        {
+            public int Totaltests { get; set; }
+            public int Passedtests { get; set; }
+            public int Failedtests { get; set; }
+            public int Skippedtests { get; set; }
+            public TimeSpan timespan { get; set; }
```
singhsarab: Why is this lower cased?
Any reason why all these are properties ?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -489,6 +508,7 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
                 case TestOutcome.Skipped:
                     {
                         this.testsSkipped++;
```
singhsarab: Do we need to have separate counts now that we are maintaining everything per dll ?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -621,6 +643,25 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+            if (verbosityLevel == Verbosity.Quiet)
+            {
+                foreach (var sd in SummaryDictionary.ToArray())
+                {
+                    var summary = SummaryDictionary[sd.Key];
+
+                    // Failed! Pass {1} Failed {2} skipped {3} Time : 233 se ({4})
+                    if (summary.Failedtests > 0)
+                    {
+                        Output.Information(false, ConsoleColor.White, string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, summary.Passedtests,summary.Failedtests,summary.Skippedtests, GetFormattedDurationString(summary.timespan),sd.Key.Split('\\').Last()));
```
singhsarab: Nit: formatting

src/vstest.console/Resources/Resources.resx
```diff
@@ -734,4 +734,10 @@
   <data name="TestSourcesDiscovered" xml:space="preserve">
     <value>A total of {0} test files matched the specified pattern.</value>
   </data>
+  <data name="TestRunSummaryFailed" xml:space="preserve">
```
singhsarab: Why do we need two of these ?

singhsarab: Expected string: Passed! Total:{0}  Pass:{1}  Fail:{2}  Skip:{3} Time:{4} ({5})

vagisha-nidhi:
>                 if (testsFailed > 0)

Can we figure out testsFailed, totalTests etc from SummaryDictionary? Class level variables won't be needed then.

---
Refers to: src/vstest.console/Internal/ConsoleLogger.cs:698 in 5bda2ee. [](commit_id = 5bda2eefaeb300ea9ca3b346d19fcc91537c7073, deletion_comment = False)

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -155,6 +155,8 @@ protected static IOutput Output
         /// </summary>
         public Verbosity VerbosityLevel => verbosityLevel;
 
+        public ConcurrentDictionary<string, SourceSummary> SummaryDictionary { get; private set; }
```
vagisha-nidhi: >SummaryDictionary [](start = 59, length = 17)

nit: Add comment

vagisha-nidhi: >public [](start = 8, length = 6)

Is this needed to be public?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -621,6 +634,25 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+            if (verbosityLevel == Verbosity.Quiet)
+            {
+                foreach (var sd in SummaryDictionary.ToArray())
+                {
+                    var summary = SummaryDictionary[sd.Key];
+
+                    // Failed! Pass {1} Failed {2} skipped {3} Time : 233 se ({4})
+                    if (summary.FailedTests > 0)
+                    {
+                        Output.Information(false, ConsoleColor.Red, string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, summary.TotalTests, summary.PassedTests, summary.FailedTests, summary.SkippedTests, GetFormattedDurationString(summary.TimeSpan), sd.Key.Split('\\').Last()));
```
vagisha-nidhi: >Information [](start = 31, length = 11)

Can you also attach a screenshot of the summary?

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: resolved

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -572,6 +571,44 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
+                loggerEvents.RaiseTestResult(new TestResultEventArgs(testResult));
+            }
+            loggerEvents.RaiseTestRunComplete(new TestRunCompleteEventArgs(new Mock<ITestRunStatistics>().Object, false, false, null, new Collection<AttachmentSet>(), TimeSpan.FromSeconds(1)));
+            loggerEvents.WaitForEventCompletion();
+
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, 5, 1, 1, 1, "1h 6m", "TestSource"), OutputLevel.Information), Times.Once);
+        }
+
+        [TestMethod]
+        public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForQuietVebosity()
```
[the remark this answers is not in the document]
hvinett: added

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: added

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -467,10 +479,17 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
         {
             ValidateArg.NotNull<object>(sender, "sender");
             ValidateArg.NotNull<TestResultEventArgs>(e, "e");
+            SummaryDictionary.TryGetValue(e.Result.TestCase.Source, out var summary);
```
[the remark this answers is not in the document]
hvinett: corrected

src/vstest.console/Resources/Resources.resx
hvinett: corrected

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: formatted

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: added

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: removed seperate counts

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: changed to private

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,7 +627,39 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+                    }
+                    else
+                    {
+                        Output.Information(false, ConsoleColor.Green, string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryPassed, summary.TotalTests, summary.PassedTests, summary.FailedTests, summary.SkippedTests, GetFormattedDurationString(summary.TimeSpan), sd.Key.Split('\\').Last()));
+                    }
+                }
+                return;
+            }
+
+            foreach (var sd in this.sourceSummaryDictionary.ToArray())
```
vagisha-nidhi: > foreach ( [](start = 11, length = 10)

Use one loop over the dictionary to calculate total information

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,7 +627,39 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+            {
+                summary = this.sourceSummaryDictionary[sd.Key];
+                testsPassed = summary.PassedTests;
```
vagisha-nidhi: >testsPassed [](start = 16, length = 11)

testsPassed += summary.PassedTests

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -572,6 +571,87 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
+            loggerEvents.WaitForEventCompletion();
+
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, 5, 1, 1, 1, "1h 6m", "(TestSource)"), OutputLevel.Information), Times.Never);
+        }
+
+
+        [TestMethod]
+        [DataRow("noraml")]
+        [DataRow("detailed")]
+        public void TestResultHandlerShouldWriteToConsoleShouldShowformatedPassesdTestsForOtherThanQuietVebosity(string verbosityLevel)
```
vagisha-nidhi: >TestResultHandlerShouldWriteToConsoleShouldShowformatedPassesdTestsForOtherThanQuietVebosity [](start = 20, length = 92)

nit: Check test name

hvinett: changed

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: corrected

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: changed

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,8 +627,41 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+            var testsPassed = 0;
+            var testsFailed = 0;
+            var testsSkipped = 0;
+            var testsTotal = 0;
             Output.WriteLine(string.Empty, OutputLevel.Information);
 
+            foreach (var sd in this.sourceSummaryDictionary.ToArray())
+            {
+                summary = this.sourceSummaryDictionary[sd.Key];
+                summary = this.sourceSummaryDictionary[sd.Key];
```
vagisha-nidhi: >summary = this.sourceSummaryDictionary[sd.Key]; [](start = 16, length = 47)

Remove. This is repeated

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -572,6 +571,60 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
+
+            loggerEvents.RaiseTestRunComplete(new TestRunCompleteEventArgs(new Mock<ITestRunStatistics>().Object, false, false, null, new Collection<AttachmentSet>(), TimeSpan.FromSeconds(1)));
+            loggerEvents.WaitForEventCompletion();
+
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryPassed, 2, 1, 0, 1, "1m 2s", "TestSourcePassed"), OutputLevel.Information), Times.Once);
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, 5, 1, 1, 1, "1h 6m", "TestSource"), OutputLevel.Information), Times.Once);
+        }
+
+        [TestMethod]
+        [DataRow("noraml")]
```
vagisha-nidhi: >noraml [](start = 18, length = 6)

normal*

hvinett: changed

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: removed

hvinett: removed

hvinett:
@adamralph Can you please review the change? I attached the screenshot too.

adamralph:
@hvinett I assume this addresses #1836?

I don't have the bandwidth to review the diff, but the results look good. I just have two comments:

- Is the blank line between the output lines deliberate? Can we remove it?
- There seem to be two spaces betweeen some of the items in each line. Is that deliberate? Can we try to see how it looks with only space used consistently?

adamralph:
Oh, two more things:

- the target framework is not currently displayed
- time unit (e.g. `ms`) should be preceded with a space

Taking all my comments into consideration, I would like to see this:

```
Failed! Total: 1. Pass: 1. Fail: 0. Skip: 0. (239 ms) UnitTestProject2.dll (netcoreapp3.0).
Failed! Total: 1. Pass: 1. Fail: 0. Skip: 0. (239 ms) UnitTestProject4.dll (netcoreapp3.0).
```

singhsarab:
@hvinett Can you please take into consideration the feedback shared by @adamralph and push the changes.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,8 +634,40 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+                testsFailed += summary.FailedTests;
+                testsSkipped += summary.SkippedTests;
+                testsTotal += summary.TotalTests;
+
+                if (verbosityLevel == Verbosity.Quiet)
+                {
+                    if (summary.FailedTests > 0)
+                    {
+                        // Failed! Pass {1} Failed {2} skipped {3} Time : 233 se ({4})
+                        Output.Information(false, ConsoleColor.Red, string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, summary.TotalTests, summary.PassedTests, summary.FailedTests, summary.SkippedTests, GetFormattedDurationString(summary.TimeSpan), sd.Key.Split('\\').Last(), targetFramework));
```
vagisha-nidhi: 1. What happens when target framework is not there? Do we always have a target framework?
2. Directly appending target framework will append the full name of target framework. eg: .NetCoreApp, Version 3.0.
Do we want to consider appending short folder name (netcoreapp3.0) as suggested by @adamralph ?
cc: @singhsarab

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,8 +634,40 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+                    }
+                    else
```
vagisha-nidhi: Also, please attach a screenshot of the current output

hvinett: scrrenshot attached

adamralph:
This seems to be progressing nicely!

My current feedback:

- I would prefer to have the short TFM rather than the long framework name. I.e. `netcoreapp3.0` instead of `NetCoreApp, Version 3.0`. The short TFM matches what we specify in the project file, and is less verbose (i.e. better for "quiet" output).
- There should be a space between the time value and the unit. E.g. `93 ms` instead of `93ms`.
- I would prefer to have only a single space between the results, instead of two spaces. E.g.
  ```
  Passed! Total: 1. Pass: 1. Fail: 0. Skip: 0.
   ```
  instead of
  ```
  Passed!  Total: 1.  Pass: 1.  Fail: 0.  Skip: 0.
  ````
- I like the red colouring for when there is a failure. Should we also have yellow colouring for when there is no failure, but some tests are skipped?
---
Again, taking all my points into account, I'd like to see:

```
// in yellow, because Skip != 0
Passed! Total: 2. Pass: 1. Fail: 0. Skip: 1. (239 ms) UnitTestProject2.dll (netcoreapp3.0).
```

hvinett:
> This seems to be progressing nicely!
> 
> My current feedback:
> 
> * I would prefer to have the short TFM rather than the long framework name. I.e. `netcoreapp3.0` instead of `NetCoreApp, Version 3.0`. The short TFM matches what we specify in the project file, and is less verbose (i.e. better for "quiet" output).
> * There should be a space between the time value and the unit. E.g. `93 ms` instead of `93ms`.
> * I would prefer to have only a single space between the results, instead of two spaces. E.g.
>   ```
>   Passed! Total: 1. Pass: 1. Fail: 0. Skip: 0.
>   ```
>   
>   
>   instead of
>   ```
>   Passed!  Total: 1.  Pass: 1.  Fail: 0.  Skip: 0.
>   ```
> * I like the red colouring for when there is a failure. Should we also have yellow colouring for when there is no failure, but some tests are skipped?
> 
> Again, taking all my points into account, I'd like to see:
> 
> ```
> // in yellow, because Skip != 0
> Passed! Total: 2. Pass: 1. Fail: 0. Skip: 1. (239 ms) UnitTestProject2.dll (netcoreapp3.0).
> ```

If the time variable is of two bits do you fine with the  spacing between time value and time unit.
eq :
suggested: (1 s 2 ms).
old : (1s 2ms).
you wanted the suggested one anyway?
Are you okay with the following output?
console_suggested
  cc:  @adamralph .

adamralph:
@hvinett a space should always be used between the quantity and the unit symbol as stated in most scientific notation style guides, (e.g. https://www.nist.gov/pml/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values-quantities).

The standard SI unit for time is seconds (s), and "milli" is a valid SI prefix, to give millseconds (ms). I would not mix the two together in a single value. Rather, I would restrict the precision to three significant digits and switch between seconds (s) and milliseconds (ms) appropriately. E.g.

```
723 ms
1.45 s
```

The remaining question is what to do when the time is more than one minute. One option is to stick with seconds (s), but 584 seconds is not as intuitive as 9 m 44 s. So perhaps using two units for the >1 m case case is acceptable. And for larger than one hour, the number of seconds can also be dropped, as it's not significant.

Incidentally, here is some code which I wrote somewhere else which does exactly this:

```c#
private static string ToString(double milliseconds, IFormatProvider provider)
{
    // less than one millisecond
    if (milliseconds < 1D)
    {
        return "<1 ms";
    }

    // milliseconds
    if (milliseconds < 1_000D)
    {
        return milliseconds.ToString("G3", provider) + " ms";
    }

    // seconds
    if (milliseconds < 60_000D)
    {
        return (milliseconds / 1_000D).ToString("G3", provider) + " s";
    }

    // minutes and seconds
    if (milliseconds < 3_600_000D)
    {
        var minutes = Math.Floor(milliseconds / 60_000D).ToString("F0", provider);
        var seconds = ((milliseconds % 60_000D) / 1_000D).ToString("F0", provider);
        return seconds == "0"
            ? $"{minutes} m"
            : $"{minutes} m {seconds} s";
    }

    // minutes
    return (milliseconds / 60_000d).ToString("N0", provider) + " m";
}
```

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -225,6 +229,12 @@ public void Initialize(TestLoggerEvents events, Dictionary<string, string> param
                 bool.TryParse(noprogress, out DisableProgress);
             }
 
+            if (this.verbosityLevel == Verbosity.Quiet)
```
vagisha-nidhi: >if (this.verbosityLevel == Verbosity.Quiet) [](start = 12, length = 43)

This is not needed. You can always get the targetframework despite verbosity being quite.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -619,8 +635,47 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+                testsSkipped += summary.SkippedTests;
+                testsTotal += summary.TotalTests;
+
+                if (verbosityLevel == Verbosity.Quiet)
+                {
+                    if (summary.FailedTests > 0)
+                    {
+                        // Failed! Pass {1} Failed {2} skipped {3} Time : 233 se ({4})
+                        Output.Information(false, ConsoleColor.Red, string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryFailed, summary.TotalTests, summary.PassedTests, summary.FailedTests, summary.SkippedTests, GetFormattedDurationString(summary.TimeSpan), sd.Key.Split('\\').Last(), targetFramework));
+                    }
```
vagisha-nidhi: if targetframework is null, will this throw exception?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -225,6 +229,12 @@ public void Initialize(TestLoggerEvents events, Dictionary<string, string> param
                 bool.TryParse(noprogress, out DisableProgress);
             }
 
+            if (this.verbosityLevel == Verbosity.Quiet)
+            {
+                this.targetFramework = parameters[DefaultLoggerParameterNames.TargetFramework];
+                this.targetFramework = NuGetFramework.Parse(this.targetFramework).GetShortFolderName();
```
vagisha-nidhi: > NuGetFramework.Parse [](start = 38, length = 21)

This should be only done when framework is not null.

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -563,15 +564,75 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
+
+            foreach (var testResult in this.GetPassedTestResultsObject())
+            {
+                loggerEvents.RaiseTestResult(new TestResultEventArgs(testResult));
+            }
+
+            loggerEvents.RaiseTestRunComplete(new TestRunCompleteEventArgs(new Mock<ITestRunStatistics>().Object, false, false, null, new Collection<AttachmentSet>(), TimeSpan.FromSeconds(1)));
+            loggerEvents.WaitForEventCompletion();
+
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummaryPassed, 2, 1, 0, 1, "1 m 2 s", "TestSourcePassed", "net451"), OutputLevel.Information), Times.Never);
```
vagisha-nidhi: Can we have a verify on mockOutput.Information that includes the console color as parameter?

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: we are only doing this in quiet case for now.

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: No, Ioutput doesnot contain informatin member.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -65,6 +65,11 @@ internal class ConsoleLogger : ITestLoggerWithParameters
         /// </summary>
         public const string ExtensionUri = "logger://Microsoft/TestPlatform/ConsoleLogger/v1";
 
+        /// <summary>
+        /// Represents failed tests. 
+        /// </summary>
+        private const string Failed = "Failed";
```
singhsarab: How will this get localized ?
if you don't want them localized, why dint you use the TestOutcome.Failed instead ?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -155,6 +161,13 @@ protected static IOutput Output
         /// </summary>
         public Verbosity VerbosityLevel => verbosityLevel;
 
+        /// <summary>
+        /// Dictionary of summary of each source. 
+        /// </summary>
+        private ConcurrentDictionary<string, SourceSummary> sourceSummaryDictionary { get; set; }
+
+        private string targetFramework;
```
singhsarab: >private [](start = 8, length = 7)

Docs? And also this should move next to the rest of the private members on the top

src/Microsoft.TestPlatform.ObjectModel/Logging/SourceSummary.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging
+{
+    using System;
+    public class SourceSummary
```
singhsarab: public class with no docs?

adamralph:
BTW I believe this behaviour should occur for verbosity `minimial`, not `quiet`.

`quiet` should be no output at all, unless there is an error.

`minimal` should be... well, minimal. I.e. what this PR is doing.

singhsarab:
> BTW I believe this behaviour should occur for verbosity `minimial`, not `quiet`.
> 
> `quiet` should be no output at all, unless there is an error.
> 
> `minimal` should be... well, minimal. I.e. what this PR is doing.

So the difference between minimal and quiet is with respect to the individual test cases.

In minimal mode: We print only the failed test cases along with the summary.
In quite mode: No test case information is printed. Only the summary is shown.

Summary I feel needs to be displayed in quite mode as well. It indicates the end of the test run, leaving things blank seems like a stretch.

adamralph:
@singhsarab OK, I don't mind the summary being shown in quiet mode as well.

The most important thing for me is that this new, compact, summary is shown in `minimal` mode. Not just in `quiet` mode.

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -63,10 +63,14 @@ public void InitializeShouldNotThrowExceptionIfEventsIsNotNull()
         [TestMethod]
         public void InitializeWithParametersShouldThrowExceptionIfEventsIsNull()
         {
+            var parameters = new Dictionary<string, string>
+            {
+                { "param1", "value" },
+                { DefaultLoggerParameterNames.TargetFramework , "net451"}
```
singhsarab: DRY!
We should have just created a test initialize for default dictionary and then add the required parameter to that.

singhsarab:
@vagisha-nidhi Can you please review this.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -225,6 +232,9 @@ public void Initialize(TestLoggerEvents events, Dictionary<string, string> param
                 bool.TryParse(enableProgress, out EnableProgress);
             }
 
+            var noFrameworkExists = parameters.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out this.targetFramework);
```
vagisha-nidhi: >noFrameworkExists [](start = 16, length = 17)

Where are we using this?

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -63,10 +63,13 @@ public void InitializeShouldNotThrowExceptionIfEventsIsNotNull()
         [TestMethod]
         public void InitializeWithParametersShouldThrowExceptionIfEventsIsNull()
         {
+            var parameters = new Dictionary<string, string>
```
vagisha-nidhi: > var parameters = new Dictionary<string, string> [](start = 11, length = 48)

Create in test initialize

vagisha-nidhi:
@hvinett Please attach the new screenshot and resolve the comments which are already addressed.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -155,6 +156,11 @@ protected static IOutput Output
         /// </summary>
         public Verbosity VerbosityLevel => verbosityLevel;
 
+        /// <summary>
+        /// Dictionary of summary of each source.
```
vagisha-nidhi: >Dictionary of summary of each source.  [](start = 12, length = 38)

nit: Source level summary of test result

vagisha-nidhi (approved):
:shipit:

adamralph:
> Please attach the new screenshot...

👍 yes please.

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: As we are doing it in single shot, there won't be much difference i guess.

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: changed

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: removed

adamralph:
Regarding the screenshot which was recently updated in the description, the only reservation I still have is around the use of time units as I described in https://github.com/microsoft/vstest/pull/2191#issuecomment-544910983

adamralph (approved):

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -636,6 +656,29 @@ private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
...
+                {
+                    var frameworkString = string.IsNullOrEmpty(targetFramework) ? string.Empty : string.Concat('(', targetFramework, ')');
+                    var resultString = summary.FailedTests > 0 ? CommandLineResources.Failed : CommandLineResources.Passed;
+                    var color = summary.FailedTests > 0 ? ConsoleColor.Red : summary.SkippedTests > 0 ? ConsoleColor.Yellow : ConsoleColor.Green;
+                    var outputLine = string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummary, resultString, summary.TotalTests, summary.PassedTests, summary.FailedTests, summary.SkippedTests, GetFormattedDurationString(summary.Duration), sd.Key.Split('\\').Last(), frameworkString);
+                    Output.Information(false, color, outputLine);
+                }
+            }
+
+            if (verbosityLevel == Verbosity.Quiet || verbosityLevel == Verbosity.Minimal)
```
singhsarab: What happens during cancel/abort ?

src/vstest.console/Resources/Resources.resx
```diff
@@ -734,4 +734,13 @@
   <data name="TestSourcesDiscovered" xml:space="preserve">
     <value>A total of {0} test files matched the specified pattern.</value>
   </data>
+  <data name="TestRunSummary" xml:space="preserve">
```
singhsarab: This is summary per source and not test run summary

src/vstest.console/Resources/Resources.resx
```diff
@@ -734,4 +734,13 @@
...
+    <value>{0}! Total: {1}. Pass: {2}. Fail: {3}. Skip: {4}. ({5}) {6} {7}.</value>
+  </data>
+  <data name="Failed" xml:space="preserve">
```
singhsarab: FailedString

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -1022,13 +1103,17 @@ public void GetTestMessagesShouldWriteMessageAndStackTraceToConsole()
             this.mockOutput.Verify(o => o.WriteLine(" AdditionalInfoCategory AnotherAdditionalInfoCategory", OutputLevel.Information), Times.Once());
         }
 
+        [DataRow("quiet")]
+        [DataRow("Normal")]
```
singhsarab: Why is normal special ? with a capital letter I mean

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -225,6 +232,9 @@ public void Initialize(TestLoggerEvents events, Dictionary<string, string> param
                 bool.TryParse(enableProgress, out EnableProgress);
             }
 
+            parameters.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out this.targetFramework);
```
singhsarab: >TryGetValue [](start = 23, length = 11)

Why use TryGetValue when you are not using the bool ?

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -563,15 +573,79 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
             this.mockOutput.Verify(o => o.Write(FailedTestIndicator, OutputLevel.Information), Times.Once);
-            this.mockOutput.Verify(o => o.WriteLine("TestName [4m 5s]", OutputLevel.Information), Times.Once());
+            this.mockOutput.Verify(o => o.WriteLine("TestName [4 m 5 s]", OutputLevel.Information), Times.Once());
             this.mockOutput.Verify(o => o.Write(SkippedTestIndicator, OutputLevel.Information), Times.Exactly(3));
             this.mockOutput.Verify(o => o.WriteLine("TestName", OutputLevel.Information), Times.Exactly(3));
             this.mockProgressIndicator.Verify(pi => pi.Pause(), Times.Exactly(5));
             this.mockProgressIndicator.Verify(pi => pi.Start(), Times.Exactly(5));
         }
 
+        [DataRow(".NETFramework,version=v4.5.1", "(net451)", "quiet")]
```
singhsarab: What happens when the framework looks something like Framework40 ?

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -563,15 +573,79 @@ public void TestResultHandlerShouldWriteToConsoleShouldShowPassedTestsForNormalV
...
+            loggerEvents.WaitForEventCompletion();
+
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummary, CommandLineResources.Passed, 2, 1, 0, 1, "1 m 2 s", "TestSourcePassed", expectedFramework), OutputLevel.Information), Times.Once);
+            this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestRunSummary, CommandLineResources.Failed, 5, 1, 1, 1, "1 h 6 m", "TestSource", expectedFramework), OutputLevel.Information), Times.Once);
+        }
+
+        [TestMethod]
+        [DataRow("normal")]
+        [DataRow("detailed")]
+        public void TestResultHandlerShouldNotShowformattedFailedTestsAndPassedTestsForOtherThanQuietVebosity(string verbosityLevel)
```
singhsarab: >TestResultHandlerShouldNotShowformattedFailedTestsAndPassedTestsForOtherThanQuietVebosity [](start = 20, length = 89)

nit: quiet and minimal

Do we tests to verify what gets printed in that case ?

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -347,14 +356,15 @@ public void TestRunErrorMessageShowShouldTestRunFailed()
...
+        public void InQuietModeTestErrorMessageShouldShowTestRunFailed()
         {
             // Setup
             var loggerEvents = new InternalTestLoggerEvents(TestSessionMessageLogger.Instance);
             loggerEvents.EnableEvents();
             var parameters = new Dictionary<string, string>
             {
-                { "verbosity", "quiet" }
+                { "verbosity", "quiet" },
+                { DefaultLoggerParameterNames.TargetFramework , "abc" }
```
singhsarab: Why do we need this change ? I thought we accept framework as null

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: changed

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: yes

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: yes, we accept.

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
hvinett: FrameworkArgumentProcessor will convert framework45 to  net45  before setting target framework in default logger parameters

singhsarab: But still maker sure that you code accommodates framework values that are not .NET framework

nohwnd:
Reviewing this code. It first needs to be merged with the changes done on master where we are counting the tests in a new way because of hierarchical tests (I am working on that.)

For the behavior I would like to discuss those points (but be aware that I am not very familiar with the results console runner produces at the moment, so please correct me where my behavior would not align with the current behavior 🙂): 

- I find it confusing that there is indicator for the overall pass/fail (Passed!, Failed!), but the color of the whole line is determined based on the count. This leads to result that is Passed!, but it is printed in yellow because at least one test was skipped. 

  I personally don't care about skipped tests most of the time, because any sizable test base has at least few of them. And if I cared about not skipping any tests, the test framework should (and I think it does) allow me to translate skipped / indeterminate to fail. So this leads me to believe that Skipped test count is a minor information and should not overpower Passed. 

- Seperating items by one space is nice as long as we only have single digit count of tests in each container, once we get past that, the output won't be very readable anymore.


   I think we should pad left to 5 spaces which would give a nicely aligned output for most cases (with a nice spacing up to 9999, and aligned up to 99999 tests per container), and still keep the output reasonably compact.

```
Passed! Fail:0, Pass:210, Skip:0, Total:10, 
Failed! Fail:17, Pass:10, Skip:0, Total:10, 
Failed! Fail:15, Pass:4410, Skip:20, Total:10,
Failed! Fail:44, Pass:10, Skip:0, Total:10, 
Failed! Fail:24, Pass:10, Skip:0, Total:10, 
Passed! Fail:0, Pass:25, Skip:0, Total:10, 


Passed! Fail:    0, Pass:  210, Skip:    0, Total:   10, 
Failed! Fail:   17, Pass:   10, Skip:    0, Total:   10, 
Failed! Fail:   15, Pass: 4410, Skip:   20, Total:   10,
Failed! Fail:   44, Pass:   10, Skip:    0, Total:   10, 
Failed! Fail:   24, Pass:   10, Skip:    0, Total:   10, 
Passed! Fail:    0, Pass:   25, Skip:    0, Total:   10, 
```

- I'd use comma as separator instead of dot, in my mind it just reads incorrectly when dots are used, and read `Failed: first` instead of `Failed: one`.

- I'd use Failed, Passed, and Skipped instead, of fail, pass and skip.

- I'd add few dashes to make the output easier to separate out visually. 

Here are two concepts of what would be a nice output to me. First is marking items that are non-zero with color, second one determines the color for the whole line. With Failed and passed being the most important status, and none and skipped are only used when there are no passed / failed tests.  

 I find the second approach to be easier to overview visually. 

color-item
whole-line

adamralph:
I'm fine with either of @nohwnd's suggestions.

adamralph:
Is this going to get finished? We had some good momentum here but it seems to have halted.

nohwnd:
Yes, but not this sprint.

adamralph:
@nohwnd what is the cadence of your sprints?

nohwnd:
16.7.0 starts on monday. This is not high priority, and a rather big change, but I still hope I will be able to finish it 🙂

nohwnd:
Had to re-write it a bit, because the original PR would output the results after everything was done. But now we have a problem that the public events changed. Will have to talk with others if that is a plroblem, and if solving it via `virtual` event, instead of abstract as I did here is a good solution. Otherwise the output looks nice. 

image

nohwnd:
Removed the empty line that was removed previously in https://github.com/microsoft/vstest/pull/2039

nohwnd:
Hmm, since the minimal is the default, and Normal shows the passed tests, this should really affect just the Quiet output. But I would expect quiet to be, well, quiet. But since this was approved above I am just going to go with it, because I see no other option, that would not be a breaking change.

adamralph:
It looks like some great progress has been made here!

Can we please update the issue description to show the final rendering and make it clear at what levels this output will be shown? Specifically, I would expect `minimal` to have this new, _minimal_ output rather than the current multi-line output. I would also expect minimal to omit "Starting test execution, please wait..." and "A total of x test files matched the specified pattern.".

`minimal` is supposed to be _minimal_, i.e. just show the minimum required to be able to know what happened, so I would expect that to be the single line with the results.

I would expect `quiet` to be _quiet_, i.e. don't show anything unless there is a problem.

nohwnd:
I would expect the same, but the problem is that Minimal is the default output, and changing the default is always problematic. :/

adamralph:
But that's exactly what happened in https://github.com/microsoft/vstest/issues/1836, when the default was changed from single line to multi-line. Now we have an opportunity to change it back again.

nohwnd:
I am more concerned about not being able to see which tests actually failed. Because this output does not report the tests anymore, just the per-source summary.

adamralph:
OK, then I misunderstood the current proposal. To summarise what I would like to see:

- `minimal` — This is the level that I really care about. It should show the one-line summary that this PR is introducing, plus any test failures, and nothing else. This will effectively restore `minimal` to what it should be (prior to #1836) and also improves it. `minimal` should never have been changed from a single line summary to a multi-line summary. And as I said above, it should not include "Starting test execution, please wait...", "A total of x test files matched the specified pattern.", etc. That is not _minimal_ output.
- `quiet` — I'm not too bothered what happens here, but I guess, in order to be "quiet", it shouldn't show anything, and the non-zero return code from `dotnet test` would indicate failure. I guess it _could_ show the one-line summary if there are failures, but as I said, I don't really care.

nohwnd:
> minimal — This is the level that I really care about. It should show the one-line summary that this PR is introducing, plus any test failures, and nothing else. This will effectively restore minimal to what it should be (prior to #1836) and also improves it. minimal should never have been changed from a single line summary to a multi-line summary. And as I said above, it should not include "Starting test execution, please wait...", "A total of x test files matched the specified pattern.", etc. That is not minimal output.


Aaah okay, reverting back to where I was and just adding the failed results reporting. That is why it was writing the summary after the whole run. not after every assembly. Could have saved myself a lot of work if this would have become clearer before :)

adamralph:
@nohwnd would it be possible to get an up-to-date screenshot which shows the current output from this PR? I'm finding it difficult to infer from the various changes that are being pushed.

nohwnd:
image
Minimal output.

adamralph:
Thanks @nohwnd.

image

☝️ I assume `--nologo` will still remove these lines (including the blank line)?

image

☝️ Can we please remove these lines from _minimal_ output?

nohwnd:
Yes, --nologo will remove those lines. 

Removing those lines would be more difficult because the logger parses the verbosity, but in this case the runner is what writes  those two lines, and they are also a nice indication that something happens. So I'd rather keep it to get his finally merged.

adamralph:
OK. It's still a big improvement.

BTW, has the blank line which was removed [here](https://github.com/microsoft/vstest/pull/2039) and [here](https://github.com/microsoft/vstest/pull/2191#issuecomment-638106894) come back _yet again_?

nohwnd:
You are right I reverted to the point where it worked basically as it should, added the change for removing the new line again. Now it should be gone.

src/vstest.console/Internal/ConsoleLogger.cs
nohwnd: Fixed

nohwnd:
- Minimal
image

- Minimal abort:
image

- quiet: 
image

- normal: 
image

adamralph:
Thanks for the latest screenshots @nohwnd. It's really good to see how the output is progressing.

One obvservation I have is that the blank line usage seems to be a bit inconsistent:

- In `Minimal`, there is a blank line after the skipped message but not in `Minimal abort`.
- Similarly, that same blank line serves as a blank line _before_ the summary in `Minimal` but not in `Minimal abort`
- There is a blank line after "A total of 2 test files..." in `quiet`, but not in `normal`

Perhaps it's worth doing a full assessment of where blank lines are used and are not used, in order to align everything?

As I mentioned, my chief concern is `minimal`. My preference would be:

- No blank line between "A total of 2 test files..." and the summary when there is no other content betwee them.
- A blank line after "A total of 2 test files..." and before the summary when there is some other content between those two lines, e.g. failures, skips, or abort messages

nohwnd:
That output is coming from approximately 3 places in quite random order, especially in the abort case, there we have output from console logger, collector logger, and blame logger. Sadly that is how the architecture is done. I also don't like the output in the abort cases, it looks messy, but there is not much I can do about it.

Sanan07 (approved):

adamralph:
🎉 great to see this merged! Is there a rought timeline for 16.8?

PR #2199, opened 2019-09-26
Create NuGet.config for blob feed publishing

ViktorHofer:
That file is required to restore the packagereference correctly from the feed.

@singhsarab

### DIFF
--- a/src/package/publish/NuGet.config
+++ b/src/package/publish/NuGet.config
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <fallbackPackageFolders>
+    <clear />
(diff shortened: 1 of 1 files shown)


src/package/publish/NuGet.config
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <fallbackPackageFolders>
```
singhsarab: Do we need to specify fallbackPackageFolder ?

singhsarab (approved):
Approving with suggestion.

ViktorHofer:
No fallbacks are needed here. That should suffice.

Issue #2196, opened 2019-09-25
TRX logger not including Timers in TestResultType


robdalsanto:
This is related to an enhancement request in [testfx](https://github.com/microsoft/testfx), namely #[504](https://github.com/microsoft/testfx/issues/504).

While `%VSINSTALLDIR%\xml\Schemas\vstst.xsd` has a `Timers` collection within `TestResultType`, it isn't clear that anything in the TRX logger will populate those elements if any timer values are present.

So the request is to expand the `TestResult.cs` and other classes to include this data in the TRX.

PR #2208, opened 2019-10-01
Add timers to the TestResult, and update the TRX logger to output the timer information

robdalsanto:
Add timers to the TestResult, and update the TRX logger output the timer 
information to the TRX file if any timers are in the collection, as per 
the XML schema of the TRX from
%VSINSTALLDIR%\xml\Schemas\vstst.xsd

It looks almost like an oversight that the TestResult did not previously include
the named timers.

## Related issues
- The primary motivation for this connects back to [testfx#504](https://github.com/microsoft/testfx/issues/504)
- To support testfx#504, the TRX logger has to be updated, as per #2196.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
+++ b/src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
@@ -286,10 +286,6 @@
     <value>  {0}:
 {1}</value>
   </data>
-  <data name="TestResultTimerFormat" xml:space="preserve">
-    <value>  {0} : {1} : {2}</value>
-    <comment>Timer name, start-time, and duration</comment>
-  </data>
   <data name="TestResultPropertyComputerNameLabel" xml:space="preserve">
     <value>Computer Name</value>
     <comment>The label of the test property ComputerName for test result.</comment>
--- a/src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
+++ b/src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
@@ -286,6 +286,10 @@
     <value>  {0}:
 {1}</value>
   </data>
+  <data name="TestResultTimerFormat" xml:space="preserve">
+    <value>  {0} : {1} : {2}</value>
+    <comment>Timer name, start-time, and duration</comment>
+  </data>
   <data name="TestResultPropertyComputerNameLabel" xml:space="preserve">
     <value>Computer Name</value>
     <comment>The label of the test property ComputerName for test result.</comment>
@@ -324,6 +328,10 @@
   </data>
   <data name="TestResultTextMessagesFormat" xml:space="preserve">
     <value>  Test Messages:
+{0}</value>
+  </data>
+  <data name="TestResultTimerMessagesFormat" xml:space="preserve">
+    <value>  Test Timers:
 {0}</value>
   </data>
   <data name="UnexpectedTypeOfProperty" xml:space="preserve">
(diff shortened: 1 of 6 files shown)


robdalsanto:
Added fix for failing unit test by adding commit 'df5213197b38378a712a544a6d095975cd737e31' to the pull request.

src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
```diff
@@ -286,6 +286,10 @@
     <value>  {0}:
 {1}</value>
   </data>
+  <data name="TestResultTimerFormat" xml:space="preserve">
```
singhsarab: This does have to be a resources.
Resources are strings that need to be localized. We should simply use a const for this.

src/Microsoft.TestPlatform.ObjectModel/TestResult.cs
```diff
@@ -309,6 +332,26 @@ public string Text
         }
     }
 
+    [DataContract]
```
singhsarab: Docs/Comments

src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
robdalsanto: Understood and I agree. I'll make that update.

src/Microsoft.TestPlatform.ObjectModel/TestResult.cs
robdalsanto: Yes, added them.

src/Microsoft.TestPlatform.ObjectModel/Resources/Resources.resx
robdalsanto: Done in commit 1cbbb23d5b40eebcf18cecfdfcb89f32680ba93f

robdalsanto:
@singhsarab I addressed your review comments. Thanks for the input. 
It looks like there was some problem in the build unrelated to my changes (see [here](https://dev.azure.com/vstestplatform/TestPlatform/_build/results?buildId=860&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&taskId=7376a427-0fb6-5027-9e88-fb5bb628b2d5&lineStart=277&lineEnd=301&colStart=1&colEnd=43)). I can't find a way to re-queue a build. Can you provide me any guidance so this doesn't stall ?

singhsarab:
@robdalsanto Could you please attach a sample trx with the change ?

singhsarab:
We need some tests, acceptance and unit for this change.

HeroMaxPower:
@robdalsanto could you make some acceptance tests?

robdalsanto:
Yes, I am busy now until June. So while I plan to get to this, it won't be
right away.

nohwnd:
Looks like a nice change, that is only missing tests, adding for consideration in 16.7.0.

Evangelink:
@robdalsanto We know it's been a really long time but would you be able to work on the tests or do you want us to take it over?

robdalsanto:
Hi,
Sorry to punt, but yes please take over and create the tests. I thought I'd
have time to write them, and agree they are needed. But I don't see an
opportunity to fit the task into my schedule anytime soon.
Thanks,
Rob

Evangelink:
@robdalsanto No worries, thanks for letting us know!

MarcoRossignoli:
I'm going to close for stale state.

PR #2211, opened 2019-10-03
A couple minor source-build fixes

crummel:
- Add binlogging to the build script.  This makes issues easier to diagnose for us.
- Rename the MS.Extensions.FileSystemGlobbing version property to match the usual convention.
- Remove a reference to System.Runtime.CompilerServices.Unsafe.  We originally removed this to get rid of a prebuilt but I built VSTest standalone and it doesn't appear to actually be needed.  If it is, I can change it to only be included outside of source-build instead.

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Microsoft.TestPlatform.TestHostProvider.csproj
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Microsoft.TestPlatform.TestHostProvider.csproj
@@ -26,7 +26,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" Condition="'$(TargetFramework)' == 'net451'" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" Condition="'$(OS)' == 'Windows_NT'" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'net451' ">
     <Reference Include="System" />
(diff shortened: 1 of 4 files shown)


singhsarab:
@crummel The build broke,  System.Runtime.CompilerServices.Unsafe seems to be required.

crummel:
I switched this to be conditioned on `DotNetBuildFromSource`.  It looked like the Linux build did work, is that something you expect?  I could condition it on OS instead.

singhsarab:
@crummel I guess it is only required for net451 framework then, can we use that as a condition instead?
```
 '$(TargetFramework)' == 'net451'
```

src/Microsoft.TestPlatform.TestHostProvider/Microsoft.TestPlatform.TestHostProvider.csproj
```diff
@@ -26,7 +26,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" Condition="'$(OS)' == 'Windows_NT'" />
```
singhsarab: This one is interesting.
/cc: @mayankbansal018

crummel: Yeah, it seems odd that netcoreapp2.1 Windows fails without it but netcoreapp2.1 Linux doesn't need it.

singhsarab: Are we going with the OS check then ?

singhsarab: @crummel Can you please reply on this

Haplois:
Closing because we have a newer sourcebuild PR pending.

PR #2216, opened 2019-10-07
Users/shreyas r msft/added env var support for blame results directory path

ShreyasRmsft:
## Description
Added  environment variable support to blame collector results directory
Ensured test host process is killed on timeout even if there is an error collecting the dump


singhsarab (approved):
:shipit:

Issue #2120, opened 2019-08-08
Since SDK 2.2.300 test performance is magnitudes slower


davidruhmann:
## Description
Tests take a long time to run on detached hosts build servers (CircleCI)

Similar issue was closed, but the issue still happens. https://github.com/microsoft/vstest/issues/2078

## Steps to reproduce
Repo which can reproduce the issue https://github.com/davidruhmann/vstest2080

## Expected behavior
Tests to complete `< 1s`

## Actual behavior
Tests take over `> 20s`

Each line of `Test run in progress.` output takes several seconds while the test takes `< 2 ms`.

## Diagnostic logs
https://circleci.com/gh/davidruhmann/vstest2080/3

## Environment

```
.NET Core SDK (reflecting any global.json):
 Version:   2.2.401
 Commit:    729b316c13

Runtime Environment:
 OS Name:     debian
 OS Version:  9
 OS Platform: Linux
 RID:         debian.9-x64
 Base Path:   /usr/share/dotnet/sdk/2.2.401/

Host (useful for support):
  Version: 2.2.6
  Commit:  7dac9b1b51

.NET Core SDKs installed:
  2.2.401 [/usr/share/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.2.6 [/usr/share/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.2.6 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.2.6 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
```

ShreyasRmsft:
@davidruhmann does this repro on your local devbox as well?

JSkimming:
@davidruhmann Try separating the package restore and build from the test steps. you're doing all three in a single step that takes 27 seconds; I usually do something like the following:

```bash
dotnet restore
dotnet build --no-restore
dotnet test --no-restore --no-build
```

If you do this in distinct CirlceCI steps it will give you a more accurate reflection of where the time is being taken.

davidruhmann:
@JSkimming separated steps as requested to show that dotnet test is super slow by itself.

https://circleci.com/gh/davidruhmann/vstest2080/7

[...]

PR #2234, opened 2019-10-29
Remove default progress indicator

singhsarab:
## Description
Add support for an arg to enable progress indicator, disabling by default.

## Related issue
https://github.com/microsoft/vstest/issues/2120

### DIFF
--- a/test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
+++ b/test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
@@ -133,14 +133,14 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.UnitTests.Internal
         {
             var parameters = new Dictionary<string, string>();
 
-            Assert.IsFalse(ConsoleLogger.DisableProgress);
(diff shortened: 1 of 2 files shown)


mayankbansal018:
>         public void InitializeWithParametersShouldSetNoProgress()

Add test case for default "Progress" value

---
Refers to: test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs:132 in dffe58d. [](commit_id = dffe58dcfa6da674a08eb1e8f3c0b78375bf1e73, deletion_comment = False)

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -133,14 +133,14 @@ public void InitializeWithParametersShouldSetNoProgress()
         {
             var parameters = new Dictionary<string, string>();
 
-            Assert.IsFalse(ConsoleLogger.DisableProgress);
+            Assert.IsFalse(ConsoleLogger.EnableProgress);
```
singhsarab: This assert takes care of the default case.

mayankbansal018 (approved):

PR #2259, opened 2019-11-29
Trx changes for fqdn mapping in test method name

vagisha-nidhi:
Fixing data driven xunit method name.

image
image

The execution id and the test Id is linking attributes between the 2 nodes. The result node will remain as is and it today contains the display name for the result which translates to test run title in the pipeline. 
In the Unit Test definition node the method name of xunit test is "XUnitTestProject.UnitTest1.Test1(a: 1)" which technically speaking is not correct. It has to be "XUnitTestProject.UnitTest1.Test1" and we are not losing any data this way as the result display name still remains the same.

This fix is taken as part of running the data driven xunit tests in pipeline

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestMethod.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestMethod.cs
@@ -24,8 +24,7 @@ namespace Microsoft.TestPlatform.Extensions.TrxLogger.ObjectModel
         {
             Debug.Assert(!string.IsNullOrEmpty(name), "name is null");
             Debug.Assert(!string.IsNullOrEmpty(className), "className is null");
-            this.name = name.StartsWith(className) ? name.Remove(0, $"{className}.".Length) : name;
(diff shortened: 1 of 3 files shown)


src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestMethod.cs
```diff
@@ -24,7 +24,8 @@ public TestMethod(string name, string className)
         {
             Debug.Assert(!string.IsNullOrEmpty(name), "name is null");
             Debug.Assert(!string.IsNullOrEmpty(className), "className is null");
-            this.name = name;
+            this.name = name.StartsWith(className) ? name.Remove(0, $"{className}.".Length) : name;
```
singhsarab: Should we do this in the Converter itself ?

vagisha-nidhi: Yes. done

singhsarab (approved):

Issue #2260, opened 2019-12-03
TestHost.dll discovery does not work on the Mac


mrward:
## Steps to reproduce

Visual Studio for Mac uses the test discovery protocol to discover tests. It currently uses an old 15.5 version but we were looking to update. Using vstest 16.4.0 in VS Mac and test discovery no longer works. It seems that the testhost.dll path is not determined when not on Windows and dotnet fails since no testhost.dll path is being used.

## Expected behavior

.NET Core xUnit tests can be discovered and run in VS Mac.

## Actual behavior

Test discovery fails.

Test Discovery Console output in VS Mac has:

```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Testhost process exited with error: The application to execute does not exist: ''
. Please check the diagnostic logs for more information.
  at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.ThrowExceptionOnConnectionFailure (System.Int32 connTimeout) [0x00074] in <6ce2e22cf5c344a4a7f798b25098ce9c>:0 
  at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.SetupChannel (System.Collections.Generic.IEnumerable`1[T] sources, System.String runSettings) [0x0028d] in <6ce2e22cf5c344a4a7f798b25098ce9c>:0 
  at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyDiscoveryManager.DiscoverTests (Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.DiscoveryCriteria discoveryCriteria, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestDiscoveryEventsHandler2 eventHandler) [0x00015] in <6ce2e22cf5c344a4a7f798b25098ce9c>:0 
```

The code in 16.4 compared with 16.2 has been changed so the test host path is no longer resolved.

https://github.com/microsoft/vstest/blob/7f90ebbd9e2ab63538a0d16d11a5915c67632a0a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L225

## Diagnostic logs

failed: https://gist.github.com/mrward/caed5cbbd4633be5c1c2674549cb5162

## Environment

macOS 10.15.1

Mobrockers:
Possibly the same issue for #2253 
[...]

PR #2266, opened 2019-12-05
Fix for discovery not working on Mac machines

vagisha-nidhi:
#2260

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -259,6 +259,10 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting
             if (!testHostExeFound)
             {
                 var currentProcessPath = this.processHelper.GetCurrentProcessFileName();
+                if (testHostPath == string.Empty)
+                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -259,6 +259,10 @@ public async Task<bool> LaunchTestHostAsync(TestProcessStartInfo testHostStartIn
             if (!testHostExeFound)
             {
                 var currentProcessPath = this.processHelper.GetCurrentProcessFileName();
+                if (testHostPath == string.Empty)
```
singhsarab: Always string.IsNullOrEmpty

singhsarab: How are you planning to test this ?

singhsarab: Also think about converging the two calls, if not possible I am okay with as well

vagisha-nidhi: Converted to string.IsNullOrEmpty.
Also added an extra assert in existing test to veryfy testhost.dll path is determined in non windows scenario.

singhsarab (approved):

PR #2267, opened 2019-12-05
Remove duplicate counting of test results in Consolelogger

hvinett:
## Description
Current behavior:
Recognizing parent result as passed ,failed or skipped and adding to respective counts.By this we are showing duplicate count.
ex:
Scenario:
if we have a parent test result(passed) and two child test results in  which one is passed and the other is failed.
current behavior:
total test =3
passed tests =1
Failed =2

expected behavior:
total tests =2
passed tests =1
failed tests =1

Approach :
Storing every result and it's outcome in dictionary. if it is recognized as parent then removing it from dictionary. By this way we end up with the dictionary that has non parent results and it's outcome. Finally counting the passed,failed, skipped and total in the resulting dictionary will give the expected behavior mentioned above

Assumed parent test result comes first and then the child test result comes.

 

## Related issue

### DIFF
--- a/src/vstest.console/Internal/ConsoleLogger.cs
+++ b/src/vstest.console/Internal/ConsoleLogger.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
     using System.Diagnostics;
@@ -85,6 +86,16 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
         /// </summary>
         public const string NoProgressParam = "noprogress";
 
+        /// <summary>
+        ///  Property Id storing the ParentExecutionId.
+        /// </summary>
+        public const string ParentExecutionIdPropertyIdentifier = "ParentExecId";
+
+        /// <summary>
+        ///  Property Id storing the ExecutionId.
+        /// </summary>
+        public const string ExecutionIdPropertyIdentifier = "ExecutionId";
+
         #endregion
 
         internal enum Verbosity
@@ -107,11 +118,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
         private Verbosity verbosityLevel = Verbosity.Minimal;
 #endif
 
-        private int testsTotal = 0;
-        private int testsPassed = 0;
-        private int testsFailed = 0;
-        private int testsSkipped = 0;
         private bool testRunHasErrorMessages = false;
+        private ConcurrentDictionary<Guid, TestOutcome> leafExecutionIdAndTestOutcomePairDictionary = new ConcurrentDictionary<Guid, TestOutcome>();
 
         #endregion
 
@@ -371,6 +379,29 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
             }
         }
 
+        private Guid GetParentExecutionId(TestResult testResult)
+        {
+            var parentExecutionIdProperty = testResult.Properties.FirstOrDefault(property =>
+                property.Id.Equals(ParentExecutionIdPropertyIdentifier));
+            return parentExecutionIdProperty == null
+                ? Guid.Empty
+                : testResult.GetPropertyValue(parentExecutionIdProperty, Guid.Empty);
+        }
+
+        private Guid GetExecutionId(TestResult testResult)
+        {
+            var executionIdProperty = testResult.Properties.FirstOrDefault(property =>
+                property.Id.Equals(ExecutionIdPropertyIdentifier));
+            var executionId = Guid.Empty;
+
+            if (executionIdProperty != null)
+            {
+                executionId = testResult.GetPropertyValue(executionIdProperty, Guid.Empty);
+            }
+
+            return executionId.Equals(Guid.Empty) ? Guid.NewGuid() : executionId;
+        }
+
         #endregion
 
         #region Event Handlers
@@ -468,9 +499,6 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
             ValidateArg.NotNull<object>(sender, "sender");
             ValidateArg.NotNull<TestResultEventArgs>(e, "e");
 
-            // Update the test count statistics based on the result of the test. 
-            this.testsTotal++;
-
             var testDisplayName = e.Result.DisplayName;
 
             if (string.IsNullOrWhiteSpace(e.Result.DisplayName))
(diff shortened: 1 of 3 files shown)


src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -484,11 +512,20 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
                 testDisplayName = string.Format("{0} [{1}]", testDisplayName, formattedDuration);
             }
 
+            var executionId = GetExecutionId(e.Result);
+            var parentExecutionId = GetParentExecutionId(e.Result);
+
+            if (parentExecutionId != null && leafExecutionIdAndTestOutcomePairDictionary.ContainsKey(parentExecutionId))
+            {
+                leafExecutionIdAndTestOutcomePairDictionary.TryRemove(parentExecutionId, out var value);
```
vagisha-nidhi: >value [](start = 97, length = 5)

Are we using this anywhere?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -371,6 +379,29 @@ private static void DisplayFullInformation(TestResult result)
             }
         }
 
+        private Guid GetParentExecutionId(TestResult testResult)
```
vagisha-nidhi: > private Guid GetParentExecutionId(TestResult testResult) [](start = 7, length = 57)

Add comments

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -1046,6 +1046,51 @@ public void AttachmentInformationShouldBeWrittenToConsoleIfAttachmentsArePresent
             this.mockOutput.Verify(o => o.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.AttachmentOutputFormat, uriDataAttachment1.Uri.LocalPath), OutputLevel.Information), Times.Once());
         }
 
+        [TestMethod]
+        public void ResultscountInhierarchicalOrder()
```
vagisha-nidhi: >ResultscountInhierarchicalOrder [](start = 20, length = 31)

ResultsInHeirarchichalOrderShouldReportCorrectCount

test/vstest.console.UnitTests/Internal/ConsoleLoggerTests.cs
```diff
@@ -1046,6 +1046,51 @@ public void AttachmentInformationShouldBeWrittenToConsoleIfAttachmentsArePresent
...
+        {
+            var loggerEvents = new InternalTestLoggerEvents(TestSessionMessageLogger.Instance);
+            loggerEvents.EnableEvents();
+            var parameters = new Dictionary<string, string>();
+            parameters.Add("verbosity", "normal");
+            this.consoleLogger.Initialize(loggerEvents, parameters);
+
+            TestCase testCase1 = CreateTestCase("TestCase1");
+            TestCase testCase2 = CreateTestCase("TestCase2");
```
vagisha-nidhi: >CreateTestCase [](start = 33, length = 14)

Does any test cover scenario for non heirarchichal tests as well?

hvinett: yes,
TestRunCompleteHandlerShouldWriteToConsoleIfTestsFail
TestRunCompleteHandlerShouldWriteToConsoleIfTestsPass
These cover  non heirarchichal tests  scenario

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: No, but tryRemove has that parameter as mandatory.
Any suggestions ?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -371,6 +379,39 @@ private static void DisplayFullInformation(TestResult result)
...
+        /// <summary>
+        /// Returns the parent Execution id of given test result. 
+        /// </summary>
+        /// <param name="testResult"></param>
+        /// <returns></returns>
+        private Guid GetParentExecutionId(TestResult testResult)
+        {
+            var parentExecutionIdProperty = testResult.Properties.FirstOrDefault(property =>
+                property.Id.Equals(ParentExecutionIdPropertyIdentifier));
+            return parentExecutionIdProperty == null
```
singhsarab: GetPropertyValue itself should take care of this.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -371,6 +379,39 @@ private static void DisplayFullInformation(TestResult result)
...
+        /// Returns execution id of given test result
+        /// </summary>
+        /// <param name="testResult"></param>
+        /// <returns></returns>
+        private Guid GetExecutionId(TestResult testResult)
+        {
+            var executionIdProperty = testResult.Properties.FirstOrDefault(property =>
+                property.Id.Equals(ExecutionIdPropertyIdentifier));
+            var executionId = Guid.Empty;
+
```
singhsarab: Log error if execution id is null

src/vstest.console/Internal/ConsoleLogger.cs
nohwnd: `GetPropertyValue` will throw if the property is null. This line is imho needed.

src/vstest.console/Internal/ConsoleLogger.cs
nohwnd: @singhsarab I am not sure what you mean here. What behavior should be logged as error? 

The `executionIdProperty` is null often because it regularly is not present in `testResult.Properties`. On the other hand `executionId` is never null because it is a ` Guid`.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -484,11 +522,20 @@ private void TestResultHandler(object sender, TestResultEventArgs e)
                 testDisplayName = string.Format("{0} [{1}]", testDisplayName, formattedDuration);
             }
 
+            var executionId = GetExecutionId(e.Result);
+            var parentExecutionId = GetParentExecutionId(e.Result);
+
+            if (parentExecutionId != null && leafExecutionIdAndTestOutcomePairDictionary.ContainsKey(parentExecutionId))
```
nohwnd: @hvinett  `GetParentExecutionId` returns non-nullable guid, removing the nullability check, as well as replacing it with `!= Guid.Empty` check passes all the tests in `ConsoleLoggerTests` what is the intent here?

src/vstest.console/Internal/ConsoleLogger.cs
singhsarab: Yep for xunit it would always be

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -617,6 +662,11 @@ private string GetFormattedDurationString(TimeSpan duration)
         /// </summary>
         private void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
         {
+            //System.Diagnostics.Debugger.Launch();
```
singhsarab: Remove

src/vstest.console/Internal/ConsoleLogger.cs
nohwnd: Okay, should I then make any change or keep it as is? We don't always log an error right?

src/vstest.console/Internal/ConsoleLogger.cs
nohwnd: Replacing it with discard `out _` gets rid of the warning. Fixed.

src/vstest.console/Internal/ConsoleLogger.cs
singhsarab: I think we don't need to log an error given the context

src/vstest.console/Internal/ConsoleLogger.cs
hvinett: I kept parentExecutionId != null, because .containskey() throws error if key is null

nohwnd: Yup. Let me quickly patch that, because I have the code open, and merge if that passes (it should, I tried it before).

hvinett: ok

nohwnd (approved):

PR #2272, opened 2019-12-11
Use RunSettingsFilePath from project file when using dotnet test

### LINKED ISSUE #2265: RunSettingsFilePath property isn't honored by VSTest
The property `RunSettingsFilePath` which points to the local .runsettings file for the test session should be honored by VSTest and not just by Visual Studio. Doc that points to that property: https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019#visual-studio-2019-version-164-and-later

We need this in the runtime repo (converged repo of coreclr, corefx and core-setup) to configure our test sessions without always needing to type the path to the runsettings file. Our end-goal is that `dotnet test` is sufficient to run our tests.

cc @ManishJayaswal @singhsarab

nohwnd:
## Description

Uses the RunSettingsFilePath from project file when specified and falls
back to the one specified in console, or the default setting if none are
provided.

The path to the settings file can be specified in multiple ways,
for example as a path relative to the project file. This enables
`dotnet test` to be invoked against the project file directly.

```xml
<RunSettingsFilePath>$(ProjectDir)..\example.runsettings</RunSettingsFilePath>
```

Alternatively path can also be specified relatively to solution file, but that will
only work when running dotnet test against the solution. On the other hand the
setting will be the same in all projects.

```xml
<RunSettingsFilePath>$(SolutionDir)\example.runsettings</RunSettingsFilePath>
```

The setting file can be overridden by specifying `--settings` from the command line
to give flexibility in providing alternative runsettings.

## Related issue
Fixes #2265

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -213,6 +213,19 @@ function Invoke-Build
     Write-Log "Invoke-Build: Complete. {$(Get-ElapsedTime($timer))}"
 }
 
+function Publish-PatchedDotnet { 
+    Write-Log "Publish-PatchedDotnet: Copy local dotnet installation to testArtifacts"
+    $dotnetPath = "$env:TP_TOOLS_DIR\dotnet\"
+    
+    $dotnetTestArtifactsPath = "$env:TP_TESTARTIFACTS\dotnet\" 
+    $dotnetTestArtifactsSdkPath = "$env:TP_TESTARTIFACTS\dotnet\sdk\$env:DOTNET_CLI_VERSION\"
+    Copy-Item $dotnetPath $dotnetTestArtifactsPath -Force -Recurse
+
+    Write-Log "Publish-PatchedDotnet: Copy VSTest task artifacts to local dotnet installation to allow `dotnet test` to run with it"
+    $buildArtifactsPath = "$env:TP_ROOT_DIR\src\Microsoft.TestPlatform.Build\bin\$TPB_Configuration\$TPB_TargetFrameworkNS2_0\*"
+    Copy-Item $buildArtifactsPath $dotnetTestArtifactsSdkPath -Force
+}
+
 function Publish-Package
 {
     $timer = Start-Timer
@@ -901,9 +914,11 @@ Restore-Package
 Update-LocalizedResources
 Invoke-Build
 Publish-Package
+Publish-PatchedDotnet
 Publish-Tests
 Create-VsixPackage
 Create-NugetPackages
 Generate-Manifest
 Write-Log "Build complete. {$(Get-ElapsedTime($timer))}"
 if ($Script:ScriptFailed) { Exit 1 } else { Exit 0 }
+ 
--- a/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
+++ b/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
@@ -31,7 +31,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <Microsoft.TestPlatform.Build.Tasks.VSTestTask
       TestFileFullPath="$(TargetPath)"
-      VSTestSetting="$(VSTestSetting)"
+      VSTestSetting="$([MSBuild]::ValueOrDefault($(VSTestSetting), '$(RunSettingsFilePath)'))"
       VSTestTestAdapterPath="$(VSTestTestAdapterPath)"
       VSTestFramework="$(TargetFrameworkMoniker)"
       VSTestPlatform="$(PlatformTarget)"
--- a/test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
+++ b/test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
@@ -478,6 +478,37 @@ namespace Microsoft.TestPlatform.AcceptanceTests
 
         #endregion
 
+        #region RunSettings defined in project file
+        /// <summary>
+        /// RunSettingsFilePath can be specified in .csproj and should be honored by `dotnet test`, this test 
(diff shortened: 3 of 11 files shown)


singhsarab:
@ViktorHofer You might want to take a look at this one.

scripts/build.ps1
```diff
@@ -213,6 +213,19 @@ function Invoke-Build
     Write-Log "Invoke-Build: Complete. {$(Get-ElapsedTime($timer))}"
 }
 
+function Publish-PatchedDotnet {
```
ViktorHofer: I'm not 100% sure if that's the right thing to do. Can you explain in more detail why you need the file copying?

singhsarab: This can be used to verify upfront that the integration with dotnet cli is not broken.

singhsarab: Do you have any better ideas?

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
```diff
@@ -487,46 +509,88 @@ private void Execute(string args, out string stdOut, out string stdError, out in
...
+        /// <summary>
+        /// Executes a local copy of dotnet that has VSTest task installed and possibly other modifications. Do not use this to 
+        /// do your builds or to run general tests, unless you want your changes to be reflected.
+        /// </summary>
+        /// <param name="command"></param>
+        /// <param name="args"></param>
+        /// <param name="stdOut"></param>
+        /// <param name="stdError"></param>
+        /// <param name="exitCode"></param>
+        private void ExcutePatchedDotnet(string command, string args, out string stdOut, out string stdError, out int exitCode)
```
singhsarab: Nit:Execute

ViktorHofer:
> The path to the settings file can be specified in multiple ways

My understand was that we can point to any runsettings file anywhere on disk. Your description indicates that we can only point to a project level or solution level runsettings file. Is that correct?

nohwnd:
You can point to a file anywhere on the disk.

The given examples use paths relative to a project or solution file, because I assumed that is what people will want to do most often. But you are not limited to do just that. You can use absolute path as well.

ViktorHofer:
Great, just wanted to make sure of that. Thanks.

scripts/build.ps1
nohwnd: @ViktorHofer could you address this one please? I am making a copy of the local dotnet to make sure we can detect that `dotnet test` is broken as part of our acceptance tests, instead of seeing it fail when we try to insert this dependency into the cli tooling. At the same time we don't want to fiddle with the local copy of dotnet, because that is used to do the actual build, and we don't want the developer to end up in a loop where changes to the code break the tool that builds the code. So making a copy, patching it and testing against that seemed like the best solution.

ViktorHofer: OK got it, just wanted to understand in more detail what this is for. If this provides value then great :)

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
```diff
@@ -478,6 +478,37 @@ public void EnvironmentVariablesSettingsShouldSetEnvironmentVariables(RunnerInfo
...
+        [NetFullTargetFrameworkDataSource]
+        [NetCoreTargetFrameworkDataSource]
+        public void RunSettingsAreLoadedFromProject(RunnerInfo runnerInfo)
+        {
+            AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
+
+            var projectName = "ProjectFileRunSettingsTestProject.csproj";
+            var projectPath = this.GetProjectFullPath(projectName);
+            this.InvokeDotnetTest(projectPath);
+            this.ValidateSummaryStatus(0, 1, 0);
```
AbhitejJohn: consider using named parameters so what these signify is clear.

scripts/build.ps1
AbhitejJohn: how much does this increase the build times by though?

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -31,7 +31,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <Microsoft.TestPlatform.Build.Tasks.VSTestTask
```
AbhitejJohn: Is there a doc on how dotnet test works internally? I assume that this being named a vstest target has nothing to do with `dotnet vstest` and that flows through the VSTestTask invoke. Hence your change would solve both dotnet test and dotnet vstest?

AbhitejJohn (approved):

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
singhsarab: dotnet vstest would not be affected. This is only for dotnet test flows where project file or sln file or a folder is specified. 
dotnet vstest verb is nothing but dotnet vstest.console and usage allows only dlls and not proj or sln files.

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
nohwnd: I did consider that, but ended up doing it the same way all the other tests do it to keep it consistent.

scripts/build.ps1
nohwnd: Not dramatically, copying takes about a minute, and the added tests as well adding about 2 minutes to the overall run of 35 minutes. So it is a bit of a cost but we can now better test the `dotnet test` integration, but every new test that will test that will not get that performance hit.

PR #2226, opened 2019-10-16
Only send Coverlet in proc datacollector dll to testhost

mayankbansal018:
Bug: https://developercommunity.visualstudio.com/content/problem/738856/could-not-load-file-or-assembly-microsoftintellitr.html

RC: Earlier we were loading all datacollector dll's in testhost process. In case of netcore test where dotnet is launched, we started loading FullCLR dlls inside testhost process. Now loading dlls was okay, but the moment user did 

`AppDomain.CurrentDomain.GetAssemblies().First(asm => asm.DefinedTypes.Any(t => t.FullName == "foo"))` it would fail

This was because the above line would cause dependent dll of FullCLR dll's to be loaded, which if not found would throw FileNotFoundException, & user code/test code would start failing.

We are white-listing coverlet inproc dll to be loaded in testhost process. Please see code comments to understand why, & how it should be fixed

PR #2274, opened 2019-12-13
Initialize only coverlet data collector

vagisha-nidhi:
Bug: https://developercommunity.visualstudio.com/content/problem/738856/could-not-load-file-or-assembly-microsoftintellitr.html

The last fix https://github.com/microsoft/vstest/pull/2226 did not help because the the data collector extensions (trace data collector) were still being picked up because they were default extensions.

Ideally, https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs#L108 should add default extensions only ending with the given endsWithPattern. At present, its adding all default extensions which even contains data collectors.

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -45,7 +45,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting
         private const string DotnetTestHostUri = "HostProvider://DotnetTestHost";
         private const string DotnetTestHostFriendlyName = "DotnetTestHost";
         private const string TestAdapterRegexPattern = @"TestAdapter.dll";
-        private const string DataCollectorRegexPattern = @"Collector.dll";
+        private const string CoverletDataCollectorRegexPattern = @"coverlet.collector.dll";
 
         private IDotnetHostHelper dotnetHostHelper;
         private IEnvironment platformEnvironment;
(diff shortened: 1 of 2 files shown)


mayankbansal018 (approved):

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -318,7 +318,7 @@ public IEnumerable<string> GetTestPlatformExtensions(IEnumerable<string> sources
 
             if (extensions != null && extensions.Any())
             {
-                extensionPaths.AddRange(extensions.Where(x => x.EndsWith(DataCollectorRegexPattern, StringComparison.OrdinalIgnoreCase)));
+                extensionPaths.AddRange(extensions.Where(x => x.EndsWith(CoverletDataCollectorRegexPattern, StringComparison.OrdinalIgnoreCase)));
```
AbhitejJohn: @vagisha-nidhi : Do we have a work item tracking removal of this hack? We'd want coverlet to add this into the runsettings and not do any special casing, wouldn't we?

vagisha-nidhi: I'll create an issue on vstest to clear up all coverlet specialized changes.

vagisha-nidhi: https://github.com/microsoft/vstest/issues/2275

PR #2283, opened 2019-12-31
Use patched version of Internal.TestPlatform.Extensions

nohwnd:
## Description

A quick fix for updating Microsoft.Internal.TestPlatform.Extensions that is coming from VS build, the version of the package is tied to all other external packages. The easiest approach to do this is to publish a patched version on myget and change the build script. 

This allows updating just a single binary in the package without updating all the other external dependencies. 

And avoids hardcoding the version number for just the single package because then we would have to remember it for the next time we update it.

The current package on myget contains non-signed dll at the moment.

## Related issue

None


nohwnd:
This is still waiting for the signed binary which can't be delivered till build outage is solved.

singhsarab (approved):

PR #2285, opened 2020-01-02
Cap version of VS to <16.0

### LINKED ISSUE #2284: Build.cmd fails when VisualStudio 2019 is installed
## Steps to reproduce
Install VisualStudio 2019, and VisualStudio 2017 with the workloads as defined in contributions guild. Run `build.cmd`.

## Expected behavior
To honor the dependency on VisualStudio 2017 and pass the build correctly.

## Actual behavior
Build fails with: 

```
... Update-VsixVersion: Completed.
C:\Projects\vstest\scripts\build.ps1 : The term 'C:\Program Files (x86)\Microsoft Visual
Studio\2019\Community\MSBuild\15.0\Bin\msbuild.exe' is not recognized as the name of a cmdlet, function, script file, or operable program.  Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ C:\Projects\vstest\scripts\build.ps1 ; exit $LastExitCode;
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Program File...Bin\msbuild.exe:String) [build.ps1], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException,build.ps1
```
It fails because it finds VisualStudio 2019, which does not ship with msbuild 15.0 (but it has folder for it). 

## Environment
Windows 10, latest code from master.

nohwnd:
## Description
Allows build.cmd to be used on systems that have VS2019 installed, by capping the version to VS2017. 

## Related issue
Fix #2284

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -796,13 +796,13 @@ function Locate-VsInstallPath
 
    Try
    {
-       Write-Verbose "VSWhere command line: $vswhere -latest -prerelease -products * -requires $requiredPackageIds -property installationPath"
+       Write-Verbose "VSWhere command line: $vswhere -version '(15.0,16.0]' -prerelease -products * -requires $requiredPackageIds -property installationPath"
        if ($TPB_CIBuild) {
-           $vsInstallPath = & $vswhere -latest -products * -requires $requiredPackageIds -property installationPath
+           $vsInstallPath = & $vswhere -version '(15.0,16.0]' -products * -requires $requiredPackageIds -property installationPath
        }
        else {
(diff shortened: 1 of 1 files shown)


nohwnd:
To state the context here, I am aware that the long-term goal is to not have dependency on a specific version of VS, but that would require a lot more work as well as updating the documentation. This is a quick-fix to make the dependency on VS2017 explicit, because it is there, and I need to change those lines everytime I try to build.

singhsarab (approved):

PR #709, opened 2017-04-12
TestPlugin cache refactored.

harshjain2:
1. Refactored Test Plugin Cache to support loading for individual extensions.

PR #2288, opened 2020-01-04
Pass coverlet codebase in runsettings for inproc data collector initialization

### LINKED ISSUE #2275: Pass coverlet codebase in runsettings for inproc data collector initialization
## Description
Initilization of inproc coverlet collector today needs passing the datacollector extensibility in TestPluginCache. This enables the InProcDataCollectionExtensionManager to be able to find the datacollector in all datacollector extension paths if inproc datacollector codebase path is not an absolute path.
We should directly add this path in runsettings ideally since vstest.console.exe is aware of all datacollector extension paths.
This will remove unnecessary coverlet specific changes from https://github.com/microsoft/vstest/blob/ca987de449d17284f8c9806df36f42276f13962a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs#L272 and 
https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L321

cc: @AbhitejJohn @MarcoRossignoli

MarcoRossignoli:
closes https://github.com/microsoft/vstest/issues/2275

cc: @vagisha-nidhi @AbhitejJohn @singhsarab 

If after merge you can deploy a preview package I can try to test with our new integration tests https://github.com/tonerdo/coverlet/blob/master/test/coverlet.integration.tests/Collectors.cs

I know we cannot test runner but only host, btw I'll run with my local patched sdk for the acceptance test.


MarcoRossignoli:
Checking failure.

MarcoRossignoli:
Ready to review.

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -113,16 +118,45 @@ public void Initialize(string argument)
...
             {
                 throw new SettingsException(string.Format(CommandLineResources.CollectWithTestSettingErrorMessage, argument));
             }
             AddDataCollectorToRunSettings(argument, this.runSettingsManager);
         }
 
+        /// <summary>
+        /// We try to fix inproc coverlet codebase searching coverlet.collector.dll assembly inside adaptersPaths
+        /// </summary>
+        private void FixCoverletInProcessCollectorCodeBase()
```
AbhitejJohn: Dont fully understand why we need this? Is this so older versions of coverlet still work?

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/TestPluginCacheTests.cs
```diff
@@ -459,6 +459,11 @@ protected override IEnumerable<string> GetFilteredExtensions(List<string> extens
             this.Action?.Invoke();
             return extensions;
         }
+
+        new public void SetupAssemblyResolver(string extensionAssembly)
+        {
+            base.SetupAssemblyResolver(extensionAssembly);
```
AbhitejJohn: Do we need this?

src/vstest.console/Processors/CollectArgumentProcessor.cs
MarcoRossignoli: It's not related to old version, we need to update codebase with full path to avoid the need to pass custom Coverlet adapter path to host, so my idea is to fix path during settings creation

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/TestPluginCacheTests.cs
MarcoRossignoli: Now that we don't pass Coverlet adapter path to host "before test run" to avoid assemblies resolution error inside inproc constructor we need to add coverlet folder to assembly resolver. In production is not a problem because assembly resolver is created thanks to TestPluginCache discover methods, but for testing purpose we need a way to create an assembly resolver, but inproc constructor doesn't need to call discover methods but only to add custom resolution folder. It's needed for testing purpose.

src/vstest.console/Processors/CollectArgumentProcessor.cs
AbhitejJohn: ah I see what's going on. Can we not fix the logic that generates these settings in this file?

MarcoRossignoli: Mmm it's not clear what you mean. Coverlet is injected through --collect:`XPlat Code Coverage` so no xml setting file is provided, but it's generated on the fly in this class, but the nuget folder path is dynamic and it's injected through coverlet collector target file https://github.com/tonerdo/coverlet/blob/master/src/coverlet.collector/build/netstandard1.0/coverlet.collector.targets

MarcoRossignoli: We need to do this on execute because we have to be sure that adapters element was loaded by adapter argument processor, so we can search coverlet lib inside adapters path directories.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/TestPluginCacheTests.cs
AbhitejJohn: I mean, it looks like we are just calling base in here...

MarcoRossignoli: Yep, `SetupAssemblyResolver` was private and not accessible outside `TestPluginCacheTests`, but I need to call it before pass to InProc collector for testing https://github.com/microsoft/vstest/pull/2288/files#diff-ebb20c6bd7cf65eed5a70867b1e73818R100 

So I've updated visiblity from private to protected and exposed to tests. I need a `new` because I kept same method name hiding base method. Would you like to expose `SetupAssemblyResolver` public?

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -90,11 +93,13 @@ internal class CollectArgumentProcessorCapabilities : BaseArgumentProcessorCapab
     /// <inheritdoc />
     internal class CollectArgumentExecutor : IArgumentExecutor
     {
-        private IRunSettingsProvider runSettingsManager;
+        private readonly IRunSettingsProvider runSettingsManager;
+        private readonly IFileHelper _fileHelper;
```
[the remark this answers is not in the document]
MarcoRossignoli: uh..you're right

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/TestPluginCacheTests.cs
AbhitejJohn: Nope, my bad - missed that it was protected. Thanks.

src/vstest.console/Processors/CollectArgumentProcessor.cs
AbhitejJohn: I see. That makes sense. 
There's probably another way of fixing this. All the processors have a priority associated which dictates which of them need to execute earlier. In this case the CollectArgumentProecessor should have a higher priority compared to the TestAdapterPathArgumentProcessor so you could just naturally make the change in the initialize without needing to fix it up later. 
So [this](https://github.com/MarcoRossignoli/vstest/blob/passcoverletcodebase/src/vstest.console/Processors/TestAdapterPathArgumentProcessor.cs#L89) should ideally be `ArgumentProcessorPriority.TestAdapterPath`.
@singhsarab - does that make sense?

AbhitejJohn: Ideally I'd like for this to be more general so any nuget based extension can easily setup configuration with the test platform. I've logged #2294 to track that down.

MarcoRossignoli: > So this should ideally be ArgumentProcessorPriority.TestAdapterPath.


I see, btw this is a bit weak, like the other way should be documented and a future update could silently break loading(maybe we should add a smoke test for coverlet).
Another idea could be share(expose) the portion of code that parse adapter arg and use in both place in init.

AbhitejJohn: Sure, arguments as a read only list would work as well. We'd need to be vary about re-parsing in that case.
I absolutely agree that we need a smoke test for coverlet so this continues to work. Looks like those priorities were added for scenarios like so but just haven't been used for some reason.

MarcoRossignoli: If @singhsarab agree I could opt for priority way, waiting(if necessary or give the green light).
I'll open a new issue for smoke test.

MarcoRossignoli: @AbhitejJohn @singhsarab do we go for priority+init?

AbhitejJohn: Yeah, I'd vote for that format since it already exists and is used across. We'd want to add a test to ensure that this order is maintained.

AbhitejJohn (approved):

MarcoRossignoli:
Did some manual tests
```
TpTrace Verbose: 0 : 12180, 1, 2020/01/16, 14:32:46.229, 589700868956, vstest.console.dll, CoverletDataCollector in-process codeBase path 'C:\Users\Marco\.nuget\packages\coverlet.collector\1.2.0\build\netstandard1.0\coverlet.collector.dll'
...
<RunSettings>
  <RunConfiguration>
    <ResultsDirectory>C:\git\coverletissue\vstestdebug\XUnitTestProject1\TestResults</ResultsDirectory>
    <TargetFrameworkVersion>.NETCoreApp,Version=v3.1</TargetFrameworkVersion>
    <TestAdaptersPaths>C:\Users\Marco\.nuget\packages\coverlet.collector\1.2.0\build\netstandard1.0\</TestAdaptersPaths>
  </RunConfiguration>
  <DataCollectionRunSettings>
    <DataCollectors>
      <DataCollector friendlyName="XPlat Code Coverage" enabled="True" />
    </DataCollectors>
  </DataCollectionRunSettings>
  <InProcDataCollectionRunSettings>
    <InProcDataCollectors>
      <InProcDataCollector assemblyQualifiedName="Coverlet.Collector.DataCollection.CoverletInProcDataCollector, coverlet.collector, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" friendlyName="XPlat Code Coverage" enabled="True" codebase="C:\Users\Marco\.nuget\packages\coverlet.collector\1.2.0\build\netstandard1.0\coverlet.collector.dll" />
    </InProcDataCollectors>
  </InProcDataCollectionRunSettings>
  <LoggerRunSettings>
    <Loggers>
      <Logger friendlyName="Console" uri="logger://microsoft/TestPlatform/ConsoleLogger/v1" assemblyQualifiedName="Microsoft.VisualStudio.TestPlatform.CommandLine.Internal.ConsoleLogger, vstest.console, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" codeBase="C:\git\coverletissue\dotnet\dotnet-sdk-3.1.100-win-x64\sdk\3.1.100\vstest.console.dll" enabled="True" />
    </Loggers>
  </LoggerRunSettings>
</RunSettings>
```

src/vstest.console/Processors/CollectArgumentProcessor.cs
MarcoRossignoli: Moved to init and updated proprity for `TestAdapterPathArgumentProcessor`

MarcoRossignoli:
Checking failures

MarcoRossignoli:
@AbhitejJohn @nohwnd can you review?

I moved to init and also switched two priority because `RunSettings` override all settings and so `TestAdapter` path must be processed after this one, I think that this bug was introduced with https://github.com/microsoft/vstest/pull/709 to fix some failing tests.

MarcoRossignoli:
Any news?

AbhitejJohn:
@MarcoRossignoli : Sorry I couldn't get to this earlier. This looks good to me. 
@nohwnd, I'm hoping you could take this through with any additional validation if you need to :)

PR #2300, opened 2020-01-09
Make --verbosity case insensitive

### LINKED ISSUE #2179: Handling of --verbosity is case-sensitive
As already mentioned [here](https://github.com/microsoft/vstest/issues/1590#issuecomment-392955873). This is not in sync with the rest of dotnet CLI.

nohwnd:
## Description
Makes `--verbosity` case insensitive. 

## Related issue
Fixes #2179

### DIFF
--- a/src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
+++ b/src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
@@ -247,11 +247,11 @@ namespace Microsoft.TestPlatform.Build.Tasks
                 var quietTestLogging = new List<string>() {"q", "quiet"};
 
                 string vsTestVerbosity = "minimal";
-                if (normalTestLogging.Contains(this.VSTestVerbosity))
+                if (normalTestLogging.Contains(this.VSTestVerbosity.ToLowerInvariant()))
                 {
                     vsTestVerbosity = "normal";
                 }
-                else if (quietTestLogging.Contains(this.VSTestVerbosity))
+                else if (quietTestLogging.Contains(this.VSTestVerbosity.ToLowerInvariant()))
                 {
                     vsTestVerbosity = "quiet";
                 }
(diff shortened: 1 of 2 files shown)


singhsarab:
This https://github.com/microsoft/vstest/issues/2179#issuecomment-548316069 surprises me

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -247,11 +247,11 @@ private List<string> AddArgs()
                 var quietTestLogging = new List<string>() {"q", "quiet"};
 
                 string vsTestVerbosity = "minimal";
-                if (normalTestLogging.Contains(this.VSTestVerbosity))
+                if (normalTestLogging.Contains(this.VSTestVerbosity.ToLowerInvariant()))
```
singhsarab: Were only these 2 affected ?

nohwnd:
> This [#2179 (comment)](https://github.com/microsoft/vstest/issues/2179#issuecomment-548316069) surprises me

Same here, it did not work for me, and the tests show that it does not work even on windows. I could see that passing `--logger:Console;Verbosity=normal` as well as `--logger:Console;Verbosity=Normal` to the console worked. But  `dotnet test --verbosity Normal` **DID NOT** due to the case-sensitive check above, because it returned the default `--logger:Console;Verbosity=minimal`.

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
nohwnd: Yes, that were the only two case sensitive checks I could find in that file and falsify with tests. There is also https://github.com/microsoft/vstest/blob/fa629d217b154ca729547d16249738d2ac3e9fbc/src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs#L194 but that is unrelated to the issue, and it did not look like it would easily break. We can start another issue if you think this is an improvement that should be made.

singhsarab (approved):
@AbhitejJohn Should approve too.

AbhitejJohn (approved):

PR #2315, opened 2020-02-03
Use version of external package with fixes

nohwnd:
## Description
Use version of trace data collector that has fixes for both 32-bit issues and TIA.


cvpoienaru (approved):
LGTM

PR #2327, opened 2020-02-13
update Newtonsoft.Json to 11.0.2

### LINKED ISSUE #2316: Can Newtonsoft.Json get updated to 11.x or 12.x?
The package `Microsoft.TestPlatform.TestHost` has a dependency on `Newtonsoft.Json`, for target framework `.NETCoreApp 2.1` it refers to version 9.0.1.

Would be possible to update the reference to at least version 11.x.

**Why?**
The package `Newtonsoft.Json` in version 9.0.1 has no target framework for .NETCore >= 2 or .NETStandard >= 2. When restoring packages for the solution, the direct dependency to `Microsoft.NET.Test.Sdk` causes indirect dependencies for several compatibility packages (e.g. `System.Text.Encoding`). This is because the latest target framework of `Newtonsoft.Json` is `.NETStandard 1.0`.
When forcing `Newtonsoft.Json` to version 11.0.2 in the project, the compatibility packages are not required and not restored.

We are using .net core 3.x and we want to limit the list of nuget packages in our internal package feed due to license review process. This is why we don't want to have all the small compatibility packages in our feed.

lg2de:
## Description
Update reference `Newtonsoft.Json` to a version with target framework dotnet core >= 2.
This reduces the list of implicitly referenced compatibility packages when targeting e.g. dotnet core 3.

## Related issue
Fixes #2316

### DIFF
--- a/samples/Microsoft.TestPlatform.Protocol/Microsoft.TestPlatform.Protocol.csproj
+++ b/samples/Microsoft.TestPlatform.Protocol/Microsoft.TestPlatform.Protocol.csproj
@@ -7,7 +7,7 @@
   <ItemGroup>
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
     <PackageReference Include="Newtonsoft.Json">
-      <Version>9.0.1</Version>
+      <Version>11.0.2</Version>
     </PackageReference>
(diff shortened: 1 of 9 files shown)


Haplois:
Closing because of no activity.

Issue #1824, opened 2018-10-25
Improve the cli experience for dotnet test


singhsarab:
## Description
> There has been a lot of feedback around the cli experience for dotnet test, here are a few suggestions that can be considered.

o	Show the results using unicode ✔ and ❌ with colors
o	Progress bar (only prints a dot every ~3 seconds just so the user knows we haven’t frozen.) 
o	Clean summary table at the end of run for the whole solution file. Includes Pass/Fail Unicode character, Test Name, and duration.
o	Option to output an html file.

jimmylewis:
Copying some feedback from a VS Feedback ticket:

After running tests in the console, it used to be really easy to search for “Passed”, “Skipped”, and “Failed” tests using those strings. When I used to get a result like:

>Test Run Failed.
Total tests: 6543
Passed: 6502
Failed: 1
Skipped: 40
Total time: 4.7696 Minutes

I could find the failing test just by searching in the console for "Failed".

Now the output uses ✔, !, and X instead of these. X is much more commonly found in test names or output messages (e.g. in “exception” messages or test names containing words like syntax, index. matrix - all of which are in my test bed), so searching becomes less effective to find the test failure.

Worth noting, I couldn't search for ❌ in the VS dev cmd prompt.  I had to use the letter X, which did find the match in the console contents.

jpribele:
Unicode Pass/Fail is bad for farms or other processes are that are capturing tty. 
` û TestName` example output

nohwnd:
(in reaction to PR #2333)

Personally I find the default indicators really nice to look at, but a pain to read and search for in log. I like the change but I don't think adding yet another set of options is the best way to do this. 

Maybe we could consider changing them:
- to something that is a normal character / character group, but is unlikely to turn up in log otherwise. (e.g. use `[-], [+], [!]`, but not just +,-,! or X).  
- put the indicators to the header, like Running tests (Passed √, Failed X, Skipped !). 
- use Passed, Failed, Skipped instead of cryptic signs

~Another problem is that the characters easily get mangled in output, but maybe that is just our problem, because we are likely the only ones running vstest in vstest (in our acceptance tests).~ I see the same issue mentioned above, so there not using unicode would be an advantage as well.

jpribele:
I could live with both unicode and plain text. [Passed √] TestName 
In log file that 2000 tests being able search quickly and reliably is nice. Ctrl+F "[Failed" or "[Passed

nohwnd:
[...]

PR #2333, opened 2020-02-18
Added parameters to the ConsoleLogger to change the Passed,Failed, an…

jpribele:
Added parameters to the ConsoleLogger to change the Passed,Failed, and Skipped text

## Description
Added ability to override the default text the ConsoleLogger uses for Passed, Failed, Skipped indicator text
passedTestIndicator
failedTestIndicator
skippedTestIndicator

### DIFF
--- a/src/vstest.console/Internal/ConsoleLogger.cs
+++ b/src/vstest.console/Internal/ConsoleLogger.cs
@@ -39,17 +39,17 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Internal
         /// <summary>
         /// Unicode for tick
         /// </summary>
-        private static string PassedTestIndicator = "\u221a";
+        private const string PassedTestIndicatorDefault = "\u221a";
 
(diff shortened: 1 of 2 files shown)


nohwnd:
@jpribele Could you share you motivation for this change please? 

Personally I find the default indicators really nice to look at, but a pain to read and search for in log. I like the change but I don't think adding yet another set of options is the best way to do this. Maybe we could consider changing them:
- to something that is a normal character / character group, but is unlikely to turn up in log otherwise. (e.g. use `[-], [+], [!]`, but not just +,-,! or X).  
- put the indicators to the header, like Running tests (Passed √, Failed X, Skipped !). 
- use Passed, Failed, Skipped instead of cryptic signs

Another problem is that the characters easily get mangled in output, but maybe that is just our problem, because we are likely the only ones running vstest in vstest (in our acceptance tests).

nohwnd:
Related #1824

PR #2351, opened 2020-03-03
Code Maintenance with static analysers

HeroMaxPower:
## Description
Fixed multiple messages with static analysers, your IDE should be cleaner with Errors, Warnings and Informations.

IDE0029 Null check can be simplified.
IDE0016 Null check can be simplified
IDE0017 Object initialization can be simplified.
IDE0019 Use pattern matching.
IDE0028 Collection initialization can be simplified.
IDE0034 'default' expression can be simplified.
IDE0054 Use compound assignment.
IDE1005 Delegate invocation can be simplified.

CA1507 Use nameof in place of string literal
CA1825: Avoid zero-length array allocations
CA1802 Field is declared as 'static readonly' but is initialized with a constant value. Mark this field as 'const' instead.

S1066 Merge this if statement with the enclosing one.
S2971 Drop 'Where' and move the condition into the 'Count'.
S2971 Drop 'Where' and move the condition into the 'FirstOrDefault'.
S2971 Drop this useless call to 'ToList' or replace it by 'AsEnumerable' if you are using LINQ to Entities.
S1125 Remove the unnecessary Boolean literal(s).
S1066 Merge this if statement with the enclosing one.

SA1623 The property's documentation summary text must begin with: 'Gets'
SA1611 The documentation for parameter is missing.
SA1642 Constructor summary documentation must begin with standard text.


HeroMaxPower:
Should be ready to review now 😋

nohwnd:
Thanks. Looks like you've set a goal to get on the top of contributors list in this repo, and so far you are killing it 😁 Will review this during some long running build when there is nothing super urgent to do, it's a lot of changes 🙂

HeroMaxPower:
@nohwnd would it be easier for you for me to split up these commits into separate pull requests.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/LazyExtension.cs
```diff
@@ -78,17 +68,12 @@ public LazyExtension(TestPluginInformation pluginInfo, Type metadataType)
         /// <param name="metadata">Test extension metadata</param>
         public LazyExtension(TestPluginInformation pluginInfo, TMetadata metadata)
         {
-            if (pluginInfo == null)
-            {
-                throw new ArgumentNullException("pluginInfo"); ;
-            }
-
             if (metadata == null)
```
nohwnd, line 71: the above call is reduced to ?? but this one is not. use ?? for both (also to avoid changing the order of execution, which is imho very low risk here). Same below.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -134,32 +134,33 @@ public class TestExtensions
         /// <returns> The test extensions defined the extension assembly if it is already discovered. null if not.</returns>
         internal TestExtensions GetExtensionsDiscoveredFromAssembly(string extensionAssembly)
         {
-            var testExtensions = new TestExtensions();
-
-            testExtensions.TestDiscoverers =
+            var testExtensions = new TestExtensions
```
nohwnd, line 137: Using object initializers would make this very hard to debug because the whole initializer will fail with the exception without properly indicating where it failed, please revert.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -54,7 +54,7 @@ public FilterExpressionWrapper(string filterString, FilterOptions options)
                     var regexString = options?.FilterRegEx;
                     if (!string.IsNullOrEmpty(regexString))
                     {
-                        Debug.Assert(options.FilterRegExReplacement != null ? options.FilterRegEx != null : true);
```
nohwnd, line 57: this is much harder to understand than before.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpressionWrapper.cs
```diff
@@ -127,7 +127,7 @@ public bool Evaluate(Func<string, Object> propertyValueProvider)
                 return this.fastFilter.Evaluate(propertyValueProvider);
             }
 
-            return this.filterExpression == null ? false : this.filterExpression.Evaluate(propertyValueProvider);
```
nohwnd, line 130: this is also much harder to understand than before.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/DiscoveryCompleteEventArgs.cs
```diff
@@ -22,7 +22,7 @@ public DiscoveryCompleteEventArgs(long totalTests, bool isAborted)
             // This event is always raised from the client side, while the total count of tests is maintained
             // only at the testhost end. In case of a discovery abort (various reasons including crash), it is
             // not possible to get a list of total tests from testhost. Hence we enforce a -1 count.
-            Debug.Assert((isAborted ? -1 == totalTests : true), "If discovery request is aborted totalTest should be -1.");
```
nohwnd, line 25: this was easier to understand before.

src/Microsoft.TestPlatform.ObjectModel/CodeCoverage/TestSDKAutoGeneratedCode.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.VisualStudio.TestPlatform
     /// <summary>
     /// Only to be used to Test platform, to exclude auto generated code from CodeCoverage
     /// </summary>
-    public class TestSDKAutoGeneratedCode : Attribute
+    public sealed class TestSDKAutoGeneratedCode : Attribute
```
nohwnd, line 10: this would change the public API, please remove it and all other seals as well.

src/Microsoft.TestPlatform.ObjectModel/Utilities/FilterHelper.cs
```diff
@@ -67,13 +67,10 @@ public static string Unescape(string str)
             for (int i = 0; i < str.Length; ++i)
             {
                 var currentChar = str[i];
-                if (currentChar == EscapeCharacter)
+                if (currentChar == EscapeCharacter && (++i == str.Length || !SpecialCharactersSet.Contains(currentChar = str[i])))
```
nohwnd, line 70: This condition is terrible. Pleas revert.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
     /// </summary>
     public partial class ProcessHelper : IProcessHelper
     {
-        private static readonly string ARM = "arm";
```
nohwnd, line 18: This changes the build semantics. please revert.

src/vstest.console/Processors/RunSpecificTestsArgumentProcessor.cs
```diff
@@ -379,7 +379,7 @@ private void TestRunRequest_OnRunCompletion(object sender, TestRunCompleteEventA
                 // we need to check if there are any tests executed - to try show some help info to user to check for installed vsix extensions
                 if (!e.IsAborted && !e.IsCanceled)
                 {
-                    var testsFoundInAnySource = (e.TestRunStatistics == null) ? false : (e.TestRunStatistics.ExecutedTests > 0);
```
nohwnd, line 382: This is more difficult to read.

src/vstest.console/Processors/RunTestsArgumentProcessor.cs
```diff
@@ -241,7 +241,7 @@ private void TestRunRequest_OnRunCompletion(object sender, TestRunCompleteEventA
                 // we need to check if there are any tests executed - to try show some help info to user to check for installed vsix extensions
                 if (!e.IsAborted && !e.IsCanceled)
                 {
-                    var testsFoundInAnySource = (e.TestRunStatistics == null) ? false : (e.TestRunStatistics.ExecutedTests > 0);
```
nohwnd, line 244: This is more difficult to read.

test/Microsoft.TestPlatform.AcceptanceTests/CodeCoverageTests.cs
```diff
@@ -159,7 +159,7 @@ private static string GetCoverageFileNameFromTrx(string trxFilePath, string resu
                 Assert.IsTrue(deploymentElements.Count == 1,
                     "None or more than one Deployment tags found in trx file:{0}", trxFilePath);
                 var deploymentDir = deploymentElements[0].Attributes.GetNamedItem("runDeploymentRoot")?.Value;
-                Assert.IsTrue(string.IsNullOrEmpty(deploymentDir) == false,
+                Assert.IsTrue(!string.IsNullOrEmpty(deploymentDir),
```
nohwnd, line 162: Assert.IsFalse

nohwnd (changes requested):
These are just representative issues of each category. I agree with many of the fixes like nameof, or Where().Count() -> Count(), but many of them make the code longer and harder to read, or break our compatibility. 

Please remove: 
Adding sealed to public types - this possibly breaks our contract
Adding static to public types - this also break our contract
Changing static readonly to const - const means that the value can be baked into the consuming library, static readonly does not, so when a new compatible dll is published the client does not have to rebuild to get the new value
Coverting if in if to a logical condition, this has negative impact on readability, and arguably no perf advantage. See the Debug.Assert examples or places where bussiness logic is mixed with checking if a logger is enabled just because two ifs, or ternary operator are merged


Using object initializers - they are harder to debug, and in most cases just add 2 lines of { } per init, which is unnecessary stylistic change

In other words revert: 
IDE0017 Object initialization can be simplified.
IDE0028 Collection initialization can be simplified.
CA1802 Field is declared as 'static readonly' but is initialized with a constant value. Mark this field as 'const' instead.
S1066 Merge this if statement with the enclosing one.
And the rules which public static and sealed comes from

Hope I listed all of them.

HeroMaxPower:
Sure no problem 😄

PR #2355, opened 2020-03-04
Fix splitting of test name from fully qualified name

nohwnd:
## Description

## Related issue

### DIFF
--- a/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/Utility/ConverterTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/Utility/ConverterTests.cs
@@ -116,30 +116,33 @@ namespace Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests.Utility
         public void ToTestElementShouldContainExpectedTestMethodPropertiesIfFqnIsSameAsTestName()
         {
             var expectedClassName = "TestProject1.Class1";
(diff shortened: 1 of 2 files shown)


test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/Utility/ConverterTests.cs
```diff
@@ -116,30 +116,33 @@ public void ToTestElementShouldNotFailWhenThereIsNoTestCategoreis()
...
-            ValidateTestMethodProperties(testName, fullyQualifiedName, expectedClassName);
+            ValidateTestMethodProperties(testName, fullyQualifiedName, expectedClassName, expectedTestName);
         }
 
         [TestMethod]
         public void ToTestElementShouldContainExpectedTestMethodPropertiesIfFqnDoesNotEndsWithTestName()
         {
             var expectedClassName = "TestProject1.Class1.TestMethod1(2, 3, 4";
-            var fullyQualifiedName = "TestProject1.Class1.TestMethod1(2, 3, 4.0d)";
+            var expectedTestName = "0d)";
```
nohwnd: @singhsarab I moved the test data to be more explicit, because the helper method ValidateTestMethodProperties was re- implementing the same code in the test and in the source, and both of them had the same bug. The current behavior is explicitly stating the same thing that it did before I refactored. It splits on the decimal point om `4.0d`, which I assume is not correct. What would be the correct behavior here?

jakubch1 (approved):

PR #2358, opened 2020-03-09
Fix running self-contained apps on Windows

nohwnd:
## Description
Always copy testhost.exe to the output directory to allow applications that are self contained to run. When application defines both RuntimeIdentifier and OutputType = Exe it will be packaged as self-contained and the test host from .nuget will not be able to resolve the hostfxr dll, because dotnet will try to find it next to the application. See https://github.com/dotnet/runtime/issues/3569#issuecomment-595820524 and below for more information on the behavior.


HeroMaxPower:
Your running into the same problem i was 
It appears that this test is flaky
Microsoft.VisualStudio.TraceDataCollector.UnitTests.VanguardTests.StartShouldThrowOnTimeout

test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
```diff
@@ -31,17 +33,22 @@ protected static void SetTestEnvironment(IntegrationTestEnvironment testEnvironm
 
         protected static string DeriveFrameworkArgValue(IntegrationTestEnvironment testEnvironment)
```
cvpoienaru: This can be eventually re-written using a switch-case.
```
switch (testEnvironment.TargetFramework) {
case Core21TargetFramework: return Core21FrameworkArgValue;
case Core31TargetFramework: return Core31FrameworkArgValue;
case DesktopTargetFramework: return DesktopFrameworkArgValue;
default:
    throw new NotSupportedException($"{testEnvironment.TargetFramework} is not supported TargetFramework value.");
}
```
Moreover, you can use a dictionary to associate the TargetFramework with the FrameworkArgValue it should correspond to. Then search the map and if the key is not there you throw an exception, otherwise you get the value.

nohwnd: Not exactly the same, switch does not do ordinal comparison but I agree that switch would be nicer. Translating to a dictionary has the same problem, and seems like unnecessary optimization. Can we keep it as is for now, and tackle this refactoring in next step where I will enable netcoreapp3.1 testing for all acceptance tests?

nohwnd: My bad, the default comparison is ordinal. So the switch does ordinal. Still would rather fix it later.

test/Microsoft.TestPlatform.AcceptanceTests/Extension/NetCoreTargetFrameworkDataSource.cs
```diff
@@ -23,24 +23,54 @@ public class NetCoreTargetFrameworkDataSource : Attribute, ITestDataSource
...
+            // laying the ground work here for tests to be able to run against 3.1 but not enabling it for
+            // all tests to avoid changing all acceptance tests right now
+            bool useNetCore31Target = false)
         {
             this.dataRows = new List<object[]>(6);
 
             if (useDesktopRunner)
             {
-                this.dataRows.Add(new object[]
+                if (useNetCore21Target)
```
cvpoienaru: It seems there's a lot of code duplication below. How about using something like this:

```
private void AddRunner(useTarget, runnerFramework, targetFramework) {
if (!useTarget) return;

this.dataRows.Add(new object[]
{
    new RunnerInfo(runnerFramework, targetFramework)
});
}
```

Then you can have statements like:
```
AddRunner(useNetCore21Target, IntegrationTestBase.DesktopRunnerFramework, AcceptanceTestBase.Core21TargetFramework);
AddRunner(useNetCore31Target, IntegrationTestBase.DesktopRunnerFramework, AcceptanceTestBase.Core31TargetFramework);
...
```

cvpoienaru (approved):

test/Microsoft.TestPlatform.AcceptanceTests/Extension/NetCoreTargetFrameworkDataSource.cs
nohwnd: That's not a very obvious design. The method is called Add* so it should Add. Not TryAdd (which has a different connotation), or AddIfEnabled. If anything I would refactor to:

```cs
if (useNetCore21Target) 
{
    AddRunnerDataRow(... 
}
```

PR #2359, opened 2020-03-13
Add option to specify custom test host

### LINKED ISSUE #1749: Add option to use a different DOTNET host than the one located in the path
Follow-up on the offline discussion that we had. We (the .NET team) want to enable VS Text Explorer support for corefx.

We are building the shared framework and the runtime ourselves, we do things differently as the average user. Still we see a strong demand for enabling Test Explorer like in a regular app.
- Framework
We (corefx) are the ones that build the framework itself. Therefore we can’t use the shared framework located under `%PATH%/dotnet/shared/Microsoft.NETCore.App/x.x.x/` as we want to test our live build.
- Host
We don’t take a dependency on the globally installed .NET Core SDK and its dotnet.exe host and use nightly builds instead. That’s because we rely on features that aren’t yet available and because we don’t need that dependency.

We need a way to tell the VS Text Framework (Test Explorer) to use our dotnet host under corefx instead of the global one in the path. Probably via a switch in the .runsettings file that allows to set the custom path.

> How are we running the tests? do we use vstest.console or dotnet test verb ?

On the command-line we run tests in a different way by directly invoking xunit.console.dll but with Visual Studio we just want to use the Test Explorer which I'm not sure what it uses behind the scenes. For xunit we are using its test adapter: https://www.nuget.org/packages/xunit.runner.visualstudio/

> Can’t we temporarily set the environment variable “PATH” to point to the new path in a script before invoking the test command ?

Can we do that when we just use VS Test Explorer?

cc @singhsarab @danmosemsft

nohwnd:
## Description
Add option to specify path to dotnet host

## Related issue
Fixes #1749

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -487,6 +487,11 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
         /// </summary>
         public bool CollectSourceInformationSet { get; private set; } = false;
 
+        /// <summary>
+        /// Path to dotnet executable to be used to invoke testhost.dll. Specifying this will skip looking up testhost.exe and will force usage of the testhost.dll. 
+        /// </summary>
+        public string DotnetHostPath { get; private set; }
+
         #endregion
 
         /// <inheritdoc/>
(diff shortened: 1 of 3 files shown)


nohwnd:
image

@ViktorHofer  Tested this withing VS by copying from C:\Projects\vstest\src\package\package\bin\Debug\net451\win7-x64 to C:\Program Files (x86)\Microsoft Visual Studio\2019\IntPreview\Common7\IDE\Extensions\TestPlatform

vritant24:
LGTM
Just ensure that the [documentation](https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019) is updated with `<DotnetHostPath>` available under `<RunConfiguration>`

vritant24 (approved):

PR #2371, opened 2020-03-24
Support /TestCaseFilter and /Tests arguments at the same time

NGloreous:
## Description
My team has a scenario where I want to be able to pass both the /TestCaseFilter and the /Tests arguments at the same time (yes I know this doesn't seem like it should be necessary but we do have this use case). Right now vstest throws an exception if both are provided but I can't find a reason why. Digging through the code everything is piped through properly and works as expected.


nohwnd:
I also can't think about a reason why those parameters should not be compatible. Did you look into vstest-docs for any clues?

NGloreous:
@nohwnd yes I looked at the docs. They have a note saying they can't be used together but no context on why. Also it looks like the code has been like this since this repo was created. Not sure where the code lived before that. My guess is someone thought there is no need to specify both?

nohwnd (approved):

PR #2393, opened 2020-04-08
Fixes Test Platform.

AbhitejJohn:
Also updates to the latest version of TP.Fakes bits so we can catch this in E2E tests.


AbhitejJohn:
We should have Unit tests for this. Follow up PR

vritant24 (approved):

PR #2382, opened 2020-04-02
VS Depencencies from 16.7.0 signed build

cvpoienaru:
## Description
WebTest issue #1073431 insertion

PR #2395, opened 2020-04-09
VS Depencencies from the latest signed build

nohwnd:
Update external dependencies from VS

### DIFF
--- a/src/package/external/external.csproj
+++ b/src/package/external/external.csproj
@@ -50,7 +50,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.Internal.TestPlatform.Extensions">
-      <Version>$(TestPlatformExternalsVersion)-patched7</Version>
+      <Version>$(TestPlatformExternalsVersion)</Version>
       <PrivateAssets>All</PrivateAssets>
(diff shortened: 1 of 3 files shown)


src/package/external/external.csproj
```diff
@@ -50,7 +50,7 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.Internal.TestPlatform.Extensions">
-      <Version>$(TestPlatformExternalsVersion)-patched7</Version>
```
jakubch1, line 53: Please upgrade also other packages that we get from mstest to TestPlatformExternalsVersion.

In this file we have:
Microsoft.Internal.Intellitrace
Microsoft.Internal.Dia

In other places we reference:
Microsoft.Internal.CodeCoverage
Microsoft.Internal.Dia.Interop

We should upgrade all of them together. Otherwise we can have some issues

nohwnd: Is it not? This line is just removing the suffix that was added to be able to use a different version. Are the dependencies that you mentioned using the variable that is used here?

cvpoienaru:
Duplicate of: https://github.com/microsoft/vstest/pull/2382

PR #2397, opened 2020-04-13
Changed new configurator method name

vritant24:
The old method name caused an ambiguous match exception since there is code in places we are still discovering that call the fakes configuration method with the name alone, not accounting for overloads.

Some more changes to variable names to explicitly show the difference

### DIFF
--- a/src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
+++ b/src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
@@ -20,7 +20,9 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.Utilities
     {
         private const string ConfiguratorAssemblyQualifiedName = "Microsoft.VisualStudio.TestPlatform.Fakes.FakesDataCollectorConfiguration";
 
-        private const string ConfiguratorMethodName = "GetDataCollectorSettingsOrDefault";
+        private const string NetFrameworkConfiguratorMethodName = "GetDataCollectorSettingsOrDefault";
+
+        private const string CrossPlatformConfiguratorMethodName = "GetCrossPlatformDataCollectorSettings";
 
         private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=16.0.0.0, Culture=neutral";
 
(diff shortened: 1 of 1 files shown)


AbhitejJohn (approved):

AbhitejJohn:
@vritant24 : Please create a cherry-pick to master as well.

nohwnd (approved):

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -50,17 +52,44 @@ public static string GenerateFakesSettingsForRunConfiguration(string[] sources,
...
+            // and default to FrameworkCore10 for .Net Core 
+            if (targetFramework.Name.IndexOf("netstandard", StringComparison.OrdinalIgnoreCase) >= 0 ||
+                targetFramework.Name.IndexOf("netcoreapp", StringComparison.OrdinalIgnoreCase) >= 0)
+            {
+                return FrameworkVersion.FrameworkCore10;
+            }
+
+            // Since the Datacollector is separated on the NetFramework/NetCore line, any value of NETFramework 
+            // can be passed along to the fakes data collector configuration creator. 
+            // We default to Framework40to preserve back compat
```
AbhitejJohn, line 91: nit: missing space. `Framework40 to`

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -50,17 +52,44 @@ public static string GenerateFakesSettingsForRunConfiguration(string[] sources,
                 doc.Load(xmlReader);
             }
 
-            return !TryAddFakesDataCollectorSettings(doc, sources, GetFramework(runSettingsXml)) 
-                ? runSettingsXml 
-                : doc.OuterXml;
+            var frameworkVersion = GetFramework(runSettingsXml);
+            if (frameworkVersion == null)
+            {
+                return runSettingsXml;
```
AbhitejJohn, line 58: it would be nice to add logs in this case. I dont think the EqtTraceLogs go anywhere useful today. We'd probably need a logger interface passed through. To reduce code churn in a release branch, we could do that in the master branch as a follow up change.

PR #2401, opened 2020-04-14
[master] Fix null reference

nohwnd:
## Description
Fix null reference that we catch internally but would pop up in Debug mode, because the default config is to stop at null ref

## Related issue
none


cvpoienaru (approved):
LGTM

PR #2411, opened 2020-04-21
Dev/vrbhardw/testplatformfakesupdate

vritant24:
Adds Datacollector to fix ambiguous match exception


vritant24:
Closed because took it off the wrong branch

PR #2419, opened 2020-04-23
Wrap error message/stack trace content in <pre>

### LINKED ISSUE #2414: HTML logger should honor line breaks in stack traces
## Description
Stack traces are difficult to read in HTML test results because the page doesn't have the line breaks from the raw stack trace string.

## Steps to reproduce

1. Break some tests in Roslyn or really any project
2. Run the tests with any invocation using an HTML logger, such as `dotnet test .\src\Compilers\CSharp\Test\Emit --framework netcoreapp3.1 --logger "html"`.
3. Open the test results HTML file

Let me know if you need me to try and adapt this invocation into the vstest console format somehow.

## Expected behavior
The call stack of the failing test uses the line breaks of the original stack trace string. The below image works by wrapping the `<span class="error-stack-trace">` in `<pre>`. I personally don't mind scrolling to the right to see the full call information for each line.
image

## Actual behavior
Call stacks use the automatic line breaks that are normal for HTML instead of the line breaks present in the original stack trace string, making it very difficult to read.
image

RikkiGibson:
## Description
This enhances the readability of error messages and stack traces in HTML logs by wrapping the content in the [`<pre>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) tag.

Here's a side-by-side comparison, where the first test failure shown has no `<pre>` tags and the second test failure does have `<pre>` tags.

image

## Related issue
Closes #2414

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
+++ b/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
@@ -28,6 +28,9 @@
         margin-top: 15px;
         margin-bottom:10px;
         }
+        pre {
+        white-space: pre-wrap;
+        }
         .summary {font-family:monospace;
         display: -webkit-flex; /* Safari */
         -webkit-flex-wrap: wrap; /* Safari 6.1+ */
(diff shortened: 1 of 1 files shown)


pavelhorak (approved):
Perhaps easier was modifying line 68 like: .error-stack-trace { color: brown; white-space: pre; }

RikkiGibson:
Thank you!

PR #2429, opened 2020-05-09
Detect runsettings file automatically for assembly invocation and detect coverlet data collector in runsettings

### LINKED ISSUE #2418: RunSettings auto-detection for assembly invocation scenarios
Following up on https://github.com/microsoft/vstest/issues/2229#issuecomment-571822652.

If no .runsettings file is passed to vstest.console on an assembly invocation (ie dotnet test abc.dll), it should search for one in the test assembly's directory. The cwd could be used as a probing location as well.

cc @nohwnd @vritant24

### LINKED ISSUE #2432: InProcess coverlet data collector isn't running when enabling coverlet from the runsettings file
If the coverlet data collector is configured via a .runsettings file and set to `enabled=True`, the InProcessDataCollector isn't added to the vstest.console's internal .runsettings memory representation. The impact is that the code coverage measurement runs as expected but the results aren't stable because of the missing in-proc collector: https://github.com/coverlet-coverage/coverlet/issues/834.

Example runsettings file in which coverlet data collector is enabled:

```
<RunSettings>
  <DataCollectionRunSettings>
    <DataCollectors>
      <DataCollector friendlyName="XPlat code coverage" enabled="True">
        <Configuration>
          <Include>[System.Text.RegularExpressions]*</Include>
          <ExcludeByFile>C:\git\runtime3\src\libraries\Common\src\System\SR.*,C:\git\runtime3\src\libraries\Common\src\System\NotImplemented.cs</ExcludeByFile>
          <IncludeDirectory>C:\git\runtime3\artifacts\bin\testhost\net5.0-Windows_NT-Debug-x64\shared\Microsoft.NETCore.App\5.0.0</IncludeDirectory>
          <Format>opencover</Format>
          <SingleHit>false</SingleHit>
          <UseSourceLink>true</UseSourceLink>
          <IncludeTestAssembly>false</IncludeTestAssembly>
        </Configuration>
      </DataCollector>
    </DataCollectors>
  </DataCollectionRunSettings>
</RunSettings>
```

I have a fix for that locally and will submit a PR.

cc @MarcoRossignoli

ViktorHofer:
Fixes https://github.com/microsoft/vstest/issues/2418
Fixes https://github.com/microsoft/vstest/issues/2432

Add an auto-detection mechanism that searches for a ".runsettings" in
the test app's directory if no runsettings file was passed in and only
a single test assembly is invoked.

Extend the AlwaysExecute mechanism for ArgumentProcessors so that
processors can run by default without a switch passed in without being a `SpecialCommand`.

cc @nohwnd

### DIFF
--- a/src/vstest.console/CommandLine/Executor.cs
+++ b/src/vstest.console/CommandLine/Executor.cs
@@ -126,7 +126,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine
             // Quick exit for syntax error
             if (exitCode == 1
                 && argumentProcessors.All(
-                    processor => processor.Metadata.Value.CommandName != HelpArgumentProcessor.CommandName))
+                    processor => processor.Metadata.Value.CommandName != HelpArgumentProcessor.CommandName && !processor.Metadata.Value.IsSpecialCommand))
             {
                 this.testPlatformEventSource.VsTestConsoleStop();
                 return exitCode;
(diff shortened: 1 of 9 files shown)


ViktorHofer:
This needs a slightly different approach as the runsettings is added too late, this should be done before the lazy initialization.

nohwnd:
> This needs a slightly different approach as the runsettings is added too late, this should be done before the lazy initialization.

Does that mean that this is WIP and you will continue on it, or I should change the approach? 

2) How do we prevent this from automatically taking the settings if the user does not want them to be applied?

ViktorHofer:
> Does that mean that this is WIP and you will continue on it, or I should change the approach?

Yes this WIP, just converted the PR to a draft one.

> How do we prevent this from automatically taking the settings if the user does not want them to be applied?

My assumption is that if a user puts a .runsettings file into an test app's output directory and names that file exactly `.runsettings`, then this is intentional to make use of the auto-detection. An alternate solution would be to search for a different name, ie `default.runsettings`.

ViktorHofer:
cc @MarcoRossignoli for the coverlet discovery change here.

PR #2430, opened 2020-05-11
More verbose info in datacollector log

nohwnd:
## Description
Data collector logs what is currently running even without the blame switch, but does not include the test names, this change makes it easier to figure out if there were any tests in progress from the normal diag log that we always ask for when investigating a bug.


cvpoienaru (approved):
LGTM

PR #2431, opened 2020-05-11
Added support for default exclusion merging for code coverage

cvpoienaru:
## Description
Initial support for default exclusion merging


src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,271 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Xml;
+    using System.Xml.XPath;
+
+    public class CodeCoverageRunSettingsProcessor
+    {
+        #region Members
+        private static readonly string CodeCoverageCollectorDefaultSettings =
+            @"<DataCollector uri=""datacollector://microsoft/CodeCoverage/2.0"" assemblyQualifiedName=""Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector, Microsoft.VisualStudio.TraceCollector, Version=16.0.0.0 " + @", Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"" friendlyName=""Code Coverage"">" + Environment.NewLine +
```
nohwnd, line 16: Make it a single multiline string? (or if you decide to concatenate then string array and string.join by environment.newline?)

cvpoienaru: I copy-pasted the string as is from `EnableCodeCoverageArgumentProcessor.cs` and I prefer to keep it formatted like this to give the reader a better understanding of the structure of the XML. However, I decided to go with your second option, i.e. `string.Join` on a string array due to perf considerations (also it's easier than writing by hand the concatenation of `Environment.NewLine` every time :) ). Even though the string has a constant number of concatenations and the operation is only done once because it's a static it should still be faster this way 👍

nohwnd:
Changed into a draft PR because it is still WIP.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,515 @@
...
+            /// 
+            /// <returns>A relative path built from path components.</returns>
+            public static string BuildPath(IEnumerable<string> pathComponents)
+            {
+                var path = ".";
+
+                foreach (var component in pathComponents)
+                {
+                    path += "/" + component;
+                }
```
Sanan07, line 80: This part can be improved by using StringBuilder.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,515 @@
...
+            }
+        }
+        #endregion
+
+        #region Members
+        #region Default Settings String
+        /// <summary>
+        /// Represents the default settings for the code coverage data collector.
+        /// </summary>
+        private static readonly string DefaultSettings = string.Join(
```
Sanan07, line 92: As you initialize variable it is better to change it to "const"

cvpoienaru: The const specifier doesn't work in this context and causes a compilation error (see [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/const)).

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: **TL;DR**: That's correct indeed and I did the change.

Long version:
Did it this way because the input is completely controller by us (even though the method is public the class in which it is defined is a private member of another class -- see [this](https://github.com/microsoft/vstest/pull/2431/files#diff-35b75052fa805261fcf83e696f1a45ccR23)). Moreover, the method is only called with arguments that aren't larger than 2 string components really. Felt it's more expressive this way than by using a StringBuilder at virtually no performance cost.

Sanan07 (approved):

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+
+        /// <summary>
+        /// Represents the default settings loaded as an <see cref="XmlNode"/>.
+        /// </summary>
+        private XmlNode defaultSettingsRootNode;
+
+        /// <summary>
+        /// Represents a list of exclusion types tracked by this processor.
+        /// </summary>
+        private IEnumerable<Exclusion> exclusions;
```
jakubch1, line 184: Please change it to IList<Exclusion>.
IEnumerable can be iterated only once. So it means that Process can be called once on the instance. Having it IEnumerable is not benefit in this case

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+        #endregion
+
+        #region Members
+        #region Default Settings String
+        /// <summary>
+        /// Represents the default settings for the code coverage data collector.
+        /// </summary>
+        private static readonly string DefaultSettings = string.Join(
+            Environment.NewLine,
+            @"<DataCollector uri=""datacollector://microsoft/CodeCoverage/2.0"" assemblyQualifiedName=""Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector, Microsoft.VisualStudio.TraceCollector, Version=16.0.0.0 " + @", Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"" friendlyName=""Code Coverage"">",
```
jakubch1, line 96: As part of TraceDataCollector we have already file with default settings:

https://github.com/microsoft/vstest/blob/master/src/DataCollectors/TraceDataCollector/VanguardCollector/DefaultCodeCoverageConfig.xml

I would use this file to not have any discrepancies. Class CodeCoverageRunSettingsProcessor could be changed to get this default settings as argument in constructor.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+        /// </summary>
+        /// 
+        /// <param name="currentSettings">The code coverage settings.</param>
+        /// 
+        /// <returns>An updated version of the current run settings.</returns>
+        public XmlNode Process(string currentSettings)
+        {
+            if (string.IsNullOrEmpty(currentSettings))
+            {
+                return null;
```
jakubch1, line 222: Why not to return default settings here?

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+        {
+            /// <summary>
+            /// Represents the <see cref="XPathNavigator"/> style path of the exclusion type.
+            /// </summary>
+            private string path;
+
+            /// <summary>
+            /// Gets the path components for the current exclusion type.
+            /// </summary>
+            public IEnumerable<string> PathComponents { get; }
```
jakubch1, line 33: similar like some other comment. IList is better in this case

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+            /// <summary>
+            /// Assembles a relative path from the path given as components.
+            /// </summary>
+            /// 
+            /// <returns>A relative path built from path components.</returns>
+            public static string BuildPath(IEnumerable<string> pathComponents)
+            {
+                var sb = new StringBuilder();
+                sb.Append(".");
+
```
jakubch1, line 77: the whole method can be done in 1 liner like this

String.Join("/", new [] {"."}.Concat(pathComponents))

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+        /// <see cref="true"/> if the node should be processed, <see cref="false"/> otherwise.
+        /// </returns>
+        private bool ShouldProcessCurrentExclusion(XmlNode node)
+        {
+            const string attributeName = "mergeDefaults";
+            const string attributeValue = "false";
+
+            foreach (XmlAttribute attribute in node.Attributes)
+            {
+                if (attribute.Name == attributeName && attribute.Value == attributeValue)
```
jakubch1, line 394: I would make it case insensitive by:

attribute.Value == attributeValue -> !bool.Parse(attribute.Value)

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+            // from the default settings document and return since there's nothing else to be done.
+            var codeCoveragePathComponents = new List<string>() { "CodeCoverage" };
+            var currentCodeCoverageNode = this.SelectNodeOrAddDefaults(
+                currentSettingsRootNode,
+                this.defaultSettingsRootNode,
+                codeCoveragePathComponents);
+
+            if (currentCodeCoverageNode == null)
+            {
+                return currentSettingsRootNode;
```
jakubch1, line 276: this is some error case, right? i think return null; is more readable. Please add some comment

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+            }
+
+            // Get the code coverage node from the default settings.
+            var defaultCodeCoverageNode = this.ExtractNode(
+                this.defaultSettingsRootNode,
+                Exclusion.BuildPath(codeCoveragePathComponents));
+
+            foreach (var exclusion in this.exclusions)
```
jakubch1, line 284: You are using exclusions only here so it doesn't make sense to keep reference on instance. You can just before this line generate list of exlusions and yes in this case it can be IEnumerable :)

You are modifying state of exclusions during process so if I understand it correctly creating 1 instance of this Processor and calling in parallel Process can produce issues. Lets keep it simple

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Done. You're right, it really makes more sense to downgrade from an instance member to a local scoped variable.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Yes, it's indeed an error case but we don't return `currentCodeCoverageNode` which is null, but `currentSettingsRootNode` which is not null here.

The effect of the call above, `this.SelectNodeOrAddDefaults(...)`, is that it either correctly extracts the code coverage node and returns it, or it adds defaults for it and returns null. In case null is returned we bail out by returning the root node which is not null.

You're correct that the reason for bailing out is not apparent so I added a comment there.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: I think it's better to use bool.TryParse instead just to make sure we're protected against invalid input.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
```diff
@@ -0,0 +1,484 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Merges the current settings rules with the default settings rules.
+        /// </summary>
+        /// 
+        /// <param name="node">The root node.</param>
+        /// <param name="exclusion">The exclusion rule.</param>
+        private void MergeNodes(XmlNode node, Exclusion exclusion)
```
jakubch1, line 458: I think whole merging can be done with something like this:


        private void MergeNodes(XmlNode currentNode, XmlNode defaultNode, Exclusion exclusion)
        {
            HashSet<string> existingExclusions = new HashSet(node.ChildNodes.Select(c => c.OuterXml));
       
            foreach (var child in defaultNode.ChildNodes)
            {
                if(!existingExclusions.Contains(child.OuterXml))
                {
                  var importedChild = currentNode.OwnerDocument.ImportNode(child, true);
                  currentNode.AppendChild(importedChild);
                }
            }
        }
        
        
By this you don't need at all method AddNodes and ExclusionRules map.

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
```diff
@@ -0,0 +1,308 @@
...
+            @"    <CompanyNames/>",
+            @"    <PublicKeyTokens/>",
+            @"  </CodeCoverage>",
+            @"</Configuration>");
+
+        private CodeCoverageRunSettingsProcessor processor;
+
+        private XmlDocument document;
+
+        private XmlNode ExtractNode(XmlNode node, string path)
```
jakubch1, line 98: Please put all  private methods to bottom of file with tests.

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
```diff
@@ -0,0 +1,308 @@
...
+
+                set.Remove(child.OuterXml);
+            }
+
+            Assert.AreEqual(set.Count, 0);
+        }
+
+        public CodeCoverageRunSettingsProcessorTests()
+        {
+            this.processor = new CodeCoverageRunSettingsProcessor();
```
jakubch1, line 152: remove constructor and use SetUp method so that every test starts with clean workspace

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
```diff
@@ -0,0 +1,308 @@
...
+            var processedNode = processor.Process(settings);
+            Assert.IsNotNull(processedNode);
+
+            var codeCoverageNodes = this.ExtractNodes(processedNode, expectedResultDocument.DocumentElement, "./CodeCoverage");
+
+            this.CompareResults(codeCoverageNodes.Item1, codeCoverageNodes.Item2, "./ModulePaths/Exclude");
+            this.CompareResults(codeCoverageNodes.Item1, codeCoverageNodes.Item2, "./Functions/Exclude");
+            this.CompareResults(codeCoverageNodes.Item1, codeCoverageNodes.Item2, "./Attributes/Exclude");
+            this.CompareResults(codeCoverageNodes.Item1, codeCoverageNodes.Item2, "./Sources/Exclude");
+        }
```
jakubch1, line 306: please add tests for mergeDefaults = false attribute

test/DataCollectors/TraceDataCollector.UnitTests/DynamicCoverageDataCollectorImplTests.cs
```diff
@@ -83,6 +83,7 @@ public void InitializeShouldCreateDefaultCodeCoverageSettingsIfConfigElementIsNu
         }
 
         [TestMethod]
+        [Ignore]
```
jakubch1, line 86: why we need to ignore it?

jakubch1 (changes requested):
Could you please create some acceptance tests for it?

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
jakubch1: Yes. You are right. I forgot about this :)

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Utilities/CodeCoverageRunSettingsProcessor.cs
cvpoienaru: Awesome simplification. While we're at it I completely got rid of the `Exclusion` class as it was rendered useless by this comment.

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
cvpoienaru: Done.

test/DataCollectors/TraceDataCollector.UnitTests/DynamicCoverageDataCollectorImplTests.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Utilities.UnitTests/CodeCoverageRunSettingsProcessorTests.cs
cvpoienaru: I think there's no need to do this anymore. After the code cleanup there's virtually no persistent state created between runs. The processor class only loads the default settings document in its constructor whereas the tests' constructor loads the default settings and creates the shared processor.

cvpoienaru:
Already started work on acceptance tests but will merge as is for now.

I created an issue [here](https://github.com/microsoft/vstest/issues/2451) and assigned myself to it to keep track of this task. Will shortly post a WIP PR associated with this issue that contains the changes I did for acceptance tests.

jakubch1 (approved):

PR #1682, opened 2018-07-10
Enhancing blame data collector options

kaadhina:
## Description
Adding two more options in blame collector
DumpType: To be able to optionally get full dump (The default is mini dump)
AlwaysCollectDump: To be able to optionally generate dump even if all test cases complete (The default behaviour has been changed to generate dump only if any test cases has crashed)

Eg usage:
/blame:collectDump;DumpType=full;AlwaysCollectDump=true

PR #2434, opened 2020-05-13
Add hangdump and crash dump capabilities and options

nohwnd:
## Description
First implementation of the hang and crash dump that supports hang dumps on and dotnet core on Windows, and dotnet core 3.1 and newer on linux without any additional tools. 

For crash dumps only Windows x64 and x86 are supported, but additionally ProcDump is resolved from PATH, not just PROCDUMP_PATH.

## Related issue
Related #1682


pavelhorak (approved):

ViktorHofer:
Wow fantastic. Has the usage of the blame collector in the runsettings file changed?

nohwnd:
@ViktorHofer it did not, as far as I am aware. Maybe the dumptype parameter was renamed, but I can make that backwards compatible when consumed from the settings. I am sure we will discover any discrepancies when we integrate this into dotnet/runtime. :)

PR #2438, opened 2020-05-15
Add the parameters to VSTestTask to allow dotnet test to work

nohwnd:
## Description
Add parameters for Blame dump options that dotnet test will use to pass the parameters to vstest console.


src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -258,9 +294,51 @@ private List<string> AddArgs()
...
+                    {
+                        dumpArgs.Add("CollectDump");
+                        if (!string.IsNullOrEmpty(this.VSTestBlameCrashCollectAlways))
+                        {
+                            dumpArgs.Add($"CollectAlways={this.VSTestBlameCrashCollectAlways}");
+                        }
+
+                        if (!string.IsNullOrEmpty(this.VSTestBlameCrashDumpType))
+                        {
+                            dumpArgs.Add($"DumpType={this.VSTestBlameCrashDumpType.Trim()}");
```
Sanan07, line 316: Why do use Trim() only here, not in other adds to List, for instance in above ` dumpArgs.Add($"CollectAlways={this.VSTestBlameCrashCollectAlways}");`?

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -258,9 +294,51 @@ private List<string> AddArgs()
...
             {
-                allArgs.Add("--Blame");
+                var blameArgs = "--Blame";
+
+                var dumpArgs = new List<string>();
+                if (blameCrash || blameHang)
+                {
+                    if (blameCrash)
+                    {
+                        dumpArgs.Add("CollectDump");
```
Sanan07, line 308: Is it not good to initialize that type of strings as const, and use them when needed?

src/Microsoft.TestPlatform.CoreUtilities/Utilities/TimeSpanParser.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                throw GetFormatException(value);
+            }
+
+            var suffix = match.Groups["suffix"].Value;
+            var c = StringComparison.OrdinalIgnoreCase;
+
+            // mil to distinguish milliseconds from minutes
+            // ""  when there is just the raw milliseconds value
+            if (suffix.StartsWith("ms", c) || suffix.StartsWith("mil", c) || suffix == string.Empty)
```
Sanan07, line 44: Is it not  better to check for string.Empty first?

Sanan07 (commented):
Left my thought about some parts of code, others look good to me :)

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
nohwnd: Generally yes, but this is a dll that does not have many dependencies on the rest of the project and only uses the string once, so putting it in constants class does not help me not making when I re-use the string, because there is no reuse. Won't fix.

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
nohwnd: Don't remember, fixing.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/TimeSpanParser.cs
nohwnd: Does not really matter here, this is not hot path, so the order has no real impact here. Won't fix.

Sanan07 (approved):
Looks good.

PR #2452, opened 2020-06-05
Microsoft.Diagnostics.NETCore.Client package upgrade for symbol archival.

AbhitejJohn:
Updating the package reference of "Microsoft.Diagnostics.NETCore.Client" to the latest from nuget.org so we have symbols for this from internal archival stores. 

Please let me know if there are specific end-to-end scenarios you'd want me to validate.


nohwnd:
The version numbers here are confusing, the publicly released version is 6 months old from December 2020, while the included version is from April 2020. So this would be a downgrade.

AbhitejJohn:
Oh, you probably meant December 2019. I wasn't sure where the preview package was being picked up from. Do we need the latest? I only moved to the public version so we have symbols.

nohwnd:
I did mean that. I asked them to publish a new version.

Haplois (approved):

PR #2462, opened 2020-06-15
Update to net472

nohwnd:
## Description
Update Test Platform to .NET 4.7.2


src/testhost/app.config
```diff
@@ -46,6 +46,34 @@
...
+        <assemblyIdentity name="System.IO.Pipelines" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.2.0" newVersion="4.0.2.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.VisualStudio.Threading" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-16.6.0.0" newVersion="16.6.0.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-12.0.0.0" newVersion="12.0.0.0" />
```
AbhitejJohn, line 75: What happens if a test assembly depends on a lower version of Newtonsoft.json? This was a major problem earlier IIRC.

ViktorHofer:
@nohwnd should we close this now given that you added net472 and net48 support already?

nohwnd:
those are related but not the same. this one deals with updating the whole test platform to net4.7.2, but yeah we can close this I guess. There is no push to actually lift the baseline.

PR #2475, opened 2020-07-03
Change indicators to words

### LINKED ISSUE #1824: Improve the cli experience for dotnet test

nohwnd:
## Related issue
Fix #1824


cvpoienaru (approved):
LGTM

PR #2481, opened 2020-07-13
Changes to allow special characters in parameter names

### LINKED ISSUE #2441: Allow special characters in parameters
@nohwnd I tried what you outlined above and is worked for the most part. 
The issue is that I have use a lot of code that uses parameter names with non alphanumeric characters.
Example:
`dotnet vstest tests.dll -- TestRunParameters.Parameter(name=\"MainGroup:Browser\",value=\"Chrome\")`

Any reason 
https://github.com/microsoft/vstest/blob/7eaa0a21b51c8e75d99848de3e4e4555b8334dfe/src/Microsoft.TestPlatform.Common/Utilities/RunSettingsProviderExtensions.cs#L98
couldn't be changed to:
 `var attrName = $"(?<{AttributeNameString}>\\S+)";`?
or you could also do something like this if you don't want to allow names to start or end with non alpha numeric:
`var attrName = $"(?<{AttributeNameString}>(\\w+\\S*\\w+)|(\\w+))";)`

* This would be closer aligned with what you can do in the VS Test task in Azure DevOps
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/vstest?view=azure-devops

_Originally posted by @TroyWalshProf in https://github.com/microsoft/vstest/issues/862#issuecomment-606005714_

cvpoienaru:
## Description
Changes to allow special characters in parameter names

Related issue #2441


test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
```diff
@@ -134,13 +134,7 @@ public void UpdateRunSettingsNodeShouldAddNewKeyIfNotPresent()
...
         public void UpdateTestRunParameterSettingsNodeShouldAddNewKeyIfNotPresent()
         {
-            var match = this.runSettingsProvider.GetTestRunParameterNodeMatch("TestRunParameters.Parameter(name=\"weburl\",value=\"http://localhost//abc\")");
-            var runSettingsWithTestRunParameters = "<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n<RunSettings>\r\n  <TestRunParameters>\r\n    <Parameter name=\"weburl\" value=\"http://localhost//abc\" />\r\n  </TestRunParameters>\r\n</RunSettings>";
-
-            this.runSettingsProvider.UpdateRunSettings("<RunSettings>\r\n  </RunSettings>");
-            this.runSettingsProvider.UpdateTestRunParameterSettingsNode(match);
-
-            Assert.AreEqual(runSettingsWithTestRunParameters, this.runSettingsProvider.ActiveRunSettings.SettingsXml);
+            this.CheckMatch("weburl", @"http://localhost//abc");
```
nohwnd, line 137: That CheckMatch method does not do at all what I would expect by it's name. It seems as if it would check regex match but it does something totally different. Please consider renaming it.

cvpoienaru: Done.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
```diff
@@ -230,6 +236,17 @@ public void QueryRunSettingsNodeShouldReturnCorrectValue()
             Assert.AreEqual("x86", this.runSettingsProvider.QueryRunSettingsNode("RunConfiguration.TargetPlatform"));
         }
 
+        private void CheckRunSettingsAreUpdated(string parameterName, string parameterValue)
+        {
+            var match = this.runSettingsProvider.GetTestRunParameterNodeMatch($"TestRunParameters.Parameter(name=\"{parameterName}\",value=\"{parameterValue}\")");
+            var runSettingsWithTestRunParameters = $"<?xml version=\"1.0\" encoding=\"utf-16\"?>\r\n<RunSettings>\r\n  <TestRunParameters>\r\n    <Parameter name=\"{parameterName}\" value=\"{parameterValue}\" />\r\n  </TestRunParameters>\r\n</RunSettings>";
```
nohwnd, line 242: Could you split it to lines and then string.join with environment.newline? That would make it a bit more readable.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
```diff
@@ -156,6 +150,18 @@ public void UpdateTetsRunParameterSettingsNodeShouldOverrideValueIfKeyIsAlreadyP
             Assert.AreEqual(runSettingsWithTestRunParametersOverrode, this.runSettingsProvider.ActiveRunSettings.SettingsXml);
         }
 
+        [TestMethod]
+        public void UpdateTestRunParameterSettingsNodeWithSpecialCharactersNameShouldSucceed()
```
nohwnd, line 154: Name this `TestRunParameterSettingsNodeCanContainSpecialCharacters` ?

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
```diff
@@ -156,6 +150,18 @@ public void UpdateTetsRunParameterSettingsNodeShouldOverrideValueIfKeyIsAlreadyP
...
+        {
+            this.CheckRunSettingsAreUpdated("weburl:name", @"http://localhost//abc");
+        }
+
+        [TestMethod]
+        public void UpdateTestRunParameterSettingsNodeWithSingleCharacterNameShouldSucceed()
```
nohwnd, line 160: Name this `TestRunParameterSettingsNodeCanBeJustASingleCharacter` ?

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsProviderExtensions.cs
```diff
@@ -95,7 +95,7 @@ public static void UpdateRunSettingsNode(this IRunSettingsProvider runSettingsPr
         /// <returns></returns>
         public static Match GetTestRunParameterNodeMatch(this IRunSettingsProvider runSettingsProvider, string node)
         {
-            var attrName = $"(?<{AttributeNameString}>\\w+)";
+            var attrName = $"(?<{AttributeNameString}>(\\w+\\S*\\w+)|(\\w+))";
```
nohwnd, line 98: Won't this match way more than we want, that is any non-whitespace, even repeated one?

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/RunSettingsProviderExtensionsTests.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsProviderExtensions.cs
nohwnd: Maybe we should use `[\w.:-]+` matching `A-Z a-z 0-9 _ . : -` which would allow reasonably separating the values without imposing any structure to it? This string has no semantic meaning to us, so forcing structure seems like overkill. Especially since _ is already part of the `\w` group so you should be able to do `__________hello_this_is_dog__` already.

cvpoienaru: Sounds good, I'll go ahead and changes this to `[\w.:-]+` then. Many thanks for your input ! 👍

cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsProviderExtensions.cs
```diff
@@ -95,7 +95,7 @@ public static void UpdateRunSettingsNode(this IRunSettingsProvider runSettingsPr
         /// <returns></returns>
         public static Match GetTestRunParameterNodeMatch(this IRunSettingsProvider runSettingsProvider, string node)
         {
-            var attrName = $"(?<{AttributeNameString}>\\w+)";
+            var attrName = $"(?<{AttributeNameString}>[\\w.:-]+)";
```
nohwnd, line 98: No need to do this but converting this and the line below to $@ string would make it so much easier to read.

nohwnd (approved):

PR #2491, opened 2020-07-24
Nuget.Frameworks renamed netcoreapp5.0 to net5.0

nohwnd:


cvpoienaru (approved):

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -81,7 +81,8 @@ public static string GenerateFakesSettingsForRunConfiguration(string[] sources,
             // Since there are no FrameworkVersion values for .Net Core 2.0 +, we check TargetFramework instead
             // and default to FrameworkCore10 for .Net Core 
             if (targetFramework.Name.IndexOf("netstandard", StringComparison.OrdinalIgnoreCase) >= 0 ||
-                targetFramework.Name.IndexOf("netcoreapp", StringComparison.OrdinalIgnoreCase) >= 0)
+                targetFramework.Name.IndexOf("netcoreapp", StringComparison.OrdinalIgnoreCase) >= 0 ||
```
wli3, line 84: Will that work after net5? @zkat there is a better way to do it right?

nohwnd: This won't work after net5, and I am sure there is a better way to do this, but this one was the fastest and safest way to get the fix out without running into potential conflicts with matching ".NETFramework" moniker in unintended places.

PR #2493, opened 2020-07-28
Multitarget testhost

### LINKED ISSUE #2376: xUnit under VSTest.Console does not always use the correct .NET Framework runtime - report false test failures
I have created a minimal xUnit test project demonstrating the problem (the test project use the SDK-style csproj, and target net472).
when running the test with 'dotnet test' and on some machines also VSTest.Console, it will throw the following exception:
System.ArgumentException : The specified value is not valid in the 'SslProtocolType' enumeration.
This exception is not thrown by .NET Framework 4.7.2

This issue does not occur with MSTest v2.

Project demonstrating the problem:
https://github.com/TalAloni/dotnettest-xunit

nohwnd:
## Description
Builds testhost and testhost.x86 against net451, net452, net46, net461, net462, net47, net471, net472, net48 and copies them into all the packages and vsix that we ship. Forces the DefaultProvider to use version of testhost based on the determined version of .NET Framework of the source dll. 

Fix #2376


test/TestAssets/MultitargetedNetFrameworkProject/UnitTest1.cs
```diff
@@ -0,0 +1,85 @@
...
+                case "NET462":
+                    Assert.NotNull(exception);
+                    break;
+                default:
+                    Assert.Null(exception);
+                    break;
+            }
+        }
+
+        static void ProcessInformation(IAsyncResult result)
```
Sanan07, line 81: Why there is an empty method?

test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
```diff
@@ -9,13 +9,32 @@ namespace Microsoft.TestPlatform.AcceptanceTests
...
+        public const string Net451TargetFramework = "net451";
+        public const string Net452TargetFramework = "net452";
+        public const string Net46TargetFramework = "net46";
+        public const string Net461TargetFramework = "net461";
+        public const string Net462TargetFramework = "net462";
+        public const string Net47TargetFramework = "net47";
+        public const string Net471TargetFramework = "net471";
+        public const string Net472TargetFramework = "net472";
+        public const string Net48TargetFramework = "net48";
         public const string DesktopTargetFramework = "net451";
```
Haplois, line 21: Should we use `Net451TargetFramework` instead of this?

test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
```diff
@@ -9,13 +9,32 @@ namespace Microsoft.TestPlatform.AcceptanceTests
...
         public const string Core21TargetFramework = "netcoreapp2.1";
         public const string Core31TargetFramework = "netcoreapp3.1";
 
+        public const string DesktopFrameworkArgValue = ".NETFramework,Version=v4.5.1";
```
Haplois, line 25: And should we use `Net451FrameworkArgValue` instead of this?

ViktorHofer:
cc @jaredpar

test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
nohwnd: I would rather avoid mixing those changes into this PR, because the DesktopTargetFramework is used in multiple places and in conjunction with IsDesktopRunner, which would prompt even more changes to make the code sense semantically. 

image

I am sure I will come across this again when I work with linux tests and expanding our testing to net5.0, so I can fix it then.

Haplois (approved):

PR #2508, opened 2020-08-04
Fixed TRX file overwrite in certain circumstances

### LINKED ISSUE #2489: dotnet test: warning: Overwriting results file: duplicate trx file
## Description
When we start up tests from a solution file using `dotnet test -l trx`, we encounter this warning occasionally:
`EXEC : warning : Overwriting results file: D:\...\TestResults\2020.07.17-17.41.02\...04231817_2020-07-17_17_41_25.trx [d:\dbs\sh\acaa\0717_171134\cmd\0\src\dirs.proj]`

## Problem
The failure is stemming from [SetDefaultTrxFilePath](https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs#L483) computing the next available file name through [IsValidIteration](https://github.com/microsoft/vstest/blob/master/src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs#L357) but does not reserve it.  In case two test sessions are completing at the same time, they end up with the same trx filename, and by the time they write the results, one ends up clobbering the other.

## Fix
Expect that the file gets created in IsValidIteration (and suppress the overwrite warning).

Haplois:
## Description
Checks the target file name for saving the TRX logs and insures it is distinct. This check does not occur if a filename is implicitly set, in this case file will be overwritten and a warning is shown.

## Related issue
Fixes #2489


src/Microsoft.TestPlatform.Extensions.TrxLogger/Resources/xlf/TrxResource.tr.xlf
```diff
@@ -303,6 +303,11 @@ Fehlerdetails: {1}: {2}</target>
         <target state="translated">LogFileName ve LogFilePrefix parametreleri birlikte kullanılamaz. </target>
         <note></note>
       </trans-unit>
+      <trans-unit id="Common_CannotGetNextTimestampFileName">
+        <source>Cannot get find an available filename for {0} using timestamp format '{2}' at {1}.</source>
+        <target state="translated">"{1}" altında "{2}" biçimlendirmesi ve "{0}" ismi kullanılarak müsait bir dosya adı bulunamadı.</target>
```
nohwnd, line 308: I don't think that is necessary, there might be "standards" that the translation team is using to keep the translations consistent, and this might circumvent them.

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs
```diff
@@ -289,6 +329,8 @@ private static bool IsReservedFileName(string fileName)
                     return tryMePath;
                 }
 
+                // Sleep a random amount before trying the next iteration
+                Thread.Sleep(rng.Next(1, 150));
```
nohwnd, line 333: This will actually block the thread for the given time, that seems unnecessary if we are able to reserve the file name.

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -389,9 +393,17 @@ internal void TestRunCompleteHandler(object sender, TestRunCompleteEventArgs e)
 
             helper.SaveObject(runSummary, rootElement, "ResultSummary", parameters);
 
-            //Save results to Trx file
-            this.DeriveTrxFilePath();
-            this.PopulateTrxFile(this.trxFilePath, rootElement);
+            EqtAssert.IsTrue(createLogFileMutex.WaitOne(MutexTimeout), "Cannot acquire the global file population mutex.");
```
nohwnd, line 396: Could we instead of using mutex to lock the system for the whole time it populates the file just simply create the file with FileMode.CreateNew instead of FileMode.Create, and handle the IOException by generating a new timestamp?

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -634,6 +635,101 @@ public void DefaultTrxFileShouldCreateIfLogFileNameParameterNotPassed()
...
+                l.TestResultHandler(new object(), pass.Object);
+                var testRunCompleteEventArgs = TrxLoggerTests.CreateTestRunCompleteEventArgs();
+                l.TestRunCompleteHandler(new object(), testRunCompleteEventArgs);
+            });
+
+            var files = loggers.Select(l => l.trxFile).Distinct().ToArray();
+
+            Assert.IsTrue(files.Length == loggers.Length, "All logger instances should get a seperate file name!");
+
+            foreach (var file in files)
```
nohwnd, line 661: This won't clean-up if the test fails.

nohwnd (changes requested):
I would avoid using pessimistic concurrency, and instead bump the timestamp when we get a conflict. Maybe pass the trx filename as an object / strategy that can easily regenerate it, or if that would change the public APIs just use our knowledge of the format (as you already do) and bump the timestamp and retry.

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs
```diff
@@ -133,6 +133,45 @@ public static string GetNextIterationFileName(string parentDirectoryName, string
...
+                string tryMePath = Path.Combine(directoryName, tryMe);
+                if (!File.Exists(tryMePath))
+                {
+                    return tryMePath;
+                }
+
+                iterationStamp = iterationStamp.AddSeconds(1);
+                ++iteration;
+            }
+            while (iteration <= ushort.MaxValue);
```
Sanan07, line 170: Why `iteration` has `uint` type, but in while loop you are comparing it with `ushort.MaxValue`?

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
Haplois: I can't see any downsides of using a Mutex, it actually feels cleaner then starting the operation, handling the exception and trying again. Am I missing something?

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs
Haplois: `uint.MaxValue` is too high. So I used `ushort` there.

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs
```diff
@@ -278,6 +316,7 @@ private static bool IsReservedFileName(string fileName)
             Debug.Assert(!string.IsNullOrEmpty(originalName), "originalName is Null");
             Debug.Assert(helper != null, "helper is null");
 
+            var rng = new Random();
```
Sanan07, line 319: Why do you need it here?

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -634,6 +635,131 @@ public void DefaultTrxFileShouldCreateIfLogFileNameParameterNotPassed()
...
+            {
+                loggers.AsParallel().ForAll(l =>
+                {
+                    var pass = TrxLoggerTests.CreatePassTestResultEventArgsMock();
+                    l.TestResultHandler(new object(), pass.Object);
+                    var testRunCompleteEventArgs = TrxLoggerTests.CreateTestRunCompleteEventArgs();
+                    l.TestRunCompleteHandler(new object(), testRunCompleteEventArgs);
+                });
+
+                files = loggers.Select(l => l.trxFile).Distinct().ToArray();
```
Sanan07, line 744: This part is the same in test methods, it is better to have it in separate method.

src/Microsoft.TestPlatform.Extensions.TrxLogger/Utility/FileHelper.cs
Haplois: Oversight.

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
nohwnd: My thinking is this:
1) Mutex is probably not behaving perfectly the same cross platform. So making sure this works on RedHat, MacOS, Linux, Windows and Windows ARM might be unnecessarily difficult.
2) the conflict only happens occasionally, but we will lock every time 
3) in case we get into conflict one process has to wait for the whole time the other is generating the whole file, not just reserving the name
( 4) there are not Mutexes in the solution so far )

All in all the solution seems overcomplicated for something that can be done reactively when the writing actually fails.  That solution is also much easier to test.

nohwnd:
I did have a look, as you requested, and it still seems very complex for what is basically a retry operation, but maybe I am not grasping the details. 

I think it might get easier if we passed in an object that contains the raw data that we used to generated the filename and provides a method to get Next, and it also indicates whether or not we should be rewriting because based on whether the user provided the name or not. At least that is how I understand from you this is determined. This would avoid the timestamp parsing code, and it looks like all the appropriate methods are internal so we should not have problem with changing public APIs.

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -33,30 +33,30 @@ namespace Microsoft.VisualStudio.TestPlatform.Extensions.TrxLogger
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TrxLogger"/> class.
         /// Constructor with Dependency injection. Used for unit testing.
         /// </summary>
         /// <param name="fileHelper">The file helper interface.</param>
-        protected TrxLogger(IFileHelper fileHelper)
+        protected TrxLogger(IFileHelper fileHelper) : this(new Utilities.Helpers.FileHelper(), new InternalFileHelper()) { }
+
+        internal TrxLogger(IFileHelper fileHelper, InternalFileHelper internalFileHelper)
```
nohwnd, line 50: Why can't those be the same class?

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -634,6 +639,95 @@ public void DefaultTrxFileShouldCreateIfLogFileNameParameterNotPassed()
...
+
+            testableTrxLogger = new TestableTrxLogger(new FileHelper(), internalFileHelper);
+            testableTrxLogger.Initialize(this.events.Object, this.parameters);
+
+            MakeTestRunComplete();
+
+            var fileName = Path.GetFileName(testableTrxLogger.trxFile);
+            var expectedName = $"{DefaultLogFilePrefixParameterValue}{time:_yyyyMMddHHmmss}.trx";
+
+            Assert.AreEqual(expectedName, fileName, "Trx file name pattern has changed. It should be in the form of prefix_yyyyMMddHHmmss.trx");
```
nohwnd, line 659: Please add the reason why we need it to be this format, which is that Azure Devops VSTest task depends on this naming.

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -642,6 +736,8 @@ public void CustomTrxFileNameShouldConstructFromLogFileParameter()
             Assert.AreEqual(Path.Combine(TrxLoggerTests.DefaultTestRunDirectory, TrxLoggerTests.DefaultLogFileNameParameterValue), this.testableTrxLogger.trxFile, "Wrong Trx file name");
         }
 
+
```
nohwnd, line 739: This change is probably unnecessary.

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
Haplois: The other class is public, and in a different assembly.

nohwnd: Hmm right. Can we call it TrxFileHelper instead of internal please? It will make it easier to know what project it belongs to.

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -634,6 +639,95 @@ public void DefaultTrxFileShouldCreateIfLogFileNameParameterNotPassed()
...
+        [TestMethod]
+        public void TrxPrefixFileNameShouldIterate()
+        {
+            this.parameters.Remove(TrxLoggerConstants.LogFileNameKey);
+            this.parameters[TrxLoggerConstants.LogFilePrefixKey] = DefaultLogFilePrefixParameterValue;
+
+            var files = TestMultipleTrxLoggers();
+
+            Assert.IsTrue(files.Length == MultipleLoggerInstanceCount, "All logger instances should get the same file name!");
```
nohwnd, line 688: One more nitpick, use equality assertions instead so we can see the actual numbers in the failure. Also is the description correct? it is the same for the positive and negative case.

Haplois: Updated the assertions. For that configuration of instances, it's true. Negative case is also tested in other method.

nohwnd (approved):

PR #2516, opened 2020-08-12
Replace NET451 compiler directives with NETFRAMEWORK

nohwnd:
## Description
Replacing all compiler directirves with netframework because we started building testhost against different frameworks, and even thought most of the project is still net451, there is probably no code that is explicitly net451, but rather is netframework (as opposed to netcoreapp).

Fixes #AB1168114


drognanar:
would multi-targeting unit test projects help to ensure that this doesn't regress?

drognanar (approved):

nohwnd:
@drognanar  it would, it is in my long term goals. :)

Issue #2380, opened 2020-04-01
Improving blame collector functionality of TestPlatform


pavelhorak:
- The blame collector should support collecting dumps on all platforms:
MacOS
Windows
Linux 
- [x] The blame collector should support killing the testhost on a hang on all platforms.

- [x] collect crash dump on Windows (without procdump) - waiting for Task exceptions to work - https://github.com/dotnet/runtime/pull/40869 ready for intergration hopefully
- [x] collect hang dumps on OSX without sudo - waiting for preview 8
- [x] revert: attach sequence even when all tests finished
- [x] ^^^ print message to console, when all tests finish, because otherwise AzDO would try to parse the sequence file
- [x] collect hang dumps on linux
- [x] collect hang dump on Windows
- [x] collect hang dumps on OSX with sudo

- [x] collect crash dumps on linux
- [x] collect crash dump on Windows (with procdump)
- [x] collect crash dumps on OSX

- [x] collect dumps from child processes  #2498

- [x] report the test that was hanging not the one that ran last #2499

- [x] add --blame syntax to dotnet test+dll and dotnet vstest
- [x] ensure we are able to run on net5.1 and newer -> (!string.IsNullOrWhiteSpace(targetFramework) && !targetFramework.Contains("v5.0"))

- [x] attach sequence even when all tests finished

- [x] add dump type for hang with netclient helper
- [x] add dump type with netclient crash dumper

~- [ ] report tests affected by testhost crash as failed, instead of NotRun https://github.com/microsoft/vstest/issues/2506~

- [x] add parameter to specify custom location for dump creation and skip adding it to attachments - added VSTEST_DUMP_PATH

- [x] report the correct test when crashing - [feedback ticket](https://developercommunity.visualstudio.com/content/problem/762293/the-active-test-run-was-aborted-reason-test-host-p.html?childToView=1152387#comment-1152387) - we don't actually know the test that crashed for sure, and we cannot know because any background thread can kill testhost, but I improved the message to bring attention to the callstack, and to point out that the failing test is the one that run, but not necessarilly the one that caused the issue.

- [x] validate crash dump on Windows (without procdump) - waiting for Task exceptions to work - https://github.com/dotnet/runtime/pull/40869 ready for intergration hopefully, waiting for this to be merged to have the dumps work correctly again, because now it won't dump testhost. https://github.com/dotnet/installer/pull/8129 - 🙌 works with 5.0.100-rc.1.20420.2

[...]

PR #2532, opened 2020-08-20
Crash dumps via net client

nohwnd:
## Description
Few more fixes, like the process tree and not suspending processes on mac and linux so we don't hang the hang dumping.

## Related issue
Related #2380


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -189,7 +185,23 @@ public BlameCollector()
         private void CollectDumpAndAbortTesthost()
         {
             this.inactivityTimerAlreadyFired = true;
-            var message = string.Format(CultureInfo.CurrentUICulture, Resources.Resources.InactivityTimeout, (int)this.inactivityTimespan.TotalMinutes);
+
+            string value;
+            string unit;
+
+            if (this.inactivityTimespan.TotalSeconds <= 90)
```
cvpoienaru, line 192: nit: Maybe use a constant instead of "90" ?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -590,6 +603,12 @@ private string GetDumpDirectory()
             var dumpDirectory = dumpDirectoryOverrideHasValue ? dumpDirectoryOverride : this.GetTempDirectory();
             Directory.CreateDirectory(dumpDirectory);
             var dumpPath = Path.Combine(Path.GetFullPath(dumpDirectory));
+
+            if (!this.uploadDumpFiles)
+            {
+                this.logger.LogWarning(this.context.SessionDataCollectionContext, $"VSTEST_DUMP_PATH is specified. Dump files will be saved in: {dumpPath}, and won't be added to attachments.");
```
cvpoienaru, line 609: nit: Should we use a resource string instead ?

cvpoienaru (approved):

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: This will be seen by virtually no-one outside of Helix infra. But in general yes.

PR #2544, opened 2020-08-27
Remove nuget frameworks dependency

nohwnd:
> ℹ️ Replaced by https://github.com/microsoft/vstest/pull/4693

## Description
Translate short and long framework names ourselves instead of depending on Nuget.Frameworks

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/Framework.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Framework.cs
@@ -3,14 +3,592 @@
 
 namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
 {
-    using NuGet.Frameworks;
-    using static NuGet.Frameworks.FrameworkConstants;
+    using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// Class for target Framework for the test container
     /// </summary>
     public class Framework
     {
+        #region Generated code
+        // Generated from Nuget.Frameworks 5.7.0.7 nuget package, 
+        // you can update it by scripts/generate/update-supported-nuget.frameworks-versions.ps1
+        internal static Dictionary<string, Framework> mapping = new Dictionary<string, Framework>
+        {
+            [".NETFramework,Version=v1.1"] = new Framework
+            {
+                Name = ".NETFramework,Version=v1.1",
+                FrameworkName = ".NETFramework",
+                Version = "1.1.0.0",
+                ShortName = "net11",
+            },
+            [".NETFramework,Version=v2.0"] = new Framework
+            {
+                Name = ".NETFramework,Version=v2.0",
+                FrameworkName = ".NETFramework",
+                Version = "2.0.0.0",
+                ShortName = "net20",
+            },
+            [".NETFramework,Version=v3.5"] = new Framework
+            {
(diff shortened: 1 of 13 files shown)


nohwnd:
@vritant24 this is what I need to do to replace Nuget.Frameworks. Are you able to validate that this would fix the LUT issue? 

Maybe we should add additional logic that will create a netcoreapp like Framework object from any net* string as a fallback? that way when net5.1 or net6.0 are finally added we don't have to update the class as long as they are in the form of net<version>. I think something similar is happening in the SDK because setting the TFM to `net7.99.123` will produce:   The current .NET SDK does not support targeting .NET Core 7.99.123.  Either target .NET Core 5.0 or lower, or use a version of the .NET SDK that supports .NET Core 7.99.123.

src/Microsoft.TestPlatform.ObjectModel/Framework.cs
```diff
@@ -3,14 +3,592 @@
```
Haplois, line 3: If I am not mistaken, this file is auto-generated. Maybe we should add an auto-gen header?

Haplois (approved):

src/Microsoft.TestPlatform.ObjectModel/Framework.cs
nohwnd: It's not autogenerated. Just part of it if generated from something else, but it won't automatically update.

src/Microsoft.TestPlatform.ObjectModel/Framework.cs
```diff
@@ -3,14 +3,592 @@
...
+                ShortName = "net461",
+            },
+            [".NETFramework,Version=v4.6.2"] = new Framework
+            {
+                Name = ".NETFramework,Version=v4.6.2",
+                FrameworkName = ".NETFramework",
+                Version = "4.6.2.0",
+                ShortName = "net462",
+            },
+            [".NETFramework,Version=v4.6.3"] = new Framework
```
nohwnd, line 96: Huh, missing .NETFramework 47*?

vritant24:
I'm not sure of the entire information that the testplatform has access to, but it looks like currently it is operating on `TargetFramework`. There is also `TargetFrameworkMoniker` that is much more consistent across the different runtimes. 

[This project system doc](https://github.com/dotnet/project-system/blob/master/docs/repo/coding-conventions.md#data) elaborates on that reasoning. For instance in the .NET 5 upgrade, the `TargetFramework` for is `net5.0` but the `TargetFrameworkMoniker` is `.NetCoreApp,Version=5.0` and didn't change from .NetCore 3.1

nohwnd:
@vritant24 thanks for the tip, but imho that is relevant only to msbuild and project properties, we don't use any of those. We look at TFM attribute in the assembly, because vstest console handles only built dlls.

raghavthind2005 (approved):

nohwnd:
@MarcoRossignoli that merge does not look that bad, might have a look on it when I am back.

MarcoRossignoli:
> @MarcoRossignoli that merge does not look that bad, might have a look on it when I am back.

Yep doesn't look so bad and we should catch issue with new tfm really soon in the dogfooding/insertion process.

PR #2553, opened 2020-09-01
Blame upload on crash even if hang dump started

nohwnd:
## Description
 In rare cases where we have a crashing tests and short hang timeout, we get into situation where hang dumper started dumping, but then session ends because of the crash, and dumps are not uploaded because the session end is only uploading when there is crash dump enabled. After this change we look for any dump in any case and upload all the dumps that hang dumper was able to create.

Remove test, because we want to upload all dumps on session end when hangdump and crash dump run into each other, otherwise hang dumps would not get uploaded and the process would just end, leaving them on the local disk.

## Related issue
Related #2380

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -46,7 +46,6 @@ namespace Microsoft.TestPlatform.Extensions.BlameDataCollector
         private IInactivityTimer inactivityTimer;
         private TimeSpan inactivityTimespan = TimeSpan.FromMinutes(DefaultInactivityTimeInMinutes);
         private int testHostProcessId;
-        private bool dumpWasCollectedByHangDumper;
         private string targetFramework;
         private List<KeyValuePair<string, string>> environmentVariables = new List<KeyValuePair<string, string>>();
         private bool uploadDumpFiles;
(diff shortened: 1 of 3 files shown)


Haplois (approved):

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -452,47 +450,37 @@ private void SessionEndedHandler(object sender, SessionEndEventArgs args)
...
-                        if (this.uploadDumpFiles)
+                        var dumpFiles = this.processDumpUtility.GetDumpFiles();
+                        foreach (var dumpFile in dumpFiles)
                         {
-                            try
+                            if (!string.IsNullOrEmpty(dumpFile))
                             {
-                                var dumpFiles = this.processDumpUtility.GetDumpFiles();
-                                foreach (var dumpFile in dumpFiles)
+                                try
```
cvpoienaru, line 462: nit: It would be nice to have the contents of the `foreach` loop in a separate method, something like `TransferDumpFile`.

cvpoienaru (approved):

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: I'd love to hear your motivation for that. Won't fix right now.

PR #2554, opened 2020-09-01
Restore netcoreapp1.0 support for testhost

sharwell:
This pull request updates the build to include netcoreapp1.0 outputs for testhost.exe and testhost.x86.exe. It does not (yet) update tests to cover these cases, or NuGet packaging to include the output binaries.


Haplois:
All looks OK, but I am converting this to draft, I'll add new targets to NuGet packages and tests and then merge.

Haplois (approved):

ViktorHofer:
@sharwell what's the reason for adding/restoring netcoreapp1.0 support? Wondering as netcoreapp1.0 is EOL for a long time.

sharwell:
@ViktorHofer It allows multi-targeting projects to move forward with adding support for the latest runtime targets without blocking adoption of new targets on the removal of an unrelated one. The issue I hit directly involved analyzer testing for Roslyn analyzers that support Visual Studio 2015, which is still supported. We could carve out slightly different cases over time, where the .NET Framework tools from 2015 keep working but the .NET Core tooling just breaks at some point, but that seemed unnecessary considering the ease with which .NET Core 1.x can be supported by the test runner.

PR #2577, opened 2020-09-18
Excluded some of already signed assemblies

Haplois:
- Refactored our `sign.proj` file exclude already signed files from signing, and fixed some issues which were causing warnings.
 - A new signature verifying script. (Which is not in use yet)

### DIFF
--- a/scripts/Test-Signatures.ps1
+++ b/scripts/Test-Signatures.ps1
@@ -29,9 +29,24 @@ $global:ValidThumbprints = @( `
 )
 
 # Allow-list for unsigned binaries
-$global:UnsignedAllowList = @( `
-    "Newtonsoft.Json.dll" `
-);
+$global:UnsignedAllowList = @{
+    # "PackageId" = relative paths
+    "Microsoft.TestPlatform" = @( `
+        ".\tools\net451\Common7\IDE\Extensions\TestPlatform\Newtonsoft.Json.dll" `
+    )
+
+    "Microsoft.TestPlatform.CLI" = @( `
+        ".\contentFiles\any\netcoreapp2.1\TestHost\Newtonsoft.Json.dll", `
+        ".\contentFiles\any\netcoreapp2.1\Newtonsoft.Json.dll"
+    )
+    
+    "Microsoft.TestPlatform.Portable" = @( `
+        ".\tools\net451\Newtonsoft.Json.dll", `
+        ".\tools\netcoreapp2.1\TestHost\Newtonsoft.Json.dll", `
+        ".\tools\netcoreapp2.1\Newtonsoft.Json.dll"
+    )
+}
+
 
 $global:Depth = 0;
 
@@ -134,6 +149,12 @@ function Test-Signatures($rootFolder) {
     $files = (Get-ChildItem $rootFolder -Recurse -Include "*.dll", "*.exe")
     Update-Operation -operation "discovered" -message "$($files.Length) binaries discovered"
 
+    $packageId = ([XML](Get-Content $rootFolder\*.nuspec)).package.metadata.id;
+    $unsignedAllowList = @();
+    if ($global:UnsignedAllowList.ContainsKey($packageId)) {
+        $unsignedAllowList = $global:UnsignedAllowList[$packageId];
+    }
+
     $files | ForEach-Object {
         $relativePath = Get-RelativePath -path $_.FullName -relativeTo $rootFolder
         Update-Operation -operation "processing" -message $relativePath -suppress $CIBuild
@@ -148,7 +169,7 @@ function Test-Signatures($rootFolder) {
                 Update-Operation -operation "inconclusive" -message "$relativePath ($($signature.SignerCertificate.Subject)) [$($signature.SignerCertificate.Thumbprint)]" -NewLine
                 $fail.Add($_.FullName, $signature)
             }
-        } elseif (-not $global:UnsignedAllowList.Contains([System.IO.Path]::GetFileName($_.FullName))) {
+        } elseif (-not $unsignedAllowList.Contains($relativePath)) {
             Undo-Operation -operation "unsigned" -message "$relativePath" -NewLine
             $fail.Add($_.FullName, $null)
         }
@@ -184,5 +205,5 @@ Pop-Location
 if($failCount -eq 0) {
     Write-Host "All binaries passed the check successfully."
 } else {
-    Write-Error "Signature verification failed on $failCount binaries!"
+    Write-Error "$failCount binaries failed signature check!"
 }
--- a/scripts/Test-Signatures.ps1
+++ b/scripts/Test-Signatures.ps1
@@ -1,10 +1,8 @@
 [CmdletBinding()]
 param (
-    [Parameter()]
     [String]
     $Path = "$PSScriptRoot\packages",
 
-    [Parameter()]
     [Switch]
     $CIBuild
 )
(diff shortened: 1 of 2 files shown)


Haplois:
@nohwnd, can you look at the `.ps1` I added? It's not part of any other script, and there to solely replace `verify-sign.ps1` when feature complete. It checks all binaries unlike `verify-sign`.

Currently it doesn't check NuGet packages themselves.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
+[CmdletBinding()]
+param (
+    [Parameter()]
```
nohwnd, line 3: Nit: this is not needed when you don't customize anything, did you mean to use `Mandatory`?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+
+    [Parameter()]
+    [Switch]
+    $CIBuild
+)
+
+Add-Type -AssemblyName System.IO.Compression.FileSystem
+
+# Allowed certificate subjects
+$global:ValidSubjects = @( `
```
nohwnd, line 15: Global is rarely needed. You should pass all your values by parameters, and assign them from output. If you really need to scope value outside of function use $script scope.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+    "CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US", `
+    "CN=Microsoft Corporation, OU=MOPR, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" `
+);
+
+# Allowed certificate thumbprints
+$global:ValidThumbprints = @( `
```
nohwnd, line 21: Grave accents `` ` `` are not needed here, and neither are commas `,`.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+    "98ED99A67886D020C564923B7DF25E9AC019DF26", `
+    "5EAD300DC7E4D637948ECB0ED829A072BD152E17", `
+    "67B1757863E3EFF760EA9EBB02849AF07D3A8080", `
+    "9DC17888B5CFAD98B3CB35C1994E96227F061675", `
+    "62009AAABDAE749FD47D19150958329BF6FF4B34", `
+    "3BDA323E552DB1FDE5F4FBEE75D6D5B2B187EEDC" `
+)
+
+# Allow-list for unsigned binaries
+$global:UnsignedAllowList = @{
```
nohwnd, line 32: nit: again, commas or grave accents not needed.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+        ".\tools\netcoreapp2.1\TestHost\Newtonsoft.Json.dll", `
+        ".\tools\netcoreapp2.1\Newtonsoft.Json.dll"
+    )
+}
+
+
+$global:Depth = 0;
+
+function Get-RelativePath($path, $relativeTo) {
+    Push-Location
```
nohwnd, line 54: Just `Join-Path $relativeTo $path` and then resolve it? No need to jump around directories.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+    Set-Location $relativeTo
+    $path = Resolve-Path $path -Relative
+    Pop-Location
+    
+    return $path
```
nohwnd, line 59: nit: return is not needed in powershell unless you return early (e.g. from `if`), all output values are written to the pipeline, not just the ones marked with return. e.g. 

```powershell
$path
return $path
```

Would return the path twice, not just once.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+function Pop-Depth {
+    $global:Depth = $global:Depth - 1
+
+    if($global:Depth -lt 0) {
+        $global:Depth = 0;
+    }
+}
+
+function Clear-Line($message = "") {
+    if(-not $CIBuild) {
```
nohwnd, line 75: If cibuild then return? to keep it consistent with the other functions?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+        $x = [System.Console]::get_CursorLeft()
+        Write-Host "`r$(' ' * $x)$message`r" -NoNewline
+    }
+}
+
+function Update-Operation($operation, $message, [switch]$NewLine, $suppress=$false) {
+    if($suppress) {
+        return;
```
nohwnd, line 83: nit: No semicolons at the end of lines are needed. You are not using it consistently everywhere anyway.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+    }
+    
+    Clear-Line;
+    Write-Host "$('  ' * $Depth)[$operation] " -ForegroundColor Yellow -NoNewline
```
nohwnd, line 87: Rather than inventing you own reporting use Write-Progress, or write Write-Verbose.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+    Write-Host $message -NoNewline
+
+    if ($newline -or $CIBuild) {
+        Write-Host;
+    }
+}
+
+function Submit-Operation($operation, $message, [switch]$NoNewLine, $suppress=$false) {
```
nohwnd, line 95: Why do you mix [switch] and [bool] assigned to $false?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+function Expand-NugetPackages {
+    $foldersToProcess = [System.Collections.Generic.HashSet[string]]@()
+
+    $nupgks = Get-ChildItem *.nupkg
+
+    Push-Depth
+    $nupgks | ForEach-Object {
+        $package = $_.FullName
+        $expand = Join-Path -Path ([System.IO.Path]::GetDirectoryName($package)) -ChildPath ([System.IO.Path]::GetFileNameWithoutExtension($package))
+        ((Test-Path -Path $expand) -and (Remove-Item $expand -Force -Recurse)) | Out-Null
```
nohwnd, line 132: What is this supposed to mean? Is this just unnecessarily clever way of if (path exitst) { delete it } ?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+        $unsignedAllowList = $global:UnsignedAllowList[$packageId];
+    }
+
+    $files | ForEach-Object {
+        $relativePath = Get-RelativePath -path $_.FullName -relativeTo $rootFolder
+        Update-Operation -operation "processing" -message $relativePath -suppress $CIBuild
+
+        $signature = Get-AuthenticodeSignature -FilePath "$($_.FullName)"
+        if ($signature.Status -eq "Valid") {
+            $valid = ($ValidSubjects.Contains($signature.SignerCertificate.Subject) -or $ValidThumbprints.Contains($signature.SignerCertificate.Thumbprint))
```
nohwnd, line 164: Pass the values by parameters, or use the variable qualification to show that this is coming in externally.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+
+        $foldersToProcess.Add($expand) | Out-Null
+    }
+    Pop-Depth
+    Write-Host;
+    return $foldersToProcess
+}
+
+function Test-Signatures($rootFolder) {
+    $fail = @{}
```
nohwnd, line 146: $failed?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+
+    $nupgks = Get-ChildItem *.nupkg
+
+    Push-Depth
+    $nupgks | ForEach-Object {
+        $package = $_.FullName
+        $expand = Join-Path -Path ([System.IO.Path]::GetDirectoryName($package)) -ChildPath ([System.IO.Path]::GetFileNameWithoutExtension($package))
+        ((Test-Path -Path $expand) -and (Remove-Item $expand -Force -Recurse)) | Out-Null
+
+        Update-Operation -operation "extracting" -message "$([System.IO.Path]::GetFileName($package))"
```
nohwnd, line 134: All parameters should start with Capital letter, see what you do with Test-Path, why do it differently for this function?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+}
+
+function Expand-NugetPackages {
+    $foldersToProcess = [System.Collections.Generic.HashSet[string]]@()
+
+    $nupgks = Get-ChildItem *.nupkg
+
+    Push-Depth
+    $nupgks | ForEach-Object {
+        $package = $_.FullName
```
nohwnd, line 130: If you do Get-Item here you can use `$package.Directory`, `$package.BaseName` etc. instead of the long winded IO.Path methods.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,209 @@
...
+        $expand = Join-Path -Path ([System.IO.Path]::GetDirectoryName($package)) -ChildPath ([System.IO.Path]::GetFileNameWithoutExtension($package))
+        ((Test-Path -Path $expand) -and (Remove-Item $expand -Force -Recurse)) | Out-Null
+
+        Update-Operation -operation "extracting" -message "$([System.IO.Path]::GetFileName($package))"
+        [System.IO.Compression.ZipFile]::ExtractToDirectory($package, $expand)
+        Submit-Operation -operation "done" -message "$([System.IO.Path]::GetFileName($package))" -suppress $CIBuild
```
nohwnd, line 136: This won't work properly if something fails. You would need to wrap that into finally, but that would suck.

nohwnd (commented):
I don't get the whole *-Operation thing. It seems like it should be replaced by Write-Verbose or Write-Progress.

Lot of the feedback is style. Further reading here if you are interested: https://github.com/PoshCode/PowerShellPracticeAndStyle

scripts/Test-Signatures.ps1
Haplois: Yep, I am trying to stop using it, but c# hooks you up :)

scripts/Test-Signatures.ps1
Haplois: No, I just used the snippet :)

scripts/Test-Signatures.ps1
Haplois: This function actually returns the relative path. This was the safest way of doing that I could think of.

Example:
```
      Path: c:\git\package\tools\netcoreapp2.1\Newtonsoft.Json.dll
RelativeTo: c:\git\package
    Output: .\tools\netcoreapp2.1\Newtonsoft.Json.dll
```

scripts/Test-Signatures.ps1
Haplois: I wanted to have a cleaner output. (Not hundreds of assemblies listed in the output unless needed) I'll try to switch that.

scripts/Test-Signatures.ps1
Haplois: Yes :) Fixing.

scripts/Test-Signatures.ps1
Haplois: I still do blame c# :)

scripts/Test-Signatures.ps1
nohwnd: Ah okay, Use try finally for the pushing and popping then?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
+[CmdletBinding()]
+param (
+    [String]
+    $Path = "$PSScriptRoot\packages",
+
+    [Switch]
+    $CIBuild
+)
+
+Add-Type -AssemblyName System.IO.Compression.FileSystem
```
nohwnd, line 10: Is this used to work with archives? There are Compress-Archive and Expand-Archive cmdlets since powershell 5, I'd use that if possible.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+    [String]
+    $Path = "$PSScriptRoot\packages",
+
+    [Switch]
+    $CIBuild
+)
+
+Add-Type -AssemblyName System.IO.Compression.FileSystem
+
+# Allowed certificate subjects
```
nohwnd, line 12: nit: those comments are just stating the same as the var names.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+    "5EAD300DC7E4D637948ECB0ED829A072BD152E17" 
+    "67B1757863E3EFF760EA9EBB02849AF07D3A8080" 
+    "9DC17888B5CFAD98B3CB35C1994E96227F061675" 
+    "62009AAABDAE749FD47D19150958329BF6FF4B34" 
+    "3BDA323E552DB1FDE5F4FBEE75D6D5B2B187EEDC" 
+    "899FA016DEE8E665FF2A315A1151C43FB96C430B"  # Microsoft 3rd Party Application Component
+)
+
+# Allow-list for unsigned binaries
+# No binaries allow to ship without a valid signature as per policy
```
nohwnd, line 31:
```suggestion
# Per policy no binaries are allowed to ship without a valid signature. 

```

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+$script:UnsignedAllowList = @{
```
nohwnd, line 32: This is not a list, what is the intent here?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+function Clear-Line($Message = "") {
+    if($CIBuild) {
+        return
+    }
+    
+    $x = [System.Console]::get_CursorLeft()
+    Write-Host "`r$(' ' * $x)$Message`r" -NoNewLine
+}
+
+function Update-Operation($Operation, $Message, [switch]$NewLine, [switch]$Suppress) {
```
nohwnd, line 75: I commented on this before. Why is this not using the built in Write-Progress functionality?

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+    Clear-Line
+    Write-Host "$('  ' * $Depth)[$Operation] " -ForegroundColor Red -NoNewline
+    Write-Host $Message -NoNewline
+
+    if (-not $NoNewLine -or $CIBuild) {
+        Write-Host
+    }
+}
+
+Function Invoke-Command ($Command, $Arguments)
```
nohwnd, line 117: nit: non-capital f

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+            $failed.Add($_.FullName, $null)
+        }
+    }
+
+    if($failed.Count -eq 0) {
+        Submit-Operation -operation "done" -message "$($nupgks.Length) packages passed signature checks."
+    } elseif ($failed.Count -eq $nupgks.Length) {
+        Undo-Operation -operation "failed" -message "$($nupgks.Length) packages failed signature checks." 
+    } else {
+        Undo-Operation -operation "partial" -message "$($nupgks.Length - $failed.Count) packages passed signature checks and $($failed.Count) packages failed!"
```
nohwnd, line 182: I don't get why this distinction is necessary, seeing that 0 packages passed is good.

scripts/Test-Signatures.ps1
```diff
@@ -0,0 +1,297 @@
...
+    if($failed.Count -eq 0) {
+        Submit-Operation -operation "done" -message "$($nupgks.Length) packages passed signature checks."
+    } elseif ($failed.Count -eq $nupgks.Length) {
+        Undo-Operation -operation "failed" -message "$($nupgks.Length) packages failed signature checks." 
+    } else {
+        Undo-Operation -operation "partial" -message "$($nupgks.Length - $failed.Count) packages passed signature checks and $($failed.Count) packages failed!" 
+    }
+
+    Write-Host
+    Pop-Depth
```
nohwnd, line 186: This will not run if anything above will fail, is that exptected?

scripts/Test-Signatures.ps1
nohwnd: You marked this as resolved but don't actually do it.

nohwnd (changes requested):
Why is it necessary to write a new script for testing signatures from scratch instead of re-factoring the current one. What makes this one one much better then the current way it is done?

scripts/Test-Signatures.ps1
Haplois: They don't work if the file extension is not a valid archive.

nohwnd: kk

Haplois:
Abandoned for now.

PR #2584, opened 2020-09-29
Early testhost startup performance work

### LINKED ISSUE #2629: Early testhost startup performance improvement for test execution
Deliver test execution performance improvements by starting the testhost early and brushing off that time from execution by decoupling the two operations.

cvpoienaru:
## Description
Early testhost startup performance improvements


src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -526,6 +518,48 @@ private void StartTestRunAttachmentsProcessing(TestRunAttachmentsProcessingPaylo
...
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="payload"></param>
+        /// <param name="testRequestManager"></param>
+        private void StartTestRunner(StartTestRunnerPayload payload, ITestRequestManager testRequestManager)
+        {
+            Task.Run(
+                delegate
```
nohwnd, line 529: Is there a reason this is delegate and not lambda `() =>`? I never saw it used like this.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -31,7 +32,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Client
     /// <summary>
     /// Implementation for TestPlatform
     /// </summary>
-    internal class TestPlatform : ITestPlatform
+    public class TestPlatform : ITestPlatform
```
nohwnd, line 35: can this be solved by IVT (Internals Visible To)?

src/Microsoft.TestPlatform.CrossPlatEngine/TestRunner/ProxyStartTestRunnerManager.cs
```diff
@@ -41,15 +43,21 @@ public void StartTestRunner(StartTestRunnerCriteria criteria, IStartTestRunnerEv
...
+                    {
+                        var operationManagerProxy = this.proxyCreator();
+                        operationManagerProxy.SetupChannel(criteria.Sources, criteria.RunSettings);
 
-                TestRunnerPool.Instance.AddProxy(session, operationManagerProxy);
+                        TestRunnerPool.Instance.AddProxy(session, operationManagerProxy);
+                    }));
             }
 
+            Task.WaitAll(taskList.ToArray());
```
Sanan07, line 59: You can create an array of the length equals to  `parallelLevel` and send it to `Task.WaitAll` instead of having list and converting it to array.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
cvpoienaru: Just for the sake of consistency, this is the way to start a Task the rest of the methods in this file adhere to.

src/Microsoft.TestPlatform.CrossPlatEngine/TestRunner/ProxyStartTestRunnerManager.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
```diff
@@ -68,6 +68,8 @@ public static class TelemetryDataConstants
 
         public static string ParallelEnabledDuringDiscovery = "VS.TestDiscovery.ParallelEnabled";
 
+        public static string ParallelEnabledDuringStartTestSession = "VS.StartTestSession.ParallelEnabled";
```
Sanan07, line 71: Why are these fields are not constants marked with 'const' keyword?

cvpoienaru: The existing fields were already marked as `static` so I marked mine `static` too for the sake of consistency.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -0,0 +1,217 @@
...
+            // Let the caller know the session has been created.
+            eventsHandler.HandleStartTestSessionComplete(testSessionInfo);
+        }
+
+        /// <inheritdoc/>
+        public void StopSession()
+        {
+            foreach (var kvp in this.proxyMap)
+            {
+                // TODO (copoiena): Do nothing for now because in the current implementation the
```
Sanan07, line 92: If do nothing for now, it is better to comment or remove foreach iteration for unnecessary looping.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -0,0 +1,217 @@
...
+                }
+
+                // Get the proxy id from the available queue.
+                var proxyId = this.availableProxyQueue.Dequeue();
+
+                // Get the actual proxy.
+                proxyContainer = this.proxyMap[proxyId];
+
+                // Mark the proxy as unavailable.
+                proxyContainer.Available = false;
```
Sanan07, line 128: Better change to IsAvailable property

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -185,7 +204,7 @@ private void ProcessRequests(ITestRequestManager testRequestManager)
                                 var testRunPayload =
                                     this.communicationManager.DeserializePayload<TestRunRequestPayload>(
                                         message);
-                                this.StartTestRun(testRunPayload, testRequestManager, skipTestHostLaunch: true);
+                                this.StartTestRun(testRunPayload, testRequestManager, shouldLaunchTesthost: true);
```
drognanar, line 207: is this correct? it seems that the parameter changed the meaning, but we didn't change the value

src/Microsoft.TestPlatform.CrossPlatEngine/EventHandlers/TestRequestHandler.cs
```diff
@@ -24,6 +24,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities
     public class TestRequestHandler : ITestRequestHandler
     {
         private int protocolVersion = 1;
+        
+        // Also check TestRequestSender.
```
Haplois: Nit: We probably need to clarify this comment. It took a while for me to understand it.

```suggestion
        // Must be in sync with the highest supported version in src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs file.
```

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -52,6 +52,7 @@ public class TestRequestSender : ITestRequestSender
         // that implies host is using version 1.
         private int protocolVersion = 1;
 
+        // Also check TestRequestHandler.
```
Haplois: Nit: We probably need to clarify this comment. It took a while for me to understand it.

```suggestion
        // Must be in sync with the highest supported version in src/Microsoft.TestPlatform.CrossPlatEngine/EventHandlers/TestRequestHandler.cs file.
```

Haplois (commented):
LGTM with some nit.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -352,47 +416,116 @@ public void ProcessTestRunAttachments(TestRunAttachmentsProcessingPayload attach
...
+            if (this.UpdateRunSettingsIfRequired(
+                payload.RunSettings,
+                payload.Sources,
+                null,
+                out string updatedRunsettings))
+            {
+                payload.RunSettings = updatedRunsettings;
+            }
+
+            if (InferRunSettingsHelper.AreRunSettingsCollectorsInCompatibleWithTestSettings(payload.RunSettings))
```
drognanar, line 452: nit: incompatible should be a single word

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -352,47 +416,116 @@ public void ProcessTestRunAttachments(TestRunAttachmentsProcessingPayload attach
...
+            {
+                throw new SettingsException(
+                    string.Format(
+                        Resources.RunsettingsWithDCErrorMessage,
+                        payload.RunSettings));
+            }
+
+            // TODO (copoiena): Collect metrics ?
+
+            lock (this.syncObject)
```
drognanar, line 462: it looks like the CreateStartTestSessionRequest isn't necessarily an immediate operation. does this mean that test sessions on our side can only be spawned sequentially? I noticed that there is just a global TestRequestManager instance.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
cvpoienaru: Good point, thanks. The meaning was wrong before but now it accurately reflects the intent. When the parameter is set on false we use no custom launcher, but we do when it's set on true.

src/Microsoft.TestPlatform.CrossPlatEngine/EventHandlers/TestRequestHandler.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
cvpoienaru: Done.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.CoreUtilities/Tracing/TestPlatformInstrumentationEvents.cs
```diff
@@ -187,5 +187,45 @@ internal class TestPlatformInstrumentationEvents
...
+
+        /// <summary>
+        /// The translation layer start test session stop event id.
+        /// </summary>
+        public const int TranslationLayerStartTestSessionStopEventId = 0x49;
+
+        /// <summary>
+        /// The stop test session start event id.
+        /// </summary>
+        public const int StopTestSessionStartEventId = 0x50;
```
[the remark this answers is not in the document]
cvpoienaru: Good point, thanks. I just checked and every other event id in this file acts like the number is not hex. For compatibility sake I won't attempt to fix those but only the new event ids that I introduced.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -161,6 +156,60 @@ public ITestRunRequest CreateTestRunRequest(IRequestData requestData, TestRunCri
             return new TestRunRequest(requestData, testRunCriteria, executionManager, loggerManager);
         }
 
+        /// <inheritdoc/>
+        public void CreateStartTestSessionRequest(
```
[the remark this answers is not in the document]
cvpoienaru: Done, renamed it to `StartTestSession`.

src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
```diff
@@ -43,86 +43,121 @@ public TestEngine() : this(TestRuntimeProviderManager.Instance, new ProcessHelpe
...
 
-                return new ProxyDiscoveryManager(requestData, new TestRequestSender(requestData.ProtocolConfig, hostManager), hostManager);
+                return new ProxyDiscoveryManager(
+                    requestData,
+                    new TestRequestSender(requestData.ProtocolConfig, hostManager),
+                    hostManager);
             };
 
-            return !testHostManager.Shared ? new ParallelProxyDiscoveryManager(requestData, proxyDiscoveryManagerCreator, parallelLevel, sharedHosts: testHostManager.Shared) : proxyDiscoveryManagerCreator();
+            return !testHostManager.Shared
```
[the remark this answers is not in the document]
cvpoienaru: Done.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSessionAsync.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        /// 
+        /// <param name="sources">The list of source assemblies for the discovery.</param>
+        /// <param name="discoverySettings">The run settings for the discovery.</param>
+        /// <param name="discoveryEventsHandler">The discovery event handler.</param>
+        /// <returns></returns>
+        Task DiscoverTestsAsync(
+            IEnumerable<string> sources,
+            string discoverySettings,
+            ITestDiscoveryEventsHandler discoveryEventsHandler);
```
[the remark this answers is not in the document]
cvpoienaru: Mainly maintaining compatibility with `IVsTestConsoleWrapper.cs`/`IVsTestConsoleWrapperAsync.cs` interface. I don't really expect the caller to use it though.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -177,75 +235,285 @@ public void CancelDiscovery()
...
+                    sources,
+                    runSettings,
+                    options,
+                    eventsHandler,
+                    testHostLauncher),
+                this);
+        }
+
+        /// <inheritdoc/>
+        public bool StopTestSession(
```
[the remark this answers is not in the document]
cvpoienaru: It's necessary to have it exposed on `IVsTestConsoleWrapper` because `ITestSession` is just a wrapper over it and the calls get forwarded to the underlying `IVsTestConsoleWrapper` object. In this way we also keep the symmetry of the API (we have both `StopTestSession` and `StartTestSession` on the same `IVsTestConsoleWrapper` object).

I decided to expose it on `ITestSession` for the convenience of the caller as well as because you'll need to actually pass a `TestSessionInfo` object to the underlying `IVsTestConsoleWrapper`, however this object is internally managed by the `ITestSession` and it's private.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
cvpoienaru: That's correct, there's only one instance of the `TestRequestManager` and all operations are sequential, including test discovery and test runs. They're not immediate operations either, meaning we cannot go ahead and execute another request while the current one running. This has been a design limitation for quite some time now and I think it's worth exploring changing this behavior in a PR of its own as I anticipate it'll require significant changes.

However, if we take only test session spawning into consideration this shouldn't be much of an issue as the plan is to spawn the test session(s) during build. Regarding multiple test sessions, is there any real use case for this on the TW side ?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
```diff
@@ -19,112 +20,307 @@ public interface IVsTestConsoleWrapper : IVsTestConsoleWrapperAsync
...
-        void RunTestsWithCustomTestHost(IEnumerable<TestCase> testCases, string runSettings, ITestRunEventsHandler testRunEventsHandler, ITestHostLauncher customTestHostLauncher);
+        /// 
+        /// <param name="sources">The list of source assemblies for the test run.</param>
+        /// <param name="runSettings">The run settings for the run.</param>
+        /// <param name="options">The test platform options.</param>
+        /// <param name="eventsHandler">The session event handler.</param>
+        /// <param name="testHostLauncher">The custom host launcher.</param>
+        /// 
+        /// <returns>A test session info object.</returns>
+        ITestSession StartTestSession(
```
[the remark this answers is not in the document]
cvpoienaru: Restarting sessions shouldn't be a thing.

The current PR enables spawning a session that has one or more testhosts attached to it and only one request per session can be served. This is because the default behavior is to terminate the testhost(s) once the request (discovery/run) was successfully completed.

The next round of changes should enable testhost sharing between discovery and run and should disable the default testhost termination behavior. This means the testhost(s) will be kept alive until the session is terminated with an explicit call from the API user (TW for example).

The only possible need for session restarting would be when the runsettings/list of sources change between discovery and execution. This will require restarting the testhosts, but at this point I think we're better off with terminating the current session and just starting a new session with the updated settings/list of sources. From the performance standpoint the cost of those operations should be about the same.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSession.cs
```diff
@@ -0,0 +1,180 @@
...
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.Interfaces;
+
+    /// <summary>
+    /// Defines a test session that can be used to make calls to the vstest.console
+    /// process.
+    /// </summary>
+    public interface ITestSession : ITestSessionAsync
+    {
+        /// <summary>
+        /// Starts test discovery.
```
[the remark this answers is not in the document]
cvpoienaru: Good point, thanks. The discovery/execution are indeed blocking but they're also cancellable. The current running operation either terminates normally or is cancelled.

What would be a better word for this ?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSession.cs
```diff
@@ -0,0 +1,180 @@
...
+        void DiscoverTests(
+            IEnumerable<string> sources,
+            string discoverySettings,
+            TestPlatformOptions options,
+            ITestDiscoveryEventsHandler2 discoveryEventsHandler);
+
+        /// <summary>
+        /// Cancels the last discovery request.
+        /// </summary>
+        new void CancelDiscovery();
```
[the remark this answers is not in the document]
cvpoienaru: The reason for this is `ITestSession` inherits from `ITestSessionAsync` and `ITestSessionAsync` declares those members first. 

The interface closely mirrors `IVsTestConsoleWrapper` and `IVsTestConsoleWrapperAsync` design and I think the original reasoning was to redeclare the members to be aware such operations exist on both interfaces. Omitting those members and reading only `ITestSession` (or `IVsTestConsoleWrapper`) would make you think the interface is complete and the Async version of the interface implements the Async equivalents, but you won't expect new members coming from the Async version only.

Would you like me to remove it though ?

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// 
+        /// <returns>The dequeued proxy.</returns>
+        public ProxyOperationManager DequeueProxy()
+        {
+            ProxyOperationManagerContainer proxyContainer = null;
+
+            lock (this.lockObject)
+            {
+                // No proxy available means the caller will have to create its own proxy.
+                if (this.availableProxyQueue.Count == 0)
```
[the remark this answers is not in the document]
cvpoienaru: The base assumption here is the runsettings/list of sources doesn't change between the moment the session was spawned and when test discovery/execution is being carried out. If there's any kind of changes to those inputs we'll need to spawn another session.

This being said, the "parallel level" is determined based on the parameters listed above. We'll spawn as many `ProxyOperationManager`s as our parallel level indicates and we'll add those to the available proxy queue. The same list of sources/runsettings should yield the same parallel level and there's a 1:1 mapping between the queued `ProxyOperationManager`s and the `ProxyDiscoveryManager`s or `ProxyExecutionManager`s that would use them.

If there's any kind of internal logic error and we end up asking for more proxies though (this is not ok, of course), this method would throw but the exception will be caught in the `TestEngine` (see `GetExecutionManager` in `TestEngine.cs`). Catching the exception there means we'll spawn a `ProxyExecutionManager` that won't try to dequeue a `ProxyOperationManager` from the pool but instead will create its very own underlying `ProxyOperationManager` independent from the session.

In fact, creating your own underlying `ProxyOperationManager` is standard procedure for the case where there's no `TestSessionInfo` being passed around between the wrapper and `vstest.console.exe`.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -116,15 +114,11 @@ public IDiscoveryRequest CreateDiscoveryRequest(IRequestData requestData, Discov
...
-        /// The create test run request.
-        /// </summary>
-        /// <param name="requestData">Request data.</param>
-        /// <param name="testRunCriteria"> The test run criteria.  </param>
-        /// <param name="options">Test platform options.</param>
-        /// <returns> The <see cref="ITestRunRequest"/>. </returns>
-        /// <exception cref="ArgumentNullException"> Throws if parameter is null. </exception>
-        public ITestRunRequest CreateTestRunRequest(IRequestData requestData, TestRunCriteria testRunCriteria, TestPlatformOptions options)
+        /// <inheritdoc/>
+        public ITestRunRequest CreateTestRunRequest(
```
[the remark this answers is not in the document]
cvpoienaru: Yes, this should be the new normal workflow when sessions are involved. You're required to create the session first (otherwise there'll be no `ITestSession` object. You could still call `RunTests` on the `VsTestConsoleWrapper` but you'll have to provide a `TestSessionInfo` object that you won't have unless you made the `StartTestSession` call first).

You can however call `RunTests` with null for the `TestSessionInfo` parameter and you'll emulate the current behavior of TP where no testhosts are pre-started and they are spawned during the time you fulfill the request.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
drognanar: our extensibility let's users specify a runsettings file to use separately for each csproj. because of that it's possible that different projects will have a different runsettings

hopefully we'll be good with these starting in sequence and we'll need to be smart about stopping/starting sessions to minimize the cases where the session isn't active at the time we need to do a discovery

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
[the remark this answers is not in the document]
drognanar: I meant stopping a session and creating a new one.

besides the settings/list of sources would changing between debugging and profiling also require us to create a new session?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManagerWithDataCollection.cs
```diff
@@ -160,7 +167,7 @@ public override int LaunchProcessWithDebuggerAttached(TestProcessStartInfo testP
         }
 
         /// <inheritdoc />
-        protected override TestProcessStartInfo UpdateTestProcessStartInfo(TestProcessStartInfo testProcessStartInfo)
+        public override TestProcessStartInfo UpdateTestProcessStartInfo(TestProcessStartInfo testProcessStartInfo)
```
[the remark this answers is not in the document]
cvpoienaru: Yes, this is called in the `ProxyOperationManager` which contains an `IBaseProxy` implementation which in turn declares this method.

This is done in order to simulate the previously polymorphic behavior of this method, which was declared in `ProxyOperationManager` as virtual. `ProxyOperationManager` used to serve as an abstract class from which `ProxyDiscoveryManager` and `ProxyExecutionManager` inherited from, but this is not the case anymore.

src/Microsoft.TestPlatform.ObjectModel/Client/TestRunCriteria.cs
```diff
@@ -329,43 +408,94 @@ public TestRunCriteria(IEnumerable<TestCase> tests, BaseTestRunCriteria baseTest
...
+        /// Gets the test containers (e.g. DLL/EXE/artifacts to scan).
         /// </summary>
         [IgnoreDataMember]
         public IEnumerable<string> Sources
         {
             get
             {
                 IEnumerable<string> sources = new List<string>();
-                return this.AdapterSourceMap?.Values.Aggregate(sources, (current, enumerable) => current.Concat(enumerable));
+                return this.AdapterSourceMap?.Values.Aggregate(
```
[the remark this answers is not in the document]
cvpoienaru: Fixed, thanks !

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -327,19 +425,349 @@ public async Task StartTestRunWithCustomHostAsync(IEnumerable<TestCase> testCase
...
+                this.communicationManager.SendMessage(
+                    MessageType.StartTestSession,
+                    payload,
+                    this.protocolVersion);
+
+                while (true)
+                {
+                    var message = this.TryReceiveMessage();
+
+                    switch (message.MessageType)
```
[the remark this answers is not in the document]
cvpoienaru: Yes, unfortunately they're mostly the same, the exception being the async parts. We could attempt splitting the repeating code in multiple methods but the way I'm thinking of if will end up hurting readability because the method encodes one large state machine.

Do you have any thoughts on how we can do this in a more elegant way ?

src/Microsoft.TestPlatform.ObjectModel/Client/Interfaces/ITestSessionEventsHandler.cs
```diff
@@ -0,0 +1,27 @@
...
+        void HandleStartTestSessionComplete(TestSessionInfo testSessionInfo);
+
+        /// <summary>
+        /// Dispatch StopTestSession complete event to listeners.
+        /// </summary>
+        /// 
+        /// <param name="stopped">
+        /// True if the session was successfully stopped, false otherwise.
+        /// </param>
+        void HandleStopTestSessionComplete(bool stopped);
```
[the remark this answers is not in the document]
cvpoienaru: Fixed, thanks !

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -161,6 +156,60 @@ public ITestRunRequest CreateTestRunRequest(IRequestData requestData, TestRunCri
             return new TestRunRequest(requestData, testRunCriteria, executionManager, loggerManager);
         }
 
+        /// <inheritdoc/>
+        public void CreateStartTestSessionRequest(
+            IRequestData requestData,
+            StartTestSessionCriteria testSessionCriteria,
+            ITestSessionEventsHandler eventsHandler)
+        {
+            if (testSessionCriteria == null)
```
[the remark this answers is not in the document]
cvpoienaru: Yes, there's indeed a large portion in the beginning that is repeated for both `CreateDiscoveryRequest` and `CreateTestRunRequest`.

We can encapsulate this code in a method along the lines of `PrepareForRunAll` as you suggested, but both `CreateDiscoveryRequest` and `CreateTestRunRequest` receive different argument types that don't implement a common interface, i.e. `DiscoveryCriteria` vs `TestRunCriteria`. We can introduce a common interface for them to implement, but I'd leave this out of the scope of this PR for now and introduce it with the next PR that will enable testhost sharing.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -497,6 +523,53 @@ private void StartTestRunAttachmentsProcessing(TestRunAttachmentsProcessingPaylo
...
+
+        private void StopTestSession(TestSessionInfo testSessionInfo)
+        {
+            Task.Run(() =>
+            {
+                var eventsHandler = new TestSessionEventsHandler(this.communicationManager);
+
+                try
+                {
+                    var stopped = TestSessionPool.Instance.RemoveSession(testSessionInfo);
```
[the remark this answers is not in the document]
cvpoienaru: Done, thanks !

Haplois (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyDiscoveryManager.cs
```diff
@@ -23,8 +23,9 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client
     /// <summary>
     /// Orchestrates discovery operations for the engine communicating with the client.
     /// </summary>
-    public class ProxyDiscoveryManager : ProxyOperationManager, IProxyDiscoveryManager, ITestDiscoveryEventsHandler2
+    public class ProxyDiscoveryManager : IProxyDiscoveryManager, IBaseProxy, ITestDiscoveryEventsHandler2
```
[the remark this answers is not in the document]
cvpoienaru: Good point, thanks.

I checked and the base class is only defined & used in `Microsoft.TestPlatform.CrossPlatEngine` project. As such no compatibility issues should arise from this.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyExecutionManager.cs
```diff
@@ -27,71 +28,121 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client
     /// <summary>
     /// Orchestrates test execution operations for the engine communicating with the client.
     /// </summary>
-    internal class ProxyExecutionManager : ProxyOperationManager, IProxyExecutionManager, ITestRunEventsHandler2
+    internal class ProxyExecutionManager : IProxyExecutionManager, IBaseProxy, ITestRunEventsHandler2
```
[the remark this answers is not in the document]
cvpoienaru: Same as above.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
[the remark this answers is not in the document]
cvpoienaru: @drognanar Debugging doesn't use the old workflow anymore (i.e. `vstest.console` asks `VS` to launch the `testhost` process and attach to it) unless we're in compatibility mode. Instead the new workflow launches the `testhost` as a standalone process and only asks `VS` to attach to it just before starting the test execution, provided `IsDebug` flag is set of course. No debugging takes place for test discovery though.

For profiling though we still use the old workflow outlined above. As such we ask `VS` to launch the `testhost` process on our behalf because there's some specific setup that needs to take place first (setting env vars, etc.). Now this behavior is achievable by passing along an object implementing `ITestHostLauncher` to the `StartTestSession` call, the downside being profiling will be set up for test discovery too in addition to test execution. If we wish to avoid this behavior though, the only way to do that would be to start the test session with no `ITestHostLauncher` (in this way test discovery won't be profiled) and initiate test execution by re-launching the `testhost` process in profiling mode.

Haplois (approved):
:shipit:

PR #2586, opened 2020-10-03
Enable Fakes Datacollector settings to be added in design mode

vritant24:
This is to enable the VSTest V2 task scenario, where the code calling into the vstest.console cannot be overridden.

TODO:
1. ~Validate vstest task with this change on existing repos~
2. ~Validate VS scenarios~

### DIFF
--- a/src/vstest.console/CommandLine/GenerateFakesUtilities.cs
+++ b/src/vstest.console/CommandLine/GenerateFakesUtilities.cs
@@ -14,6 +14,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLineUtilities
         internal static void GenerateFakesSettings(CommandLineOptions commandLineOptions, IEnumerable<string> sources, ref string runSettingsXml)
         {
             // dynamically compute the fakes datacollector settings
+            // This runs with or without design mode.
             if (commandLineOptions == null || !commandLineOptions.DisableAutoFakes)
             {
                 runSettingsXml = FakesUtilities.GenerateFakesSettingsForRunConfiguration(sources.ToArray(), runSettingsXml);
--- a/src/vstest.console/CommandLine/GenerateFakesUtilities.cs
+++ b/src/vstest.console/CommandLine/GenerateFakesUtilities.cs
@@ -14,7 +14,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLineUtilities
         internal static void GenerateFakesSettings(CommandLineOptions commandLineOptions, IEnumerable<string> sources, ref string runSettingsXml)
         {
             // dynamically compute the fakes datacollector settings
-            if (!commandLineOptions.DisableAutoFakes)
+            if (commandLineOptions == null || !commandLineOptions.DisableAutoFakes)
             {
                 runSettingsXml = FakesUtilities.GenerateFakesSettingsForRunConfiguration(sources.ToArray(), runSettingsXml);
             }
--- a/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
+++ b/src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
@@ -267,8 +267,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.TestPlatformHelpers
             }
 
             // get Fakes data collector settings
-            var skipFakesEnvVar = Environment.GetEnvironmentVariable("VSTEST_SKIP_FAKES_CONFIGURATION");
-            if (string.IsNullOrEmpty(skipFakesEnvVar) || !string.Equals(skipFakesEnvVar, "1"))
+            if (!string.Equals(Environment.GetEnvironmentVariable("VSTEST_SKIP_FAKES_CONFIGURATION"), "1"))
             {
                 // The commandline Options do not have sources in design time mode,
                 // and so we fall back to using sources instead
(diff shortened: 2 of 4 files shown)


nohwnd:
Making this a draft because it looks like work in progress.

test/Microsoft.TestPlatform.Common.UnitTests/Utilities/FakesUtilitiesTests.cs
```diff
@@ -39,5 +34,91 @@ public void FakesSettingsShouldBeNotGeneratedIfFakeConfiguratorAssemblyIsNotPres
             var generatedRunSettings = FakesUtilities.GenerateFakesSettingsForRunConfiguration(new string[] {@"C:\temp\UT.dll" }, runSettingsXml);
             Assert.AreEqual(generatedRunSettings, runSettingsXml);
         }
+
+        [TestMethod]
+        public void FakesDataCollectorSettingsShouldBeOverridden()
+        {
+            string runSettingsXml = @"<RunSettings><RunConfiguration></RunConfiguration ></RunSettings>";
```
AbhitejJohn, line 41: >  [](start = 87, length = 1)

nit: there's a space everywhere this is used.

test/Microsoft.TestPlatform.Common.UnitTests/Utilities/FakesUtilitiesTests.cs
```diff
@@ -39,5 +34,91 @@ public void FakesSettingsShouldBeNotGeneratedIfFakeConfiguratorAssemblyIsNotPres
...
+
+            var dataCollectorNode2 = new DataCollectorSettings()
+            {
+                AssemblyQualifiedName = FakesUtilities.FakesMetadata.DataCollectorAssemblyQualifiedName,
+                Uri = new Uri(FakesUtilities.FakesMetadata.DataCollectorUriV2),
+                FriendlyName = FakesUtilities.FakesMetadata.FriendlyName,
+                IsEnabled = true,
+                Configuration = doc.CreateElement("Configuration")
+            };
+            FakesUtilities.InsertOrReplaceFakesDataCollectorNode(doc, dataCollectorNode2);
```
AbhitejJohn, line 118: >InsertOrReplaceFakesDataCollectorNode [](start = 27, length = 37)

Can an incorrect runsettings format be passed in to this API? Do we have checks for that?

src/vstest.console/CommandLine/GenerateFakesUtilities.cs
```diff
@@ -14,7 +14,7 @@ public static class GenerateFakesUtilities
         internal static void GenerateFakesSettings(CommandLineOptions commandLineOptions, IEnumerable<string> sources, ref string runSettingsXml)
         {
             // dynamically compute the fakes datacollector settings
-            if (!commandLineOptions.DisableAutoFakes)
+            if (commandLineOptions == null || !commandLineOptions.DisableAutoFakes)
```
AbhitejJohn, line 17: >commandLineOptions [](start = 16, length = 18)

Can we not check for design mode instead? Would make things a little more explicit... 
At the least, I think we would want a comment.

AbhitejJohn (approved):
Are you planning to couple this with the VSUnitTesting changes for 16.9?

src/vstest.console/CommandLine/GenerateFakesUtilities.cs
vritant24: I'm not sure what you mean here. The change is to enable this to run even in the design mode, so we shouldn't be checking for it

test/Microsoft.TestPlatform.Common.UnitTests/Utilities/FakesUtilitiesTests.cs
vritant24: It is a possibility if called by an external user. Within testplatform, it isn't a possibility (other things would fail before this), and so the assumption has been that the runsettings passed in is sound.

vritant24:
> 
> 
> Are you planning to couple this with the VSUnitTesting changes for 16.9?

I planned for the VSUnitTesting side changes to go in separately, so that this can be available to users through the Microsoft.TestPlatform changes earlier

src/vstest.console/CommandLine/GenerateFakesUtilities.cs
AbhitejJohn: So when you said `commandLineOptions == null`, I'm assuming you are checking for designMode, so you can update Fakes setting then too, but that isn't explicit at all here...

vritant24: Oh the design mode check is completely gone, this is just a sanity null check since we check whether `commandLineOptions.DisableAutoFakes` is false/true.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -267,11 +266,19 @@ public void RunTests(TestRunRequestPayload testRunRequestPayload, ITestHostLaunc
                 this.LogTelemetryForLegacySettings(requestData, runsettings);
             }
 
-            if (!commandLineOptions.IsDesignMode)
+            // get Fakes data collector settings
+            if (string.IsNullOrEmpty(Environment.GetEnvironmentVariable("_Disable_Fakes_DataCollector_Configuration")))
```
nohwnd, line 270: Who would use this? We usually name env variables like this VSTEST_<component>_<action> (e.g. VSTEST_RUNNER_DEBUG`,  and spell them in uppercase, in case some OSes would consider spelling when grabbing the variable. We also consider 0 to be off, not just null.

vritant24: This is for the users in the vstest task. By Shreyas' suggestion, if something is being enabled that wasn't before, it's best to have a fallback to be able to replicate previous behavior. I can change the var name and behavior to meet the standard in the vstest repo

PR #2600, opened 2020-10-20
Ensure that a supplied vstest.console path is escape sequenced

AbhitejJohn:
## Description
A supplied path to vstest.console currently may have spaces and could break the logic we have in place. Fixed this to ensure we escape-sequence this path. This is required for a change on the Test Explorer side.

### DIFF
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
@@ -227,11 +227,7 @@ namespace Microsoft.TestPlatform.VsTestConsole.TranslationLayer
         }
 
         private string GetConsoleRunner()
-        {
-            var runnerPath = isNetCoreRunner ? ( string.IsNullOrEmpty(this.dotnetExePath) ? new DotnetHostHelper().GetDotnetPath() : this.dotnetExePath) : GetEscapeSequencedPath(this.vstestConsolePath);
-
-            return runnerPath;
-        }
+            => isNetCoreRunner ? ( string.IsNullOrEmpty(this.dotnetExePath) ? new DotnetHostHelper().GetDotnetPath() : this.dotnetExePath) : GetEscapeSequencedPath(this.vstestConsolePath);
 
         private string GetEscapeSequencedPath(string path)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -220,15 +220,24 @@ private string[] BuildArguments(ConsoleParameters parameters)
...
+                args.Insert(0, GetEscapeSequencedPath(vstestConsolePath));
             }
 
             return args.ToArray();
         }
 
         private string GetConsoleRunner()
+            => isNetCoreRunner ? ( string.IsNullOrEmpty(this.dotnetExePath) ? new DotnetHostHelper().GetDotnetPath() : this.dotnetExePath) : GetEscapeSequencedPath(this.vstestConsolePath);
+
+        private string GetEscapeSequencedPath(string path)
```
nohwnd, line 232: I never heard any path to be referred to as "escape sequenced". I would call this quoted path.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -220,15 +220,24 @@ private string[] BuildArguments(ConsoleParameters parameters)
...
         {
-            return isNetCoreRunner ? ( string.IsNullOrEmpty(this.dotnetExePath) ? new DotnetHostHelper().GetDotnetPath() : this.dotnetExePath) : vstestConsolePath;
+            if (path == null || (path.StartsWith("\"") && path.EndsWith("\"")))
```
nohwnd, line 234:
```
if (path == null)
{
    return path;
}

return $"\"{path.Trim('"')}\"";
```
This seems to have less edge cases.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
AbhitejJohn: oh looks like we have a few references to those words in this repo already: https://github.com/microsoft/vstest/search?q=Escape+sequence

AbhitejJohn:
@nohwnd : Just a friendly reminder on this. Sounds like its good to go. I've addressed all your comments.

AbhitejJohn:
@microsoft/vstest : hoping to get a review to merge this in.

Sanan07 (approved):
LGTM

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
nohwnd: They refer to "escape sequence" which is well established, and means prepending special characters with escape character. Your method just adds quotes. Quotes would sometimes be escaped, but are themselves not escape sequence. 

I am not trying to make this unnecessarily difficult. But if Sanan thinks this is good, let's merge it then.

PR #2601, opened 2020-10-21
Signing instructions for Newtonsoft.Json.dll added

Haplois:
## Description
Added signing instructions for Newtonsoft.Json.dll


nohwnd (approved):
LGTM

PR #2604, opened 2020-10-23
Temporary code to enable correct Fakes and Code Coverage integration

jakubch1:
## Description
This is temporary code to enable correct Fakes and Code Coverage integration. We will clean it when both will start using CLR IE. For now it's required to get env variables from code coverage data collector at the end.

## Related issue


cvpoienaru (approved):
LGTM

PR #2606, opened 2020-10-26
vstest.console: CommandLineOptions: preserve stacktrace on re-throw (CA2200)

tmds:


src/vstest.console/CommandLine/CommandLineOptions.cs
```diff
@@ -295,14 +295,14 @@ public void AddSource(string source)
                 // Get matching files from file pattern parser
                 matchingFiles = FilePatternParser.GetMatchingFiles(source);
             }
-            catch(TestSourceException ex)
+            catch(TestSourceException)
```
AbhitejJohn, line 298: Perhaps we can use the `catch() when` expression instead to simplify this. (doc [link](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when#when-in-a-catch-statement))

src/vstest.console/CommandLine/CommandLineOptions.cs
```diff
@@ -295,14 +295,14 @@ public void AddSource(string source)
...
             {
                 if(source.StartsWith("-") || source.StartsWith("/"))
                 {
                     throw new TestSourceException(
```
nohwnd, line 302: Shouldn't `ex` be passed as inner exception?

tmds:
@nohwnd @AbhitejJohn I forgot about this. I've addressed your feedback. ptal.

src/vstest.console/CommandLine/CommandLineOptions.cs
```diff
@@ -295,14 +295,10 @@ public void AddSource(string source)
...
             }
-            catch(TestSourceException ex)
+            catch(TestSourceException ex) when (source.StartsWith("-") || source.StartsWith("/"))
             {
-                if(source.StartsWith("-") || source.StartsWith("/"))
-                {
-                    throw new TestSourceException(
-                        string.Format(CultureInfo.CurrentUICulture, CommandLineResources.InvalidArgument, source));
-                }
-                throw ex;
```
Haplois, line 305: We won't throw when `(source.StartsWith("-") || source.StartsWith("/"))` is not satisfied anymore. This should be addressed.

Haplois (changes requested):

src/vstest.console/CommandLine/CommandLineOptions.cs
tmds: > We don't throw when (source.StartsWith("-") || source.StartsWith("/")) is not satisfied anymore.

When the condition is not met, the catch block doesn't apply and the exception just throws, right?

Haplois: Yep, you're right :)

Haplois (approved):
LGTM

Haplois:
Thank you for your contribution, @tmds.

PR #2611, opened 2020-11-02
Managed TestCase Properties implemented

Haplois:
## Description
Implemented https://github.com/microsoft/vstest-docs/blob/master/RFCs/0017-Managed-TestCase-Properties.md


src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -131,7 +213,17 @@ public string DisplayName
         {
             get
             {
-                return string.IsNullOrEmpty(this.displayName) ? this.FullyQualifiedName : this.displayName;
+                if(string.IsNullOrEmpty(this.displayName))
+                {
+                    if(string.IsNullOrWhiteSpace(managedType) || string.IsNullOrWhiteSpace(managedMethod))
```
peterwald, line 218: Consider adding a helper method `HasManagedMethodAndType` or some such that can be used in all of the places that need this check.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -234,10 +333,27 @@ private Guid GetTestId()
                 // do nothing
             }
 
-            string testcaseFullName = this.ExecutorUri + source + this.FullyQualifiedName;
+            var testcaseFullName = this.ExecutorUri + source;
```
peterwald, line 336: Note that we still need to handle parameters in the case of a Theory or TestGroup of test cases that are only distinguished by parameters. If this is to be handled in a later PR, then it would be good to note it here.

peterwald:
In general this looks good to me. A couple of overall comments in addition to the inline ones....

1. I don't know enough about the TestCase property serialization to know how these changes will interact with forward and backward version mismatches. I think it would be prudent to give this some more thought and testing.
1. Now that the spec has evolved a bit I wonder if we should consider changing the nomenclature slightly. Since FullyQualifiedNameUtilities is really about removing the reliance on FullyQualifiedName (rather than assisting with it), maybe we should consider renaming it to something along the lines of `ManagedQualifiedNameUtilities` or `ManagedNameUtilities` or some such. This may reduce some potential confusion given that these changes really have nothing to do with the existing `FullyQualifiedName` property.

peterwald (approved):

eng/Versions.props
```diff
@@ -75,5 +75,11 @@
     <MicrosoftSourceLinkVersion>1.0.0-beta2-19554-01</MicrosoftSourceLinkVersion>
     <MicrosoftDotNetSwaggerGeneratorMSBuildVersion>1.0.0-beta.20567.6</MicrosoftDotNetSwaggerGeneratorMSBuildVersion>
     <XliffTasksVersion>1.0.0-beta.20055.1</XliffTasksVersion>
+    
+    <!-- Roslyn -->
+    <!-- WARNING: Depending on Roslyn in code that also supports unmanaged scenarios (e.g. C++,
```
nohwnd, line 80: Why do we do that then?

scripts/build.ps1
```diff
@@ -119,7 +119,7 @@ $language = @("cs", "de", "es", "fr", "it", "ja", "ko", "pl", "pt-BR", "ru", "tr
 # Capture error state in any step globally to modify return code
 $Script:ScriptFailed = $false
 
-Import-Module "$($CurrentScriptDir.FullName)\verify-nupkgs.ps1"
+Import-Module -Name "$($CurrentScriptDir.FullName)\verify-nupkgs.ps1" -Scope Local
```
nohwnd, line 122: why? the other file is not even a module (not your fault).  But why local scope?

nohwnd:
Looks good :)

scripts/build.ps1
Haplois: It's a workaround. Sometimes if the `build.ps1` fails or interrupted by `CTRL+C`, subsequent run fails to import `Verify-Nuget-Packages` method this fixes that. I don't know why.

eng/Versions.props
Haplois: We ended up not referencing Roslyn, but I moved the variable anyway because we already have the code for source based discovery (in TestWindow), and this is the version used there.

PR #2641, opened 2020-11-18
[16.8] Fix collect dump always

nohwnd:
## Description
Fix how we collect dumps when CollectAlways is not specified. In previous versions procdump collected dump always, but because we controlled what files exactly will be uploaded we just skipped the dump when collect always was not specified. Now we no longer know which dumps will be collected, so we need to set procdump parameters correctly. 

Also skip warning about sequence file when we don't collect hang dump. And prefer agent temp over normal temp, to make running on azdo contained in the agent folder.

Update branding to 16.8.1.

## Related issue
Fix https://github.com/dotnet/sdk/issues/14578

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -462,7 +462,10 @@ namespace Microsoft.TestPlatform.Extensions.BlameDataCollector
                 }
                 else
                 {
-                    this.logger.LogWarning(this.context.SessionDataCollectionContext, Resources.Resources.NotGeneratingSequenceFile);
+                    if (this.collectProcessDumpOnTestHostHang)
+                    {
+                        this.logger.LogWarning(this.context.SessionDataCollectionContext, Resources.Resources.NotGeneratingSequenceFile);
+                    }
                 }
 
                 if (this.uploadDumpFiles)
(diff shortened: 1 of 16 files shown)


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -585,7 +588,15 @@ private string GetTempDirectory()
...
             {
-                this.tempDirectory = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+                // DUMP_TEMP_PATH will be used as temporary storage location
+                // for the dumps, this won't affect the dump uploads. Just the place where
+                // we store them before moving them to the final folder.
+
+                // AGENT_TEMPDIRECTORY is AzureDevops variable, which is set to path
+                // that is cleaned up after every job. This is preferable to use over
+                // just the normal temp.
+                var temp = Environment.GetEnvironmentVariable("VSTEST_DUMP_TEMP_PATH") ?? Environment.GetEnvironmentVariable("AGENT_TEMPDIRECTORY") ?? Path.GetTempPath();
```
cvpoienaru, line 598: Nice chaining of "??" operators, although it's a little harder to read 👍

cvpoienaru (approved):
LGTM

Sanan07 (approved):

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: JavaScript taught me that :D

PR #2642, opened 2020-11-18
Make sure log indicates that debug wait for connection may hang TestHost

jfleisher:
## Description
Add warning log entry in testhost launch if the testhost debug environment variables are set.
This allows inspections of the logs to reveal why the testhost may hang on start.
This issue was surface by a customer who did not know he had the environment variable set.

## Related issue

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
@@ -386,10 +386,30 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting
             }
         }
 
+        private void WarnIfDebugEnvironment()
+        {
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSTEST_HOST_NATIVE_DEBUG")))
+            {
+                EqtTrace.Warning("VSTEST_HOST_NATIVE_DEBUG is set and will cause testhost to hang.");
+            }
+
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSTEST_HOST_DEBUG")))
+            {
+                EqtTrace.Warning("VSTEST_HOST_DEBUG is set and will cause testhost to hang.");
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


Sanan07 (approved):
Looks good, just small naming suggestion.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -386,11 +386,31 @@ private void OnHostExited(HostProviderEventArgs e)
             }
         }
 
+        private void WarnIfDebugEnvironment()
+        {
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("VSTEST_HOST_NATIVE_DEBUG")))
```
nohwnd, line 391: This should check if the variable is set to 1, the same way the internal check does it.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -386,11 +386,31 @@ private void OnHostExited(HostProviderEventArgs e)
...
+            {
+                EqtTrace.Warning("VSTEST_HOST_NATIVE_DEBUG is set and will cause testhost to hang.");
```
nohwnd, line 393: I don't think "hang" is the correct word to use here. Hang suggests that this is a bad thing (at least to me). And it does not suggest how to "solve" it. I would rather see it say: and will cause testhost to wait indefinitely for a debugger to be attached.

This message should also be localized.

The same treatment should be given to datacollector, because it can do the same thing.

nohwnd (changes requested):
It would be nice to know in which situation the user is supposed to see the message, because a similar message is already present when invoking in command line.

https://github.com/microsoft/vstest/blob/31cddc5b7beb6e13ee7ee2b2a6ec86baeb968f73/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs#L150-L164 

In any case the check for the warning should be done consistently with how we check to enable the debugging to avoid misleading the user.

MarcoRossignoli:
I'm going to close this one for stale state, feel free to reopen if needed.

PR #2671, opened 2020-12-10
Change logic of code coverage attachments merging to merge only files with coverage extension

jakubch1:
## Description
Change logic of code coverage attachments merging to merge only files with *.coverage extension. This will allow code coverage data collectors to send additional files, for example logs.

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
+++ b/src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
@@ -8,7 +8,6 @@ namespace Microsoft.VisualStudio.TestPlatform.Utilities
     using System.Collections.ObjectModel;
     using System.IO;
     using System.Linq;
-    using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -16,7 +15,6 @@ namespace Microsoft.VisualStudio.TestPlatform.Utilities
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
-    using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
 
     public class CodeCoverageDataAttachmentsHandler : IDataCollectorAttachmentProcessor
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
```diff
@@ -95,6 +113,20 @@ private async Task MergeCodeCoverageFilesAsync(IList<string> files, Cancellation
                     cancellationToken).ConfigureAwait(false);
 
             coverageUtility.WriteCoverageFile(files[0], coverageData);
+
+            foreach(var file in files.Skip(1))
+            {
+                try
+                {
+                    File.Delete(file);
```
fhnaseer, line 121: Deleting original files after merge?,

jakubch1: yes, this is something that we had previously and somehow we missed this when Codrin was making changes. We want to clean those files due to logic which is inside DevOps. After running tests they are searching for all files *.coverage and merging them. If we keep original files it will be taken and merged 1 more time, which is not needed.

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
```diff
@@ -56,20 +78,16 @@ public async Task<ICollection<AttachmentSet>> ProcessAttachmentSetsAsync(ICollec
 
         private async Task<string> MergeCodeCoverageFilesAsync(IList<string> files, IProgress<int> progressReporter, CancellationToken cancellationToken)
         {
-            if (files.Count == 1)
```
cvpoienaru, line 59: By removing this we assume the merging lib properly handles merging only one file. I'm sure it does, but should we keep this protection just in case ?

We also avoid any additional overhead, however small, from calling the lib method.

cvpoienaru (approved):

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
jakubch1: Original logic was:

```
                    IList<string> filesToDelete = new List<string>(files.Count) { tempFileName };

                    for (int i = 1; i < files.Count; i++)
                    {
                        cancellationToken.ThrowIfCancellationRequested();
                        progressReporter?.Report(100 * i / files.Count);

                        cancellationToken.ThrowIfCancellationRequested();
                        methodInfo.Invoke(null, new object[] { files[i], outputfileName, tempFileName, true });

                        cancellationToken.ThrowIfCancellationRequested();
                        File.Copy(tempFileName, outputfileName, true);

                        filesToDelete.Add(files[i]);
                    }

                    cancellationToken.ThrowIfCancellationRequested();
                    foreach (string fileName in filesToDelete)
                    {
                        File.Delete(fileName);
                    }
```

I don't know how I missed this during review :)

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
jakubch1: I've added logic to function which is calling this method to invoke it only if at least 2 reports

fhnaseer (approved):

PR #2537, opened 2020-08-25
TestPluginCache and other dependencies refactoring

Sanan07:
## Description
Refactoring of TestPluginCache and other classes responsible for test discovery and test execution

Fix : #2634

PR #2674, opened 2020-12-14
Removed `TypesToLoadAttribute` from ObjectModel.

Haplois:
Removed `TypesToLoadAttribute` from ObjectModel, and moved the functionality into adapters themselves to fix compatibility issues introduced in https://github.com/microsoft/vstest/pull/2537.


src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+            return typesToLoad;
+        }
+
+        private static IEnumerable<Type> GetTypesToLoad(Attribute attribute)
+        {
+            if (attribute == null)
+                return Enumerable.Empty<Type>();
+
+            var type = attribute.GetType();
+            var typesProperty = type.GetProperty("Types");
```
Sanan07, line 35: This method contains reflection, which is slow. I assume how much improvement it could give us with having custom attribute?

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
+
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+    using System.Linq;
+
+    internal static class TypesToLoadUtilities
+    {
+        public const string TypesToLoadAttributeFullName = "Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities.TypesToLoadAttribute";
```
Sanan07, line 16: You have several `TypesToLoadAttribute` classes, but here loading the one which is in `ObjectModel`. How does it work for other types?

Sanan07 (changes requested):
I assume about performance gain which we can get from it, as the number of reflections used in code increased. And how this can solve problem with MSTestAdapter?

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
Haplois: The other implementation was also using reflection. 

Our performance gain is because this will save us from scanning the entire assembly.

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
Haplois: Public assembly-level attribute causing compatibility issues with older versions of VS, where that attribute is not present. We removed the attribute from `Microsoft.VisualStudio.TestPlatform.ObjectModel` but kept the functionality by allowing extensions to define that internally. 

We look for an attribute with the full name of `Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities.TypesToLoadAttribute` defined in target assembly. If we find it and it has a property called `Types`, which returns a `System.Type` array, we use that.

Sanan07: Yes, I see it, I got you created several `TypesToLoadAttribute` classes in `Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities.TypesToLoadAttribute` namespace.
But the main question : how will it help to solve the main problem in MSTestAdapter?

Haplois: The same way I did here. Check this https://github.com/microsoft/testfx/compare/master...Haplois:fix/removed-types-to-load.

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
Sanan07: In my change we did not have reflection, the logic before it used reflection. It used `GetTypes()` on assembly, but we changed to load only specific types. But now you included that logic in each type, where for each assembly you again call `GetType`.
So I suspicious in overall benefit, which it could give us.

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
Sanan07, line 16: Why is it public? Needs to be private.

Haplois: It's `internal` by class. I am planning to use this in E2E tests.

src/Microsoft.TestPlatform.Common/Utilities/TypesToLoadUtilities.cs
Haplois: `GetCustomAttribute<T>` does use reflection, as this is. Only extra is now, instead of matching the attribute by type, it's matching by name. This is a necessary evil to keep this implementation.

test/Microsoft.TestPlatform.AcceptanceTests/DiscoveryTests.cs
```diff
@@ -91,5 +97,33 @@ public void DiscoverTestsShouldShowProperWarningIfNoTestsOnTestCaseFilter(Runner
...
+                var assemblyFile = Path.Combine(extensionsDirectory, extension);
+                var assembly = Assembly.LoadFrom(assemblyFile);
+
+                var expected = extensionsToVerify[extension];
+                var actual = TypesToLoadUtilities.GetTypesToLoad(assembly).Select(i => i.FullName).ToArray();
+
+                CollectionAssert.AreEquivalent(expected, actual, $"Specified types using TypesToLoadAttribute in \"{extension}\" assembly doesn't match the expected.");
+            }
+
+            Assert.IsTrue(true);
```
Sanan07, line 126: Why we need this?

Haplois: This test verifies the logic works as expected.

src/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector/Properties/TypesToLoadAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    internal sealed class TypesToLoadAttribute : Attribute
+    {
+        public TypesToLoadAttribute(params Type[] types)
+        {
+            this.Types = types;
+        }
+
+        public Type[] Types { get; }
```
Sanan07, line 23: Why not have the types initialize right here?
`public Type[] Types = {typeof(EventLogDataCollector)}`. In this case no need for constructor and above statement.

Haplois: I didn't want to change the implementation and keep it somewhat consistent in all adapters. This way actual assembly-level attribute is cleaner,

Sanan07 (approved):

src/Microsoft.TestPlatform.ObjectModel/Utilities/AssemblyHelper.cs
```diff
@@ -340,6 +339,18 @@ internal static void SetNETFrameworkCompatiblityMode(AppDomainSetup setup, IRunC
                 }
             }
         }
-    }
 #endif
+
+        public static IEnumerable<Attribute> GetCustomAttribute(this Assembly assembly, string fullyQualifedName)
```
peterwald, line 344: nit: `fullyQualifiedName`

peterwald (approved):

src/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector/Properties/TypesToLoadAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+
+[assembly: TypesToLoad(typeof(EventLogDataCollector))]
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
+{
+    /// <summary>
+    /// Custom Attribute to specify the exact types which should be loaded from assembly
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    internal sealed class TypesToLoadAttribute : Attribute
```
shyamnamboodiripad, line 16: As mentioned in https://github.com/microsoft/testfx/pull/753 it would be great to change the name of the attribute to `TestExtensionTypesAttribute` and place it in `Microsoft.VisualStudio.TestPlatform` namespace.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/Properties/TypesToLoadAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+
+[assembly: TypesToLoad(typeof(BlameLogger), typeof(BlameCollector))]
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities
+{
+    /// <summary>
+    /// Custom Attribute to specify the exact types which should be loaded from assembly
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    internal sealed class TypesToLoadAttribute : Attribute
```
shyamnamboodiripad, line 17: Could we put the attribute definition in a single file in the repo and add this file as a linked file in multiple projects? That way if we need to fix / change anything in the future, we would only need to make the change in one place...

src/Microsoft.TestPlatform.ObjectModel/Utilities/AssemblyHelper.cs
```diff
@@ -340,6 +339,18 @@ internal static void SetNETFrameworkCompatiblityMode(AppDomainSetup setup, IRunC
...
 #endif
+
+        public static IEnumerable<Attribute> GetCustomAttributes(this Assembly assembly, string fullyQualifiedName)
+        {
+            ValidateArg.NotNull(assembly, nameof(assembly));
+            ValidateArg.NotNullOrWhiteSpace(fullyQualifiedName, nameof(fullyQualifiedName));
+
+            var attributes = assembly
+                .GetCustomAttributes()
+                .Where(i => i.GetType().FullName == fullyQualifiedName);
```
shyamnamboodiripad, line 351: Would `assembly.GetType(fullName)` be more efficient than `assembly.CustomAttributes()`? The latter will end up loading and iterating over all the custom attributes in an assembly which could potentially also end up loading other assemblies if for instance, base type of the traversed attribute is defined in a different assembly...

Also, if we keep `assembly.GetCustomAttributes()` call then it may be nice to early out as soon as we find the type - consider using `.FirstOrDefault(i => i.GetType().FullName == fullyQualifiedName)` in place of `.Where(i => i.GetType().FullName == fullyQualifiedName)`.

shyamnamboodiripad (commented):
👍 modulo comments

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/Properties/TypesToLoadAttribute.cs
Haplois: I was too lazy to do that. I will do it with tomorrow's PR.

src/Microsoft.TestPlatform.ObjectModel/Utilities/AssemblyHelper.cs
Haplois: We can first try `assembly.GetType(fullyQualifiedName)` and only if we can't load iterate all attributes.

shyamnamboodiripad: That's a good idea! However, it is also possible to have pathological case where someone has our attribute defined in multiple assemblies and some of the references are to the attribute in the current assembly while others are to the external ones (using extern alias). Hopefully, pathological cases like that would be very very rare 😄

Haplois: That is pathological. We can easily say that we don't support it. IMO

PR #2685, opened 2021-01-07
Loc Update

cristianosuzuki77:
Latest translation updates.


Sanan07 (approved):

Haplois:
Nitpicking on the translation of "Proxy" into Turkish. Instead of "ara sunucu" (which translates as "middle server" verbatim), "vekil" should be used. It's more representative of the original meaning.

/cc: @cristianosuzuki77

PR #2698, opened 2021-01-15
Upgrade CC components

jakubch1:
## Description

## Related issue


src/package/sign/sign.proj
```diff
@@ -131,11 +131,9 @@
 
       <!-- DataCollector related dlls -->
       <AssembliesToSign Include="$(ArtifactsDirectory)Microsoft.IntelliTrace.Core.dll" />
```
fhnaseer, line 133: Not related to this, but I think we get these from VS repo, and they should be already signed as well,

fhnaseer (approved):

src/package/sign/sign.proj
jakubch1: @Sanan07  please try to cleanup this file after you get new packages from VS

fhnaseer (approved):

src/package/sign/sign.proj
Sanan07: What do you mean by cleanup? Do you mean clean packages, like git clean -Xdf?

jakubch1: No, I mean remove files from this file (sign.proj) which are already signed by VS. Signing takes a lot of time and we don't need that

PR #2437, opened 2020-05-14
Reimplement the MSBuild VSTestTask task based on ToolTask

mcartoixa:
## Description
The [MSBuild ToolTask class](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask) was designed to help create custom tasks that wraps a command line tool, which is exactly what `VSTestTask` does (it wraps `dotnet.exe exec vstest.console.dll`). Using it allows to delete unneeded code while fixing #680.
The console output lost its colorized output (which was expected) but the task gained the ability to be logged.
A bit of maintenance was performed as well, making use of boolean properties where needed (instead of strings), and of [`ITaskItem`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.itaskitem)s for paths (instead of strings as well). Those changes should be compatible with existing usage though.
Unit tests have been refactored as well along the lines of the new implementation (and are all green, obviously). A real world test has been performed on a project of mine that was affected by #680, on Windows 10 only.
This might need further real world testing.

## Related issue
Fixes #680

Issue #2224, opened 2019-10-14
VSTest is printing using Console.WriteLine instead of using the logger


krwq:
## Description

VSTest should be using logger instead of Console.WriteLine. I.e.:
https://github.com/microsoft/vstest/blob/e6c198ca439603f0e81d723538894646032745f6/src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs#L132
https://github.com/microsoft/vstest/blob/e6c198ca439603f0e81d723538894646032745f6/src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs#L301

This is causing that in some cases output gets lost.

## Steps to reproduce

- Clone https://github.com/dotnet/iot/
- Replace following ("Test") task https://github.com/dotnet/iot/blob/a5a53b5a3a287638df0176c483fcb1f3b2fef7f0/build.proj#L59 with

```xml
  <Target Name="Test">
    <ItemGroup>
      <UnitTestProjects Include="$(MSBuildThisFileDirectory)src\devices\**\*.Tests.csproj" />
    </ItemGroup>
    <MSBuild Projects="@(UnitTestProjects)" Targets="VSTest"
        ContinueOnError="ErrorAndContinue" />
    <Error Condition="$(MSBuildLastTaskResult) != 'true'"
        Text="Unit tests failed. Please check the detailed log to find out which ones failed." />
  </Target>
```

**Replacing will not be needed once linked PR is merged**

- Run build.cmd/build.sh (in the root directory)

If any property is passed to MSBuild task then output gets printed (because of luck it gets run on master node where Console.WriteLine works correctly)

## Expected behavior

Test output is displayed

## Actual behavior

Build fails but no errors get printed

ericstj:
It also looks like this task creates a process but doesn't redirect output, so that will further be lost logging.
https://github.com/microsoft/vstest/blob/e6c198ca439603f0e81d723538894646032745f6/src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingApp.cs#L30-L51

ViktorHofer:
cc @nohwnd

nohwnd:
@ViktorHofer did you see this issue happen somewher or why are you tagging me just now?:)

ViktorHofer:
@akoeplinger a colleague of mine is hitting this issue in mono/linker. Thought you might be interested in this issue :)

krwq:
[...]

Issue #1503, opened 2018-03-24
dotnet test is incompatible with MSBuild node reuse


rainersigwald:
## Description

In the .NET SDK 2.1.300, MSBuild on .NET Core will introduce support for [node reuse](https://github.com/Microsoft/msbuild/issues/2703). This feature enable faster cycle times by keeping MSBuild worker processes alive after the first build ends, allowing them to be reused "hot" by the second and subsequent builds.

Unfortunately, this is incompatible with the output strategy used by VSTestTask, which depends on child processes of MSBuild being attached to the stdout stream of the entry-point `dotnet test` process. That can't be the case when `vstest.console` may be run from an MSBuild that has no stdout stream.

On a CLI with the new MSBuild, output from VSTest does not appear on the console.

## Steps to reproduce

This is the root cause of the test failures in https://ci.dot.net/job/dotnet_cli/job/release_2.1.2xx/job/release_windows_nt_x64_prtest/41/

On a CLI with the new MSBuild, run `dotnet test`.

## Expected behavior

Test output is displayed from `dotnet test` regardless of MSBuild internals. From preview1:

```
s:\cli\bin\3\win-x64\test\dotnet-test.Tests\1\VSTestMulti>"C:\Program Files\dotnet\dotnet.exe" test
Build started, please wait...
Build completed.

Test run for s:\cli\bin\3\win-x64\test\dotnet-test.Tests\1\VSTestMulti\bin\Debug\net46\VSTestDesktopAndNetCore.dll(.NETFramework,Version=v4.6)
Microsoft (R) Test Execution Command Line Tool Version 15.6.0-preview-20180109-01
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
Failed   VSTestFailTest
Error Message:
 Assert.Fail failed.
Stack Trace:
   at TestNamespace.VSTestTests.VSTestFailTest()


Total tests: 3. Passed: 2. Failed: 1. Skipped: 0.
Test Run Failed.
Test execution time: 1.8794 Seconds
Test run for s:\cli\bin\3\win-x64\test\dotnet-test.Tests\1\VSTestMulti\bin\Debug\netcoreapp2.1\VSTestDesktopAndNetCore.dll(.NETCoreApp,Version=v2.1)
Microsoft (R) Test Execution Command Line Tool Version 15.6.0-preview-20180109-01
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[...]

Issue #680, opened 2017-04-01
`VSTestTask` only logs to the console


dougbu:
## Description
File logging output contains information about _calling_ the `VSTestTask` but no output from the task itself. This makes it difficult to search the test's output.

## Steps to reproduce
Appears to occur with every ASP.NET repo. For example,
1. `git clone https://github.com/aspnet/HtmlAbstractions.git`
2. `cd HtmlAbstractions\test\Microsoft.Extensions.WebEncoders.Tests`
3. `dotnet restore`
4. `dotnet build`
5. `dotnet test --no-build --framework netcoreapp2.0 /flp:verbosity=diag`
6. `notepad msbuild.log`

## Expected behavior
Same information in the `msbuild.log` file as was shown on the console:
```
Test run for C:\dd\dnx\Universe\.r\HtmlAbstractions\test\Microsoft.Extensions.WebEncoders.Tests\bin\Debug\netcoreapp2.0\Microsoft.Extensions.WebEncoders.Tests.dll(.NETCoreApp,Version=v2.0)
Microsoft (R) Test Execution Command Line Tool Version 15.0.0.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:00.4636575]   Discovering: Microsoft.Extensions.WebEncoders.Tests
[xUnit.net 00:00:00.5966555]   Discovered:  Microsoft.Extensions.WebEncoders.Tests
[xUnit.net 00:00:00.6391310]   Starting:    Microsoft.Extensions.WebEncoders.Tests
[xUnit.net 00:00:00.8075514]   Finished:    Microsoft.Extensions.WebEncoders.Tests

Total tests: 6. Passed: 6. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 1.9789 Seconds
```

## Actual behavior
Information _around_ the `VSTestTask` call but nothing from the task itself:
```
12:23:02.754     1>Target "VSTest: (TargetId:2)" in file "C:\Users\dougbu\AppData\Local\Microsoft\dotnet\sdk\2.0.0-preview1-005418\Microsoft.TestPlatform.targets" from project "C:\dd\dnx\Universe\.r\HtmlAbstractions\test\Microsoft.Extensions.WebEncoders.Tests\Microsoft.Extensions.WebEncoders.Tests.csproj" (entry point):
[...]

PR #2702, opened 2021-01-16
Create a proper MSBuild ToolTask based VSTestTask

mcartoixa:
## Description
_Update from nohwnd:_ 

A new task VSTestTask2 and a logger MSBuildLogger (for internal use only) was added. And can be enabled by setting `/p:VsTestUseMSBuildOutput=true`. 

This is not enabled anywhere by default, neither in dotnet test, or under vstest, because I want to collect some feedback first. 

----

The current console forwarding task has been renamed _VSTestForwardingTask_. The new _VSTestTask_, [based on ToolTask](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask), properly integrates into the MSBuild framework.
If the `$(VSTestUseConsole)` property is set to `True` during the build, the old console forwarding _VSTestForwardTask_ is used: logs are lost but output is colorized (like today). Otherwise the new _VSTestTask_ is used: logs are fine but colorization is lost (except for errors, that are properly recognized by MSBuild).

A new PR will be submitted right away on the dotnet/sdk project so that `dotnet test` sets the `$(VSTestUseConsole)` property.

In effect:
* users that favor a colorized console output can use `dotnet test` (like today).
* users that favor a proper MSBuild integration can use `dotnet msbuild -Target:VSTest`.

This PR feels more like a workaround than a proper fix: it still feels weird to depend on a MSBuild task (_VSTestForwardingTask_) that is not entirely MSBuild friendly... But it seems like the easiest fix that is compatible with the old behavior.
A proper fix would probably center around the implementation of `dotnet test` which could call `dotnet exec vstest.console.dll` directly (in addition to MSBuild, for the build part only). This would render current workarounds like for dotnet/msbuild#3130, or `AllowFailureWithoutError` definitely obsolete.

## Related issue
Fixes (or at least gives an option around) #680, #1503, #2224, dotnet/sdk#9481, dotnet/msbuild#5387 and probably more...
This is a new attempt after #2437.


mcartoixa:
Please note that I would be happy to answer any questions and/or reservations, and/or discuss any suggestions or improvements on this PR.

Haplois:
Thank you for the contribution @mcartoixa, I'll check this during the week and ask my questions, if any.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,108 @@
...
+            Tracing.Trace("VSTest: Arguments: " + processInfo.FileName + " " + processInfo.Arguments);
+
+            using (var activeProcess = new Process { StartInfo = processInfo })
+            {
+                activeProcess.Start();
+                this.activeProcessId = activeProcess.Id;
+
+                activeProcess.WaitForExit();
+                Tracing.Trace("VSTest: Exit code: " + activeProcess.ExitCode);
+                return activeProcess.ExitCode == 0;
```
KirillOsenkov, line 92: We should only return `!Log.HasLoggedErrors`. The task should fail if and only if it has logged any errors.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,108 @@
...
+            var processInfo = new ProcessStartInfo
+                                  {
+                                      FileName = toolExe,
+                                      Arguments = this.CreateCommandLineArguments(),
+                                      UseShellExecute = false,
+                                  };
+
+            if (!string.IsNullOrEmpty(this.VSTestFramework))
+            {
+                Console.WriteLine(Resources.TestRunningSummary, this.TestFileFullPath, this.VSTestFramework);
```
KirillOsenkov, line 79: A task shouldn't use Console.WriteLine, but instead use Log.LogMessage:
https://source.dot.net/#Microsoft.Build.Tasks.Core/Copy.cs,255

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,108 @@
...
+                    Thread.Sleep(1000);
+                }
+
+                Debugger.Break();
+            }
+
+            // Avoid logging "Task returned false but did not log an error." on test failure, because we don't
+            // write MSBuild error. https://github.com/dotnet/msbuild/blob/51a1071f8871e0c93afbaf1b2ac2c9e59c7b6491/src/Framework/IBuildEngine7.cs#L12
+            var allowfailureWithoutError = BuildEngine.GetType().GetProperty("AllowFailureWithoutError");
+            allowfailureWithoutError?.SetValue(BuildEngine, true);
```
KirillOsenkov, line 68: we should really not do this. If there was an error, log an MSBuild error. If there wasn't an error, don't log it.

Let's remove this reflection logic.

MarkKharitonov:
Just to make it clear to me. The `VSTestUseConsole` build property is introduced by this PR, which has not been merged yet, right?

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
mcartoixa: This is usually the way it's done in MSBuild tasks. Does `vstest.console.exe` return an error code even in cases of success?
By the way this task has been mostly left unchanged from current behaviour and kept for compatibility. The goal of this PR is rather to introduce a new task that would be more respectful of MSBuild, see `VSTestTask`.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
mcartoixa: Agreed in general, but this has been left unchanged for compatibility with current behavior.   The goal of this PR is rather to introduce a new task that would be more respectful of MSBuild, see `VSTestTask`.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
mcartoixa: This has been left unchanged for compatibility with current behavior.  The goal of this PR is rather to introduce a new task that would be more respectful of MSBuild, see `VSTestTask`.

mcartoixa:
@MarkKharitonov Right.
And to ensure compatibility would require the reopening of dotnet/sdk#15376 to initialize the `VSTestUseConsole`  property in the `dotnet test` command.

mcartoixa:
When can we expect some kind of decision about the future of this PR? As mentioned before quite a few bugs could be affected by this.

- Is it going to be rejected, and why?
- Is it going to be merged, and when?

mcartoixa:
It took me 2 days to setup a new barely working development environment, but I think I managed to rebase this PR against `main`. Again.

mcartoixa:
Any update on the status of this request?

Nirmal4G:
nit: file should be renamed too!

test/Microsoft.TestPlatform.Build.UnitTests/**VsTestTaskTests.cs** -> **VSTestTaskTests.cs**

Nirmal4G (approved):
Overall I like it. It's a welcome change but a huge breaking change nonetheless, so its up to the owners, I think!

mcartoixa:
@Nirmal4G Thanks for the review! There is nothing more that I would like than having a discussion with the owners.

Meanwhile it seems that I will have to rebase this PR again...

Nirmal4G:
I think they are delaying this because of some internal politics. I'm sure the team wants this more than anyone. Even @KirillOsenkov is enthusiastic about this!! 😉

KirillOsenkov:
I can't speak for the MSBuild team but there are only a handful of people on the team, they're all fantastic, but super busy and overloaded with constant firefighting. I apologize for the delay but hopefully they will have time for this at some point soon. Just wanted to clarify that it's not the team's fault that their response times are like this, they're just overloaded with work.

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -24,12 +28,61 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      <CallTarget Targets="_VSTestMSBuild" Condition="!$(VSTestUseConsole)" />
+  </Target>
+
+  <!-- Used when called from dotnet msbuild command line: uses MSBuild logs but does not allow for console colorization -->
+  <Target
+    Name="_VSTestMSBuild"
+    DependsOnTargets="$(_VSTestMSBuildDependsOn)"
+    Condition="'$(IsTestProject)' == 'true'"
+  >
+    <Microsoft.TestPlatform.Build.Tasks.VSTestTask
```
Nirmal4G, line 47:
```suggestion
    <VSTestTask
```

This is more than enough!

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -24,12 +28,61 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    Name="_VSTestConsole"
+    Condition="'$(IsTestProject)' == 'true'"
+  >
     <CallTarget Condition="'$(VSTestNoBuild)' != 'true' AND '$(IsTestProject)' == 'true'" Targets="BuildProject" />
-    
+
     <CallTarget Targets="ShowCallOfVSTestTaskWithParameter" />
 
-    <Microsoft.TestPlatform.Build.Tasks.VSTestTask
+    <Microsoft.TestPlatform.Build.Tasks.VSTestForwardingTask
```
Nirmal4G, line 85:
```suggestion
    <VSTestForwardingTask
```

same here!

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -70,37 +122,36 @@ Copyright (c) .NET Foundation. All rights reserved.
     <MSBuild Projects ="$(MSBuildProjectFullPath)" />
     <Microsoft.TestPlatform.Build.Tasks.VSTestLogsTask LogType="BuildCompleted" />
```
Nirmal4G, line 123:
```suggestion
    <VSTestLogsTask LogType="BuildCompleted" />
```

here too!

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -70,37 +122,36 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   <Target Name="ShowCallOfVSTestTaskWithParameter">
-    <Message Text="Calling task Microsoft.TestPlatform.Build.Tasks.VSTestTask with following parameter..." Importance="low" />
-    <Message Text="TestFileFullPath = $(TargetPath)" Importance="low" />
-    <Message Text="VSTestSetting = $(VSTestSetting)" Importance="low" />
-    <Message Text="VSTestTestAdapterPath = $(VSTestTestAdapterPath)" Importance="low" />
-    <Message Text="VSTestFramework = $(TargetFrameworkMoniker)" Importance="low" />
-    <Message Text="VSTestPlatform = $(PlatformTarget)" Importance="low" />
-    <Message Text="VSTestTestCaseFilter = $(VSTestTestCaseFilter)" Importance="low" />
-    <Message Text="VSTestLogger = $(VSTestLogger)" Importance="low" />
+    <Message Text="Calling task Microsoft.TestPlatform.Build.Tasks.VSTestTask with following parameter..." Importance="Low" />
```
Nirmal4G, line 138:
```suggestion
    <Message Text="Calling task "Microsoft.TestPlatform.Build.Tasks.VSTestTask" with following parameters..." Importance="Low" />
```

**OR**

```suggestion
    <Message Text="Calling task "VSTestTask" with following parameters..." Importance="Low" />
```

I like this one better. Namespace is always implied, It's not like we're going to change it anytime soon!

src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets
```diff
@@ -70,37 +122,36 @@ Copyright (c) .NET Foundation. All rights reserved.
...
   </Target>
   
   <Target Name="ShowMsbuildWithParameter">
-    <Message Text="Building project $(MSBuildProjectFullPath) for TargetFramework=$(TargetFramework)" Importance="low"/>
-    <Message Text="Value passed to msbuild are..." Importance="low" />
-    <Message Text="Configuration = $(Configuration)" Importance="low" />
-    <Message Text="TargetFramework = $(TargetFramework)" Importance="low" />
-    <Message Text="Platform = $(PlatformTarget)" Importance="low" />
-    <Message Text="OutputPath = $(OutputPath)" Importance="low" />
+    <Message Text="Building project $(MSBuildProjectFullPath) for TargetFramework=$(TargetFramework)" Importance="Low"/>
```
Nirmal4G, line 129:
```suggestion
    <Message Importance="Low" Text="Building project $(MSBuildProjectFullPath) for TargetFramework=$(TargetFramework)"/>
```

Why not put importance first, that way, every message will be aligned with the lines around it?

Nirmal4G (changes requested):
Also, it's nothing major but can you split the changes into two commits:

- First commit can contain renaming the existing task.
- Second commit can contain introducing ToolTask based one.

Also, some nits; since we're overhauling the targets file, we should take this change and simplify some of it.

Nirmal4G:
I'm very interested in this. It's bit harder to review the source to check if nothing's changed, that's why I'm asking for two logical commits, one for the rename and other is the `ToolTask` based task implementation. I'll re-review the source once that's done.

mcartoixa:
@Nirmal4G I can't believe I got feedback on this PR (and on a weekend, too)... Thanks! 🥲 

My setup is a bit wonky: I can run the build but I can't compile on Visual Studio (hence the several forced commits after every merge...), so I will try and do my best under these conditions.

I think I understand your reasons for 2 commits. But as the goal of this change is to have 2 tasks that do the same thing in a different way their common behaviour has been gathered in an extension method, and it has also been slightly refactored to make better use of the MSBuild framework. So if I simply rename the existing task in the first commit, there will still be a lot going on in the second commit (interface, refactoring, extension method).
Maybe I could try to rename *and* refactor the existing task in the first commit, then introduce the new one and simply move the common behaviour in the extension method in the second commit. What do you think?

I'll also refactor the targets file as indicated (though the ultimate fate of these `Message` tasks is to disappear IMO).

mcartoixa:
@Nirmal4G I have finally settled on 3 commits. Tell me what you think.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+    public string? VSTestBlameHangTimeout { get; set; }
+    public ITaskItem? VSTestTraceDataCollectorDirectoryPath { get; set; }
+    public bool VSTestNoLogo { get; set; }
+    public string? VSTestArtifactsProcessingMode { get; set; }
+    public string? VSTestSessionCorrelationId { get; set; }
+
+    public override bool Execute()
+    {
+        var traceEnabledValue = Environment.GetEnvironmentVariable("VSTEST_BUILD_TRACE");
+        Tracing.traceEnabled = !traceEnabledValue.IsNullOrEmpty() && traceEnabledValue.Equals("1", StringComparison.OrdinalIgnoreCase);
```
Evangelink, line 50: `NullOrEmpty` check doesn't seem necessary if you use `string.Equals` syntax.

```suggestion
        Tracing.traceEnabled = string.Equals(traceEnabledValue, "1", StringComparison.OrdinalIgnoreCase);
```

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        var debugEnabled = Environment.GetEnvironmentVariable("VSTEST_BUILD_DEBUG");
+        if (!debugEnabled.IsNullOrEmpty() && debugEnabled.Equals("1", StringComparison.Ordinal))
```
Evangelink, line 53:
```suggestion
        if (string.Equals(debugEnabled, "1", StringComparison.Ordinal))
```

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            Console.WriteLine("Waiting for debugger attach...");
+
+            var currentProcess = Process.GetCurrentProcess();
+            Console.WriteLine(string.Format("Process Id: {0}, Name: {1}", currentProcess.Id, currentProcess.ProcessName));
```
Evangelink, line 58: `Console.WriteLine` has an overload with composite string so you don't need the `string.Format`.

```suggestion
            Console.WriteLine("Process Id: {0}, Name: {1}", currentProcess.Id, currentProcess.ProcessName);
```

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                Thread.Sleep(1000);
+            }
+
+            Debugger.Break();
+        }
+
+        // Avoid logging "Task returned false but did not log an error." on test failure, because we don't
+        // write MSBuild error. https://github.com/dotnet/msbuild/blob/51a1071f8871e0c93afbaf1b2ac2c9e59c7b6491/src/Framework/IBuildEngine7.cs#L12
+        var allowfailureWithoutError = BuildEngine.GetType().GetProperty("AllowFailureWithoutError");
```
Evangelink, line 70: Nit: `allowFailureWithoutError` (uppercase `F`).

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    public override bool Execute()
+    {
+        var traceEnabledValue = Environment.GetEnvironmentVariable("VSTEST_BUILD_TRACE");
+        Tracing.traceEnabled = !traceEnabledValue.IsNullOrEmpty() && traceEnabledValue.Equals("1", StringComparison.OrdinalIgnoreCase);
+
+        var debugEnabled = Environment.GetEnvironmentVariable("VSTEST_BUILD_DEBUG");
+        if (!debugEnabled.IsNullOrEmpty() && debugEnabled.Equals("1", StringComparison.Ordinal))
+        {
+            Console.WriteLine("Waiting for debugger attach...");
```
Evangelink, line 55: Probably needs to be put in a resource (to be translated).

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,109 @@
...
+            UseShellExecute = false,
+        };
+
+        if (!VSTestFramework.IsNullOrEmpty())
+        {
+            Console.WriteLine(Resources.Resources.TestRunningSummary, TestFileFullPath, VSTestFramework);
+        }
+
+        Tracing.Trace("VSTest: Starting vstest.console...");
+        Tracing.Trace("VSTest: Arguments: " + processInfo.FileName + " " + processInfo.Arguments);
```
Evangelink, line 86: Can we use interpolated string here? It's easier to read.

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -2,344 +2,98 @@
...
-    {
-        if (VSTestCLIRunSettings != null && VSTestCLIRunSettings.Length > 0)
-        {
-            allArgs.Add("--");
-            foreach (var arg in VSTestCLIRunSettings)
-            {
-                allArgs.Add(ArgumentEscaper.HandleEscapeSequenceInArgForProcessStart(arg));
-            }
-        }
+        return this.CreateCommandLineArguments();
```
Evangelink, line 62: Nit: 
```suggestion
        return CreateCommandLineArguments();
```

Evangelink (dismissed):
Did a first pass of review. I need to make some manual tests.

mcartoixa:
@Evangelink Thanks for the review. 🥲 

Unfortunately I just lost my main computer. I hope it's just an issue with the PSU 🤞🏻 I've ordered a new one but I may take some time before addressing your suggestions... Most of them are on the legacy code though 😉

src/Microsoft.TestPlatform.Build/Tasks/TestTaskExtensions.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        var isConsoleLoggerSpecifiedByUser = false;
+        var isCollectCodeCoverageEnabled = false;
+        var isRunSettingsEnabled = false;
+
+        var builder = new CommandLineBuilder();
+        builder.AppendSwitch("exec");
+        if (task.VSTestConsolePath != null && !task.VSTestConsolePath.ItemSpec.IsNullOrEmpty())
+        {
+            builder.AppendSwitchIfNotNull("", task.VSTestConsolePath);
```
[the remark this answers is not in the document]
mcartoixa: IIRC there is no [overload of `AppendSwitch`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.commandlinebuilder.appendswitch) that takes [an `ITaskItem` parameter](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.itaskitem), and furthermore I am not sure that _NotNull_ tests for empty (though I could check).
Reading the documentation, [the fairly new `AppendFileNameIfNotNull`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.commandlinebuilder.appendfilenameifnotnull) should be more appropriate here.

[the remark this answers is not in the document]
Evangelink: My bad, I didn't check the available APIs (I just assumed it existed). I am closing this comment!

Evangelink:
@mcartoixa No worries, take your time. We need to discuss and validate this change quite extensively anyways.

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
mcartoixa: This is an extension method, in order to keep as much functionality shared between the tasks as possible. So I think `this` is necessary here.
It might be reworked as a simple static method instead?

Nirmal4G: I think what @Evangelink means is this:

```suggestion
        return CreateCommandLineArguments(this);
```

But 

> It might be reworked as a simple static method instead?

I might prefer this instead!

Evangelink: Sorry I missed it was an extension method. Yeah seems like it would be better as static method.

mcartoixa:
@Evangelink I think I have addressed most of your points in the latest commits. Let me know what you think.

src/Microsoft.TestPlatform.Build/Resources/Resources.Designer.cs
```diff
@@ -1,26 +1,25 @@
 ﻿//------------------------------------------------------------------------------
 // <auto-generated>
-//     This code was generated by a tool.
-//     Runtime Version:4.0.30319.42000
+//     Ce code a été généré par un outil.
+//     Version du runtime :4.0.30319.42000
```
Nirmal4G, line 4: You need to undo these comment language changes. They should only be in English. You can update the localizations in XLF instead of putting it in source file

mcartoixa: Nice catch. Now I know why I was afraid of meddling with those resources. 😓

Nirmal4G: Been there; done that!! 😇

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        builder.AppendSwitchIfNotNull("--testCaseFilter:", task.VSTestTestCaseFilter);
+
+        if (task.VSTestLogger != null && task.VSTestLogger.Any())
+        {
+            foreach (var arg in task.VSTestLogger)
+            {
+                builder.AppendSwitchIfNotNull("--logger:", arg);
+
+                if ((arg != null) && arg.StartsWith("console", StringComparison.OrdinalIgnoreCase))
```
Evangelink, line 65: Nit: un-needed parenthesis

```suggestion
                if (arg != null && arg.StartsWith("console", StringComparison.OrdinalIgnoreCase))
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        // vstest.console only support x86 and x64 for argument platform
+        if (!task.VSTestPlatform.IsNullOrEmpty() && !task.VSTestPlatform.Contains("AnyCPU"))
+        {
+            builder.AppendSwitchIfNotNull("--platform:", task.VSTestPlatform);
+        }
+
+        builder.AppendSwitchIfNotNull("--testCaseFilter:", task.VSTestTestCaseFilter);
+
+        if (task.VSTestLogger != null && task.VSTestLogger.Any())
```
Evangelink, line 59: Not sure if the `Any` check brings anything as we are iterating with foreach. I'd say we are even doing 1 "iteration" for nothing.

```suggestion
        if (task.VSTestLogger != null)
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            var normalTestLogging = new List<string>() { "n", "normal", "d", "detailed", "diag", "diagnostic" };
+            var quietTestLogging = new List<string>() { "q", "quiet" };
+
+            string vsTestVerbosity = "minimal";
+            if (normalTestLogging.Contains(task.VSTestVerbosity.ToLowerInvariant()))
+            {
+                vsTestVerbosity = "normal";
+            }
+            else if (quietTestLogging.Contains(task.VSTestVerbosity.ToLowerInvariant()))
```
Evangelink, line 101: I would introduce a variable to avoid calling twice `ToLowerInvariant`.

```suggestion
            var taskVsTestVerbosity = task.VSTestVerbosity.ToLowerInvariant();
            if (normalTestLogging.Contains(taskVsTestVerbosity))
            {
                vsTestVerbosity = "normal";
            }
            else if (quietTestLogging.Contains(taskVsTestVerbosity))
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+                vsTestVerbosity = "quiet";
+            }
+
+            builder.AppendSwitchUnquotedIfNotNull("--logger:", $"Console;Verbosity={vsTestVerbosity}");
+        }
+
+        if (task.VSTestBlame || task.VSTestBlameCrash || task.VSTestBlameHang)
+        {
+            var dumpArgs = new List<string>();
+            if (task.VSTestBlameCrash || task.VSTestBlameHang)
```
Evangelink, line 112: This outer if seems un-needed as it only contains 2 ifs checking each side of this outer if condition.

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+            {
+                builder.AppendSwitchIfNotNull("--Blame:", string.Join(";", dumpArgs));
+            }
+            else
+            {
+                builder.AppendSwitch("--Blame");
+            }
+        }
+
+        if (task.VSTestCollect != null && task.VSTestCollect.Any())
```
Evangelink, line 154: The check to `Any` is not needed.

```suggestion
        if (task.VSTestCollect != null)
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+                    // Split the argument with ';' and compare first token value.
+                    var tokens = arg.Split(';');
+
+                    if (arg.Equals(codeCoverageString, StringComparison.OrdinalIgnoreCase) ||
+                        tokens[0].Equals(codeCoverageString, StringComparison.OrdinalIgnoreCase))
+                    {
+                        isCollectCodeCoverageEnabled = true;
+                    }
+                }
+                builder.AppendSwitchIfNotNull("--collect:", arg);
```
Evangelink, line 170: It seems it could be simplified by inverting the condition.

```suggestion
                if (arg == null)
                {  
                    continue;
                }

                // For collecting code coverage, argument value can be either "Code Coverage" or "Code Coverage;a=b;c=d".
                // Split the argument with ';' and compare first token value.
                var tokens = arg.Split(';');

                if (arg.Equals(codeCoverageString, StringComparison.OrdinalIgnoreCase) ||
                    tokens[0].Equals(codeCoverageString, StringComparison.OrdinalIgnoreCase))
                {
                    isCollectCodeCoverageEnabled = true;
                }

                builder.AppendSwitchIfNotNull("--collect:", arg);
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+            }
+            else
+            {
+                if (isCollectCodeCoverageEnabled)
+                {
+                    // Not showing message in runsettings scenario, because we are not sure that code coverage is enabled.
+                    // User might be using older Microsoft.NET.Test.Sdk which don't have CodeCoverage infra.
+                    task.Log.LogWarning(Resources.Resources.UpdateTestSdkForCollectingCodeCoverage);
+                }
+            }
```
Evangelink, line 195: If could be merged with outer else

```suggestion
            else if (isCollectCodeCoverageEnabled)
            {
                // Not showing message in runsettings scenario, because we are not sure that code coverage is enabled.
                // User might be using older Microsoft.NET.Test.Sdk which don't have CodeCoverage infra.
                task.Log.LogWarning(Resources.Resources.UpdateTestSdkForCollectingCodeCoverage);
            }
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        if (!task.VSTestArtifactsProcessingMode.IsNullOrEmpty() && task.VSTestArtifactsProcessingMode.Equals("collect", StringComparison.OrdinalIgnoreCase))
+        {
+            builder.AppendSwitch("--artifactsProcessingMode-collect");
+        }
+
+        builder.AppendSwitchIfNotNull("--testSessionCorrelationId:", task.VSTestSessionCorrelationId);
+
+        // VSTestCLIRunSettings should be last argument as vstest.console ignore options after "--"(CLIRunSettings option).
```
Evangelink, line 210: Nit:

```suggestion
        // VSTestCLIRunSettings should be last argument as vstest.console ignore options after "--" (CLIRunSettings option).
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -0,0 +1,222 @@
...
+                }
+            }
+        }
+
+        if (task.VSTestNoLogo)
+        {
+            builder.AppendSwitch("--nologo");
+        }
+
+        if (!task.VSTestArtifactsProcessingMode.IsNullOrEmpty() && task.VSTestArtifactsProcessingMode.Equals("collect", StringComparison.OrdinalIgnoreCase))
```
Evangelink, line 203: Nit:

```suggestion
        if (string.Equals(task.VSTestArtifactsProcessingMode, "collect", StringComparison.OrdinalIgnoreCase))
```

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    public override bool Execute()
+    {
+        var traceEnabledValue = Environment.GetEnvironmentVariable("VSTEST_BUILD_TRACE");
+        Tracing.traceEnabled = string.Equals(traceEnabledValue, "1", StringComparison.OrdinalIgnoreCase);
+
+        var debugEnabled = Environment.GetEnvironmentVariable("VSTEST_BUILD_DEBUG");
+        if (string.Equals(debugEnabled, "1", StringComparison.Ordinal))
+        {
+            Console.WriteLine("Waiting for debugger attach...");
```
Evangelink, line 56: Maybe this can be part of a follow-up PR but I feel like we should not be using `Console.WriteLine`.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,110 @@
...
+    public string? VSTestBlameHangDumpType { get; set; }
+    public string? VSTestBlameHangTimeout { get; set; }
+    public ITaskItem? VSTestTraceDataCollectorDirectoryPath { get; set; }
+    public bool VSTestNoLogo { get; set; }
+    public string? VSTestArtifactsProcessingMode { get; set; }
+    public string? VSTestSessionCorrelationId { get; set; }
+
+    public override bool Execute()
+    {
+        var traceEnabledValue = Environment.GetEnvironmentVariable("VSTEST_BUILD_TRACE");
```
Evangelink, line 50: Could we add:

1. An internal ctor taking `IEnvironmentVariableHelper` as argument.
2. Save this argument as a field
3. A public parameterless ctor calling the internal ctor with `new EnvironmentVariableHelper()`
4. Use the field instead of the static `Environment.GetEnvironmentVariable` here and below

This would help with testing via mocking.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
```diff
@@ -0,0 +1,110 @@
...
+        // write MSBuild error. https://github.com/dotnet/msbuild/blob/51a1071f8871e0c93afbaf1b2ac2c9e59c7b6491/src/Framework/IBuildEngine7.cs#L12
+        var allowFailureWithoutError = BuildEngine.GetType().GetProperty("AllowFailureWithoutError");
+        allowFailureWithoutError?.SetValue(BuildEngine, true);
+
+        var processInfo = new ProcessStartInfo
+        {
+            FileName = DotnetExe,
+            Arguments = TestTaskUtils.CreateCommandLineArguments(this),
+            UseShellExecute = false,
+        };
```
Evangelink, line 79: Shall we redirect outputs?

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -2,344 +2,98 @@
...
-                    isCollectCodeCoverageEnabled = true;
-                }
-
-                allArgs.Add("--collect:" + ArgumentEscaper.HandleEscapeSequenceInArgForProcessStart(arg));
-            }
-        }
+        //TODO: https://github.com/dotnet/sdk/issues/20 Need to get the dotnet path from MSBuild?
 
-        if (isCollectCodeCoverageEnabled || isRunSettingsEnabled)
+        var dhp = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
```
Evangelink, line 74: Same as for base class, please use `IEnvironmentVariableHelper`.

src/Microsoft.TestPlatform.Build/Tasks/VSTestTask.cs
```diff
@@ -2,344 +2,98 @@
...
             }
         }
 
-        if (!VSTestNoLogo.IsNullOrWhiteSpace())
-        {
-            allArgs.Add("--nologo");
-        }
-
-        if (!VSTestArtifactsProcessingMode.IsNullOrEmpty() && VSTestArtifactsProcessingMode.Equals("collect", StringComparison.OrdinalIgnoreCase))
+        if (File.Exists(ToolExe))
```
Evangelink, line 84: Can we also take `IFileHelper` as ctor argument and use it here?

Evangelink:
I have run some tests locally and so far I have only seen some "problem" when there are errors running tests.

For example,

Before:

```
Test run for C:\XXX\TestProject1\TestProject3\bin\Debug\net6.0\TestProject3.dll (.NETCoreApp,Version=v6.0)
Microsoft (R) Test Execution Command Line Tool Version 17.3.0-preview-20220523-03 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

A total of 1 test files matched the specified pattern.
Html Logger Error : '￾', hexadecimal value 0xFFFE, is an invalid character. Line 1, position 461.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: < 1 ms - TestProject3.dll (net6.0)
```

After:

```
C:\XXX\artifacts\testArtifacts\dotnet\sdk\7.0.100-preview.4.22252.9\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(216,5): message NETSDK1057: You are using a preview v
ersion of .NET. See: https://aka.ms/dotnet-support-policy [C:\XXX\TestProject1\TestProject3\TestProject3.csproj]
  TestProject3 -> C:\XXX\TestProject1\TestProject3\bin\Debug\net6.0\TestProject3.dll
C:\XXX\artifacts\testArtifacts\dotnet\sdk\7.0.100-preview.4.22252.9\Microsoft.TestPlatform.targets(41,5): error : Html Logger Error : '?', hexadecimal value 0xFFFE, is an invalid character. Line 1, pos
ition 461. [C:\XXX\TestProject1\TestProject3\TestProject3.csproj]
```

mcartoixa:
@Evangelink This behaviour does not surprise me. I assume you are testing the `dotnet test` command here.
In this PR, legacy behaviour is guarantied only if the `$(VSTestUseConsole)` property is set to `true` (cf. the https://github.com/dotnet/sdk/pull/15376 PR). This is obviously debatable, but the correct behaviour should be by default and IMHO the correct behaviour of a MSBuild task is to conform to the MSBuild standard: less workarounds and more predictability. If you `SET VSTestUseConsole=True` before running your tests you should see no difference in the output. As I indicated (a long time ago), the `$(VSTestUseConsole)` property should then be properly initialized inthe `dotnet test` command itself.
MSBuild is very limited in its ability to colorize outputs. But its ouputs are very clearly organized and you can see more or less by playing with the verbosity level. Errors are the highest level, which is why you can still see them in the output.

So to recap: in the short term, dotnet test should set the `$(VSTestUseConsole)` property to `true` to kee the current behaviour. In the long term it would be a good idea IMHO if `dotnet test` did not depend on MSBuild but on `dotnet vstest.console.dll` directly. This way you can keep the colorized output, and the MSBuild task can be used when it is beneficial. In the context of Continuous Integration for instance.

Evangelink:
@mcartoixa It's not easy to have the full picture at first so thank you so much for the explanation! I will make more manual tests and ping the team to have more reviews.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
mcartoixa: The whole purpose of this class is to send output to the console, an the whole purpose of this PR is to have an alternative class that does **not** do that. So, yeah 😁

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
mcartoixa: It is useful when `VSTestBlame` is `true` and the others are `false`: it adds the `--Blame` switch.

src/Microsoft.TestPlatform.Build/Tasks/VSTestForwardingTask.cs
mcartoixa: I'd rather not mess too much with this legacy (IMO) code...

mcartoixa:
[This commit fixed the integration tests](https://github.com/microsoft/vstest/pull/2702/commits/8dcc625896109cf7694d17af5fa08776565ff417) btw 😉 
Anyone knows why the _microsoft.vstest.pr_ checks don't run anymore?

Evangelink:
@mcartoixa Not sure but from bot message it looks like this is due to merge conflicts (we had to touch the file in some other PR).

mcartoixa:
Thanks. The first 18 months were the hardest (it felt pretty lonely around here) 😉

mcartoixa:
So it happens that using `IFileHelper` and `IEnvironmentHelper` requires a new dependency on `Microsoft.TestPlatform.CoreUtilities` which messes things up too much with the current build. Plus I still can't compile on my computer so I'll revert those changes.

mcartoixa:
@krwq Thanks for the push.

I'd rather not mess up too much with this PR myself: it is very old now, my development setup is very unstable (I haven't been able to compile the project for at least a year), and I already split up the original commit at someone's request a long time ago.
If I remember correctly, the first 3 commits are splits from the original PR, the rest are various fixes and enhancements that have been requested along the way.

To summarize:

- the current, console based MSBuild task is renamed _VSTestForwardingTask_ (and kept for compatibility reasons).
- it is used by the build, provided `$(VSTestUseConsole)` is set to `true` (not the default value, so that the build does the right thing by default).
- _VSTestTask_ is a new task that uses the MSBuild logging infrastructure instead of the console, as it should. A side effect is that the colorized output provided by _VSTestForwardingTask_ is lost (we can only have yellow for warnings and red for errors).

If this PR is merged, another PR should be made in _dotnet/sdk_ so that `dotnet test` internally sets `$(VSTestUseConsole)=true` (cf. dotnet/sdk#15376). Users who want the MSBuild logging can execute `dotnet msbuild -Target:VSTest`.

joperezr:
@nohwnd @Evangelink Can we prioritize getting this pull request in? We are having some test failures that are very hard to reproduce and when it does we can't even know which one failed due to this very old bug. We'd really appreciate assistance here.

nohwnd:
@joperezr we are discussing what we should do with our PM.

mcartoixa:
> we are discussing what we should do with our PM.

@nohwnd Any update? 🐌

nohwnd:
Yup. planning to do this in net9. Now I just need to find a bit of time to work on this. :)

nohwnd:
@mcartoixa  I've renamed the original task back to what it was so we don't change the public api that much, inverted the flag, renamed the tests, and now I am adding some tests for this new switch.  Please see #4706 I want to avoid pushing  extensive changes into your branch but will in the end push here.

When I am trying to use this task I don't get much useful output. Either I run with minimal output and all I see is: 

image


When I increase the build verbosity to normal, I see the normal console log as a message, but it is dispersed through the build so it makes it a bit more confusing.  Do you see an easy way to write 1 error per failed test, kinda like a compiler does when there is failed build?

image

nohwnd:
I need to write to error output, of course. Sorry for spamming you,  next time I will try to think first and then ask  😁

mcartoixa:
> I've renamed the original task back to what it was so we don't change the public api that much, inverted the flag, renamed the tests, and now I am adding some tests for this new switch.

My choices were made on the assumption that the build should do the right by default. As we are implementing the build with MSBuild, it should be compatible with MSBuild by default (think _[Pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/)_). Most of the bugs related to this PR originate from users and tools that use MSBuild and assume MSBuild behaviour. In my mind the use of  `dotnet test` is a particular case, with specific requirements (even though most users will use it this way), which are in essence incompatible with MSBuild. In fact, if I were involved in the project I would push very hard to **not** use MSBuild anymore in `dotnet test` (though I know this is difficult as dotnet currently assumes the use of MSBuild).
But it's your project, so I guess it's your call.

> When I am trying to use this task I don't get much useful output.

`dotnet test`, like other _dotnet_ commands, sets a very minimal level of output by default. You should use `dotnet msbuild` to get the default MSBuild behaviour.
Also, I think the following lines in the constructor should get you were you want to be:
```
    public VSTestTask()
    {
        LogStandardErrorAsError = true;
        StandardOutputImportance = "Normal";
    }
```

mcartoixa:
> I need to write to error output, of course. Sorry for spamming you, next time I will try to think first and then ask 😁

No worries. I am glad there is some movement on this topic after all this time.

nohwnd:
I've added a new logger, which should log an error for each failed test. (pushed to the prototype PR #4706). 

image

The output has few problems: 

In the logger we are getting output that has some newlines in it, the tool task seem to represent every new line as a new error, so even though I have 1 failed test I see way more errors.  Some of the output is already formatted by the test framework and assertion library so we cannot fully avoid newlines, unless we replace them all. 

What would you expect to see in the output? Normally there would be the error message, which test failed, the full callstack, any captured standard output, error output, and dbg output. 

Is expected to just show one line result message, and store the rest of the info in trx? 

@KirillOsenkov in case you have some ideas or tips or tricks. @AArnott in case you also have some suggestions or expectations.

mcartoixa:
> What would you expect to see in the output?

If I'm not mistaken, the default behaviour of a MSBuild task is to output any standard error output (by the test tool in this case) as an error. This is just fine to me: I don't need anything else.

KirillOsenkov:
The output that is there right now seems like the right amount of information, but we need to find a way to make it a single error message with line breaks in it, instead of splitting it into error per line. I'm not sure how to do that, probably need to read the source of the ToolTask or whoever is creating these `BuildErrorEventArgs` instances.

KirillOsenkov:
If all we get from the runner is a multi-line string in the Error stream, and no other information via other channels, then one hacky way would be to parse those and coalesce into one error per failure. Seems fragile.

Is there really no better API to connect to the runner and receive the error information in a more structured way?

nohwnd:
> Is there really no better API to connect to the runner and receive the error information in a more structured way?

We could use the vstest.console wrapper to connect to the testhost as we do under VS. And use that to communicate with the process. We then get callbacks in the original process where we can collect the information and report them via msbuild api.

mcartoixa:
I don't think the stack trace is very relevant at the error level in MSBuild... MSBuild output needs to be simple. IMHO the "Assert failed" message is relevant, the rest can be output as standard. I will try and find what the venerable old .NET used to output, for reference.

Also, instead of a specific logger you could just disable the default error logging (`LogStandardErrorAsError = false;`) and override [the _LogEventsFromTextOutput_ method](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask.logeventsfromtextoutput?view=msbuild-17-netcore). In it use the `Log` property, which returns [a _TaskLoggingHelper_](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper?view=msbuild-17-netcore), to log messages, warnings and errors as you wish. Example from a very old project of mine:
```csharp
protected override void LogEventsFromTextOutput(string singleLine, MessageImportance messageImportance)
{
    if (string.IsNullOrWhiteSpace(singleLine))
        return;

    // Most probably an info
    if (char.IsWhiteSpace(singleLine[0]))
        messageImportance=MessageImportance.Low;

    base.LogEventsFromTextOutput(singleLine, messageImportance);
}
```

nohwnd:
> > Is there really no better API to connect to the runner and receive the error information in a more structured way?
> 
> We could use the vstest.console wrapper to connect to the testhost as we do under VS. And use that to communicate with the process. We then get callbacks in the original process where we can collect the information and report them via msbuild api.

Okay so this is also a no-go. vstest.console can accept only runsetting through the client and not arguments. But dotnet test allows most of the parameters that vstest.cosole would take, and we would either have to: 
- translate them to runsettings: complicated
- run the argument processors at the start of a request to translate them to runsettings, with the exception of any action processors, while substituting all static dependencies: even more difficult

So commandline communication it is for now.

mcartoixa:
> So commandline communication it is for now.

The current problem to solve is no output from MSBuild. So _any_ kind of output is an order of magnitude better than that.
I just checked old projects on .NET 4.0, and everything was on standard output at the time. So any improvement on this front is indeed... an improvement.
image

But the use case of executing tests with `dotnet msbuild` is different than the use case of `dotnet test`. The latter is targeted to a developer using a console, the former is meant to be used in relation to other scripts and tools (like CI). This usually means relying on logs and result files (cf. _VSTestResultsDirectory_) for troubleshooting.

Also, for reference, _xunit_ generates an output that is very suited for MSBuild (and its users), with a single line describing the error and the rest as standard output:
image

AArnott:
> But the use case of executing tests with dotnet msbuild is different than the use case of dotnet test. The latter is targeted to a developer using a console, the former is meant to be used in relation to other scripts and tools (like CI). This usually means relying on logs and result files (cf. VSTestResultsDirectory) for troubleshooting.

This is news to me. I _never_ have used `dotnet msbuild`, local, CI or otherwise. I just use `dotnet test` in CI. Is there a doc anywhere that advocates for using `dotnet msbuild` for anything? I don't even know how it differs from `dotnet build`, except (just guessing here) that `dotnet msbuild` won't restore by default?

> I don't think the stack trace is very relevant at the error level in MSBuild... MSBuild output needs to be simple. IMHO the "Assert failed" message is relevant, the rest can be output as standard.

+1 to @mcartoixa. The 'error' log message should be one line. It could say "test X failed" or it could include the exception message, up to the first newline. The rest of the details should be emitted with MessageImportance.Normal (or Low, perhaps), IMO.

mcartoixa:
> This is news to me. I _never_ have used `dotnet msbuild`, local, CI or otherwise. I just use `dotnet test` in CI. Is there a doc anywhere that advocates for using `dotnet msbuild` for anything? I don't even know how it differs from `dotnet build`, except (just guessing here) that `dotnet msbuild` won't restore by default?

Well, right now it's essentially the same thing: `dotnet test` is `dotnet msbuild` in disguise [with some adequate default values](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-test/Program.cs#L125-L130) (like every other `dotnet` command, [including `dotnet build`](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-build/BuildCommand.cs#L42)). 😁 This is why we have a problem, by the way: the output of tests that was tailored for end users is essentially at odds with MSBuild and requires some _tricks_ to _work_...
If this PR were merged, `dotnet test` would keep the same behaviour, but the select people who encounter a bug related to these tricks (no logs, or no output, or output in a strange order...) would be able to execute their tests with `dotnet msbuild` (and the proper arguments) and expect a proper MSBuild behaviour.
This is also why I advocated earlier that using `dotnet msbuild` should not require specific parameters and behave like it should by default. Specific parameters can be silently added to `dotnet test` to trigger its behaviour.

To me this means that `dotnet msbuild` will naturally find its place (again) in the core of CI environments. Also, as a side note, [I took the habit](https://dev.to/mcartoixa/continuous-integration-for-net-core-1fk4) a looong time ago to describe my entire build process (build, tests, analysis, packaging...) using MSBuild. This makes this process testable locally, largely independent of any CI vendor and it simplifies greatly the (usually YAML) configuration.

nohwnd:
New approach base on what Kirill suggested, where the error is output formatted, and then reported using MSBuild api. 

image

I take the location from the first stack frame that I get from the filtered StackTrace from each respective framework. Here is how it looks like for mstest, xunit and nunit.

One problem is when we don't have symbols, or the library that throws does not have it. Like fluent assertions in the first line.

KirillOsenkov:
Looks great! How does it look in the binlog?

nohwnd:
Like this,  full log attached.

image

[msbuild.zip](https://github.com/microsoft/vstest/files/13306072/msbuild.zip)

KirillOsenkov:
A couple small questions about that:

1. should we have a "nologo" mode that skips printing these three messages?
 
image

2. should we log a low-importance message with the full call stack in addition to the error?
3. would it be possible to embed the file path and line number of the .cs file that threw the exception? Right now if we press Space on the error node in the binlog viewer, it opens Microsoft.TestPlatform.targets line 41 (which is where the `<VSTestTask2` is declared in XML). It's good, but it'd be more helpful to go to the C# code that threw the exception. However you're saying the symbols might be missing, so we might not have source locations to figure out where that .cs line of code is.

nohwnd:
1. yup, 
2. I think I can do that.
3. the error in Fluent.csproj points to the targets, because I don't have symbols, and so I don't have a better place to point it to. But the other three errors point at the source code. And pressing space on them works correctly for me. Could you re-check?

image

KirillOsenkov:
Indeed! Awesome!

KirillOsenkov:
I think this is great!

nohwnd:
1. /p:vstestnologo=true was already there and it suppresses the first 2 messages. I think there is an issue for printing nothing when verbosity=quiet is provided, but I don't think I am ready to implement it here.

2.  
image


3. this is solved by picking the first location we can find in the first 20 stack frames

nohwnd (approved):

mcartoixa:
Great to see something done after 3 years! 👏🏻 

As a user, I still can't [wrap my head around this line](https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets#L18) though: if I use MSBuild, I have to set `VsTestUseMSBuildOutput=True` to _gain_ MSBuild compatibility?

nohwnd:
Yes, for now, I want to get feedback first before we ship as the default.

-jakub
________________________________
Od: Mac ***@***.***>
Odesláno: Friday, November 17, 2023 12:01:15 PM
Komu: microsoft/vstest ***@***.***>
Kopie: Jakub Jares ***@***.***>; Mention ***@***.***>
Předmět: Re: [microsoft/vstest] Create a proper MSBuild ToolTask based VSTestTask (#2702)


Great to see something done after 3 years! 👏🏻

As a user, I still can't wrap my head around this line<https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Build/Microsoft.TestPlatform.targets#L18> though: if I use MSBuild, I have to set VsTestUseMSBuildOutput=True to gain MSBuild compatibility?

KirillOsenkov:
Thank you both so much @mcartoixa and @nohwnd! Lots of hard work went into this. Very happy to see it merged.

nohwnd:
Thanks, I will update here once we release it to somewhere. :)

nohwnd:
Not yet in 9.0.100-alpha.1.23564.26.

nohwnd:
@mcartoixa @KirillOsenkov this is now default (as well as terminal logger, and reusing nodes) in [net9 9.0.100-alpha.1.24073.1 and newer](https://github.com/dotnet/installer#table). Feedback is welcome, and collected here #4843 😊

Issue #2708, opened 2021-01-21
Move FQN related code into a separate NuGet package 


Haplois:
## Description
As part of supporting new TestCase properties like `ManagedType` and `ManagedClass`, some changes introduced to `Microsoft.TestPlatform.ObjectModel`. (https://github.com/microsoft/vstest/pull/2611) These changes broke the compatibility of the Test Platform with earlier Visual Studio releases. 

We need to move new TestCase properties, Managed Name Utilities, Modern Test Id generation logic to a separate assembly and ultimately into a new NuGet package, preferably without any dependencies into the Test Platform itself.

We need to consider supporting the source-based test discovery support extension and providing support in a second new NuGet package since it depends on Roslyn.

This change should fix https://github.com/microsoft/testfx/issues/748.

## Steps
 - [x] Move the `Microsoft.TestPlatform.ObjectModel/ManagedNameUtilities` into the new NuGet package. The New NuGet package is called `Microsoft.TestPlatform.AdapterUtilities`; it has no dependencies to the Test Platform itself. It only contains constants and helper methods.
 - [x] Remove hardcoded support and .ctors from TestCase, and TestFx, and move it to the new NuGet package
 - [x] Provide a way to customize TestId generation by the adapter
 - [x] Make sure the TestPlatform is compatible with Visual Studio 16.8 (https://github.com/microsoft/testfx/pull/765)
 - [x] Implement the new feature in MSTest (https://github.com/microsoft/testfx/pull/765)
 - [x] Figure out how to support hierarchies in TestCase (Namespace, Class, TestGroup, DisplayName, ~~DataRow~~) (https://github.com/microsoft/vstest/pull/2724)
 - ~~Update the documentation, provide implementation samples and guidance for adapter authors, etc.~~ (We will add this after hierarchy support is merged into https://github.com/microsoft/testfx)
 
## Remarks
We will do test Id customization in the adapters. Adapters will create and assign the test id, and We will provide a sample here once it is checked into https://github.com/microsoft/testfx.

AB#1268457

PR #2714, opened 2021-01-24
Move FQN related code into a separate NuGet package

Haplois:
## Description
As part of https://github.com/microsoft/vstest/issues/2708:
 - [x] Moved ManagedNameUtilities into a new project.
 - [x] Removed hardcoded support from TestCase class to restore compatibility with older versions of TestPlatform.
 - [x] Create a new NuGet package
 - [x] Determine what to call `Microsoft.TestPlatform.ManagedNames` package. We're calling it `Microsoft.TestPlatform.AdapterUtilities`.
 - [x] Fix Ubuntu build


src/Microsoft.TestPlatform.ManagedNames/Extensions/ReflectionExtensions.cs
```diff
@@ -1,20 +1,20 @@
...
+    internal static partial class ReflectionExtensions
     {
-        private static void AssertSupport<T>(T obj, string methodName, string className) 
+        private static void AssertSupport<T>(T obj, string methodName, string className)
             where T : class
         {
             if (obj == null)
             {
-                throw new NotImplementedException(string.Format(Resources.MethodNotImplementedOnPlatform, className, methodName));
+                throw new NotImplementedException(string.Format(ManagedNameMessages.MethodNotImplementedOnPlatform, className, methodName));
```
Sanan07, line 17: Better to change to ManagedNameResources to be consistent with other resources.

src/Microsoft.TestPlatform.ManagedNames/ManagedNameUtilities/Contants.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.ManagedNameUtilities
+{
+    public static class Contants
+    {
+        public const string ManagedTypeLabel = "ManagedType";
+        public const string ManagedMethodLabel = "ManagedMethod";
+
+        public const string ManagedTypePropertyId = "TestCase." + ManagedTypeLabel;
```
Sanan07, line 11: Why do not have value equal to "TestCase.ManagedType" as it is const?

src/Microsoft.TestPlatform.ManagedNames/Extensions/TestCase2Extensions.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.VisualStudio.TestPlatform.Extensions
+{
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+
+    public static class TestCase2Extensions
+    {
+        /// <summary>
+        /// Returns <c>true</c> if both <see cref="ITestCase2.ManagedType" /> and <see cref="ITestCase2.ManagedMethod" /> are not null or whitespace.
+        /// </summary>
+        public static bool HasManagedMethodAndType(this ITestCase2 testCase)
```
Sanan07, line 13:
```suggestion
        public static bool ContainsManagedMethodAndType(this ITestCase2 testCase)
```

src/Microsoft.TestPlatform.ManagedNames/Extensions/ReflectionExtensions.cs
Haplois: You're right, but this resource is moved from `ObjectModel`, and it's already translated. It's unnecessarily long refactoring for no apparent gain IMO. (All `.resx`, `.xlf` and generated code needs to be updated)

src/Microsoft.TestPlatform.ManagedNames/Extensions/TestCase2Extensions.cs
cvpoienaru: Assuming we cannot change the `ITestCase2` interface, maybe we should consider introducing `ITestCase3` that inherits `ITestCase2` and adds the new methods/properties instead of working around with extension methods ?

src/Microsoft.TestPlatform.ManagedNames/ManagedNameUtilities/ManagedNameParser.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.ManagedNameUtilities
 
     public class ManagedNameParser
```
cvpoienaru, line 13: Maybe this class should be internal too ?

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -450,6 +329,35 @@ protected override void ProtectedSetPropertyValue(TestProperty property, object
...
+        string ITestCase2.ManagedMethod
+        {
+            get => GetPropertyValue<string>(TestCaseProperties.ManagedMethod, null);
+            set => SetPropertyAndResetId(TestCaseProperties.ManagedMethod, value);
+        }
+
+        /// <inheritdoc />
+        string ITestCase2.GetDisplayName()
+        {
+            if (this.HasManagedMethodAndType())
```
cvpoienaru: nit: This method's body can be re-written using the ternary operator, but I acknowledge it's a matter of personal taste.

src/Microsoft.TestPlatform.ManagedNames/Extensions/TestCase2Extensions.cs
Haplois: `ITestCase2` interface is new. `ContainsManagedMethodAndType` is not included there; I'll include it. The interface is explicitly implemented, so these extension methods provide ease of access to those members.

src/Microsoft.TestPlatform.ManagedNames/ManagedNameUtilities/ManagedNameParser.cs
Haplois: We like to provide these helper methods mainly for adapter authors. That's the reason they are public.

peterwald (approved):

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
...
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
-    using System.Reflection;
+    using Microsoft.VisualStudio.TestPlatform.Extensions;
 
     /// <summary>
     /// Stores information about a test case.
     /// </summary>
     [DataContract]
-    public sealed class TestCase : TestObject
+    public sealed class TestCase : TestObject, ITestCase2
```
peterwald, line 19: It seems like there could be a failure in the case where a new test adapter (implementing the Managed properties), was used with an older version of the test platform.  Say, for instance, that vstest.console is run from the command line on an older install, but it is referencing a new adapter that supports the new properties. I think it would build and load properly, but the extension methods that access ITestCase2 would fail. Can we protect against that?

Haplois: The problem is that we can't access `TestProperty` from this new package. The only way I can see is not providing this interface at all and doing everything in the test adapters; in that case, I don't see a way to customize `TestId` generation. Any ideas?

peterwald: If we do everything in the test adapters, it seems to me that would work. TestId's are ideally created in the test adapter, so it seems like they could use the new code in the nuget package to generate it. Am I missing something?

Haplois: > TestId's are ideally created in the test adapter

In this case, no. I'll move the functionality into the test adapter.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
Haplois: Method removed.

shyamnamboodiripad:
@Haplois @peterwald Can we now delete the Reflection-based implementation of FQN from the VSUnitTesting repo?

I would still like to ensure that the Reflection and Roslyn based implementations continue to align - so it would be great to preserve unit tests for the FQN functionality in VSUnitTesting repo (i.e. replace the Reflection implementation testted by the unit tests with the new MS.TP.AdapterUtilities nuget - but continue to reference Roslyn based implementation from the VSUnitTesting repo).

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
+  </PropertyGroup>
+  <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
+  
+  <PropertyGroup>
+    <TargetFrameworks>netstandard1.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);uap10.0;</TargetFrameworks>
```
shyamnamboodiripad, line 9: Just an observation on something that is likely to come up in the future: Since managed test frameworks will probably require this reference to be added in their framework layer (as opposed to the VS / adapter layer) and since the framework layer will likely be consumed in older versions of VS (as well as outside of VS), test frameworks may end up needing to conditionally add package reference for `MS.TP.AdapterUtilities` to only a subset of their supported targets and / or conditionally opt in to the FQN functionality in their framework layer only in cases where they know that the caller is VS. Alternatively, if they can't consume the package they may need to maintain a parallel implementation of the FQN spec or copy the Reflection code from the current repo...

shyamnamboodiripad (commented):
👍🏾 modulo comments

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -139,7 +139,7 @@ public static MethodBase GetManagedName(Assembly assembly, string fullTypeName,
 
             if (type == null)
```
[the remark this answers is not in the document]
Haplois: I refactored the parameters on the first method, as you suggested. However, for the second method (one that returns MethodBase), the argument names were actually misleading. `fullTypeName` is actually the type name of `Assembly.GetName`, and not our `ManagedTypeName`. I renamed `fullTypeName` to `typeName`, and `fullMethodName` to `managedMethodName` on that method.

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
```diff
@@ -0,0 +1,53 @@
...
+    <TargetFrameworkIdentifier Condition="'$(OS)' != 'Windows_NT'">.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion Condition="'$(OS)' != 'Windows_NT'">v4.5.1</TargetFrameworkVersion>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'uap10.0' ">
+    <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="5.1.0" />
+
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
+    <PackageReference Include="System.Diagnostics.FileVersionInfo" Version="4.3.0" />
+    <PackageReference Include="Microsoft.Win32.Primitives" Version="4.3.0" />
```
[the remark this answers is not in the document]
Haplois: The last time I used this, it was called something else (win8 era :)). I don't know where these come from. I copied those from another project. I will try to remove and compile.

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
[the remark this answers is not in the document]
Haplois: TP only supports net451 and newer. It looks like netstandard1.0 maps into net45 (https://dotnet.microsoft.com/platform/dotnet-standard#versions). I added net20 and net35 support. The code is uglier now :)

scripts/build.ps1
```diff
@@ -805,16 +815,19 @@ function Create-NugetPackages
...
-                     "TestPlatform.Build.nuspec",
-                     "TestPlatform.Extensions.TrxLogger.nuspec", 
-                     "Microsoft.NET.Test.Sdk.nuspec",
-                     "Microsoft.TestPlatform.nuspec",
-                     "Microsoft.TestPlatform.Portable.nuspec",
-                     "Microsoft.CodeCoverage.nuspec")
+    $nuspecFiles = @(
+        "TestPlatform.TranslationLayer.nuspec",
+        "TestPlatform.ObjectModel.nuspec",
+        "Microsoft.TestPlatform.AdapterUtilities.nuspec",
```
AbhitejJohn, line 821: nit nit nit: could we order them on name?

src/Microsoft.TestPlatform.AdapterUtilities/Extensions/ReflectionExtensions.MethodBase.cs
```diff
@@ -1,12 +1,12 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Extensions
+namespace Microsoft.TestPlatform.AdapterUtilities.Extensions
 {
     using System;
     using System.Reflection;
 
-    public static partial class ReflectionExtensions
+    internal static partial class ReflectionExtensions
```
AbhitejJohn, line 9: Have we checked the VS repo to ensure this does not have any references that could break?

src/Microsoft.TestPlatform.AdapterUtilities/Extensions/ReflectionExtensions.Type.cs
```diff
@@ -1,12 +1,12 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Extensions
+namespace Microsoft.TestPlatform.AdapterUtilities.Extensions
 {
     using System;
     using System.Reflection;
 
-    public static partial class ReflectionExtensions
+    internal static partial class ReflectionExtensions
```
AbhitejJohn, line 9: For members here as well...

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
```diff
@@ -0,0 +1,53 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'uap10.0'">
+    <CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>
+    <NugetTargetMoniker>UAP,Version=v10.0</NugetTargetMoniker>
+    <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
+    <TargetPlatformVersion>10.0.14393.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.10240.0</TargetPlatformMinVersion>
+    <TargetFrameworkIdentifier>.NETPortable</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
AbhitejJohn, line 23: Is this required? (<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>)

src/Microsoft.TestPlatform.CrossPlatEngine/Microsoft.TestPlatform.CrossPlatEngine.csproj
```diff
@@ -21,7 +21,7 @@
     <ProjectReference Include="..\Microsoft.TestPlatform.CommunicationUtilities\Microsoft.TestPlatform.CommunicationUtilities.csproj" />
     <ProjectReference Include="..\Microsoft.TestPlatform.ObjectModel\Microsoft.TestPlatform.ObjectModel.csproj" />
     <ProjectReference Include="..\Microsoft.TestPlatform.Common\Microsoft.TestPlatform.Common.csproj" />
-	<ProjectReference Include="..\Microsoft.TestPlatform.Utilities\Microsoft.TestPlatform.Utilities.csproj" />
+	  <ProjectReference Include="..\Microsoft.TestPlatform.Utilities\Microsoft.TestPlatform.Utilities.csproj" />
```
AbhitejJohn, line 24: nit: formatting.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -66,64 +63,6 @@ public TestCase(string fullyQualifiedName, Uri executorUri, string source)
...
-        /// <param name="executorUri">
-        /// The Uri of the executor to use for running this test.
-        /// </param>
-        /// <param name="source">
-        /// Test container source from which the test is discovered.
-        /// </param>
-        /// <remarks>
-        /// If <paramref name="method"/> is specified, TestId will be calculated based on that instead of <paramref name="fullyQualifiedName"/>.
-        /// </remarks>
-        public TestCase(string fullyQualifiedName, MethodBase method, Uri executorUri, string source)
```
AbhitejJohn, line 88: Isn't this breaking?

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -165,34 +104,6 @@ public Guid Id
...
         }
 
-        /// <summary>
-        /// Gets or sets the fully specified type name metadata format.
-        /// </summary>
-        /// <example>
-        ///     <code>NamespaceA.NamespaceB.ClassName`1+InnerClass`2</code>
-        /// </example>
-        [DataMember]
-        public string ManagedType {
```
AbhitejJohn, line 175: Does the serializer in TP work well with deleted members?

AbhitejJohn (commented):
Concerned with the changes in OM. What does our test matrix look like?

src/package/nuspec/Microsoft.TestPlatform.AdapterUtilities.nuspec
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0"?>
+<package>
+    <metadata>
+        <id>Microsoft.TestPlatform.AdapterUtilities</id>
+        <version>15.0.0</version>
```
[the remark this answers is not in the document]
Haplois: The version is overridden during the build process. It's coming from the CI. Something like `16.9.0-preview-20210106-01` for the previews and `16.8.3` for the releases.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
Haplois: This was introduced as part of FQN changes. The current release doesn't have these.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
Haplois: These members don't exist on older releases, so removed to keep compatibility.

src/Microsoft.TestPlatform.AdapterUtilities/Extensions/ReflectionExtensions.MethodBase.cs
Haplois: The class itself was introduced as part of FQN, never used anywhere since it didn't ship.

src/Microsoft.TestPlatform.AdapterUtilities/Extensions/ReflectionExtensions.Type.cs
Haplois: done.

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
Haplois: I actually don't know.  Without those properties, the UAP build fails. I am not an expert on that platform so just copied over from ObjectModel.

src/Microsoft.TestPlatform.CrossPlatEngine/Microsoft.TestPlatform.CrossPlatEngine.csproj
Haplois: fixed

scripts/build.ps1
Haplois: done.

Haplois:
We're removing new FQN features from ObjectModel since they were breaking compatibility. So changes on ObjectModel should affect any older releases. They didn't ship on a release yet.  /cc: @AbhitejJohn

AbhitejJohn:
@Haplois : Sounds great, thanks.

AbhitejJohn (approved):

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
AbhitejJohn: Weird, cause we are setting TFM to a different version below. So does this project target 4.5 or 4.5.1?

PR #2715, opened 2021-01-24
Update `azure-pipelines.yml` to publish test results

Haplois:
Updated `azure-pipelines.yml` to publish test results.

- [x] `test.sh` needs to be updated to generate `.trx` logs.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,6 +33,13 @@ jobs:
       modifyEnvironment: false
       failOnStandardError: true
 
+  - task: PublishTestResults@2
+    displayName: 'Publish Test Results **\*.trx'
+    inputs:
+      testResultsFormat: VSTest
+      testResultsFiles: '**\*.trx'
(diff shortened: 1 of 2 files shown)


jakubch1 (approved):

azure-pipelines.yml
```diff
@@ -124,16 +131,25 @@ jobs:
...
+
   - script: ./tools/dotnet-linux/dotnet build -c $(buildConfiguration) ./test/TestAssets/TestAssets.sln
     displayName: 'Build test assets'
+
   - script: ./test.sh -c $(buildConfiguration) -p Acceptance
     displayName: 'Acceptance tests'   
-
-   
+  
+  - task: PublishTestResults@2
```
Sanan07, line 150: Why there are two same tasks here? The same as above PublishTestResults@2

Haplois: They are two different jobs. One is windows acceptance build, and the other one is for macOS and Linux.

PR #2716, opened 2021-01-25
Added git branch and commit NuGet packages

Haplois:
## Description
Added git branch and commit into NuGet packages.


scripts/build.ps1
```diff
@@ -56,6 +56,29 @@ Param(
...
+$TPB_BRANCH = "LOCALBRANCH"
+$TPB_COMMIT = "LOCALBUILD"
+
+try {
+    $TPB_BRANCH = $env:BUILD_SOURCEBRANCH -replace "^refs/heads/"  
+    if ([string]::IsNullOrWhiteSpace($TPB_BRANCH)) { 
+        $TPB_BRANCH = git -C "." rev-parse --abbrev-ref HEAD
+    }
+}
+catch { }
```
cvpoienaru, line 71: Should we log some error message here ?

scripts/build.ps1
```diff
@@ -56,6 +56,29 @@ Param(
...
+
+try {
+    $TPB_COMMIT = $env:BUILD_SOURCEVERSION
+    if ([string]::IsNullOrWhiteSpace($TPB_COMMIT)) { 
+        $TPB_COMMIT = git -C "." rev-parse HEAD
+    }
+}
+catch { }
```
cvpoienaru, line 79: Same as above.

Sanan07:
What is the reason to add this info to Nuget packages?

scripts/build.ps1
Haplois: It's not important; an error here should only happen locally if git is not installed already, we can safely ignore the error.

Haplois:
> What is the reason to add this info to Nuget packages?

To make debugging easier, especially in symbol issues.

Sanan07 (approved):

cvpoienaru (approved):
LGTM

PR #2721, opened 2021-01-29
Remove duplicate dll's from signing

Sanan07:
Remove duplicate dll's from signing as currently we get [warning](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4421663&view=results).
Update Readme file with VS 2019

### DIFF
--- a/src/package/sign/sign.proj
+++ b/src/package/sign/sign.proj
@@ -222,8 +222,8 @@
       <AssembliesToSign Include="$(ArtifactsDirectory)SettingsMigrator.exe" />
 
       <!-- NetFullExtensions -->
-      <AssembliesToSign Include="$(ArtifactsDirectory)Extensions\Microsoft.VisualStudio.TestPlatform.Extensions.Trx.TestLogger.dll" />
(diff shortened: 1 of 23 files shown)


src/package/sign/sign.proj
```diff
@@ -233,7 +233,7 @@
       <AssembliesToSign Include="$(ArtifactsDirectory)Extensions\%(ResxLang.Identity)\*.*" />
 
       <!-- Third Party assemblies to sign -->
-      <ThirdPartyAssembliesToSign Include="$(ArtifactsDirectory)Newtonsoft.Json.dll" />
+      <!--<ThirdPartyAssembliesToSign Include="$(ArtifactsDirectory)Newtonsoft.Json.dll" />-->
```
Haplois, line 236: Even though this assembly is already signed with the Newtonsoft certificate, we need to sign it again with the Microsoft 3rdParty certificate. Make sure we're doing that.

Haplois:
Running `TestPlatform.CI.Real` to make sure we're signing everything. Please wait for that to complete before merging.

src/package/sign/sign.proj
Sanan07: We do sign it. This is duplicate of the same dll signing. You can find the same in 178 line.

Haplois (approved):

PR #2724, opened 2021-02-01
Added constants for hierarchical naming

Haplois:
- Added constants for hierarchical naming.
 - Reduced dependency surface of the AdapterUtilities package.


Haplois:
We're adding a TestProperty to store multiple levels of a name. I am calling it `HierarchicalName`. Any recommendations about the naming?

`HierarchicalName` property is an array, which holds these for a `TestCase`: 
  - Namespace 
  - Class
  - TestGroup
  - DisplayName
  - Theory

peterwald:
> `HierarchicalName` property is an array, which holds these for a `TestCase`:
> 
> * Namespace
> * Class
> * TestGroup
> * DisplayName
> * Theory / DataRow

I'm ok with the name `HierarchicalName` or `HierarchyName`. We shouldn't need an entry for Theory/DataRow. That is captured by TestGroup & DisplayName.

In the case of a "Fact" test, you would have a Namespace/Class/<blank>/DisplayName and in the case of a "Theory" test you would have a Namespace/Class/TestGroup/DisplayName where TestGroup captures the "test method" level and DisplayName would be different for each data driven test case. The way we handle this today in the test hierarchy is if there is only one DisplayName in a TestGroup, then the TestGroup level is removed.

peterwald:
What about using just the name `Hierarchy` for the property.

src/Microsoft.TestPlatform.AdapterUtilities/HierarchicalNamingConstants.cs
```diff
@@ -0,0 +1,59 @@
...
+            /// Index of the test group element of the array.
+            /// </summary>
+            public const int TestGroupIndex = 2;
+
+            /// <summary>
+            /// Index of the display name element of the array.
+            /// </summary>
+            public const int DisplayNameIndex = 3;
+
+            /// <summary>
```
peterwald, line 52: Per my comment on the PR, this level isn't needed.

peterwald (changes requested):

Haplois:
> What about using just the name `Hierarchy` for the property.

Sounds better.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/Contants.cs
```diff
@@ -3,12 +3,29 @@
 
 namespace Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities
 {
+    /// <summary>
+    /// Contants to help declare ManagedType and ManagedMethod test properties.
+    /// </summary>
     public static class Contants
```
peterwald, line 9: Spelling...should be `Constants`

src/Microsoft.TestPlatform.AdapterUtilities/HierarchyConstants.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.TestPlatform.AdapterUtilities
+{
+    /// <summary>
+    /// Contants to help declare Hierarchy test property.
+    /// </summary>
+    public static class HierarchyConstants
```
peterwald, line 9: I noticed this type name is slightly inconsistent with the naming of `ManagedNameUtilities.Constants`, should we move the managed names up a level in the hierarchy and make the names consistent (esp. since this is a public API).

peterwald (approved):
Looks good with a few minor changes.

src/Microsoft.TestPlatform.AdapterUtilities/HierarchyConstants.cs
Haplois: I was thinking about the same thing. At first `Constants` name is too general; I like to change it. My hesitation is this package is technically released as a preview. What do you think?

peterwald: We can change it while it's still a preview. I think we should go ahead and change it to ManagedNameConstants and move it up a level.

src/Microsoft.TestPlatform.AdapterUtilities/HierarchyConstants.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Property id to use on Hierarchy test property.
+        /// </summary>
+        public const string HierarchyPropertyId = "TestCase." + HierarchyLabel;
+
+        /// <summary>
+        /// Meanings of the indices in the Hierarchy array.
+        /// </summary>
+        public static class Levels
```
AbhitejJohn, line 24: Could the HierarchyName property be a dictionary instead, so both setting/retrieving is easier to understand and uniqueness of values is guaranteed? We could serialize/deserialise this the way we want internally,

Haplois: Do you mean like a `<string, string>` directory with set keys on the adapter?

AbhitejJohn: Gosh, I dont have a good way of responding to notifications from github. Yes, I meant a <string,string> property bag.

PR #2743, opened 2021-02-09
NuGet Feed Update

Haplois:
# NuGet Feed Update
This pull request updates the usage of NuGet.org in cases where it is used in conjunction with other feeds.


Haplois:
This is fixed on #2741 and #2742,

PR #2746, opened 2021-02-12
Don't swallow stacktrace from adapter exception when running in thread

### LINKED ISSUE #2729: Improve logging when adapter fails
## Description
When an adapter has internal failure we don't get the stacktrace in log which makes finding the root cause difficult. 

AB#1274107

nohwnd:
## Description
Ensure that we re-throw the exception coming from thread without losing the stacktrace. This path is used in STA thread on net451, netcore runner mostly avoids it.

## Related issue
Fix #2729


Haplois (approved):
LGTM

PR #2748, opened 2021-02-15
Modified integration tests to use temp folder

Haplois:
## Description
Modified the integration tests to use a temporary folder to store test results and `.trx` logs, so when we report test results to the CI pipeline, they wouldn't falsely be reported.

### DIFF
--- a/test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
+++ b/test/Microsoft.TestPlatform.AcceptanceTests/AcceptanceTestBase.cs
@@ -4,8 +4,8 @@
 namespace Microsoft.TestPlatform.AcceptanceTests
 {
     using System;
-    using System.Collections.Generic;
-    using System.Linq;
+    using System.IO;
+
     using Microsoft.TestPlatform.TestUtilities;
 
     public class AcceptanceTestBase : IntegrationTestBase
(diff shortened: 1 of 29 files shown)


Sanan07 (approved):

nohwnd:
Maybe you should use agent temp if present, rather than the real temp folder. Agent temp is automatically cleaned up by the agent, normal temp is not. https://github.com/microsoft/vstest/blob/f2256b89a7177481e9c54dc538dd253653dde5a1/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs#L590-L602

Haplois:
>      // DUMP_TEMP_PATH will be used as temporary storage location 
>      // for the dumps, this won't affect the dump uploads. Just the place where 
>      // we store them before moving them to the final folder. 
>   
>      // AGENT_TEMPDIRECTORY is AzureDevops variable, which is set to path 
>      // that is cleaned up after every job. This is preferable to use over 
>      // just the normal temp. 
>      var temp = Environment.GetEnvironmentVariable("VSTEST_DUMP_TEMP_PATH") ?? Environment.GetEnvironmentVariable("AGENT_TEMPDIRECTORY") ?? Path.GetTempPath(); 
>      this.tempDirectory = Path.Combine(temp, Guid.NewGuid().ToString());

I have shamelessly stolen your code.

PR #2756, opened 2021-02-16
Use reported adapters for telemetry

nohwnd:
## Description
Count the actual reported adapters in the telemetry instead of the adapter that is the default for the assembly. This is useful for adapters that are delegating the work to other adapters, such as when MSTest v1 actually uses MSTest v0 to execute the tests when the conditions are right. In this case we will now see v0 in the telemetry, instead of v1.

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Execution/TestRunCache.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Execution/TestRunCache.cs
@@ -35,6 +35,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution
         /// </summary>
         private long totalExecutedTests;
 
+        private Dictionary<string, TestResultMetric> totalExecutedTestsPerAdapter = new Dictionary<string, TestResultMetric>();
+
         /// <summary>
         /// Callback used when cache is ready to report some test results/case.
         /// </summary>
@@ -163,6 +165,17 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution
             }
         }
 
+        public Dictionary<string, TestResultMetric> TestsPerAdapter
+        {
+            get
+            {
+                lock (this.syncObject)
+                {
+                    return this.totalExecutedTestsPerAdapter;
+                }
+            }
+        }
+
         /// <summary>
         /// Gets the test run stats
         /// </summary>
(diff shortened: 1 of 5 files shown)


test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestableImplementations/TestableTestRunCache.cs
```diff
@@ -9,7 +9,7 @@ namespace TestPlatform.CrossPlatEngine.UnitTests.TestableImplementations
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
 
-    public class TestableTestRunCache : ITestRunCache
+    internal class TestableTestRunCache : ITestRunCache
```
Haplois, line 12: I assume this change won't hurt us? Since we're closing down a publicly available class?

Haplois (approved):
LGTM

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestableImplementations/TestableTestRunCache.cs
nohwnd: It should be fine. It's a class in test project, and I had to reduce it's visibility because the TestResultMetric class would otherwise be less visible.

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/TestRunCache.cs
```diff
@@ -35,6 +35,8 @@ internal class TestRunCache : ITestRunCache
         /// </summary>
         private long totalExecutedTests;
 
+        private Dictionary<string, TestResultMetric> totalExecutedTestsPerAdapter = new Dictionary<string, TestResultMetric>();
```
AbhitejJohn, line 38: Could we make this a Concurrent Dictionary instead?

AbhitejJohn (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/TestRunCache.cs
nohwnd: I just copied the approach other the total tests property was using. I will consider it.

nohwnd:
Converting to draft, not sure if this is a good way of doing this, and it would be nice to have a more general way of passing metrics data from the adapter.

nohwnd:
Obsolete. Will do it differently.

PR #2762, opened 2021-02-19
Dynamic corelib.net loading

fhnaseer:


src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
```diff
@@ -106,15 +112,25 @@ private async Task<string> MergeCodeCoverageFilesAsync(IList<string> files, IPro
...
-
-            var coverageData = await coverageUtility.MergeCoverageFilesAsync(
-                    files,
-                    cancellationToken).ConfigureAwait(false);
-
-            coverageUtility.WriteCoverageFile(files[0], coverageData);
-
-            foreach(var file in files.Skip(1))
+            cancellationToken.ThrowIfCancellationRequested();
+
```
jakubch1, line 116: Maybe add comment why we are invoking it dynamically. 
Also add try catch block inside this method or when you call MergeCodeCoverageFilesAsync - return original attachments in case of error

jakubch1 (approved):

PR #2765, opened 2021-02-22
Mark early testhost startup APIs as internal

cvpoienaru:
Mark early testhost startup APIs as internal


Haplois (approved):
Should we change those internal APIs to `internal`?

cvpoienaru:
The APIs are already released in version 16.9.0-preview-20210106-01 because I split the early testhost startup work into two separate PRs for better management. Due to the project dependency graph, actually marking the APIs as internal will mean negating most of the work done in [this PR](https://github.com/microsoft/vstest/pull/2584).

However, a disclaimer should be enough to prevent external users from using those APIs.

Haplois:
This work is continuing on #2768.

PR #2768, opened 2021-02-23
Mark early testhost startup APIs as internal for TP 16.9

cvpoienaru:
Mark early testhost startup APIs as internal for TP 16.9


Haplois (approved):
LGTM

PR #2799, opened 2021-03-04
Update outdated doc link

vritant24:
Addresses [this devcom feedback](https://developercommunity.visualstudio.com/t/Runsettings-documentation-leads-you-down/1353353)

The previous link pointed to an outdated doc link where the "Recommended Version" led to a dead end. 
I have updated the link to point to the same doc in a new location


Haplois:
It might be easier to update where https://go.microsoft.com/fwlink/?LinkID=236877&clcid=0x409 is pointing. Let me first try that.

Haplois (changes requested):
Trying to update the aka.ms link.

Haplois (approved):
Instead of updating the old fwlink, just created a new one and used that. https://aka.ms/tp/vstest/multitargetingdoc?view=vs-2019

PR #2829, opened 2021-03-25
Add support for loading Testplatform.fakes in Dev 17

vritant24:
We ship v 17.0 of the Testplatform.Fakes assembly in Dev 17. This change accounts for it.


vritant24:
@cvpoienaru @AbhitejJohn Ready for review and merge. Testing it locally with the changes, the Fakes tests were able to be run with the TestPlatform.Fakes 17.0 and 16.0 versions.

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -24,7 +24,8 @@ public static class FakesUtilities
 
         private const string CrossPlatformConfiguratorMethodName = "GetCrossPlatformDataCollectorSettings";
 
-        private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=16.0.0.0, Culture=neutral";
+        private const string FakesConfiguratorAssemblyDev16 = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=16.0.0.0, Culture=neutral";
```
AbhitejJohn, line 27: Do we still need a 16.0 version?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -278,6 +279,23 @@ private static void EnsureSettingsNode(XmlDocument settings, TestRunSettings set
...
+        /// <summary>
+        /// Testplatform.Fakes has version 16.0 in Dev 16
+        /// and version 17.0 in Dev 17.
+        /// </summary>
+        /// <returns></returns>
+        private static Assembly LoadTestPlatformAssembly()
+        {
+            try
+            {
+                return Assembly.Load(new AssemblyName(FakesConfiguratorAssemblyDev16));
```
AbhitejJohn, line 291: If we know where the Fakes assembly is on disk always, we could use Assembly.LoadFrom and skip the version altogether...

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
vritant24: I left both in there because currently we will not be shipping the 17.0 TestPlatform.Fakes in the Microsoft.Testplatform nuget package. We will be able to remove that once we update the TestRunnerHarness project. Do you think there may be any problems in doing so?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
vritant24: I didn't know that the API existed. Giving it a try, I found out that the API doesn't exist for .NetStandard 1.3 that the project builds for, so we will have to use this API

cvpoienaru (approved):
LGTM

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
AbhitejJohn: Oh wouldn't we be able to ship that in the 17.0 package of the TestPlatform? Its a little confusing that we have both versions supported here though in reality we'd only expect one of the versions to be present based on the version the TP package.

vritant24: Ah I did see that we have a 17.0 package already shipped. Updated and removed 16.0

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -277,6 +278,18 @@ private static void EnsureSettingsNode(XmlDocument settings, TestRunSettings set
 
             return null;
         }
+        private static Assembly LoadTestPlatformAssembly()
```
AbhitejJohn, line 281: github does not allow me comment on the right line here but does the version at ln 333 need an update too?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -24,7 +24,8 @@ public static class FakesUtilities
 
         private const string CrossPlatformConfiguratorMethodName = "GetCrossPlatformDataCollectorSettings";
 
-        private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=16.0.0.0, Culture=neutral";
+        // Must be kept in sync with version being generated by VSUnitTesting
+        private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=17.0.0.0, Culture=neutral";
```
AbhitejJohn, line 28: consider: There seems to be a lot of places in the repo where the string `16.0.0.0` is used. I think it would be great to pull this from a central .cs file that can be updated for every release. Since it's just a .cs file, the consumers wouldn't have an additional dll reference.
/cc: @cvpoienaru

AbhitejJohn (approved):
Looks good, except for the few comments added in.

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
vritant24: Thanks for catching that, updated

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
vritant24: Would this be something I add here? It sounds like it would have to be a repo wide change that I wouldn't want to make. We could adopt to a standard if it's set in vstest. For now I've shared the versions in this file.
@cvpoienaru do you have recommendations on whether I could use a central file to pull the version string from?

AbhitejJohn: yes, I was recommending a repo wide file. We could put it in the eng folder where version updates are usually made.

vritant24: The versions are located in 2 places. in the eng folder exists version for packages, and in scripts\build folder exists versions for dependencies. The Fakes version currently exists in scripts\build, and so I added a central .cs file there

AbhitejJohn: sounds good

cvpoienaru: Thanks for taking the time to make the changes Abhitej proposed, @vritant24. Indeed I think the `scripts\build` folder works better overall for placing the new `.cs` file.

On a side note I see the PR has a merge conflict so the build hasn't run yet. Does this change build locally ? I'm asking because it's not completely clear to me how the `.cs` file ends up being pulled in from `scripts\build` & compiled with the project in case, i.e `Microsoft.TestPlatform.Common.csproj`. Should we add a compile directive for that file in the `.csproj` ?

vritant24: Looks like I forgot to commit the csproj file. added.
Thanks for pointing that out!

cvpoienaru (approved):
LGTM, approved !

scripts/build/ExternalAssemblyVersions.cs
```diff
@@ -0,0 +1,12 @@
+namespace Microsoft.VisualStudio.TestPlatform
+{
+    internal class ExternalAssemblyVersions
+    {
+        /// <summary>
+        /// Refers to the versions of the assemblies retrieved from the 
+        /// Microsoft.QualityTools.Testing.Fakes.TestRunnerHarness package.
+        /// The Package version can be found in "scripts\build\TestPlatform.Dependencies.props"
+        /// </summary>
+        internal const string MicrosoftFakesAssemblyVersion = "17.0.0.0";
```
Sanan07, line 10: Why not to add this version value inside `TestPlatform.Dependencies.props` file?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -24,7 +24,9 @@ public static class FakesUtilities
 
         private const string CrossPlatformConfiguratorMethodName = "GetCrossPlatformDataCollectorSettings";
 
-        private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=16.0.0.0, Culture=neutral";
+        // Must be kept in sync with version being generated by VSUnitTesting
+        private const string version = ExternalAssemblyVersions.MicrosoftFakesAssemblyVersion;
+        private const string FakesConfiguratorAssembly = "Microsoft.VisualStudio.TestPlatform.Fakes, Version=" + version + ", Culture=neutral";
```
Sanan07, line 29: I assume it will not compile as `FakesConfiguratorAssembly` is const and `version`'s  value is not known as compile time.

Sanan07 (approved):

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
Sanan07: Actually, it built fine :)

vritant24: It's known during compile time as the assembly version is a hard coded const string, and not really acquired through reflection. We need to do this as the `TestPlatform.Fakes` assembly is dropped by the VSUnitTesting build into VS and not from this repo

scripts/build/ExternalAssemblyVersions.cs
vritant24: This is so this can be referenced inside .cs files. Values from within `TestPlatform.Dependencies.props` cannot be referenved by .cs files.

For more context: There seems to be hardcoded versions of assemblies in many .cs files (refer to abhitej's comment above) and in instances like this where a bunch of them are moving from 16.0.0.0 to 17.0.0.0, it would be better to have a central file so the gaps can be caught earlier on.

PR #2830, opened 2021-03-25
Removed & updated old interops

cvpoienaru:
Removed & updated old interops


NuGet.config
```diff
@@ -15,6 +15,7 @@
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
     <add key="vssdk@Local" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vssdk%40Local/nuget/v3/index.json" />
```
Sanan07, line 17: There is already `vssdk@Local` feed above where you can pull Interop package, no need to the new vssdk feed.

scripts/build/TestPlatform.Dependencies.props
```diff
@@ -34,6 +34,7 @@
     <MoqVersion>4.7.63</MoqVersion>
     <TestPlatformExternalsVersion>16.10.0-preview-2-31108-246</TestPlatformExternalsVersion>
     <CodeCoverageExternalsVersion>16.10.0-beta.21175.5</CodeCoverageExternalsVersion>
+    <InteropExternalsVersion>17.0.0-preview-1-30928-1111</InteropExternalsVersion>
```
Sanan07, line 37: New version `17.0.0-preview-1-30928-1112` just uploaded recently, it is better to use that, it contains some new fixes.

scripts/build.ps1
```diff
@@ -800,6 +801,12 @@ function Create-VsixPackage
     # Copy CUIT Related depedencies
     $legacyDir = Join-Path $env:TP_PACKAGES_DIR "Microsoft.VisualStudio.CUIT\$testPlatformExternalsVersion\tools\net451"
     Copy-Item -Recurse $legacyDir\* $packageDir -Force
+    $fileToRemove = Join-Path $packageDir "Microsoft.VisualStudio.Ole.Interop.dll"
+    Remove-Item $fileToRemove -Force
```
Sanan07, line 805: This should be temporarily code for removal. When we release new CUIT nuget package, which does not contains this old interop, so it will not appear in artifacts, so no need to remove it.

Sanan07 (changes requested):

scripts/build.ps1
cvpoienaru: Agreed, it's temporary until we fix CUIT. But we can fix it later though as the priority is to have the updated interops for the insertion, so we should keep this here for now.

NuGet.config
cvpoienaru: Done, thanks !

scripts/build/TestPlatform.Dependencies.props
cvpoienaru: Version `17.0.0-preview-1-30928-1112` fails on package restore due to a dependency on `Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime` so we have to stick to this one for now.

Sanan07 (approved):

PR #2837, opened 2021-03-30
Merging servicing fixes for 16.9.4

Haplois:
merging servicing fixes.

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
+++ b/src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
@@ -8,19 +8,25 @@ namespace Microsoft.VisualStudio.TestPlatform.Utilities
     using System.Collections.ObjectModel;
     using System.IO;
     using System.Linq;
+    using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
 
-    using Microsoft.VisualStudio.Coverage.CoreLib.Net;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+    using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
 
     public class CodeCoverageDataAttachmentsHandler : IDataCollectorAttachmentProcessor
     {
         private const string CoverageUri = "datacollector://microsoft/CodeCoverage/2.0";
         private const string CoverageFileExtension = ".coverage";
         private const string CoverageFriendlyName = "Code Coverage";
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
```diff
@@ -106,15 +112,25 @@ private async Task<string> MergeCodeCoverageFilesAsync(IList<string> files, IPro
...
-                    cancellationToken).ConfigureAwait(false);
-
-            coverageUtility.WriteCoverageFile(files[0], coverageData);
-
-            foreach(var file in files.Skip(1))
+            cancellationToken.ThrowIfCancellationRequested();
+
+            var assemblyPath = Path.Combine(Path.GetDirectoryName(typeof(CodeCoverageDataAttachmentsHandler).GetTypeInfo().Assembly.GetAssemblyLocation()), CodeCoverageCoreLibNetAssemblyName + ".dll");
+            
+            // Get assembly, type and methods
```
cvpoienaru, line 119: Should we guard against exceptions with a try catch maybe ?

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
```diff
@@ -106,15 +112,25 @@ private async Task<string> MergeCodeCoverageFilesAsync(IList<string> files, IPro
...
+            Assembly assembly = new PlatformAssemblyLoadContext().LoadAssemblyFromPath(assemblyPath);
+            var classType = assembly.GetType($"{CodeCoverageCoreLibNetAssemblyName}.{CoverageFileUtilityTypeName}");
+            var classInstance = Activator.CreateInstance(classType);
+            var mergeMethodInfo = classType?.GetMethod(MergeMethodName, new[] { typeof(IList<string>), typeof(CancellationToken) });
+            var writeMethodInfo = classType?.GetMethod(WriteMethodName);
+            
+            // Invoke methods
+            var task = (Task)mergeMethodInfo.Invoke(classInstance, new object[] { files, cancellationToken });
```
Sanan07, line 127:
```suggestion
            var task = mergeMethodInfo.Invoke(classInstance, new object[] { files, cancellationToken }) as Task;
```

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
Haplois: It's already cached where this method is called.

src/Microsoft.TestPlatform.Utilities/CodeCoverageDataAttachmentsHandler.cs
Haplois: This might change behavior, code is cherrypicked from master. We should do this on master IMO.

cvpoienaru (approved):

Issue #2686, opened 2021-01-08
Vstest hangs after "Cancelling the operation as requested." message


richardwerkman:
## Description
*Note* This** is a follow up issue for #2578. Lowering the timeout does _not_ fix our issue. Stryker keeps crashing thanks to the error coming out of vstest.

Original issue:
When working on Stryker mutator I noticed vstest is reporting some sort of _error_ when running a session with two assemblies. At some point in stryker we know all mutations have been tested and we can abort/cancel the testrun. This goes right most of the times. Some times however a error with the message `Cancelling the operation as requested.` comes from vstest. After that the session seems to have crashed. No clear exception is thrown and it's not clear how to handle the error. 

This might be a bug in vstest.

## Steps to reproduce
The error occurs when running stryker against two test projects.
1. 
a. Clone [Stryker-net](https://github.com/stryker-mutator/stryker-net/tree/1136_multi_test_project_bug)
b. Rename dir to stryker-net-main
c. Checkout branch 1136_multi_test_project_bug
2.
a. Clone [Stryker-net ](https://github.com/stryker-mutator/stryker-net/tree/multi_test_project_example)
b. Checkout branch multi_test_project_example
3. Open `stryker-net-main` (Solution `Stryker.CLI.sln`)
3. Run project Stryker.CLI in that solution with the following settings (Project Stryker.CLI -> Properties -> Debug)
Working directory: `{path where you cloned}\stryker-net\src\Stryker.Core\Stryker.Core`
Application arguments: `-f`
4. Set a breakpoint on [this line](https://github.com/stryker-mutator/stryker-net/blob/8bc53cd3858d1d76aa4a65d668ed32589b968728/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/RunEventHandler.cs#L154) in the clone from step 1
5. Run clone 1 with the settings as in step 3. The breakpoint should go off in a few minutes.

windows example for step 1&2:
```
git clone https://github.com/stryker-mutator/stryker-net.git
ren stryker-net stryker-net-main
cd ./stryker-net-main
git checkout 1136_multi_test_project_bug
cd ..
git clone https://github.com/stryker-mutator/stryker-net.git
cd ./stryker-net
git checkout multi_test_project_example
```

## Expected behavior
I expect a clear error to occur or no error at all. 

## Actual behavior
[...]

PR #2857, opened 2021-04-16
Abort when init and run messages fail to process

nohwnd:
## Description

Failing to deserialize incoming request will be logged into log, but won't kill testhost. Vstest console won't know that testhost failed to process and will wait for it indefinitely, or till the whole run times out. The deserialization error happens from time to time (investigating why), and this will prevent the hang if it happens by aborting the run and passing back the error.


Haplois (approved):
LGTM

rouke-broersma:
@nohwnd could this possibly be related to #2686?

PR #2863, opened 2021-04-21
Signing `Microsoft.VisualStudio.Interop.dll`.

Haplois:


src/package/sign/sign.proj
```diff
@@ -232,6 +232,7 @@
       <AssembliesToSign Include="$(ArtifactsDirectory)testhost.net48.exe" />
       <AssembliesToSign Include="$(ArtifactsDirectory)testhost.net48.x86.exe" />
       <AssembliesToSign Include="$(ArtifactsDirectory)SettingsMigrator.exe" />
+      <AssembliesToSign Include="$(ArtifactsDirectory)Microsoft.VisualStudio.Interop.dll" />
```
Sanan07, line 235: It is coming from public feed, so it is already signed.

Haplois: No it's not. See: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4674195

cvpoienaru (approved):

PR #2881, opened 2021-04-29
CancelDiscovery bug fix

Sanan07:
## Description
Cancel all ongoing discoveries when Abort requested
## Related issue
Bug in AzDo - 1302270


src/Microsoft.TestPlatform.ObjectModel/Client/Events/DiscoveryCompleteEventArgs.cs
```diff
@@ -38,6 +38,15 @@ public DiscoveryCompleteEventArgs(long totalTests, bool isAborted)
         /// </summary>
         public bool IsAborted { get; private set; }
 
+        // List of sources which were fully discovered
+        public IList<string> FullyDiscoveredSources { get; set; } = new List<string>();
```
shyamnamboodiripad, line 42: It may be better to use something like IReadonlyCollection / ImmutableArray as the return type as we don't want the caller to mutate the returned collection.

shyamnamboodiripad, line 42: Also the property should probably be read-only (i.e. setter can probably be omitted).

shyamnamboodiripad: (Although it is surprising that the Metrics property below does have a public setter 😞)

src/Microsoft.TestPlatform.Common/Interfaces/Engine/TesthostProtocol/IDiscoveryManager.cs
```diff
@@ -28,6 +28,6 @@ public interface IDiscoveryManager
         /// <summary>
         /// Aborts the test discovery.
         /// </summary>
-        void Abort();
+        void Abort(ITestDiscoveryEventsHandler2 eventsHandler);
```
AbhitejJohn, line 31: Would this affect compat between vstest.console and testhost*.exe? I think irrespective it would be good to validate against older versions of Test SDK that would not understand the new messages. Also please validate that UWP works as well cause it uses its own host process.

src/Microsoft.TestPlatform.CommunicationUtilities/Messages/DiscoveryCompletePayload.cs
```diff
@@ -4,31 +4,61 @@
...
     using System.Collections.Generic;
+    using System.Runtime.Serialization;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
     /// <summary>
     /// The discovery complete payload.
     /// </summary>
     public class DiscoveryCompletePayload
     {
+        [DataMember]
```
AbhitejJohn, line 15: Why weren't these required earlier?

Sanan07: I am not sure, it works fine now, but we decided to add it as "good to have", because payload is used in serialization.

src/Microsoft.TestPlatform.Common/Interfaces/Engine/TesthostProtocol/IDiscoveryManager.cs
Sanan07: In TestRequestHandler we have switch case with default behaviour which will log "Invalid Message types".
When vstest.console is old, but testhost is new : means that vstest.console just will not send new message
When vstest.console is new, but testhost is old : will log "Invalid Message type."

But obviously we will ask vendor team to test all scenarious and also UWP, which you mention

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -248,5 +276,19 @@ private void DiscoverTestsOnConcurrentManager(IProxyDiscoveryManager proxyDiscov
                 EqtTrace.Verbose("ProxyParallelDiscoveryManager: No sources available for discovery.");
             }
         }
+
+        /// <summary>
+        /// Gets all sources which were not discovered(skipped), because of cancelation/abortion of discovery
+        /// </summary>
+        private void getAllNotDiscoveredSources()
```
cvpoienaru, line 283: Nit pick: this should be named according to Pascal Case, i.e. `GetAllNotDiscoveredSources`.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -248,5 +276,19 @@ private void DiscoverTestsOnConcurrentManager(IProxyDiscoveryManager proxyDiscov
...
+        {
+            lock (enumeratorLockObject)
+            {
+                while (this.sourceEnumerator != null && this.sourceEnumerator.MoveNext())
```
cvpoienaru, line 287: Why do we MoveNext before we get to process the current event ? Doesn't that mean we'll skip the current element from processing ?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryEventsHandler.cs
```diff
@@ -151,5 +163,21 @@ private void ConvertToRawMessageAndSend(string messageType, object payload)
             var rawMessage = this.dataSerializer.SerializePayload(messageType, payload);
             this.actualDiscoveryEventsHandler.HandleRawMessage(rawMessage);
         }
+
+        /// <summary>
+        /// Filters dictionary based on condition if source was fully discovered or not
+        /// </summary>
+        /// <param name="discoveryStatus">discoveryStatus indicates if source was fully or partially discovered</param>
+        /// <returns></returns>
+        private IReadOnlyCollection<string> getFilteredSources(DiscoveryStatus discoveryStatus)
```
cvpoienaru, line 172: Nit pick: this should be named according to Pascal Case, i.e. `GetFilteredSources`.

src/Microsoft.TestPlatform.Common/Interfaces/Engine/ClientProtocol/IParallelProxyDiscoveryManager.cs
```diff
@@ -25,4 +29,11 @@ public interface IParallelProxyDiscoveryManager : IParallelOperationManager, IPr
             IEnumerable<TestCase> lastChunk,
             bool isAborted);
     }
+
+    public enum DiscoveryStatus
```
cvpoienaru, line 33: Could we make this internal ? If not, please add documentation for the enum and for each of its values (I know the names are self descriptive but some additional context would be nice).

cvpoienaru (changes requested):

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
Sanan07: Based on this [docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.movenext?view=net-5.0#remarks) initial pointer points before first element, so we should call `MoveNext` to get first element.

src/Microsoft.TestPlatform.Common/Interfaces/Engine/ClientProtocol/IParallelProxyDiscoveryManager.cs
Sanan07: I use it in different project, so should be public. Added more comments.

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscovererEnumerator.cs
```diff
@@ -161,6 +165,8 @@ private void LoadTestsFromAnExtension(string extensionAssembly, IEnumerable<stri
                         return;
                     }
 
+                    // First mark sources as partially discovered before starting discovery
```
shyamnamboodiripad, line 168: Comment may need updates

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscovererEnumerator.cs
```diff
@@ -255,6 +268,9 @@ private void CollectTelemetryAtEnd(double totalTimeTakenByAdapters, double total
             }
             catch (Exception e)
             {
+                // If exception happens mark sources as not discovered
```
shyamnamboodiripad, line 271: Comment may need updates

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -215,7 +234,16 @@ private void DiscoverTestsOnConcurrentManager(IProxyDiscoveryManager proxyDiscov
                             EqtTrace.Verbose("ParallelProxyDiscoveryManager: Discovery started.");
                         }
 
+                        // First marking source as "partially discovered"
+                        DiscoveredSources[nextSource] = DiscoveryStatus.PartiallyDiscovered;
```
shyamnamboodiripad, line 238: AS we discussed, not high pri - but can we start off with NotDiscovered and change to PartiallyDiscovered only on exception?

Sanan07:
Not doing this work for now

PR #2887, opened 2021-05-04
Abort when init and run messages fail to process

Haplois:
* Abort when init and run messages fail to process, usually on deserialization, to prevent the host from hanging
* Send the message via the abort handler when we abort the run

Cherrypicked from #2857 with an extra sauce of localization.


cvpoienaru (approved):
LGTM

PR #2890, opened 2021-05-07
Upgrade Arcade version to 6.0.0-beta.21256.3.

crummel:


Haplois:
This is no longer needed after #2913. Thank you for the contribution @crummel.

PR #2924, opened 2021-05-27
Fix legacy tests

Sanan07:
Fixing failing legacy tests after updating to 17.0 external dependecies


Sanan07:
No need to this PR, as we already updated externals and finished CUIT interop work.

PR #2931, opened 2021-06-01
Updating externals to 17.0

Sanan07:
## Description
Updating externals to 17.0 for Code Coverage use and commenting legacy tests for now, cause interop changes for them is still not finished my engineering team. After they finish that work we can enable that tests. All of that tests are legacy, so do not affect current logic.

**DO NOT MERGE IT YET. WE CAN MERGE AFTER OUR PIPLINE FIXED AND AFTER FIX FOR INTEROPS MERGED INTO VS.**


test/Microsoft.TestPlatform.AcceptanceTests/CUITTest.cs
```diff
@@ -11,6 +11,7 @@ public class CUITTest : AcceptanceTestBase
     {
         [TestMethod]
         [TestCategory("Windows-Review")]
+        [Ignore]
```
nohwnd, line 14: Please consider adding info to the Ignore, I am not sure if the attribute can do it directly, or if you would need to add it as a TestCategory (if we can have multiple of those), so we know why this was ignored and when to re-visit it.

Haplois (approved):

PR #2932, opened 2021-06-02
Update externals to 16.11

nohwnd:
## Description
Update externals to 16.11


nohwnd:
It's a wrong base, it should be based on preview/16.11, but it does not exist.

Haplois (approved):

PR #2940, opened 2021-06-11
Adapts TrxLogger constructor to actually use passed variable

schibu007:
## Description
The description of the constructor says it's with DI for unit testing. But the passed variable isn't actually used.


Haplois (approved):

Haplois:
Thank you for the contribution @schibu007! Approved.

PR #2947, opened 2021-06-22
Add .NET 6.0 TestHost for HotReload support in TestWindow

vritant24:
## Description
1. Add .NET6.0 Test Host 
2. Add `TestHotReload` capability to indicate support

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -58,11 +58,9 @@
 
       <!-- 
         Test projects are not discovered in test window without TestContainer capability.
-        Test projects are marked with TestHotReload to indicate presence of .NET6.0 test host
-        which is needed for hot reloading. 
       -->
       <ItemGroup>
-        <ProjectCapability Include="TestContainer;TestHotReload" />
+        <ProjectCapability Include="TestContainer;" />
       </ItemGroup>
 
       <!-- Test project references -->
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -60,7 +60,7 @@
         Test projects are not discovered in test window without TestContainer capability.
       -->
       <ItemGroup>
-        <ProjectCapability Include="TestContainer;" />
+        <ProjectCapability Include="TestContainer" />
       </ItemGroup>
 
       <!-- Test project references -->
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -56,9 +56,13 @@
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
       </PropertyGroup>
 
-      <!-- Test projects are not discovered in test window without test container capability -->
+      <!-- 
+        Test projects are not discovered in test window without TestContainer capability.
+        Test projects are marked with TestHotReload to indicate presence of .NET6.0 test host
+        which is needed for hot reloading. 
+      -->
       <ItemGroup>
-        <ProjectCapability Include="TestContainer" />
+        <ProjectCapability Include="TestContainer;TestHotReload" />
       </ItemGroup>
 
       <!-- Test project references -->
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.props
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.props
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectCapability Include="TestContainer" />
+    <ProjectCapability Include="TestContainer;TestHotReload" />
   </ItemGroup>  
  
 </Project>
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.props
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.props
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectCapability Include="TestContainer;TestHotReload" />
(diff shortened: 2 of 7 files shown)


scripts/build/TestPlatform.Settings.targets
```diff
@@ -56,9 +56,11 @@
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
       </PropertyGroup>
 
-      <!-- Test projects are not discovered in test window without test container capability -->
+      <!-- 
+        Test projects are not discovered in test window without TestContainer capability.
+      -->
       <ItemGroup>
-        <ProjectCapability Include="TestContainer" />
+        <ProjectCapability Include="TestContainer;" />
```
AbhitejJohn, line 63: curious, is the `;` needed?

src/package/nuspec/TestPlatform.TestHost.nuspec
```diff
@@ -93,6 +98,24 @@
...
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\testhost.dll" target="lib\net6.0\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\testhost.deps.json" target="lib\net6.0\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\x86\msdia140.dll" target="lib\net6.0\x86\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\x64\msdia140.dll" target="lib\net6.0\x64\" />
+
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\Microsoft.TestPlatform.PlatformAbstractions.dll" target="build\net6.0\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\win7-x64\testhost.dll" target="build\net6.0\x64\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\win7-x64\testhost.exe" target="build\net6.0\x64\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\win7-x64\Microsoft.TestPlatform.PlatformAbstractions.dll" target="build\net6.0\x64\" />
+    <file src="Microsoft.TestPlatform.TestHost\net6.0\win7-x86\testhost.x86.dll" target="build\net6.0\x86\" />
```
AbhitejJohn, line 113: The `win7-*` runtime switch seems weird for .NET Core. I see that's what we have for netcoreapp2.1 too but I'm trying to understand how this works in a cross-plat runtime. Any thoughts?

src/package/nuspec/Microsoft.NET.Test.Sdk.props
```diff
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectCapability Include="TestContainer" />
+    <ProjectCapability Include="TestContainer;TestHotReload" />
```
AbhitejJohn, line 21: nit: `HotReloadForTest` maybe.

AbhitejJohn (approved):
Also tagging @nohwnd in case he has any concerns with this change.

Haplois (approved):
LGTM. Have you verified that this package works under linux?

src/package/nuspec/TestPlatform.TestHost.nuspec
nohwnd: It was added here: https://github.com/microsoft/vstest/pull/217/files and it looks like it's coming from a "template" to migrate to csproj. Imho this does not affect how our libs are built, it would just package them with appropriate runtime, which we then ignore because we copy only our dlls (which are still anycpu) to the result folder -> nuget package. So imho this is totally okay, even if not the cleanest way probably.

nohwnd (changes requested):
@vritant24 Can you confirm the hot reload does not work with the current testhost? I had similar situation where I thought I need to ship another testhost, when I was working on Blame, but it turned out that building it with net6.0 SDK while targetting netcoreapp2.1 will still keep it working, because later we provide deps.json a runtimeconfig.json that force it to run as the actual project TFM, e.g. net6.0. 

I am not sure how exactly it works internally, but imho newer SDK builds the exe / dll, and then just gives it runtimeconfig that forces it to run as netcoreapp2.1 with all those compat switches on. But the exe / lib is the same as if you built it for net5.0 or net6.0. That is why are are then able to push it back to net6.0 using the correct runtimeconfig.json. 

Please try it, and let me know how I can try for myself to validate. I want to avoid shipping multiple testhosts for .NET if we can avoid it.

vritant24:
> @vritant24 Can you confirm the hot reload does not work with the current testhost? I had similar situation where I thought I need to ship another testhost, when I was working on Blame, but it turned out that building it with net6.0 SDK while targetting netcoreapp2.1 will still keep it working, because later we provide deps.json a runtimeconfig.json that force it to run as the actual project TFM, e.g. net6.0.
> 
> I am not sure how exactly it works internally, but imho newer SDK builds the exe / dll, and then just gives it runtimeconfig that forces it to run as netcoreapp2.1 with all those compat switches on. But the exe / lib is the same as if you built it for net5.0 or net6.0. That is why are are then able to push it back to net6.0 using the correct runtimeconfig.json.
> 
> Please try it, and let me know how I can try for myself to validate. I want to avoid shipping multiple testhosts for .NET if we can avoid it.

I had previously tried it and it didn't work, but on trying it out on the latest VS 2022 main I have seen no issues and have not been able to reproduce the issues I encountered previously with a non .NET 6.0 Test Host. The issue was building the test project with a previous preview of the .NET 6 SDK that didn't have the required APIs, and VS is now updated to use a version higher than the required one. 
Will be abandoning this PR since no changes are needed 👍 @Haplois

nohwnd:
Cool, good to know my intuition of how this works is correct :)

PR #2951, opened 2021-06-24
Stop prefering 32-bit for vstest.console

nohwnd:
## Description
vstest.console should stop prefering 32-bit, and instead run as 64-bit.


Haplois (approved):

nohwnd:
Wait with merging this until we agree upon it in the internal ticket where I tagged you :)

nohwnd:
The CPP tests are fixed by running the correct version of the test in process. 

The CUIT tests are still ignored.

PR #2954, opened 2021-06-25
Add support for s390x processor architecture

uweigand:
## Description
We have recently added support for the s390x processor architecture to the dotnet runtime, which included adding a new value for the `System.Runtime.InteropServices.Architecture` enum: https://github.com/dotnet/runtime/issues/52909

However, code in `Microsoft.VisualStudio.TestPlatform.PlatformAbstractions` does not recognize this new value, which currently causes any test run on s390x to immediately abort on startup.  To fix this, this patch adds a new value to `Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.PlatformArchitecture` as well, and maps the s390x Architecture value to that new value.

## Related issue
n/a

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/PlatformArchitecture.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/Interfaces/System/PlatformArchitecture.cs
@@ -11,6 +11,7 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
         X86,
         X64,
         ARM,
-        ARM64
+        ARM64,
+        S390x
     }
 }
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/netcore/System/PlatformEnvironment.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/netcore/System/PlatformEnvironment.cs
@@ -28,6 +28,12 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
                         return PlatformArchitecture.ARM;
                     case System.Runtime.InteropServices.Architecture.Arm64:
                         return PlatformArchitecture.ARM64;
+
+                    // The symbolic value is only available with .NET 6
+                    // preview 6 or later, so use the numerical value for now.
+                    // case System.Runtime.InteropServices.Architecture.S390x:
+                    case (Architecture)5:
+                        return PlatformArchitecture.S390x;
                     default:
                         throw new NotSupportedException();
                 }


nohwnd:
@uweigand I wonder how we can validate this. Are there any online builds where you are running the tests that we can see, or add to our build?

uweigand:
Unfortunately, I wasn't able to run the vstest test suite itself so far.  The problem is that the tests seem to require the .NET Core 2.1.0 framework, but we only support .NET 6 on our platform.  If you have any suggestion how to run tests in that environment, that would be appreciated!

However, using a vstest package including this patch, I can now successfully run e.g. the roslyn test suite, where all tests would abort on startup without this patch.

nohwnd:
@Haplois being able to run Roslyn test suite seems good enough for start, while we investigate how we could run our own tests on that platform. Does that sound good to you? I think we do more than just a single switch on that enum that is being changed, but the "basic" scenarios probably don't run into it.

nohwnd (approved):

Haplois (approved):
@uweigandit LGTM.

nohwnd:
We are pushing to dotnet-public feed. You should be able to consume the package even before we release.  It is version 17.0.0-preview-20210628-05 and newer :) 

https://github.com/dotnet/sdk/blob/main/NuGet.config#L7

uweigand:
> We are pushing to dotnet-public feed. You should be able to consume the package even before we release. It is version 17.0.0-preview-20210628-05 and newer :)
> 
> https://github.com/dotnet/sdk/blob/main/NuGet.config#L7

Thanks for quick merge!   I can confirm that adding
```
    <MicrosoftNETTestSdkVersion>17.0.0-preview-20210628-05</MicrosoftNETTestSdkVersion>
```
to `eng/Versions.props` in roslyn allows me run run the roslyn test suite successfully on s390x.  (Roslyn already consumes the dotnet-public feed.)

uweigand:
We just noticed that projects built using  "dotnet new xunit"  still crash on startup.   This is because the .csproj files automatically generated for those still contain
```
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.4" />
```
which is too old and doesn't contain this PR.

If I understand correctly how this all works, that version number ultimately comes from https://github.com/dotnet/test-templates, right?

Now I'm wondering how this can get fixed ...   I assume we'd have to wait for an official vstest release containing the fix, and then get test-templates updated to refer to that version?

Haplois:
This support will go live with Visual Studio 2022. Right now this support is on our daily feed, it will be release with next Visual Studio 2022 preview. 

Templates will be updated after `Microsoft.NET.Test.Sdk v17.0.0` release.

uweigand:
> This support will go live with Visual Studio 2022. Right now this support is on our daily feed, it will be release with next Visual Studio 2022 preview.
> 
> Templates will be updated after `Microsoft.NET.Test.Sdk v17.0.0` release.

Got it, thanks for the explanation.  What I'm still wondering is how this lines up with the .NET 6 release ...

Specifically, is it the plan that the (copy of the) test templates included with the initial .NET 6 release (e.g. in the installer tarball) will already have been updated to refer to Microsoft.NET.Test.Sdk v17.0.0, or will they still refer to a v16.x release?

Haplois:
We generally update the templates with the preview stable release. So when when we publish `17.0.0`, templates will use `16.11.x`.

uweigand:
> We generally update the templates with the previous stable release. So when when we publish `17.0.0`, templates will use `16.11.x`.

I see.  In that case, would it be possible to backport this PR to the 16.11 branch so it will be available once the templates are updated to 16.11 ?  The goal would be to make the templates delivered with .NET 6 work out of the box on s390x ...

PR #2981, opened 2021-07-12
Adjust SettingsMigrator and build.sh for source-build

MichaelSimons:
SettingsMigrator failed to build in the offical source-build builds because it was targetting a windows RID.  This type of issue will get caught in the vstest repo's source-build leg when prebuilt detection is enabled (the source-build team will be enabling this soon).  To fix this, I added a condition to the RID.

I also made a couple tweaks to the binlog parameters within build script.


MichaelSimons:
cc @Haplois

crummel (approved):

Haplois (approved):

PR #2983, opened 2021-07-14
Update externals to 17.0.0-previews-3-31514-041

nohwnd:
## Description
Update externals to version that is using the fixed AssemblyResolver which links all 17.0 libs instead of 16.0.


nohwnd:
Re-build tomorrow, those nugets are not published yet. But this is the minimal build we need. https://devdiv.visualstudio.com/DevDiv/DevDiv%20Team/_build/results?buildId=4977851&view=results 31514.41

cvpoienaru (approved):

Haplois (approved):

PR #2984, opened 2021-07-16
Changed externals version to 17.0.0-previews-3-31516-026

cvpoienaru:
Changed externals version to 17.0.0-previews-3-31516-026

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -346,15 +346,15 @@ function Publish-Package
 
     ################################################################################
     # Publish msdia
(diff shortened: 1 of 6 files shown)


fhnaseer (approved):

nohwnd:
Retriggered to see if thw version became available, or a newer one. Is the one mentioned in the PR the earliest one to use?

cvpoienaru:
Unfortunately the minimal version to use for *.Dia is lower than the one that contains the changes we need. I'm watching the VS feeds for when a newer version becomes available.

PR #2987, opened 2021-07-20
Respect VSLANG env variable

nohwnd:
Grab language settings from VSLANG env var that VS sets to respect their language pack. Doing this the same way dotnet cli does it. https://github.com/dotnet/cli/pull/7021/files

### DIFF
--- a/src/Microsoft.TestPlatform.CoreUtilities/Utilities/UILanguageOverride.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/Utilities/UILanguageOverride.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.VisualStudio.TestPlatform.CoreUtilities.Utilities
+{
+    public static class UILanguageOverride
+    {
+        private const string DOTNET_CLI_UI_LANGUAGE = Constants.DotNetUserSpecifiedCulture;
+        private const string VSLANG = nameof(VSLANG);
+        private const string PreferredUILang = nameof(PreferredUILang);
+
+        public static void SetCultureSpecifiedByUser()
+        {
+            CultureInfo language = GetOverriddenUILanguage();
+            if (language == null)
+            {
+                return;
+            }
+
+            ApplyOverrideToCurrentProcess(language);
+            FlowOverrideToChildProcesses(language);
+        }
+
+
+        private static void ApplyOverrideToCurrentProcess(CultureInfo language)
+        {
+            CultureInfo.DefaultThreadCurrentUICulture = language;
+        }
+
+        private static CultureInfo GetOverriddenUILanguage()
+        {
+            // DOTNET_CLI_UI_LANGUAGE=<culture name> is the main way for users to customize the CLI's UI language.
+            string dotnetCliLanguage = Environment.GetEnvironmentVariable(DOTNET_CLI_UI_LANGUAGE);
+            if (dotnetCliLanguage != null)
+            {
+                try
+                {
+                    return new CultureInfo(dotnetCliLanguage);
+                }
+                catch (CultureNotFoundException) { }
+            }
+
+            // VSLANG=<lcid> is set by VS and we respect that as well so that we will respect the VS 
+            // language preference if we're invoked by VS. 
+            string vsLang = Environment.GetEnvironmentVariable(VSLANG);
+            if (vsLang != null && int.TryParse(vsLang, out int vsLcid))
+            {
+                try
+                {
+                    return new CultureInfo(vsLcid);
+                }
+                catch (ArgumentOutOfRangeException) { }
+                catch (CultureNotFoundException) { }
+            }
+
+            return null;
+        }
+
+        private static void FlowOverrideToChildProcesses(CultureInfo language)
+        {
+            // Do not override any environment variables that are already set as we do not want to clobber a more granular setting with our global setting.
+            SetIfNotAlreadySet(DOTNET_CLI_UI_LANGUAGE, language.Name);
+            SetIfNotAlreadySet(VSLANG, language.LCID.ToString()); // for tools following VS guidelines to just work in CLI
+            SetIfNotAlreadySet(PreferredUILang, language.Name); // for C#/VB targets that pass $(PreferredUILang) to compiler
+        }
+
+        private static void SetIfNotAlreadySet(string environmentVariableName, string value)
+        {
+            string currentValue = Environment.GetEnvironmentVariable(environmentVariableName);
+            if (currentValue == null)
+            {
+                Environment.SetEnvironmentVariable(environmentVariableName, value);
+            }
+        }
+    }
+}
--- a/src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
+++ b/src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
@@ -1,0 +1,71 @@
+﻿using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
+using Microsoft.VisualStudio.TestPlatform.Utilities;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.VisualStudio.TestPlatform.Execution
+{
+    internal static class DebuggerBreakpoint
+    {
+        internal static void WaitForDebugger(string environmentVariable)
+        {
+            if (string.IsNullOrWhiteSpace(environmentVariable))
+            {
+                throw new ArgumentException($"'{nameof(environmentVariable)}' cannot be null or whitespace.", nameof(environmentVariable));
+            }
+
+            var debugEnabled = Environment.GetEnvironmentVariable(environmentVariable);
+            if (!string.IsNullOrEmpty(debugEnabled) && debugEnabled.Equals("1", StringComparison.Ordinal))
+            {
+                ConsoleOutput.Instance.WriteLine("Waiting for debugger attach...", OutputLevel.Information);
+
+                var currentProcess = Process.GetCurrentProcess();
+                ConsoleOutput.Instance.WriteLine(
+                    string.Format("Process Id: {0}, Name: {1}", currentProcess.Id, currentProcess.ProcessName),
+                    OutputLevel.Information);
+
+                while (!Debugger.IsAttached)
+                {
+                    Thread.Sleep(1000);
+                }
+
+                Debugger.Break();
+            }
+        }
+
+        internal static void WaitForNativeDebugger(string environmentVariable)
+        {
+            if (string.IsNullOrWhiteSpace(environmentVariable))
+            {
+                throw new ArgumentException($"'{nameof(environmentVariable)}' cannot be null or whitespace.", nameof(environmentVariable));
+            }
+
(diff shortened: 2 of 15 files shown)


src/Microsoft.TestPlatform.CoreUtilities/Utilities/UILanguageOverride.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.VisualStudio.TestPlatform.CoreUtilities.Utilities
+{
+    public static class UILanguageOverride
```
nohwnd, line 9: Frankly I would rather see this shared via a shared project, as the code is duplicated among projects but they don't have any dependency on each other, and this just adds a possible point of failure when the object model mismatches.

@Haplois @jakubch1  @pavelhorak thoughts?

src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (!string.IsNullOrEmpty(debugEnabled) && debugEnabled.Equals("1", StringComparison.Ordinal))
+            {
+                ConsoleOutput.Instance.WriteLine("Waiting for debugger attach...", OutputLevel.Information);
+
+                var currentProcess = Process.GetCurrentProcess();
+                ConsoleOutput.Instance.WriteLine(
+                    string.Format("Process Id: {0}, Name: {1}", currentProcess.Id, currentProcess.ProcessName),
+                    OutputLevel.Information);
+
+                while (!Debugger.IsAttached)
```
Haplois, line 32: This line sometimes hand forever. (Especially when the user doesn't know that they are debugging.) Should we try attaching to a debugger using `Debugger.Attach`?

src/Microsoft.TestPlatform.Execution.Shared/Microsoft.TestPlatform.Execution.Shared.projitems
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MSBuildAllProjects Condition="'$(MSBuildVersion)' == '' Or '$(MSBuildVersion)' &lt; '16.0'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <HasSharedItems>true</HasSharedItems>
+    <SharedGUID>7f26eda3-c8c4-4b7f-a9b6-d278c2f40a13</SharedGUID>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <Import_RootNamespace> Microsoft.VisualStudio.TestPlatform.Execution</Import_RootNamespace>
```
Haplois, line 9:
```suggestion
    <Import_RootNamespace>Microsoft.VisualStudio.TestPlatform.Execution</Import_RootNamespace>
```

nit.

src/datacollector/DataCollectorMain.cs
```diff
@@ -63,7 +63,7 @@ internal DataCollectorMain(IProcessHelper processHelper, IEnvironment environmen
 
         public void Run(string[] args)
         {
-            WaitForDebuggerIfEnabled();
+            DebuggerBreakpoint.WaitForDebugger("VSTEST_DATACOLLECTOR_DEBUG");
```
Haplois, line 66: Since we are moving code... Should we move these constants to shared place as well?

src/datacollector/datacollector.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
   </PropertyGroup>
+  <Import Project="..\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```
Haplois, line 6:
```suggestion
  <Import Project="$(TestPlatformRoot)src\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```

src/testhost.x86/Program.cs
```diff
@@ -52,8 +52,9 @@ public static void Main(string[] args)
         // In UWP(App models) Run will act as entry point from Application end, so making this method public
         public static void Run(string[] args)
         {
-            WaitForDebuggerIfEnabled();
-            SetCultureSpecifiedByUser();
+            DebuggerBreakpoint.WaitForNativeDebugger("VSTEST_HOST_NATIVE_DEBUG");
```
Haplois, line 55: Some comment about possibly moving the names for these variables to sharedproj too..

src/testhost.x86/testhost.x86.csproj
```diff
@@ -5,6 +5,7 @@
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
+  <Import Project="..\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```
Haplois, line 8:
```suggestion
  <Import Project="$(TestPlatformRoot)src\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```

src/testhost/testhost.csproj
```diff
@@ -5,6 +5,7 @@
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
   </PropertyGroup>
+  <Import Project="..\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```
Haplois, line 8:
```suggestion
  <Import Project="$(TestPlatformRoot)src\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```

src/vstest.console/Program.cs
```diff
@@ -20,44 +17,10 @@ public static class Program
...
-                try
-                {
-                    CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.CreateSpecificCulture(userCultureSpecified);
-                }
-                catch(Exception)
-                {
-                    ConsoleOutput.Instance.WriteLine(string.Format("Invalid Culture Info: {0}", userCultureSpecified), OutputLevel.Information);
-                }
-            }
+            DebuggerBreakpoint.WaitForDebugger("VSTEST_RUNNER_DEBUG");
```
Haplois, line 21: So... Should we move?

src/vstest.console/vstest.console.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
   </PropertyGroup>
+  <Import Project="..\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```
Haplois, line 6:
```suggestion
  <Import Project="$(TestPlatformRoot)src\Microsoft.TestPlatform.Execution.Shared\Microsoft.TestPlatform.Execution.Shared.projitems" Label="Shared" />
```

src/Microsoft.TestPlatform.Execution.Shared/Microsoft.TestPlatform.Execution.Shared.projitems
Haplois: Are we using this property anywhere?

Haplois (approved):
I like this approach. I have some comments thou.

src/testhost.x86/Program.cs
nohwnd: Why would you move those variables to shared project when they are not shared?

src/vstest.console/Program.cs
nohwnd: What do you mean?

src/datacollector/DataCollectorMain.cs
nohwnd: I don't think so. This is only place where it is used. No reason to "share" that.

src/Microsoft.TestPlatform.Execution.Shared/Microsoft.TestPlatform.Execution.Shared.projitems
nohwnd: It is used to define the default namespace. otherwise the namespace would be the same as the name of the project.

src/testhost.x86/Program.cs
Haplois: Just to have them in one place.

src/vstest.console/Program.cs
Haplois: > Some comment about possibly moving the names for these variables to sharedproj too..

:) Just marking to spot so we don't miss it if we decide move it.

src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
nohwnd: I don't think so. In most cases you don't have anything to attach to, or you don't want to start VS for every start of the executable.

src/testhost.x86/Program.cs
nohwnd: I don't agree with that. This would publish those constants to other projects that don't care about them. And those constants are used literally once in the whole code base. I don't want to have another ".common" library.

src/vstest.console/Program.cs
nohwnd: I am for not moving.

src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
Haplois: I think it doesn't start a new instance of VS. It will show this dialog from which you can an select an existing instance or launch a new one.

image

nohwnd: I just moved the code, I don't want to also change the behavior in this PR. If you want to do it, please start a new PR. I will give it a try.

src/Microsoft.TestPlatform.Execution.Shared/UILanguageOverride.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        private static void ApplyOverrideToCurrentProcess(CultureInfo language)
+        {
+            CultureInfo.DefaultThreadCurrentUICulture = language;
+        }
+
+        private static CultureInfo GetOverriddenUILanguage()
+        {
+            // DOTNET_CLI_UI_LANGUAGE=<culture name> is the main way for users to customize the CLI's UI language.
+            string dotnetCliLanguage = Environment.GetEnvironmentVariable(DOTNET_CLI_UI_LANGUAGE);
```
AbhitejJohn, line 36: @nohwnd : Looking at the env variable name, just wanted to confirm that this works for .NET Framework and VS scenarios as well? Sounds like it does from your offline responses but thought I'll put it out here.

AbhitejJohn: Never mind, looks like that is covered by the code below.

PR #2990, opened 2021-07-27
Respect VSLANG env variable

Haplois:
Grab language settings from VSLANG env var that VS sets to respect their language pack. Doing this the same way dotnet cli does it. https://github.com/dotnet/cli/pull/7021

Cherry-picked from #2987


Haplois:
We decided against this fix, since it might be a breaking change.

PR #2991, opened 2021-07-27
Controlling code coverage externals version using Darc

jakubch1:
## Description
This PR is changing way how CC version is stored to enable automatic updates using Darc.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
+    <Dependency Name="Microsoft.Internal.CodeCoverage" Version="17.0.0-beta.21377.6">
+      <Uri>https://dev.azure.com/devdiv/DevDiv/_git/vs-code-coverage</Uri>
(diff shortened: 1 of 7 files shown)


eng/Version.Details.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
+    <Dependency Name="Microsoft.Internal.CodeCoverage" Version="17.0.0-beta.21377.6">
```
Sanan07, line 4: Why don't you use `MicrosoftInternalCodeCoverageVersion` parameter from Versions.props here?

jakubch1: This version is controlled by Darc. This will be automatically changed.

jakubch1: Check here: https://github.com/dotnet/arcade/blob/main/Documentation/Darc.md#updating-dependencies-in-your-local-repository

jakubch1: This is example of such automatic update: https://devdiv.visualstudio.com/DevDiv/_git/vs-code-coverage/pullrequest/340431?_a=files&path=%2Feng%2FVersions.props
You can see that version is changed in both places

Haplois (approved):

PR #2996, opened 2021-07-30
16.11 External Packages Insertion

Haplois:
## Description
Updated VS Externals


cvpoienaru (approved):
Tentatively approving this but please check if all the localization changes should've been included as well before merging.

Haplois:
They are just formatting changes - I checked. This way it'll be easier on us if we need to service it again.

PR #2998, opened 2021-08-02
Fix run in no isolation to use x64 architecture

nohwnd:
## Description
Run in no isolation is decided based on the inferred architecture, and if the architecture is not the same it will offload to a host. vstest.console is now a 64-bit executable for the desktop case, so we need to non-x64 bit workloads outside of the executable.

### DIFF
--- a/test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/Properties/launchSettings.json
+++ b/test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/Properties/launchSettings.json
@@ -1,8 +1,0 @@
-{
-  "profiles": {
-    "WSL": {
-      "commandName": "WSL2",
-      "distributionName": ""
-    }
-  }
-}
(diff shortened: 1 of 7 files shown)


test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "WSL": {
+      "commandName": "WSL2",
+      "distributionName": ""
+    }
+  }
+}
```
Haplois, line 8: Did you mean to include this?

nohwnd: Nope, removed.

Haplois (approved):

nohwnd:
The status is not reported but the build passed. https://dev.azure.com/dnceng/public/_build/results?buildId=1272941&view=results

PR #3025, opened 2021-08-31
Updating tests in TP based on recent work

Sanan07:
1. Resuming previously commented tests.
2. Removing ordered and generic tests.
3. Updating externals


Haplois (approved):
LGTM

Issue #3005, opened 2021-08-08
--blame-hang causing crash with AccessViolationException


nick-beer:
I'm using --blame-hang in an invocation to `dotnet vstest` (via a runsettings file) on an AzDO pipeline.  In the logs, I can see that when the blame collector identifies a hang, it tries to create a dump of the hung process, but it's failing with the following `AccessViolationException`:

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.WindowsHangDumper+NativeMethods.MiniDumpWriteDump(IntPtr, UInt32, Microsoft.Win32.SafeHandles.SafeFileHandle, MINIDUMP_TYPE, MINIDUMP_EXCEPTION_INFORMATION ByRef, IntPtr, IntPtr)
--------------------------------
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.WindowsHangDumper.CollectDump(System.Diagnostics.Process, System.String, Microsoft.TestPlatform.Extensions.BlameDataCollector.DumpTypeOption)
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.WindowsHangDumper.Dump(Int32, System.String, Microsoft.TestPlatform.Extensions.BlameDataCollector.DumpTypeOption)
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.ProcessDumpUtility.HangDump(Int32, System.String, Microsoft.TestPlatform.Extensions.BlameDataCollector.DumpTypeOption, System.String, System.Action`1<System.String>)
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.ProcessDumpUtility.StartHangBasedProcessDump(Int32, System.String, Boolean, System.String, System.Action`1<System.String>)
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.BlameCollector.CollectDumpAndAbortTesthost()
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.InactivityTimer+<>c__DisplayClass1_0.<.ctor>b__0(System.Object)
   at System.Threading.TimerQueueTimer+<>c.<.cctor>b__23_0(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
[...]

PR #3028, opened 2021-09-01
Blame fixes

nohwnd:
## Description
Fix few issues with Blame collector. 

1) Crash collector was catching FailFast because in the past it was done by always collecting the dump, and only sending it back when user enabled AlwaysCollect. With the new approach where crashes can produce multiple dumps, including dumps from child processes we can't really tell if the process crashed or exited normally. So I am using the test count to determine if the testhost crashed while executing tests. It would definitely be nicer to get a message with the actual exit status after testhost exits, but that would mean extending the IDataCollector interface, which needs to be done with a bit more thinking. 

The native .NET Crash monitor still don't seem to catch FailFast, but there is internal option `VSTEST_DUMP_FORCEPROCDUMP=1` that forces the use of ProcDump even where the built it net dump would be used. 

2) This now also applies the hang dumps, because on some systems the PInvoke I do to not need ProcDump throws AccessViolationException, so this is a way to force that to use ProcDump. 

3) There is also `VSTEST_DUMP_FORCENETDUMP=1` internal option which is forcing the net dumper to be used on Windows so we can test it a bit better.

4) I also added `VSTEST_DUMP_PROCDUMPARGUMENTS` and `VSTEST_DUMP_PROCDUMPADDITIONALARGUMENTS` internal option to fully or partially override what arguments procdump will use, because I found it quite useful when debugging crashes. And there is sadly no way to collect both "managed" and "native" dumps at the same time. So when I can't really see the crash I need to change how proc dump runs a lot.

Not sure how to fix this properly without relying on those internal env variables to set the dumpers. Maybe I could try looking if procdump can be found when on windows and try using that, otherwise try the other dumpers based on the platform. 

## Related issue
Workaround for #3005 by using `VSTEST_DUMP_FORCEPROCDUMP=1`
Workaround for https://github.com/dotnet/sdk/issues/14578#issuecomment-910168503 also by using `VSTEST_DUMP_FORCEPROCDUMP=1`

Easiest way to install procdump is by using this task: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/vstest?view=azure-devops
 
Easiest way to deploy latest version of TestPlatform is by using this task: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/vstest-platform-tool-installer?view=azure-devops

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -214,6 +214,14 @@ namespace Microsoft.TestPlatform.Extensions.BlameDataCollector
                 EqtTrace.Verbose("Inactivity timer is already disposed.");
             }
 
+            if (this.collectProcessDumpOnTrigger)
+            {
+                // Detach procdump from the testhost process to prevent testhost process from crashing
+                // if/when we try to kill the existing proc dump process.
+                // And also prevent collecting dump on exit of the process.
+                this.processDumpUtility.DetachFromTargetProcess(this.testHostProcessId);
+            }
+
             try
             {
                 Action<string> logWarning = m => this.logger.LogWarning(this.context.SessionDataCollectionContext, m);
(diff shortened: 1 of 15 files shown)


nohwnd:
> The native .NET Crash monitor still don't seem to catch FailFast,

Actually it catches it when I run on net6, and net5, in a console app. Will have to test again, and codify that into my pipeline. Because there are more and more scenarios. 😁

AArnott:
I'm happy to test when you have a package for me.

Haplois (approved):

Issue #2362, opened 2020-03-17
Review visibility of public interfaces


cvpoienaru:
### Motivation

Some of the public interfaces inside this repository should not be public by default (e.g.: IDesignModeClient, etc.) as they are only used by components implemented by us that aren't meant to be replaceable by third party components.

In order to avoid exposing unneeded interfaces and make it convenient for us to mutate them without having to worry about breaking compatibility, we should review which public interfaces should remain public and which ones can be made internal.

nohwnd:
Taking note of this in a design doc, this needs huge refactoring of the extensibility model, and dependency model to have any impact.

PR #3042, opened 2021-09-08
Add public api analyzers for ObjectModel and dependencies

nohwnd:
## Description
Add public API analyzers for ObjectModel and it's dependencies to avoid changing public apis that we shipped. 

Adding only to ObjectModel, because de definitions need to be duplicated for every TFM. Once/if this is merged, we can add it to all projects. Even those where changing the public api is not critical because they are used only by us. 

Easiest way to introduce changes is by using the drop down TFM selector in VS, and fixing it for every TFM.

Related #2362


nohwnd:
@Haplois we don't normally build for net6.0 only for source build, which now fails on the server. I think it's okay if I just don't analyze for that tfm?

Sanan07:
What does this analyzer do? It checks when we add some new public API and checks if we have it in that txt file?
So then we need manually add new API in txt file?

nohwnd:
@Sanan07 For any public API it will check the shipped and unshipped files. If it does not find it, then RSxxx build error occurs and you will get the option to add the api using the normal lightbulb quick fix. The annoying thing for now is that you need to apply the fix for every TFM that it applies to. You can copy the line that was generated into Unshipped.txt into the other files, or select each tfm in VS and apply the same fix.

This can also be done for all occurrences of the error, per TFM. But sadly not for all TFMs at the same time. This will get better hopefully when the above fix is merged, because then we can have a central file with the common api, and separate files with TFM specific API. So you only need to change it in one file when it applies to all TFMs. 

Additionally I will add a check to GH to require approval when PR changes any of the shipped and unshipped files. And before release, or someone in the release cycle we should review the apis that we would newly ship.

Sanan07 (approved):

PR #3050, opened 2021-09-10
New aborting flow

### LINKED ISSUE #2902: Fix discovery abort scenario
## Description
Fix current aborting logic with correct handling it on testhost side.
That is already implemented here #2892.
Need more debugging, testing and code review.

[AB#1338987](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1338987)

Sanan07:
## Description
New aborting flow with event handling on testhost side

## Needs to be done 

- [x] Add new protocol version
- [x] Add constraints for new flow to be used in new protocol version
- [x] Add unit and acceptance tests
- [x] Change naming in flow starting from VstestConcoleWrapper to remove confusion about `CancelDiscovery` and `AbortDiscovery`

## Related issue
Fixes #2902


nohwnd:
If we are not doing it already, it would be nice to document the changes in the protocol. Ideally incremental changes from version 1. I think this is how it went:

1) Initial version of the protocol. Is very verbose and slow because it serializes all bag properties in very verbose way.
2) Same as 1 but serializes known bag properties more efficiently. 
3) Adds ???. But is incorrectly handled and reverts serializer back to v1 serializer, making it really slow. Do not use. Vstest.console automatically downgrades the usage to 2 if it receives 3.
4) Same as 3, but uses serializer for v2 which is fast.
5) ???

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -226,7 +227,16 @@ private void ProcessRequests(ITestRequestManager testRequestManager)
 
                         case MessageType.CancelDiscovery:
                             {
-                                testRequestManager.CancelDiscovery();
+                                // If testhost has old version, we should call old cancel discovery logic
+                                // to be consistent and not create regression issues
+                                if (this.protocolConfig.Version < ObjectModel.Constants.DefaultProtocolConfig.Version)
```
nohwnd, line 232: Shouldn't this change to the literal version where the change is introduced, that is 6? If the default version changes to 7 in the future, you would be calling the old api for 6 as well.

Sanan07: That is correct, will fix it.

src/Microsoft.TestPlatform.ObjectModel/Client/Interfaces/IDiscoveryRequest.cs
```diff
@@ -49,5 +49,10 @@ DiscoveryCriteria DiscoveryCriteria
         /// Aborts the discovery request
         /// </summary>
         void Abort();
+
+        /// <summary>
+        /// Aborts the discovery request with event handler
+        /// </summary>
+        void AbortWithEventHandler();
```
nohwnd, line 56: Why is this important to be exposed as a separate operation than `Abort`? Is there any point in doing `Abort` instead of `AbortWithEventHandler` if both side support protocol version 6 or newer?

nohwnd: In other words, can we just use Abort and switch the behavior on the protocol version instead of exposing a new operation?

Sanan07: This new method was introduced to fix aborting flow and make it complete. New `AbortWithEventHandler ` flow sends message to testhost and in the end `HandleDiscoveryComplete` method is executed. I added new method to be easier to understand and easier separate behaviour based on protocol version.

nohwnd: > I added new method to be easier to understand and easier separate behaviour based on protocol version. 

Okay, but this is our public api. Is there any reason to call Abort on a test host that supports AbortWithEventHandler? 

I am asking because we cannot remove Abort method, and if there is no reason to use it in the future, it would be just hanging there. It seems better to just do this bug fix inside, and if the client supports it, we would call AbortWithEventHandler (protocol >= 6), otherwise we would call Abort (protocol  <6). But we would not expose an extra method for it.

Sanan07: I couldn't find a way to check protocol version inside `DiscoveryRequest` to add new method only there and check protocol version there. So that why added new api to be consistent in interfaces.

PR #3067, opened 2021-09-22
Suppress warnings

Haplois:
## Description
Suppresses warnings about `netcoreapp2.1` end of life.


nohwnd:
Wouldn't it be better to put this into the files with common targets? Or that is not common to all projects?

Haplois:
> Wouldn't it be better to put this into the files with common targets? Or that is not common to all projects?

It would.

Haplois:
Replaced by #3078.

Issue #1811, opened 2018-10-22
Add support for merging coverage files from multiple test projects using dotnet test --collect:"Code Coverage"


lomholdt:
## Description
When collecting code coverage using `dotnet test --collect:"Code Coverage"` I would like to be able to merge multiple coverage files into one if I have multiple test projects in a solution. 

[See #981 for reference](https://github.com/Microsoft/vstest/issues/981#issuecomment-431771521)

## Expected behavior
A single coverage file that contains coverage for all test projects in a solution.

## Actual behavior
Multiple coverage files are created in each test project

[AB#1272814](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1272814)

kennethkryger:
+1 🤞

KLuuKer:
Any decent size application (heck even small ones) will have many projects each with there own tests, so this is a must!

mayankbansal018:
@cltshivash , @PBoraMSFT can you check where it can be fit in our backlog

michael-colclough:
This is very important to our team. Without it, the code coverage results are meaningless. Please could this be implemented as soon as possible

cheenamalhotra:
We are also waiting for this support to be added, it's a basic requirement in order to gather cumulative code coverage. Is there an ETA to this request?

ManishJayaswal:
bumping the priority and also assigning to @jakubch1 as he is working on code coverage. Let's see if we can get to this in 16.6 release.

nohwnd:
For what it's worth, the types that can merge the code coverage files already ship with the test platform. You can parse the coverage files out from the output and merge them yourself. The only hiccup here is that the interop library is built for specific architecture, so you need the correct bitness. Both 32-bit and 64-bit versions are shipped in TP.

Here is an example that works for solution that mixes netcoreapp2.1 and 3.1 and produces two coverage files as a result.


```powershell
dotnet build
dotnet test --collect:"Code Coverage" | Tee-Object -Variable output

$coverage = if ($null -ne $output) { 
    $output | 
    Select-String \.coverage | 
    ForEach-Object { $_.Line.Trim() }
}

if ($null -ne $coverage){ 
    if (1 -eq $coverage.Count) {
        $coverage
    }
    else {
        $architecture = if (64 -eq [System.IntPtr]::Size * 8) { "Amd64" } else { "X86" }

[...]

PR #3161, opened 2021-11-09
Complete the attachment processors extension

### LINKED ISSUE #2379: Possible confusing logging on ProxyOperationManager.cs
Usually when we've some issue with coverlet msbuild or .NET tool driver we enable logging to understand if vstest plat killed process after famous 100ms.
BTW seem that logging is wrong and also in case of clean shutdown timeout log is emitted

https://github.com/microsoft/vstest/blob/d10bcbb28cc3999bcc12758a41a04b998eb9595b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs#L202-L236

I think that log emitted in line 228 should be emitted only if `testHostExited` event was not signaled.

MarcoRossignoli:
Complete attachment processors extension

fixes https://github.com/microsoft/vstest/issues/2379

Missing part unrelated to this repo:
* Update documentation for attachment processor discovery
* Move coverage attachment processor to coverage repo
* VS integration

contributes to https://github.com/microsoft/vstest/issues/1811


src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+        private const string CoverageFriendlyName = "Code Coverage";
+
+        public IReadOnlyDictionary<string, IConfigurableDataCollectorAttachmentProcessor> Create(InvokedDataCollector[] invokedDataCollectors)
+        {
+            IDictionary<string, Tuple<string, IConfigurableDataCollectorAttachmentProcessor>> datacollectorsAttachmentsProcessors = new Dictionary<string, Tuple<string, IConfigurableDataCollectorAttachmentProcessor>>();
+            bool addCodeCoverageAttachmentProcessors = true;
+
+            if (invokedDataCollectors != null)
+            {
+                // We order files by filename descending so in case of the same collector from the same nuget but with different versions, we'll run the newer version.
```
AbhitejJohn, line 27: > We order files by filename descending so in case of the same collector from the same nuget but with different versions, we'll run the newer version. [](http://example.com/codeflow?start=18&length=149)

When would this normally happen? (where we have different version of coverlet referenced)
Is this across projects in the solution?

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+                // i.e. C:\Users\xxx\.nuget\packages\coverlet.collector
+                // /3.0.2
+                // /3.0.3
+                // /3.1.0
+                foreach (var invokedDataCollector in invokedDataCollectors.OrderByDescending(d => d.FilePath))
+                {
+                    EqtTrace.Info($"DataCollectorAttachmentsProcessorsFactory: Analyzing data collector attachment processor Uri: {invokedDataCollector.Uri} AssemblyQualifiedName: {invokedDataCollector.AssemblyQualifiedName} FilePath: {invokedDataCollector.FilePath} HasAttachmentProcessor: {invokedDataCollector.HasAttachmentProcessor}");
```
AbhitejJohn, line 34: >AssemblyQualifiedName [](http://example.com/codeflow?start=157&length=21)

What format is this in? Would be nice to add a comment in code. Same for fie path as well.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                    // We'll merge using only one AQN in case of more "same processors" in different assembly.
+                    if (!invokedDataCollector.HasAttachmentProcessor || datacollectorsAttachmentsProcessors.ContainsKey(invokedDataCollector.AssemblyQualifiedName))
+                    {
+                        continue;
+                    }
+
+                    var dataCollectorExtensionManager = DataCollectorExtensionManager.Create(invokedDataCollector.FilePath, TestSessionMessageLogger.Instance);
```
AbhitejJohn, line 42: >DataCollectorExtensionManager.Create [](http://example.com/codeflow?start=56&length=36)

Do we cache this within the extension manager today?

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+                    {
+                        Type attachmentProcessorType = ((DataCollectorConfig)dataCollectorExtension.TestPluginInfo).AttachmentsProcessorType;
+                        IConfigurableDataCollectorAttachmentProcessor dataCollectorAttachmentProcessorInstance = null;
+                        try
+                        {
+                            dataCollectorAttachmentProcessorInstance = TestPluginManager.CreateTestExtension<IConfigurableDataCollectorAttachmentProcessor>(attachmentProcessorType);
+                        }
+                        catch (Exception ex)
+                        {
+                            EqtTrace.Error($"DataCollectorAttachmentsProcessorsFactory: Failed during the creation of data collector attachment processor '{attachmentProcessorType.AssemblyQualifiedName}'\n{ex}");
```
AbhitejJohn, line 54: >EqtTrace.Error [](http://example.com/codeflow?start=28&length=14)

curious, where do these get printed? Do they go into the test platform diagnostic log file? How about to the output->Tests pane if it is driven through VS...

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+                        }
+
+                        if (dataCollectorAttachmentProcessorInstance != null && !datacollectorsAttachmentsProcessors.ContainsKey(attachmentProcessorType.AssemblyQualifiedName))
+                        {
+                            datacollectorsAttachmentsProcessors.Add(attachmentProcessorType.AssemblyQualifiedName, new Tuple<string, IConfigurableDataCollectorAttachmentProcessor>(dataCollectorExtension.Metadata.FriendlyName, dataCollectorAttachmentProcessorInstance));
+
+                            if (invokedDataCollector.Uri.AbsoluteUri == CoverageUri.AbsoluteUri)
+                            {
+                                EqtTrace.Info($"DataCollectorAttachmentsProcessorsFactory: Attachment data processor for data collector with friendly name '{CoverageFriendlyName}' found '{attachmentProcessorType.AssemblyQualifiedName}' inside '{invokedDataCollector.FilePath}'");
```
AbhitejJohn, line 63: >EqtTrace.Info($"DataCollectorAttachmentsProcessorsFactory: Attachment data processor for data collector with friendly name '{CoverageFriendlyName}' found '{attachmentProcessorType.AssemblyQualifiedName}' inside '{invokedDataCollector.FilePath}'"); [](http://example.com/codeflow?start=32&length=247)

hmm, this is interesting. Could we add comments on when this would happen and why we would want to overwrite our default processor if we see an external one with the same uri. Shouldn't we ideally throw and ignore the external one?

src/Microsoft.TestPlatform.ObjectModel/InvokedDataCollector.cs
```diff
@@ -0,0 +1,99 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
+{
+    public sealed class InvokedDataCollector : IEquatable<InvokedDataCollector>
```
AbhitejJohn, line 9: >public [](http://example.com/codeflow?start=4&length=6)

Does this need to be public? Can we make it internal?

src/Microsoft.TestPlatform.ObjectModel/DataCollector/IConfigurableDataCollectorAttachmentProcessor.cs
```diff
@@ -0,0 +1,51 @@
...
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using System.Xml;
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+    /// <summary>
+    /// Interface for data collectors add-ins that choose to reprocess generated attachments
+    /// </summary>
+    public interface IConfigurableDataCollectorAttachmentProcessor
```
AbhitejJohn, line 16: >IConfigurableDataCollectorAttachmentProcessor [](http://example.com/codeflow?start=21&length=45)

So the one difference between this and the existing `IDataCollectorAttachmentProcessor` interface is that we now pass in the data collector configuration in the ProcessAttachmentSetsAsync method? Do we know if this interface is (or even can be) used by anyone outside? I wonder if we can just change the existing interface if this wasn't usable externally anyway. Also would we need the whole runsettings blob instead of just that data collectors config. Would someone be interested in reading anything in the runconfiguration element for instance?

src/Microsoft.TestPlatform.ObjectModel/InvokedDataCollector.cs
AbhitejJohn: Never mind I see that we reference this in other public interfaces.

test/TestAssets/AttachmentProcessorDataCollector/TestExtensionTypesAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+            this.Types = types;
+        }
+
+        public Type[] Types { get; }
+    }
+
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = true)]
+    internal sealed class TestExtensionTypesV2Attribute : Attribute
+    {
+        public string ExtensionType { get; }
```
AbhitejJohn, line 29: >ExtensionType [](http://example.com/codeflow?start=22&length=13)

I'm assuming this attribute would make it into the product after a review. Would be nice to add comments and example formats of what each of these signify.

src/Microsoft.TestPlatform.Common/Utilities/MetadataReaderHelper.cs
```diff
@@ -0,0 +1,156 @@
...
+                            {
+                                // Expected ctor shape ctor(string,string,System.Type,Int32)
+                                // [assembly: TestExtensionTypesV2(ExtensionMetadata.ExtensionType, ExtensionMetadata.Uri, typeof(ExtensionImplementation), 1)]
+
+                                // If the parameter kind is string, (middle line in above diagram) then the blob contains
+                                // a SerString – a PackedLen count of bytes, followed by the UTF8 characters.If the
+                                // string is null, its PackedLen has the value 0xFF(with no following characters).If
+                                // the string is empty(“”), then PackedLen has the value 0x00(with no following
+                                // characters).
+                                string extension = valueReader.ReadSerializedString();
```
AbhitejJohn, line 94: >extension [](http://example.com/codeflow?start=39&length=9)

what are the values that this can take?

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: Yep if for instance in VS we have more than one project with different version(we're using the coverage collector as use case)  the idea is to use the up to date one.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: Assembly qualified name is the standard type name format https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=net-6.0#remarks

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: They will go to diagnostic files.

> How about to the output->Tests pane if it is driven through VS...

At the moment looks like this messages are not sent to tests pane in VS. I'll verify the output story there.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: The idea here is the opposite, today we're shipping attachment processor for coverage inside TP. 
Now we're moving the implementation to coverage repo and we won't move on future implementation there. 
So we cannot remove current implementation for back-compat reason but we want to use implementation coming from datacollector if present.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: Yes this call will cache the extension.

src/Microsoft.TestPlatform.ObjectModel/DataCollector/IConfigurableDataCollectorAttachmentProcessor.cs
MarcoRossignoli: >  wonder if we can just change the existing interface if this wasn't usable externally anyway. 

We cannot update the existing one(we can do addition of new method but looks weird because for the implementer only one out of the two method will be called so it's not clear imo from user perspective) because if we implement new one but we load collector extension inside old SDK hosts we'll throw a missing implementation exception.

> Also would we need the whole runsettings blob instead of just that data collectors config. Would someone be interested in reading anything in the runconfiguration element for instance?

We had this discussion with @jakubch1 and @fhnaseer and we can do it, but again it could look strange from user perspective, inside DataCollector we're flowing only collector section config and not all runsettings so it's a bit asymmetric. And we expect that settings for an attachment processor that is "attached" to specific collector should handle that "context" and so settings should come from datacollector section.

Anyway we're open to discuss this one if you strongly feel the need.

test/TestAssets/AttachmentProcessorDataCollector/TestExtensionTypesAttribute.cs
MarcoRossignoli: The idea is to "not ship" the attribute inside the object model. 
We will ship the "documentation" of it very similar to what we're doing inside coverlet for code coverage exclusion https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Instrumentation/Instrumenter.cs#L66

In this way we won't risk to ship a version of attribute and in future change the signature and come back to the same "load issue" we have today.

src/Microsoft.TestPlatform.Common/Utilities/MetadataReaderHelper.cs
MarcoRossignoli: First two value for the attribute are for "future use".
We want to be able in future to understand from the attribute the "type" of extension and the "identifier" for better performance and to avoid to "get the type" if we're searching for a different type of extension(i.e. DataCollector, TestAdapter etc...).
We didn't define yet the precise "type name" for all extensions.

MarcoRossignoli: For now we're using Type and version.

jakubch1 (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: We changed here, we cannot cache this extension because in VS scenario if a user for instance update the runsettings we should load and run the new attachment processor.

src/Microsoft.TestPlatform.ObjectModel/DataCollector/IConfigurableDataCollectorAttachmentProcessor.cs
jakubch1: Let's break existing interface as nobody is using it now and regarding runsettings I think we should be consistent and pass only part for this collector. If some additional options are needed we will need to add options to this data collector.

src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/DataCollectionResult.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.DataCollection
+{
+    /// <summary>
+    /// Information returned after data collection.
+    /// </summary>
+    public class DataCollectionResult
```
MarcoRossignoli, line 12: check if works internals

MarcoRossignoli: We cannot, is used by object with public visibility

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -391,6 +391,14 @@ private void HandleAfterTestRunEnd(Message message)
 
             var attachmentsets = this.dataCollectionManager.SessionEnded(isCancelled);
             var invokedDataCollectors = this.dataCollectionManager.GetInvokedDataCollectors();
+            if (invokedDataCollectors != null && invokedDataCollectors.Count > 0)
+            {
+                foreach (var invokedDataCollector in invokedDataCollectors)
```
MarcoRossignoli, line 396: @AbhitejJohn this is the place where we want to send an "object metric"

cc: @jakubch1

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -391,6 +391,14 @@ private void HandleAfterTestRunEnd(Message message)
...
+                {
+                    this.requestData.MetricsCollection.Add(TelemetryDataConstants.InvokedDataCollector, invokedDataCollector);
```
MarcoRossignoli, line 398: @AbhitejJohn this is the place where we want to send an "object metric"

cc: @jakubch1

src/Microsoft.TestPlatform.Common/Utilities/MetadataReaderHelper.cs
```diff
@@ -0,0 +1,191 @@
...
+                    ExtensionType = extensionType;
+                    ExtensionIdentifier = extensionIdentifier;
+                    ExtensionImplementation = extensionImplementation;
+                    Version = version;
+                }
+            }
+        }
+
+    */
+    internal class MetadataReaderExtensionsHelper
```
MarcoRossignoli, line 43: TODO: Add units

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
MarcoRossignoli: We won't add telemetry in this PR, we need a better way to send "object data", primitive type are not so useful or easy to query.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
MarcoRossignoli: I've added message logger, it will flow to VS thanks to ITestRunAttachmentsProcessingEventsHandler

src/Microsoft.TestPlatform.Common/Interfaces/Engine/IDataCollectorAttachmentsProcessorsFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    internal interface IDataCollectorAttachmentsProcessorsFactory
+    {
+        /// <summary>
+        /// Creates and return a list of available attachments processor
+        /// </summary>
+        /// <param name="invokedDataCollector">List of invoked data collectors</param>
+        /// <param name="logger">Message logger</param>
+        /// <returns>List of attachments processors</returns>
+        DataCollectorAttachmentProcessor[] Create(InvokedDataCollector[] invokedDataCollectors, IMessageLogger logger);
```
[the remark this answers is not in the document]
MarcoRossignoli: `IEnumerable` express deferred evaluation...it means that if you run methods two time you can get different result. Public contract without streaming capabilities should return "immutable" objects of possible.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -339,28 +342,35 @@ internal IEnumerable<string> DefaultExtensionPaths
...
             }
+            else
+            {
+                // Check if extensions from this assembly have already been discovered.
+                var extensions = this.TestExtensions?.GetExtensionsDiscoveredFromAssembly<TPluginInfo>(
+                    this.TestExtensions.GetTestExtensionCache<TPluginInfo>(),
+                    extensionAssembly);
 
-            var pluginInfos = this.GetTestExtensions<TPluginInfo, TExtension>(new List<string>() { extensionAssembly });
+                if (extensions != null && extensions.Count > 0)
```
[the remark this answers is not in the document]
MarcoRossignoli: Discussed offline

src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -9,20 +9,17 @@ const Microsoft.VisualStudio.TestPlatform.Common.TestPlatformDefaults.EnableBoun
...
 Microsoft.VisualStudio.TestPlatform.Common.DataCollection.BeforeTestRunStartResult
 Microsoft.VisualStudio.TestPlatform.Common.DataCollection.BeforeTestRunStartResult.BeforeTestRunStartResult(System.Collections.Generic.IDictionary<string, string> environmentVariables, int dataCollectionEventsPort) -> void
 Microsoft.VisualStudio.TestPlatform.Common.DataCollection.BeforeTestRunStartResult.DataCollectionEventsPort.get -> int
 Microsoft.VisualStudio.TestPlatform.Common.DataCollection.BeforeTestRunStartResult.EnvironmentVariables.get -> System.Collections.Generic.IDictionary<string, string>
 Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException
 Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException.InvalidLoggerException(string message) -> void
 Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException.InvalidLoggerException(string message, System.Exception innerException) -> void
-Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.DataCollectorMetadata
```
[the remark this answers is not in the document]
MarcoRossignoli: Will revert

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
```diff
@@ -26,6 +26,8 @@ public static class TelemetryDataConstants
 
         public static string DataCollectorsEnabled = "VS.TestRun.DataCollectorsEnabled";
 
+        internal const string InvokedDataCollectors = "VS.TestRun.InvokedDataCollectors";
```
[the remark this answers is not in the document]
MarcoRossignoli: Thanks will check how to classify

src/Microsoft.TestPlatform.Common/ExtensionFramework/DataCollectionExtensionManager.cs
```diff
@@ -51,20 +51,45 @@ internal class DataCollectorExtensionManager : TestExtensionManager<DataCollecto
...
+
+            return new DataCollectorExtensionManager(unfilteredTestExtensions, filteredTestExtensions, messageLogger);
+        }
     }
 
     /// <summary>
     /// Hold data about the Data Collector.
     /// </summary>
-    public class DataCollectorMetadata : IDataCollectorCapabilities
+    internal class DataCollectorMetadata : IDataCollectorCapabilities
```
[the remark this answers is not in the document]
MarcoRossignoli: I will revert, tests run but actually I don't know if are used around by someone also if are our internal services.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/TestRunAttachmentsProcessingManager.cs
```diff
@@ -8,63 +8,64 @@
...
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
 
 namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestRunAttachmentsProcessing
 {
     /// <summary>
     /// Orchestrates test run attachments processing operations.
     /// </summary>
-    public class TestRunAttachmentsProcessingManager : ITestRunAttachmentsProcessingManager
+    internal class TestRunAttachmentsProcessingManager : ITestRunAttachmentsProcessingManager
```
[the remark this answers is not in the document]
MarcoRossignoli: Like the above, it's internal implementation not used around because attachment processor is new and until this PR is not possible for users register an attachment processor, but I agree that we can start to cleanup public api with more numbers in follow-up PRs

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
```diff
@@ -158,25 +158,48 @@ private void AddKnownExtensions(ref IEnumerable<string> extensionPaths)
...
+                var typesToLoad = TypesToLoadUtilities.GetTypesToLoad(assembly);
+                if (typesToLoad?.Any() == true)
+                {
+                    types.AddRange(typesToLoad);
+                }
 
                 if (!types.Any())
                 {
-                    types = assembly.GetTypes().Where(type => type.GetTypeInfo().IsClass && !type.GetTypeInfo().IsAbstract);
+                    types.AddRange(assembly.GetTypes().Where(type => type.GetTypeInfo().IsClass && !type.GetTypeInfo().IsAbstract));
```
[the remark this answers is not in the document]
MarcoRossignoli: Yep, this update is needed for the new "versioning" attribute.

src/Microsoft.TestPlatform.Common/Utilities/MetadataReaderHelper.cs
```diff
@@ -0,0 +1,190 @@
...
+    using System.Reflection;
+    using System.Reflection.Metadata;
+    using System.Reflection.PortableExecutable;
+    using System.Text;
+
+    using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+
+    /* Expected attribute shape
+
+        namespace Microsoft.VisualStudio.TestPlatform
```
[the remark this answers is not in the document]
MarcoRossignoli: I took a deeper look and the namespace is needed because we're searching for it , so I think it's good have a copy/paste sample.

```cs
private const string TestExtensionTypesAttributeV2 = "Microsoft.VisualStudio.TestPlatform.TestExtensionTypesV2Attribute";
```

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/TestRunAttachmentsProcessingManager.cs
```diff
@@ -91,52 +92,87 @@ private async Task<Collection<AttachmentSet>> InternalProcessTestRunAttachmentsA
             {
                 EqtTrace.Error("TestRunAttachmentsProcessingManager: Exception in ProcessTestRunAttachmentsAsync: " + e);
 
-                eventHandler?.HandleLogMessage(TestMessageLevel.Error, e.Message);
+                eventHandler?.HandleLogMessage(TestMessageLevel.Error, "TestRunAttachmentsProcessingManager: Exception in ProcessTestRunAttachmentsAsync: " + e);
```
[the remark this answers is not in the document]
MarcoRossignoli: Moved back to `e.Message` and aligned the other exception too so we can avoid translation here, we have the logs above also in case of issue and the message of exception will be already translated.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/TestRunAttachmentsProcessingManager.cs
```diff
@@ -91,52 +92,87 @@ private async Task<Collection<AttachmentSet>> InternalProcessTestRunAttachmentsA
...
+                                    attachments.Add(attachment);
+                                }
+                            }
                         }
                     }
                 }
+                catch (Exception e)
+                {
+                    EqtTrace.Error("TestRunAttachmentsProcessingManager: Exception in ProcessAttachmentsAsync: " + e);
+                    logger.SendMessage(TestMessageLevel.Error, "TestRunAttachmentsProcessingManager: Exception in ProcessAttachmentsAsync: " + e);
```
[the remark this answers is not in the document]
MarcoRossignoli: moved to `e.Message`

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
```diff
@@ -157,6 +166,17 @@ public Exception GetAggregatedException()
                 if (exception != null) Exceptions.Add(exception);
                 if (executorUris != null) this.executorUris.AddRange(executorUris);
                 if (testRunStats != null) testRunStatsList.Add(testRunStats);
+
+                if (invokedDataCollectors != null && invokedDataCollectors.Count > 0)
```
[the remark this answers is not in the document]
MarcoRossignoli: I took a deeper look, it's risky here, I prefer to keep as-is, if in future someone add code below and lose that return could miss his code if tests are not good(could return before the new code). But I'll improve with `invokedDataCollectors?.Count > 0`

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/net/PublicAPI.Shipped.txt
```diff
@@ -68,7 +68,7 @@ Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.IDataCollectionSi
...
-Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.IDataCollectorAttachmentProcessor.ProcessAttachmentSetsAsync(System.Collections.Generic.ICollection<Microsoft.VisualStudio.TestPlatform.ObjectModel.AttachmentSet> attachments, System.IProgress<int> progressReporter, Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging.IMessageLogger logger, System.Threading.CancellationToken cancellationToken) -> System.Threading.Tasks.Task<System.Collections.Generic.ICollection<Microsoft.VisualStudio.TestPlatform.ObjectModel.AttachmentSet>>
+Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.IDataCollectorAttachmentProcessor.ProcessAttachmentSetsAsync(System.Xml.XmlElement configurationElement, System.Collections.Generic.ICollection<Microsoft.VisualStudio.TestPlatform.ObjectModel.AttachmentSet> attachments, System.IProgress<int> progressReporter, Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging.IMessageLogger logger, System.Threading.CancellationToken cancellationToken) -> System.Threading.Tasks.Task<System.Collections.Generic.ICollection<Microsoft.VisualStudio.TestPlatform.ObjectModel.AttachmentSet>>
```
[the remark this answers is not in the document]
MarcoRossignoli: Discussed off-line, we'll break current api because is not used today by anyone.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
AbhitejJohn: Would be nice if we could add that detail to the comment as well.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
AbhitejJohn: How is that different from test adapters though? Dont we refresh the cache based on the adapter version?

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentsProcessorsFactory.cs
AbhitejJohn: So the attachment processor would be coming in from the Code Coverage data collector nuget package eh?

src/Microsoft.TestPlatform.ObjectModel/DataCollector/IConfigurableDataCollectorAttachmentProcessor.cs
AbhitejJohn: yeah, adding additional runsettings data would need to break the interface again, so I was hoping we could just do that once.

src/Microsoft.TestPlatform.Common/Utilities/MetadataReaderHelper.cs
AbhitejJohn: Could we add that as a comment to code too please?

cvpoienaru (approved):

cvpoienaru (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -103,10 +103,6 @@ Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.GetDiscoveryManager() -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.TesthostProtocol.IDiscoveryManager
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.GetExecutionManager() -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.TesthostProtocol.IExecutionManager
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.TestHostManagerFactory(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> void
-Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestRunAttachmentsProcessing.TestRunAttachmentsProcessingManager
```
MarcoRossignoli, line 106: @jakubch1 I saw that this class implements an internal interface and it's an internal detail, I think we can move the implementation to internal too and remove from public surface

jakubch1: sure, let's do it

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -339,14 +343,15 @@ Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestRunChangedEventArgs.N
...
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestRunChangedEventArgs.TestRunStatistics.get -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestRunStatistics
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestRunCompleteEventArgs
+Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestRunCompleteEventArgs.TestRunCompleteEventArgs(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestRunStatistics stats, bool isCanceled, bool isAborted, System.Exception error, System.Collections.ObjectModel.Collection<Microsoft.VisualStudio.TestPlatform.ObjectModel.AttachmentSet> attachmentSets, System.Collections.ObjectModel.Collection<Microsoft.VisualStudio.TestPlatform.ObjectModel.InvokedDataCollector> invokedDataCollectors, System.TimeSpan elapsedTime) -> void
```
nohwnd, line 346: This adds the invokedDataCollectors but does not keep the previous constructor, which is not a backwards compatible change in public API. Is there a good reason we should do it this way?

src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -20,9 +21,10 @@ Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException
 Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException.InvalidLoggerException(string message) -> void
 Microsoft.VisualStudio.TestPlatform.Common.Exceptions.InvalidLoggerException.InvalidLoggerException(string message, System.Exception innerException) -> void
 Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.DataCollectorMetadata
-Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.DataCollectorMetadata.DataCollectorMetadata(string extension, string friendlyName) -> void
+Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.DataCollectorMetadata.DataCollectorMetadata(string extension, string friendlyName, bool hasAttachmentProcessor) -> void
```
nohwnd, line 24: This would break the exitsting public api. Is there a reason why this is okay?

MarcoRossignoli: DataCollectorMetadata is an internal type for the datacollector extensione...I don't expect external usage outside platform. I don't see usage around https://grep.app/search?q=DataCollectorMetadata&regexp=true http://index/?query=DataCollectorMetadata
The way the metadata object is created is by convention(Type, metadata) 

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/LazyExtension.cs#L161

I don't see back-compat issue here, but if you're concerned about it we can keep old ctor and set false there. That value is used to have better performance during attachment processor creation, if hasAttachmentProcessor is false we skip dll loading/inspection.

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Shipped.txt
MarcoRossignoli: I did some test with old SDK and this event is created/used inside runner and serialized/deserialized with translator inside TE.
I see few usage inside UTs 
http://index/?query=TestRunCompleteEventArgs&rightProject=VS(Repo)&file=src%5Cvset%5CAgile%5CTestPlatform%5Csuitesrc%5CCore%5CTDD%5CLogging%5CTestRunCompleteEventArgsBehaviors.cs&rightSymbol=ltzdatsjdorb 
https://grep.app/search?q=TestRunCompleteEventArgs&regexp=true&filter[repo][0]=stryker-mutator/stryker-net

Do you see possible runtime issue?

Anyway as like the below one, I'll add overload, I liked the idea to start to narrow down the public api, but if you feel the risk here it's ok to me.

MarcoRossignoli: Overload added

src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
MarcoRossignoli: Overload added

nohwnd (approved):

PR #3165, opened 2021-11-11
Aggregate api files

nohwnd:
## Description
Use the new feature in the public api analyzer that allows us to add (actually rewrite entries, but we don't use that) in additional files. 
This adds two baseline files, because there is big difference between netstandard1.0 and others. The file in root is common to all tfms, and you should 
try to add your api there. If it is not an api common to all tfms, try adding it to the one in the `net` folder. And only after that fall back to the tfm specific files.


nohwnd:
I used this code to split the files: 

```cs
// See https://aka.ms/new-console-template for more information


using System.Text;

var dir = @"C:\p\vstest\src\Microsoft.TestPlatform.PlatformAbstractions\PublicAPI";

// Find the common baseline
var files = Directory.GetFiles(dir, "*.Shipped.txt", SearchOption.AllDirectories);
List<string>? previousLines = null;
foreach (var file in files)
{
    var lines = File.ReadAllLines(file, Encoding.UTF8).ToList();
    if (!lines.Any())
    {
        continue;
    }
    if (previousLines == null)
    {
        previousLines = lines.ToList();
    }
    else
    {
        previousLines = lines.Intersect(previousLines).ToList();
        Console.WriteLine($"There are {previousLines.Count} common lines - {file}");
    }
}

File.WriteAllLines(@$"{dir}\PublicAPI.Shipped.txt", previousLines, Encoding.UTF8);

// find common baseline for all projects that are not netstandard1.0
// because that is where the biggest difference is (and call it net baseline)
List<string>? netLines = null;
foreach (var file in files.Where(f => !f.Contains("netstandard1.0")))
{
    var lines = File.ReadAllLines(file, Encoding.UTF8).ToList();
    if (!lines.Any())
    {
        continue;
    }
    if (netLines == null)
    {
        netLines =  lines.ToList();
    }
    else
    {
        netLines = lines.Intersect(netLines).Except(previousLines).ToList();
        Console.WriteLine($"There are {netLines.Count} net common lines - {file}");
    }
}

Directory.CreateDirectory(@$"{dir}\net");
File.WriteAllLines(@$"{dir}\net\PublicAPI.Shipped.txt", netLines, Encoding.UTF8);

// Save each file as diff from the baseline + net baseline (or just from baseline if you are netstandard1.0 tfm)
foreach (var file in files)
{
    var lines = File.ReadAllLines(file, Encoding.UTF8).ToList();
    if (!lines.Any())
    {
        continue;
    }

    if (file.Contains("netstandard1.0"))
    {
        lines = lines.Except(previousLines).ToList();
    }
    else
    {
        lines = lines.Except(previousLines).Except(netLines).ToList();
    }

    Console.WriteLine($"There are {lines.Count} framework specific lines - {file}");

    File.WriteAllLines(file, lines, Encoding.UTF8);
}
```

Haplois (approved):

PR #3166, opened 2021-11-11
Make older uwp work

nohwnd:
## Description
Downgrade UWP dependencies so they can be consumed from all supported versions of UWP adapter.

## Related issue
This is pre-requisite for testfx being able to downgrade because it depends on object model: https://github.com/microsoft/testfx/issues/927


nohwnd:
@Haplois could you run this through our official pipeline for me please so I can create a PR on testfx that consumes the offical preview package before committing this to master? I tried it locally and it downgraded to the versions we need, hopefully, but I am not sure so I want to give it to Luke to try first.

fhnaseer (approved):

nohwnd:
Confirmed by Luke to work. Merging so I can update testfx.

PR #3174, opened 2021-11-18
Cleanup tmp code for the --arch feature

MarcoRossignoli:
Cleanup tmp code and remove VSTEST_WINAPPHOST_DOTNET_ROOT from host process environment variables

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
@@ -44,6 +44,10 @@ namespace Microsoft.VisualStudio.TestPlatform.PlatformAbstractions
                     }
                 }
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -44,6 +44,10 @@ public object LaunchProcess(string processPath, string arguments, string working
                     }
                 }
 
+                // VSTEST_WINAPPHOST_DOTNET_ROOT is made to flow by SDK to support apphost and --arch feature
+                // we remove to not pollute host process environment variables
+                process.StartInfo.RemoveEnvironmentVariable("VSTEST_WINAPPHOST_DOTNET_ROOT");
```
nohwnd, line 49: I don't think this is worth adding another public API to VSTest, users won't try to re-use this variable name inside of their process, and having additional env variable has no impact on performance.

MarcoRossignoli:
@nohwnd PTAL

nohwnd (approved):

PR #3192, opened 2021-11-25
Fixing .net framework parallel issue

Sanan07:
## Description
We already changed behaviour and now .net framework can also run in parallel and can have separate testhosts for each assembly.
And we faced issue when `ProxyDiscoveryManager` instance was already closed before sending some request to testhost.
So I removed that condition and now `ProxyDiscoveryManager` instance will assign new object as it was previously only for .netcore.

IsAborted condition is already not mandatory, because we need to execute that part of code any time we come here.

## Related issue

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
@@ -7,7 +7,6 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.Parallel
     using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
-
     using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities;
     using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Interfaces;
     using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.ObjectModel;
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -145,27 +144,27 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
-                {
-                    EqtTrace.Verbose("ParallelProxyDiscoveryManager: HandlePartialDiscoveryComplete: Replace discovery manager. Shared: {0}, Aborted: {1}.", this.SharedHosts, isAborted);
-                }
+                EqtTrace.Verbose("ParallelProxyDiscoveryManager: HandlePartialDiscoveryComplete: Replace discovery manager. Shared: {0}, Aborted: {1}.", this.SharedHosts, isAborted);
+            }
 
-                this.RemoveManager(proxyDiscoveryManager);
+            this.RemoveManager(proxyDiscoveryManager);
 
-                proxyDiscoveryManager = this.CreateNewConcurrentManager();
```
cvpoienaru, line 160: We should probably take advantage of the fact that .net fw testhost is shared and re-use it instead of killing it in the parallel discovery/execution case.

E.g.: `parallelLevel = 4`, .net fw test assmeblies = 8.

What happens now is that we spawn 4 testhosts, discover 4 .net fw test assmeblies, kill the 4 testhosts and then spawn another 4 testhosts to do the same for the remaining 4 test assemblies.

cvpoienaru (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
Sanan07: Actually, after discovery 4 test assemblies, it will spawn next 4 testhosts one by one, not simultaniously as the first time.
So it will spawn 4 testhosts and when some testhost finishes discovering it will start new one for the next source.

As we discussed, I just created new issue for that idea : #3193.
Feel free to add or edit description for the issue.

PR #3197, opened 2021-11-26
Attach to VS automatically

nohwnd:
## Description
Add a tool to attach sub-processes to an instance of Visual Studio, so we can run vstest.console via wrapper and get it attached, or similarly run testhost, and get it automatically attached to VS.

 This tool needs to be built as net472, because netcoreapp2.1 and netcoreapp3.1 don't support the used api. So the same executable is re-used for everything. This has the added benefit of not packing it into the nugets we create.

Adds `VSTEST_RUNNER_DEBUG_ATTACHVS`, `VSTEST_HOST_DEBUG_ATTACHVS`, `VSTEST_DATACOLLECTOR_DEBUG_ATTACHVS` environment variables, which take value 0, 1, and `<number>`. 

0 disables attaching to VS and is the default, 1 enables it and a parent process of Visual Studio is looked up, or the first Visual Studio that started. 

Any number larger than 1 is considered a PID of the process to be inspected, and same logic is used. The current process and all it's parents are checked, and if none of them is VS, then the first VS that started is used. You can use this to define exactly which VS should we attach to.

`VSTEST_DEBUG_NOBP` is also added to prevent stopping on breakpoints after attaching.

The AttachVS tool is located in 3 ways, by looking at `VSTEST_DEBUG_ATTACHVS_PATH` env variable, by searching PATH, and by walking up the directory tree, and looking for special path (this is to locate the tool when you start VS, without and additional env variables).  Our build script also adds the correct path to PATH. This allows you to debug both stuff from artifacts, as well as arbitrary `dotnet test` run.

For general use outside of our repo, put AttachVS to your PATH, or specify `VSTEST_DEBUG_ATTACHVS_PATH`.

The tool can be observed using DebugView++, logs are written to Trace.


src/AttachVS/AttachVS.csproj
```diff
@@ -0,0 +1,16 @@
...
+  <PropertyGroup>
+	<TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
+	<TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
+  </PropertyGroup>
+  <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
+	
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net472</TargetFrameworks>
+	  <LangVersion>preview</LangVersion>
```
MarcoRossignoli, line 11: nit: looks off

src/vstest.console/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
+{
```
MarcoRossignoli, line 1: to remove this one?

src/datacollector/DataCollectorMain.cs
```diff
@@ -63,7 +63,10 @@ internal DataCollectorMain(IProcessHelper processHelper, IEnvironment environmen
 
         public void Run(string[] args)
         {
+            DebuggerBreakpoint.AttachVisualStudioDebugger("VSTEST_DATACOLLECTOR_DEBUG_ATTACHVS");
             DebuggerBreakpoint.WaitForDebugger("VSTEST_DATACOLLECTOR_DEBUG");
+
+
```
MarcoRossignoli, line 69: nit: new line or maybe is on purpose

nohwnd:
> Our build script also adds the correct path to PATH. 

Actually no. Build.cmd is generally used, so the build will run in a sub-process, and won't affect the current session. And to set it to user context we need admin, and I don't want to do that. So just add it to your PATH I guess?

MarcoRossignoli (approved):
LL(ovely)TM

MarcoRossignoli:
It's great, maybe follow-up with docs update

src/vstest.console/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
...
+  "profiles": {
+    "vstest.console": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "VSTEST_RUNNER_DEBUG_ATTACHVS": "1",
```
nohwnd, line 6: Actually this is on purpose, I want to use this in the end to attach to testhost and datacollector when you press debug on vstest.console in VS. But currently it does not work (because the vstest.console we build under VS is not complete and is missing the testhost runtime provider), so this just attaches to vstest.console if you press run without debugging :)
This is more like how it would look later. 
```
        "VSTEST_HOST_DEBUG_ATTACHVS": "1",
        "VSTEST_DATACOLLECTOR_DEBUG_ATTACHVS": "1",
```
I can remove this if you feel strongly about this :)

src/vstest.console/Properties/launchSettings.json
nohwnd: > It's great, maybe follow-up with docs update

I would wait a bit with it, it's useful mostly for our team, and I was using simplified version of this for a while, but not this one, so maybe there will be some more feedback.

src/AttachVS/AttachVS.csproj
nohwnd: not sure why, but lately pressing format in .csproj totally wrecks the file...

src/datacollector/DataCollectorMain.cs
nohwnd: fixed

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+                                    if (pn == pid)
+                                    {
+                                        lpn.Current.GetType().InvokeMember("Attach", BindingFlags.InvokeMethod, null, lpn.Current, null);
+                                        return true;
+                                    }
+                                }
+                            }
+                            catch (COMException ex)
+                            {
+                                Trace($"ComException: Tetrying in 250ms.\n{ex}");
```
cvpoienaru, line 141: nit: You spelled retrying wrong.

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+                    string dn;
+
+                    moniker[0].GetDisplayName(bindCtx, null, out dn);
+
+                    if (dn.StartsWith("!VisualStudio.DTE.") && dn.EndsWith(dteSuffix))
+                    {
+                        object dte, dbg, lps;
+                        runninObjectTable.GetObject(moniker[0], out dte);
+
+                        for (var i = 0; i < 10; i++)
```
cvpoienaru, line 120: It's not clear to me why we do this 10 times ?

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+                    Trace($"Process {process.ProcessName} ({process.Id}) is VS.");
+                }
+                else
+                {
+                    Trace($"Process {process.ProcessName} ({process.Id}) is not VS.");
+                }
+
+                return isVs;
+            }
+            catch
```
cvpoienaru, line 216: Why would the above `try` block throw anything ? Looks like we're only doing a string comparison in there and some tracing and we made sure to check `process` is not null first.

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+        {
+            try
+            {
+                // Parent needs to start before the child, otherwise it might be a different process 
+                // that is just reusing the same PID.
+                if (parent.StartTime <= currentProcess.StartTime)
+                {
+                    return true;
+                }
+                else
```
cvpoienaru, line 232: nit: No need for `else` if you return in the above `if` block.

src/AttachVS/AttachVs.cs
nohwnd: When the com object is busy doing something else, or blocked by modal window, or attaching other window this will return a com error, that the object is not ready and we need to retry, 10 is arbitrary.

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+            {
+                Trace($"Starting with pid '{pid}', and vsPid '{vsPid}'");
+                if (pid == null)
+                {
+                    Trace($"FAIL: Pid is null.");
+                    return false;
+                }
+                var process = Process.GetProcessById(pid.Value);
+                Trace($"Using pid: {pid} to get parent VS.");
+                var vs = GetVsFromPid(vsPid != null
```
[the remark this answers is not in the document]
nohwnd: Right, or `GetVsFromPid(Process.GetProcessById((vsPid ?? process.Id))`;

src/AttachVS/AttachVs.cs
```diff
@@ -0,0 +1,309 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+using System.Threading;
+
+namespace Nohwnd.AttachVS
```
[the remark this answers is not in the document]
nohwnd: Yeah true.

src/AttachVS/AttachVs.cs
nohwnd: You are right. It's probably pointless now.

PR #3200, opened 2021-11-29
Add playground project

nohwnd:
## Description
Add playground project that is setup for one end-to-end flow, that can be debugged under VS directly without rebuilding and re-packing the whole solution all the time. To keep it simple I am using .NET Framework, because that does not need microsoft.net.test.sdk installed.


Haplois (approved):

Haplois (approved):
Nit: I think we should add a Readme file to this folder to document what it does.

LGTM

README.md
```diff
@@ -1,58 +1,17 @@
-### VSTest
```
MarcoRossignoli, line 1: I think this is the main readme right? Should we put the new one inside playground folder?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
```diff
@@ -71,7 +71,7 @@ public string LogFilePath
                 var directoryPath = Path.GetDirectoryName(value);
                 if (!string.IsNullOrEmpty(directoryPath) && !fileHelper.DirectoryExists(Path.GetDirectoryName(value)))
                 {
-                    throw new ArgumentException("LogFilePath must point to a valid directory for logging!");
+                    Directory.CreateDirectory(Path.GetDirectoryName(value));
```
MarcoRossignoli, line 74: maybe use `directoryPath` everywhere for clarity

PR #3205, opened 2021-11-30
Add PublicAPI analyzer to all public projects

nohwnd:
## Description
Add Roslyn PublicAPI analyzer to all our public projects.

## Related issue
Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1271895


nohwnd:
Please check that all Unshipped files are empty (all our public api is already shipped). 

And that I did not forget the explicit reference to the analyzer package in any csproj.

There is no need to review the Shipped files, build fails if there are not all entries, or some extra entries.

src/Microsoft.TestPlatform.CrossPlatEngine/Microsoft.TestPlatform.CrossPlatEngine - Copy.csproj
```diff
@@ -0,0 +1,72 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Sanan07, line 1: Why there is CrossPlatform - Copy project?

nohwnd: It's a mistake when I was copy pasting the empty PublicAPI files. Removing.

src/Microsoft.TestPlatform.Extensions.HtmlLogger/Microsoft.TestPlatform.Extensions.HtmlLogger.csproj
```diff
@@ -44,5 +44,13 @@
     <PropertyGroup Label="Configuration">
         <RootNamespace>Microsoft.VisualStudio.TestPlatform.Extensions.HtmlLogger</RootNamespace>
     </PropertyGroup>
+    <ItemGroup>
+        <!-- API that is common to all frameworks that we build for. -->
+        <AdditionalFiles Include="PublicAPI/PublicAPI.Shipped.txt" />
+        <AdditionalFiles Include="PublicAPI/PublicAPI.Unshipped.txt" />
+        <!-- Framework specific API -->
+        <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Shipped.txt" />
```
Haplois, line 52: Instead of adding empty PublicAPI* files for all targets, should we add `Condition="!Exists('PublicAPI/$(TargetFramework)/PublicAPI.Shipped.txt')"` here remove those files?

Sanan07 (approved):

Haplois (approved):
LGTM with a comment.

src/Microsoft.TestPlatform.Extensions.HtmlLogger/Microsoft.TestPlatform.Extensions.HtmlLogger.csproj
nohwnd: We could but I already did the work of adding the empty files, and while they are not strictly needed, if you would want to add framework specific api you can do it right now without having to decipher the csproj, and adding files. The roslyn analyzer also has a code fix, which will output into the specific file for the tfm. So I would vote for keeping it as is.

nohwnd: **they are not strictly needed** for most of the projects, because most of them (luckily) does not have tfm specific api.

nohwnd:
Acceptance tests were broken in unrelated place, because runtime could not be found.

PR #3230, opened 2021-12-25
Improve AttachVS gem output

MarcoRossignoli:
I found useful have some output to understand if everything is good.

Let me know @nohwnd 

We could ship this gem as nuget+source file with `DebuggerHelper.AttachVisualStudioDebugger(string environmentVariable)`, inner loop for a project that spawns processes would be straightforward to debug(easier than https://devblogs.microsoft.com/devops/introducing-the-child-process-debugging-power-tool/ here we can jump into new repo/vs instance), I'm using it also with TE(I will propose `AttachVisualStudioDebugger` also there for our stuff) and it's a bit "magic" put bp on datacollector process and that's all 💘 💘 💘 

image

### DIFF
--- a/src/AttachVS/AttachVs.cs
+++ b/src/AttachVS/AttachVs.cs
@@ -15,13 +15,13 @@ namespace Microsoft.TestPlatform.AttachVS
         {
             try
             {
-                Trace($"Starting with pid '{pid}', and vsPid '{vsPid}'");
                 if (pid == null)
                 {
                     Trace($"FAIL: Pid is null.");
                     return false;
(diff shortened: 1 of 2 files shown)


nohwnd (approved):

nohwnd:
Sure, I am usually using DebugView++ and the Debug output in VS to observe it, when it does not work as expected. But this might help as well.

nohwnd:
> we could ship this gem as nuget+source file with

Sure, I tried not to overcomplicate it. But there is room for improvement for sure.

PR #3236, opened 2022-01-05
Check exitcodes in build script

nohwnd:
## Description
Check exit codes of executables we call and write the actual call into verbose rather than a copy of the command that is sometimes not updated.

## Related issue

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -146,10 +146,10 @@ function Invoke-TestAssetsBuild
 
     Write-Log ".. .. Build: Source: $TPB_TestAssets_Solution"
     Write-Log ".. .. Build: Source: $TPB_TestAssets_Solution -- add NuGet source"
-    Invoke-Exe -IgnoreExitCode 1 $nugetExe "sources add -Name ""locally-built-testplatform-packages"" -Source ""$env:TP_TESTARTIFACTS\packages\"" -ConfigFile ""$nugetConfig"""
+    Invoke-Exe -IgnoreExitCode 1 $nugetExe "sources add -Name ""locally-built-testplatform-packages"" -Source $env:TP_TESTARTIFACTS\packages\ -ConfigFile ""$nugetConfig"""
     Invoke-Exe $dotnetExe "build $TPB_TestAssets_Solution --configuration $TPB_Configuration -v:minimal -p:CIBuild=$TPB_CIBuild -p:LocalizedBuild=$TPB_LocalizedBuild -bl:""$($env:TP_ROOT_DIR)\TestAssets.binlog"""
     Write-Log ".. .. Build: Source: $TPB_TestAssets_Solution -- remove NuGet source"
-    Invoke-Exe $nugetExe "sources remove -Name ""locally-built-testplatform-packages"" -ConfigFile ""$nugetConfig"""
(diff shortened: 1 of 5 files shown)


scripts/common.lib.ps1
```diff
@@ -51,15 +53,25 @@ $env:DOTNET_CLI_VERSION = $GlobalJson.tools.dotnet
...
+    )
+    
     $currentColor = $Host.UI.RawUI.ForegroundColor
-    $Host.UI.RawUI.ForegroundColor = "Green"
-    if ($message)
-    {
-        Write-Output "... $message"
+    try {
+        $Host.UI.RawUI.ForegroundColor = if ("Success" -eq $Level) { "Green" } else { "Red" }
+        if ($message)
```
cvpoienaru, line 67: Could we maybe do some kind of early return here and basically do nothing, not even change the `ForegroundColor` if the `$message` variable is empty ? Something along the lines of:

```
function Write-Log {
    param (...)

    if (!$message)
    {
        return
    }

    try {
        ...
    }
    finally {
        ...
    }
```

scripts/build.ps1
```diff
@@ -54,6 +54,8 @@ Param(
     [String[]] $Steps = @("InstallDotnet", "Restore", "UpdateLocalization", "Build", "Publish", "PrepareAcceptanceTests")
 )
 
+$ErrorActionPreference = 'Stop'
```
cvpoienaru, line 57: Would there be any use case in which we'd like to have this error action preference configurable ?

cvpoienaru (approved):

scripts/common.lib.ps1
```diff
@@ -203,4 +214,18 @@ function PrintAndExit-OnError([System.String] $output)
...
+}
+
+function Invoke-Exe {
+    param (
+        [Parameter(Mandatory)]
+        [string] $Command,
+        [string] $Arguments,
+        [int[]] $IgnoreExitCode
+    )
+    Write-Verbose "Invoking: $Command $Arguments"
```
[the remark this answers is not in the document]
nohwnd: $Arguments will be defined and will be $null, so it will make no difference to the output, and won't fail even if we were running in strict mode.

scripts/common.lib.ps1
```diff
@@ -51,15 +53,25 @@ $env:DOTNET_CLI_VERSION = $GlobalJson.tools.dotnet
 $env:VSWHERE_VERSION = "2.0.2"
 $env:MSBUILD_VERSION = "15.0"
```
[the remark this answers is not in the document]
nohwnd: IDK I updated the build tools because they were using parameter that is not present in the referenced version, and because the exit code was not checked it just "silently" failed. I don't see that environment variable used anywhere, and if it has significance to msbuild, then it is not breaking, and I don't want to fiddle with this any more than I need to 😁

scripts/common.lib.ps1
nohwnd: We might, but it won't save any time. Thanks for the feedback, but I won't do it now.

scripts/build.ps1
nohwnd: Nope. This is a best practice. The common.lib.ps1 file actually sets this as well. Setting this to Stop, and optionally overriding it to Continue on commands that are allowed to fail is how this is done.

PR #3244, opened 2022-01-11
Multi-request, split messages to multiple handlers.

nohwnd:
## Description
The idea here is that we extend the protocol on message to carry recipient as well, together with message type and content. (Similar to HTTP headers). This tells us which handler is waiting for the message. And on the wrapper side we dispatch the messages to the correct handler. 
This extension will be used only between vstest.console and the wrapper, because that is where we need to respond to multiple requests over the same channel, and will be considered only by the design mode client. Other handlers are safe to ignore it, as long as they play by the 
old rules of only using single request at the time.

This way we avoid changing every message that can be sent. And allow the vstest console wrapper to register multiple handlers that will receive the responses that are only meant for them. 

One "special" case is raw message. Raw messages are used to forward the text of the message over a hop without parsing it. (e.g. message from testhost goes via vstest.console to TestWindowStoreHost). This approach still supports this, it just wraps the message into another Message
and sends it as a non-raw message so it can have a recipient. Technically we could modify the recipient in the raw message, but that would require us to parse the message, which is what we want to avoid.

## Related issue


playground/TestPlatform.Playground/Program.cs
```diff
@@ -49,42 +53,97 @@ static void Main(string[] args)
...
+            foreach (var (source, handler, launcher) in sources)
+            {
+                var options = new TestPlatformOptions { };
+                tasks2.Add(Task.Run(() =>
+                {
+                    r.DiscoverTests(new[] { source }, sourceSettings, options, handler);
+                    var tests = handler.DiscoveredTests;
+                    if (!tests.Any())
+                    {
+                        // REview: the second discovery can get cancelled too early. This is task that will probably be solved by fixing the cancellation in the translation layer sender and how it receives messages.
```
MarcoRossignoli, line 79:
```suggestion
                        // Review: the second discovery can get cancelled too early. This is task that will probably be solved by fixing the cancellation in the translation layer sender and how it receives messages.
```

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeTestEventsRegistrar.cs
```diff
@@ -4,22 +4,21 @@
...
-    /// Registers the discovery and test run events for design mode flow
-    /// </summary>
-    public class DesignModeTestEventsRegistrar : ITestDiscoveryEventsRegistrar, ITestRunEventsRegistrar
+    internal class IdentifiableDesignModeTestEventsRegistrar : ITestDiscoveryEventsRegistrar, ITestRunEventsRegistrar
     {
+        private string testRunId;
         private IDesignModeClient designModeClient;
 
-        public DesignModeTestEventsRegistrar(IDesignModeClient designModeClient)
+        public IdentifiableDesignModeTestEventsRegistrar(IDesignModeClient designModeClient, string testRunId)
```
MarcoRossignoli, line 16: I'd keep `recipient`  for clarity or do you have a reason to change the name?

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
```diff
@@ -28,6 +28,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Client.DesignMode
     using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces;
 
     using CommunicationUtilitiesResources = CommunicationUtilities.Resources.Resources;
+    using System.Linq.Expressions;
```
MarcoRossignoli, line 31: cannot compile with eyes :) is it needed?

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeTestEventsRegistrar.cs
```diff
@@ -4,22 +4,21 @@
...
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
     using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
 
-    /// <summary>
-    /// Registers the discovery and test run events for design mode flow
-    /// </summary>
-    public class DesignModeTestEventsRegistrar : ITestDiscoveryEventsRegistrar, ITestRunEventsRegistrar
+    internal class IdentifiableDesignModeTestEventsRegistrar : ITestDiscoveryEventsRegistrar, ITestRunEventsRegistrar
     {
+        private string testRunId;
```
MarcoRossignoli, line 13: nit: `readonly`, also the other one maybe

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -484,36 +491,50 @@ public void InitializeExtensions(IEnumerable<string> pathToAdditionalExtensions)
                     TestPlatformOptions = options
                 };
 
+                var id = Id.Next().ToString();
+
+                if (!this.sessionEventHandlers.TryAdd(id, eventsHandler))
+                {
+                    throw new InvalidOperationException($"Adding session hancler under unique id {id} did not succeed, this should never happen.");
```
MarcoRossignoli, line 498:
```suggestion
                    throw new InvalidOperationException($"Adding session handler under unique id {id} did not succeed, this should never happen.");
```

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeTestEventsRegistrar.cs
nohwnd: I thought I need to rename this type, but I don't need to do that, so I will rename that back. And the testRunId was a previous design. So I will change that as well.

playground/TestPlatform.Playground/Program.cs
nohwnd: Yeah, that task in the comment needs to be completed before merging.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeTestEventsRegistrar.cs
Evangelink: Happy to enable the related rule as warning after we merge the `.editorconfig` change if we agree on it.

src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
Evangelink: Happy to enable the related rule as warning after we merge the `.editorconfig` change if we agree on it.

PR #3246, opened 2022-01-11
Fixed manifest build and publishing 

Haplois:
- Fixed a small issue in `Invoke-Exe`
 - Fixed a trailing backslash issue in manifest publishing


scripts/common.lib.ps1
```diff
@@ -220,12 +222,148 @@ function Invoke-Exe {
...
+        [switch] $CaptureOutput
     )
-    Write-Verbose "Invoking: $Command $Arguments"
-    & $Command ($Arguments -split ' ')
-    if ($IgnoreExitCode -notcontains $LASTEXITCODE) {
+    Write-Verbose "Invoking: > $Command $Arguments"
+    Write-Verbose "          > Ignored exit-codes: 0, $($IgnoreExitCode -join ', ')"
+
+    $workingDirectory = [System.IO.Path]::GetDirectoryName($Command)
+    $process = Start-InlineProcess -Path $Command -Arguments "$Arguments" -WorkingDirectory $workingDirectory -SuppressOutput:$CaptureOutput.IsPresent
```
nohwnd, line 233: Why do you need to split the arguments into array of strings in all the Invoke-Exe calls when you are then joining it to a single string and passing that to Start-InlineProcess?

Haplois: I forgot to revert that change, thank for the heads up!

nohwnd (approved):

PR #3264, opened 2022-01-18
Apply modern code styles

Evangelink:
## Description

Allows to use latest C# syntax across the codebase.

Fix suggestions about (their default severity is and remains as info, let me know if you prefer warning):
- collection initialization
- object initialization

Set simplify using rule severity to warning and apply code fix.


.editorconfig
```diff
@@ -0,0 +1,4 @@
+﻿[*.cs]
```
Evangelink, line 1: If we agree on some styles, we could start defining expectations in this file (force empty line at the end of the file, number of tabs/spaces per file type...).

scripts/build/TestPlatform.Settings.targets
```diff
@@ -6,6 +6,7 @@
     at the start of new iteration to the goal number. This is also used to version the local packages. This version needs to be statically
     readable when we read the file as xml, don't move it to a .props file, unless you change the build server process -->
     <TPVersionPrefix>17.2.0</TPVersionPrefix>
+    <LangVersion>latest</LangVersion>
```
Evangelink, line 9: We could either stick with a specific version of C#, use `latest` or even use `preview`.

.editorconfig
nohwnd: We want to use the standard .editorconfig (dotnet new editorconfig), but roslyn/dotnet format cannot fix naming rules for all. I did not look much further into that, but there was no simple solution to apply all rules to all files, unfortunately.

scripts/build/TestPlatform.Settings.targets
nohwnd: I vote for `preview`.

test/Microsoft.TestPlatform.TestUtilities/TestPluginCacheHelper.cs
```diff
@@ -54,8 +54,10 @@ public static void SetupMockExtensions(string[] extensions, Action callback, Moc
 
             mockFileHelper.Setup(fh => fh.DirectoryExists(It.IsAny<string>())).Returns(true);
 
-            var testableTestPluginCache = new TestableTestPluginCache(extensions.ToList());
-            testableTestPluginCache.Action = callback;
+            var testableTestPluginCache = new TestableTestPluginCache(extensions.ToList())
```
nohwnd, line 57: Using object initializer is IMHO less readable than the original code. And populating multiple members in initializer leads to errors that are harder to debug than if object initializer is not used. This would probably have benefit only if we used `init;`.

.editorconfig
Evangelink: A big part of the styling rules now have code fixes associated, might be worth trying again at some point (happy to have a look anyways).

scripts/build/TestPlatform.Settings.targets
MarcoRossignoli: are we sure that `preview` won't emit possible bugged IL that we ship?

MarcoRossignoli: https://github.com/dotnet/runtime/blob/main/Directory.Build.props#L284-L287 runtime repo does some distinction but for cs is preview...but I don't know if there's some override inside csproj, we could take a fast look to the most used lib(HttpClient ie).

Evangelink: > are we sure that `preview` won't emit possible bugged IL that we ship?

I can't say 100% sure but the same config you are mentioning is set for `roslyn` and `roslyn-analyzers` so it feels relatively safe to me:
https://github.com/dotnet/roslyn-analyzers/blob/main/Directory.Build.props#L21-L22
https://github.com/dotnet/roslyn/blob/main/eng/targets/Settings.props#L198
https://github.com/dotnet/roslyn/blob/main/eng/targets/Settings.props#L176

MarcoRossignoli: Ok let's go with it so

.editorconfig
nohwnd: They even had when I looked into that last time, it was only 2 weeks ago 😁 

But some of the fixes cannot be applied in bulk (e.g. naming style violation we get for every private field because we don't prefix them with `_` as the default style dictates), this is a limitation of dotnet format / Roslyn as mentioned here: https://github.com/dotnet/format/issues/348#issuecomment-767044552

dotnet format also has trouble formatting our solution. It fails with: 

```
Unhandled exception: System.Exception: Restore operation failed.
   at Microsoft.CodeAnalysis.Tools.CodeFormatter.OpenMSBuildWorkspaceAsync
```
And the fixes end up with a lot of code being commented out in "unmerged changes" blocks, which prevents the codebase from building.

I am now looking into ways of doing the formatting automatically, so we can use the default editor config, and also automatically reformat PRs. Because there won't ever be the right time to merge otherwise :)

nohwnd (approved):

PR #3268, opened 2022-01-20
Enable parallelization of acceptance tests

Evangelink:
## Description

Update acceptance tests to enable parallelization.

## Related issue

Fixes [AB#1463179](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1463179)


test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
```diff
@@ -8,27 +8,19 @@ namespace Microsoft.TestPlatform.TestUtilities
...
     public class Workspace : IDisposable
     {
-        public Workspace(string path)
+        /// <summary>
+        /// Creates a workspace object around the given directory path. Pass null to use a temp unique directory.
+        /// </summary>
+        public Workspace(string path = null)
         {
-            Path = path;
+            Path = path ?? IntegrationTestBase.GetResultsDirectory();
```
Evangelink, line 16: It's simpler than having most callers passing the same arg.

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
```diff
@@ -167,7 +166,7 @@ public void DiscoverTestsUsingSourceNavigation(RunnerInfo runnerInfo)
             }
             else
             {
-                Assert.AreEqual(22, testCase.FirstOrDefault().LineNumber);
+                Assert.AreEqual(24, testCase.FirstOrDefault().LineNumber);
```
Evangelink, line 169: I have added pragma around 2 usings hence this +2.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -54,13 +66,13 @@ private static bool IsFilesContentEqual(string filePath1, string filePath2)
...
             var content2 = File.ReadAllText(filePath2);
-            Assert.IsTrue(string.Equals(content1, content2, StringComparison.Ordinal), "Content miss match file1 content:{2}{0}{2} file2 content:{2}{1}{2}", content1, content2, Environment.NewLine);
+            Assert.IsTrue(string.Equals(content1, content2, StringComparison.Ordinal), "Content mismatch{2}- file1 content:{2}{0}{2}- file2 content:{2}{1}{2}", content1, content2, Environment.NewLine);
             return string.Equals(content1, content2, StringComparison.Ordinal);
         }
 
         private string GetInProcDataCollectionRunsettingsFile(bool disableAppDomain)
         {
-            var runSettings = Path.Combine(Path.GetTempPath(), "test_" + Guid.NewGuid() + ".runsettings");
+            var runSettings = Path.Combine(GetTempPath(), "test_" + Guid.NewGuid() + ".runsettings");
```
nohwnd, line 75: Possible improvement would be adding BannedApiAnalyzer to this project and banning System.IO.Path.GetTempPath.

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
```diff
@@ -60,17 +47,23 @@ public void BlameDataCollectorShouldGiveCorrectTestCaseName(RunnerInfo runnerInf
         [NetCoreTargetFrameworkDataSource]
         public void BlameDataCollectorShouldOutputDumpFile(RunnerInfo runnerInfo)
         {
-            Environment.SetEnvironmentVariable("PROCDUMP_PATH", Path.Combine(this.testEnvironment.PackageDirectory, @"procdump\0.0.1\bin"));
```
nohwnd, line 63: Probably banning this call as well would be a good idea.

test/Microsoft.TestPlatform.AcceptanceTests/ExecutionTests.cs
```diff
@@ -55,39 +54,30 @@ public void RunMultipleTestAssembliesWithoutTestAdapterPath(RunnerInfo runnerInf
...
             // based on the target framework
-            int expectedNumOfProcessCreated = this.IsDesktopRunner() ? 2 : 3;
-            var testhostProcessNames = new[] { "testhost.x86", "dotnet" };
-
-            var cts = new CancellationTokenSource();
-            var numOfProcessCreatedTask = NumberOfProcessLaunchedUtility.NumberOfProcessCreated(
-                cts,
-                testhostProcessNames);
+            int expectedNumOfProcessCreated =
+                this.testEnvironment.TargetFramework == IntegrationTestBase.CoreRunnerFramework ? 1 : 2;
```
nohwnd, line 73: This seems odd. Why would the .net framework have more processes running? I would expect to have the same number as we have dlls, or less for the .net framework, but not the other way around.

test/Microsoft.TestPlatform.AcceptanceTests/TestPlatformNugetPackageTests.cs
```diff
@@ -18,11 +18,11 @@ public class TestPlatformNugetPackageTests : CodeCoverageAcceptanceTestBase
         private string resultsDirectory;
 
         [ClassInitialize]
-        public static void ClassInit(TestContext testContext)
+        public static void ClassInit(TestContext _)
         {
             var packageLocation = Path.Combine(IntegrationTestEnvironment.TestPlatformRootDirectory, "artifacts", IntegrationTestEnvironment.BuildConfiguration, "packages");
             var nugetPackage = Directory.EnumerateFiles(packageLocation, "Microsoft.TestPlatform.*.nupkg").OrderBy(a => a).FirstOrDefault();
-            nugetPackageFolder = Path.Combine(packageLocation, Path.GetFileNameWithoutExtension(nugetPackage));
+            nugetPackageFolder = Path.Combine(GetResultsDirectory(), Path.GetFileNameWithoutExtension(nugetPackage));
```
nohwnd, line 25: In most places you are using workspace.Path instead of this, is this correct?

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/CodeCoverageTests.cs
```diff
@@ -22,21 +22,21 @@ namespace Microsoft.TestPlatform.AcceptanceTests.TranslationLayerTests
     [TestCategory("Windows-Review")]
     public class CodeCoverageTests : CodeCoverageAcceptanceTestBase
     {
-        private IVsTestConsoleWrapper vstestConsoleWrapper;
+        private TestConsoleWrapperContext wrapperContext;
```
nohwnd, line 25: Not sure about this name, context suggests something that we would insert into the class being used (the vstestconsole wrapper) to share some context with it. This seems more like a "container" or something. 

I don't see the additional properties of the `TestConsoleWrapperContext` used below. Maybe this field could remain IVSTestConsoleWrapper by grabbing the actual wrapper from the object, and all the additional changes in this class could be rolled back? (VSCode has nice "Revert changes in selected range" feature to do that when you right click selected text in a file diff).

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
nohwnd, line 169: Why? Did the file change?

test/TestAssets/SimpleDataCollector/Class1.cs
```diff
@@ -40,7 +40,7 @@ public void TestSessionStart(TestSessionStartArgs testSessionStartArgs)
             Console.WriteLine(testSessionStartArgs.Configuration);
             File.WriteAllText(this.fileName, "TestSessionStart : " + testSessionStartArgs.Configuration + "\r\n");
 #if NETFRAMEWORK
-            var appDomainFilePath = Path.Combine(Path.GetTempPath(), "appdomain_datacollector.txt");
+            var appDomainFilePath = Environment.GetEnvironmentVariable("AppDomainDataCollectorFilePath");
```
nohwnd, line 43: This should probably be consistent with `Environment.GetEnvironmentVariable("SampleDataCollectorTempPath") ?? Path.GetTempPath();` and use the temp path as fallback. This would help when you have to debug the test itself in VS.

test/TestAssets/SimpleTestProject/UnitTest1.cs
```diff
@@ -33,7 +35,7 @@ public void FailingTest()
         {
 #if NETFRAMEWORK
             // current App domain should be write to file to test DisableAppDomain acceptance test.
-            var appDomainFilePath = Path.Combine(Path.GetTempPath(), "appdomain_test.txt");
+            var appDomainFilePath = Environment.GetEnvironmentVariable("AppDomainTestFilePath");
```
nohwnd, line 38: Same, add fallback to temp path.

test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
```diff
@@ -8,27 +8,19 @@ namespace Microsoft.TestPlatform.TestUtilities
 {
     public class Workspace : IDisposable
```
nohwnd, line 9: I know there is no change, but imho calling this TempDirectory and giving it a doc comment that says it represents a dictionary in temporary path that attempts to delete itself once disposed would be nice. So we know what it is when we hover over it.

nohwnd (commented):
Nice. I think we jumped too ahead of ourselves with the WrapperContext, the additional property is used in only 1 or two tests, and we initialize it in the constructor, so I would prefer having the tests use the IVSTestConsoleWrapper directly, to minimize the changes. And where it is needed you can add additional property to the test class that would hold the additional info. Maybe GetVstestConsoleWrapper could return out parameter that would have an object holding that info?

test/Microsoft.TestPlatform.AcceptanceTests/TestPlatformNugetPackageTests.cs
Evangelink: Correct. I forgot to add a comment suggesting if I could rename `GetResultsDirectory` into something like `CreateUniqueDirectory` which makes more sense for what it does.

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/DiscoverTests.cs
Evangelink: See comment above (https://github.com/microsoft/vstest/pull/3268/files#r788654002). I forgot to change the part for the release (will push the fix soon).

test/TestAssets/SimpleDataCollector/Class1.cs
Evangelink: I am actually wondering whether we should fail when the path is not set?

src/vstest.console/Publisher/TextFileTelemetryPublisher.cs
```diff
@@ -51,7 +51,8 @@ public void Dispose()
         /// </param>
         internal void LogToFile(string eventName, IDictionary<string, object> metrics, IFileHelper fileHelper)
         {
-            string resultDirectory = Path.GetTempPath() + "TelemetryLogs";
+            string resultDirectory = Environment.GetEnvironmentVariable("VSTEST_LOGTELEMETRY_PATH")
```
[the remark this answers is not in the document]
MarcoRossignoli: No familiar with this part is it used this class only for testing purpose? cc: @nohwnd

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -24,28 +25,39 @@ public void RunTestExecutionWithDisableAppDomain(RunnerInfo runnerInfo)
         {
             AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
 
-            var testResults = GetResultsDirectory();
-            var testAppDomainDetailFileName = Path.Combine(Path.GetTempPath(), "appdomain_test.txt");
-            var dataCollectorAppDomainDetailFileName = Path.Combine(Path.GetTempPath(), "appdomain_datacollector.txt");
+            using var workspace = new Workspace();
```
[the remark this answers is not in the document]
MarcoRossignoli: credits to @jakubch1 😃

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -24,28 +25,39 @@ public void RunTestExecutionWithDisableAppDomain(RunnerInfo runnerInfo)
...
                 runsettingsFilePath,
                 this.FrameworkArgValue,
                 runnerInfo.InIsolationValue,
-                testResults);
+                workspace.Path);
+
+            // Sets the environment variables used by the test project and test data collector.
+            var env = new Dictionary<string, string>
+            {
+                ["AppDomainTestFilePath"] = testAppDomainDetailFileName,
```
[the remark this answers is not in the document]
MarcoRossignoli: I like uppercase _ separated for env var, won't oppose to different choice.

test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
Evangelink: I think workspace is quite generic and is common name but I am fine either way. 

@MarcoRossignoli What's your opinion?

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
MarcoRossignoli: cool I vote for it

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
```diff
@@ -749,17 +748,40 @@ protected string BuildMultipleAssemblyPath(params string[] assetNames)
             return string.Join(" ", assertFullPaths);
         }
 
+        protected static string GetDiagArg(string rootDir)
```
Evangelink, line 751: Most of these helpers that don't rely on the class state could potentially be moved to a `TestHelper` or `TestUtils` class as it feels a little weird to see/read `IntegrationTestBase.XXX`.

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
```diff
@@ -290,14 +311,14 @@ private void ValidateDump(int expectedDumpCount = 1)
             }
         }
 
-        private void VaildateOutput(string testName, bool validateDumpFile = false)
+        private void VaildateOutput(Workspace workspace, string testName, bool validateDumpFile = false)
```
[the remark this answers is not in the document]
MarcoRossignoli: if we keep `Workspace` most likely future tests writer will use workspace. I bet on dev laziness 😄

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
Evangelink: Done.

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
Evangelink: Done.

test/Microsoft.TestPlatform.AcceptanceTests/DataCollectionTests.cs
```diff
@@ -141,10 +143,16 @@ public void DataCollectorAttachmentProcessor(RunnerInfo runnerInfo)
...
+            {
+                ["SampleDataCollectorTempPath"] = workspace.Path,
+            };
+
+            this.InvokeVsTest(arguments, env);
+
             this.ValidateSummaryStatus(2, 2, 2);
 
-            string mergedFile = Directory.GetFiles(resultsDir, "MergedFile.txt", SearchOption.AllDirectories).Single();
+            string mergedFile = Directory.GetFiles(workspace.Path, "MergedFile.txt", SearchOption.AllDirectories).Single();
```
MarcoRossignoli, line 155: we can also think to add GetFiles to workspace...it's a common task search for files.

test/Microsoft.TestPlatform.AcceptanceTests/Properties/AssemblyInfo.cs
```diff
@@ -19,3 +20,6 @@
 
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 [assembly: Guid("755996fa-672a-4272-9776-7f707a520058")]
+
+// Enable IAP at method level with as many threads as possible based on CPU and core count.
+[assembly: Parallelize(Workers = 0, Scope = ExecutionScope.MethodLevel)]
```
MarcoRossignoli, line 25: 💘

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
```diff
@@ -573,63 +554,17 @@ private string GetRunsettingsFilePath(Dictionary<string, string> runConfiguratio
...
 
             //cleanup
             if (!string.IsNullOrWhiteSpace(runsettingsPath))
             {
                 File.Delete(runsettingsPath);
             }
-            TryRemoveDirectory(resultsDir);
-        }
-
-        private int GetExpectedNumOfProcessCreatedForWithoutParallel()
```
MarcoRossignoli, line 600: I don't know if I would remove this one...looks a good "piece of documentation"

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
```diff
@@ -214,12 +211,6 @@ public void RunSettingsWithParallelAndPlatformX64(RunnerInfo runnerInfo)
             var targetPlatform = "x64";
             var testhostProcessName = new[] { "testhost", "dotnet" };
             var expectedProcessCreated = 2;
-            if (!this.IsDesktopRunner())
-            {
-                // this creates dotnet hosted vstest console and 2 testhosts one of which is hosted 
-                // in dotnet, so we have two dotnet + 1 testhost.exe
-                expectedProcessCreated = 3;
```
MarcoRossignoli, line 221: is it ok to remove?

test/Microsoft.TestPlatform.AcceptanceTests/TranslationLayerTests/CodeCoverageTests.cs
MarcoRossignoli: Agree

src/vstest.console/Publisher/TextFileTelemetryPublisher.cs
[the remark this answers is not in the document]
nohwnd: This is for testing and debugging when you develop telemetry / do telemetry overview.

test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
MarcoRossignoli: Not a strong opinion, my 2c I think that `Workspace` is one layer above the directory concept...workspace contains also some storage but can contains also other stuff.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -24,28 +25,39 @@ public void RunTestExecutionWithDisableAppDomain(RunnerInfo runnerInfo)
...
                 this.GetSampleTestAssembly(),
                 this.GetTestAdapterPath(),
                 runsettingsFilePath,
                 this.FrameworkArgValue,
                 runnerInfo.InIsolationValue,
-                testResults);
+                workspace.Path);
+
+            // Sets the environment variables used by the test project and test data collector.
+            var env = new Dictionary<string, string>
```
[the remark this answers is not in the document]
nohwnd: This looks fine to me. Normal ProcessStartInfo also uses dictionary.

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
Evangelink: It's useful info but we are no longer looking at the processes to find how many we started, instead we ask each process to write down some logs and then we count them. Main difference is that we used to count some processes "wrongly" depending on whether they were run in process or not.

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
MarcoRossignoli: we should have already some inheritance we can add one more on top of `IntegrationTestBase` 
```cs
 public class AcceptanceTestBase : IntegrationTestBase
    {
```

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
nohwnd: I would also prefer SCREAMING_SNAKE_CASE for environment variables. Possibly prefixed with something like `TEST_` or `PARENT_`.

src/vstest.console/Publisher/TextFileTelemetryPublisher.cs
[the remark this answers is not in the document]
MarcoRossignoli: thx it's ok so with env var.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
Evangelink: I was more talking about the magic env var string not the dictionary per say (i.e. instead of having the various places using "AppDomainTestFilePath" we could have a class with a static/const string).

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
nohwnd: Weird I've came up with the same idea without seeing his code. Great minds think alike?:D I want some credit as well 😁

test/Microsoft.TestPlatform.AcceptanceTests/TestPlatformNugetPackageTests.cs
nohwnd: Yes that sounds like a better name.

test/TestAssets/SimpleDataCollector/Class1.cs
nohwnd: I was considering that as well, but if the test fails for some weird reason (e.g. it seems like the task that the test should do did not happen at all), it's easier to just open the test solution and run the test there interactively to debug it. In such case the value would be empty and this would throw.

test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
nohwnd: I was thinking in the same direction as you Marco. Workspace simply seemed like a distracting concept that I should grasp, rather than just a helper for creating and cleaning temp directories.

MarcoRossignoli: (I like a lot the way we're doing inside code coverage repo and used also around, we have a way to create an solution/project object model add source files update project file restore/build/run etc...everything inside that "workspace". Looks more powerful that use test assets build to me.)

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
nohwnd: I don't want to gold plate this too much. But it seems that the getResultsDirectory and TryRemoveDirectory could become part of the `Workspace` class. And it looks like a good fit especially if we renamed it to TempDirectory, and changed the tests that currently grab a random temp path as a string, and then later try to clean that up (either in test body - which will fail if test fails, or in some cases in cleanup - which we would just replace with a call on a workspace / tempdirectory instance). 

But none of this seems necessary to do in this PR.

test/Microsoft.TestPlatform.AcceptanceTests/DataCollectionTests.cs
nohwnd: Yup, I had that in my original proposal as well. So maybe as a second step?

test/Microsoft.TestPlatform.AcceptanceTests/RunsettingsTests.cs
nohwnd: Yes, this was a special case for the thing described in the comment. We no longer count processes by observing which process started during run. But rather by counting the number of diagnostic logs for host that were produced during run. This makes it possible to run in parallel, and is less fragile in case where vstest.console and testhost both run under dotnet.exe apphost (which is the special case above where we have 2 dlls, running in 2 dotnet testhost.dll, but also 1 dotnet vstest.console.dll, and we could not distinguish which dotnet.exe is actually hosting testhost, and which is hosting vstest.console, so we counted 3 in total).

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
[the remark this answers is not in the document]
Evangelink: I am assuming you mean that's better, right?

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
Evangelink: Renamed with `TEST_ASSET_` prefix.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
nohwnd: I'd say duplicate the code. Tests should not be DRY.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -54,13 +66,13 @@ private static bool IsFilesContentEqual(string filePath1, string filePath2)
...
             var content1 = File.ReadAllText(filePath1);
             var content2 = File.ReadAllText(filePath2);
-            Assert.IsTrue(string.Equals(content1, content2, StringComparison.Ordinal), "Content miss match file1 content:{2}{0}{2} file2 content:{2}{1}{2}", content1, content2, Environment.NewLine);
+            Assert.IsTrue(string.Equals(content1, content2, StringComparison.Ordinal), "Content mismatch{2}- file1 content:{2}{0}{2}- file2 content:{2}{1}{2}", content1, content2, Environment.NewLine);
             return string.Equals(content1, content2, StringComparison.Ordinal);
         }
 
         private string GetInProcDataCollectionRunsettingsFile(bool disableAppDomain)
         {
-            var runSettings = Path.Combine(Path.GetTempPath(), "test_" + Guid.NewGuid() + ".runsettings");
```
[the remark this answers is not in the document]
nohwnd: Not sure where my comment about using banned api analyzer went. But we might also want to use banned api analyer for System.IO.Path.GetTempPath and for System.Environment.SetEnvironmentVariable.

test/Microsoft.TestPlatform.TestUtilities/Workspace.cs
nohwnd: That is next step. Right now we need to get our tests running in parallel. And then we can look at tightening the developer loop.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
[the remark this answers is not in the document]
Evangelink: That's done, thx.

test/Microsoft.TestPlatform.AcceptanceTests/ExecutionTests.cs
nohwnd: If you look at the logs in standard output in Test Explorer you will see that it outputs two .host. files every time. You are then checking the line to match testhost and dotnet, but the line is testhost.exe or testhost.dll or testhost.x86.dll, so the core runner is not matching because you are trying to match `dotnet`. 

Looking at how this is used, there is potential false positive in the cases where `testhost` is matched because that is a substring of `testhost.x86`. The `Contains` check needs to be improved.

test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -25,28 +25,28 @@ public void RunTestExecutionWithDisableAppDomain(RunnerInfo runnerInfo)
         {
             AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
 
-            using var workspace = new Workspace();
-            var testAppDomainDetailFileName = Path.Combine(workspace.Path, "appdomain_test.txt");
-            var dataCollectorAppDomainDetailFileName = Path.Combine(workspace.Path, "appdomain_datacollector.txt");
+            using var workingDir = new TempDirectory();
```
nohwnd, line 28: Why not name both tempDirectory?

test/Microsoft.TestPlatform.AcceptanceTests/ArgumentProcessorTests.cs
```diff
@@ -39,8 +39,8 @@ public void PassingInvalidArgumentsToVsTestConsoleShouldNotPrintHelpMessage(Runn
         {
             AcceptanceTestBase.SetTestEnvironment(this.testEnvironment, runnerInfo);
 
-            using var workspace = new Workspace();
-            var arguments = PrepareArguments(this.GetSampleTestAssembly(), this.GetTestAdapterPath(), string.Empty, this.FrameworkArgValue, resultsDirectory: workspace.Path);
+            using var workingDir = new TempDirectory();
```
nohwnd, line 42: Same here, and probably further below as well.

test/Microsoft.TestPlatform.AcceptanceTests/DotnetArchitectureSwitchTests.cs
```diff
@@ -36,7 +37,7 @@ public static void ClassCleanup()
         {
             try
             {
-                Directory.Delete(privateX64Installation, true);
+                new DirectoryInfo(privateX64Installation).Parent.Delete(true);
```
nohwnd, line 40: Why Parent?

test/Microsoft.TestPlatform.AcceptanceTests/BannedSymbols.txt
```diff
@@ -0,0 +1,3 @@
+﻿M:System.IO.Path.GetTempPath(); Use 'IntegrationTestBase.GetTempPath()' instead
+M:System.Environment.SetEnvironmentVariable(System.String,System.String); Use one of the overload accepting a dictionary of environment variable instead
+M:System.Environment.SetEnvironmentVariable(System.String,System.String,System.EnvironmentVariableTarget); Use one of the overload accepting a dictionary of environment variable instead
```
nohwnd, line 3:
```suggestion
﻿M:System.IO.Path.GetTempPath(); Use 'IntegrationTestBase.GetTempPath()' instead
M:System.Environment.SetEnvironmentVariable(System.String,System.String); Use one of the overload accepting a dictionary of environment variables instead
M:System.Environment.SetEnvironmentVariable(System.String,System.String,System.EnvironmentVariableTarget); Use one of the overload accepting a dictionary of environment variables instead
```

test/Microsoft.TestPlatform.AcceptanceTests/DotnetArchitectureSwitchTests.cs
Evangelink: In the init we do a temp dir with a subfolder named `x64` and we keep a reference to the sub-folder, so I want to clean the right folder (i.e. the parent).

nohwnd (approved):

PR #3275, opened 2022-01-25
Add diag log env variable

nohwnd:


nohwnd:
@MarcoRossignoli  what do you think? Should we make it more configurable? (verbosity?) and change the naming some more?

MarcoRossignoli:
Looks good...but yep maybe we should add the verbosity so we can ask to start with Error and after Verbose during the analysis, improving a bit the perf.

nohwnd:
Maybe I should only call it VSTEST_DIAG, to keep the same name as the argument?

MarcoRossignoli:
> Maybe I should only call it VSTEST_DIAG & VSTEST_DIAG_VERBOSITY to keep the same name as the argument?

Yep sorry misread

MarcoRossignoli:
:shipit:

src/vstest.console/CommandLine/Executor.cs
```diff
@@ -114,6 +115,24 @@ internal int Execute(params string[] args)
...
+                {
+                    var verbosity = TraceLevel.Verbose;
+                    if (diagVerbosity != null)
+                    {
+                        if (Enum.TryParse<TraceLevel>(diagVerbosity, ignoreCase: true, out var parsedVerbosity))
+                        {
+                            verbosity = parsedVerbosity;
+                        }
+                    }
+                    args = args.Concat(new[] { $"--diag:{diag};TraceLevel={verbosity}" }).ToArray();
```
nohwnd, line 132: The parameter is actually called TraceLevel, which was a bit surprising to me. Because all the other places I can think about use Verbosity, and imho we should standardize on that? So the env variable name now differs, because I won't be able to remember it is called trace level when I want to use it (and I don't have a place to log it into because logs are not  setup at that time.).

MarcoRossignoli: Agree, I was also convinced about verbosity

MarcoRossignoli (approved):
LGTM

src/vstest.console/CommandLine/Executor.cs
```diff
@@ -114,6 +115,24 @@ internal int Execute(params string[] args)
...
+            if (!isDiag)
+            {
+                var diag = Environment.GetEnvironmentVariable("VSTEST_DIAG");
+                var diagVerbosity = Environment.GetEnvironmentVariable("VSTEST_DIAG_VERBOSITY");
+                if (!string.IsNullOrWhiteSpace(diag))
+                {
+                    var verbosity = TraceLevel.Verbose;
+                    if (diagVerbosity != null)
+                    {
+                        if (Enum.TryParse<TraceLevel>(diagVerbosity, ignoreCase: true, out var parsedVerbosity))
```
Evangelink, line 127: [minor] This could be merged with previous `if`.

src/vstest.console/CommandLine/Executor.cs
```diff
@@ -114,6 +115,24 @@ internal int Execute(params string[] args)
                 exitCode = 1;
             }
 
+            if (!isDiag)
+            {
+                var diag = Environment.GetEnvironmentVariable("VSTEST_DIAG");
```
Evangelink, line 120: Also I am wondering if it would be better to have a file (or files) with env var constants instead of hardcoding them everywhere. It would help to ensure we use the same thing, would help discoverability and also would help to see patterns we have already used.

Evangelink:
@nohwnd I have added some minor comments, feel free to disregard them all.

PR #3277, opened 2022-01-25
Fix encoding when writing updated dependencies in build

nohwnd:
Use the correct encoding when writing updated dependencies in build so we always use utf8 with BOM.


Evangelink (approved):
LGTM

PR #3283, opened 2022-01-27
Log callbacks to delegates better

nohwnd:
## Description
Log what is being called when event is invoked.

Example of the log: TpTrace Verbose: 0 : 20708, 3, 2022/01/27, 13:26:24.691, 2456941827563, vstest.console.exe, MulticastDelegateUtilities.SafeInvoke: InternalTestLoggerEvents.SendTestRunStart: Invoking callback 1/1 for Microsoft.VisualStudio.TestPlatform.CommandLine.Internal.ConsoleLogger.TestRunStartHandler, took 0 ms.

--- 

TpTrace Verbose: 0 : 84508, 7, 2022/01/27, 13:33:05.983, 2460954751207, vstest.console.exe, MulticastDelegateUtilities.SafeInvoke: TestRun.TestRunComplete: Invoking callback 2/2 for Microsoft.VisualStudio.TestPlatform.CommandLine.TestRunResultAggregator.TestRunCompletionHandler, took 0 ms.

TpTrace Verbose: 0 : 84508, 7, 2022/01/27, 13:33:05.983, 2460954753814, vstest.console.exe, MulticastDelegateUtilities.SafeInvoke: TestRun.TestRunComplete: Invoking callback 2/2 for Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.RunTestsArgumentExecutor+TestRunRequestEventsRegistrar.TestRunRequest_OnRunCompletion, took 0 ms.

--- 

TpTrace Verbose: 0 : 62012, 5, 2022/01/27, 13:46:07.014, 2468765066033, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: Polling on remoteEndPoint: 127.0.0.1:1055 localEndPoint: 127.0.0.1:1054
TpTrace Verbose: 0 : 62012, 5, 2022/01/27, 13:46:07.014, 2468765066212, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: NotifyDataAvailable remoteEndPoint: 127.0.0.1:1055 localEndPoint: 127.0.0.1:1054
TpTrace Verbose: 0 : 62012, 5, 2022/01/27, 13:46:07.014, 2468765066380, vstest.console.exe, LengthPrefixCommunicationChannel.NotifyDataAvailable: New data are waiting to be received, but there is no subscriber to be notified. Not reading them from the stream.


src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
```diff
@@ -63,6 +63,13 @@ public Task Send(string data)
         /// <inheritdoc />
         public Task NotifyDataAvailable()
         {
+            // TODO: Review the comment below, because it says something different than what is
```
Evangelink, line 66: Is it a todo for you before the PR gets merged or a later change that needs to be done? If the 2nd, could we create a ticket and link it to the todo?

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
```diff
@@ -138,7 +138,11 @@ public async Task AcceptClientAsync()
         /// <returns>True if Client is connected, false otherwise</returns>
         public bool WaitForClientConnection(int clientConnectionTimeout)
         {
-            return this.clientConnectedEvent.WaitOne(clientConnectionTimeout);
+            Stopwatch stopWatch = Stopwatch.StartNew();
```
Evangelink, line 141: [cosmetic] You could use `var`.

```suggestion
            var stopWatch = Stopwatch.StartNew();
```

Evangelink, line 141: It seems weird to me that we start a timer and never really pause/stop it. Is it expected?

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
```diff
@@ -138,7 +138,11 @@ public async Task AcceptClientAsync()
...
+            var result = this.clientConnectedEvent.WaitOne(clientConnectionTimeout);
+            EqtTrace.Verbose("SocketCommunicationManager.WaitForClientConnection took: {0} ms, with {1} ms timeout, and finished with {2}", stopWatch.ElapsedMilliseconds, clientConnectionTimeout, result);
```
Evangelink, line 143: Nit: We could add a period at the end of the sentence.

```suggestion
            EqtTrace.Verbose("SocketCommunicationManager.WaitForClientConnection took: {0} ms, with {1} ms timeout, and finished with {2}.", stopWatch.ElapsedMilliseconds, clientConnectionTimeout, result);
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
+        /// <summary>
+        /// Invokes each of the subscribers of the event and handles exceptions which are thrown
+        /// ensuring that each handler is invoked even if one throws.
+        /// </summary>
+        /// <param name="delegates">Event handler to invoke.</param>
+        /// <param name="sender">Sender to use when raising the event.</param>
+        /// <param name="args">Arguments to provide.</param>
+        /// <param name="traceDisplayName">Name to use when tracing out errors.</param>
+        // Using [CallerMemberName] for the traceDisplayName is a possibility, but in few places we call through other
+        // methods until we reach here. And it would change the public API.
```
Evangelink, line 41: For this one, you are adding the new method so I don't think that the comment `And it would change the public API.` makes total sense.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
+        public static void SafeInvoke(this Delegate delegates, object sender, object args, string traceDisplayName)
```
Evangelink, line 42: Could we make the method generic instead of taking `object`?

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
         {
             if (args == null)
             {
-                throw new ArgumentNullException(Resources.CannotBeNullOrEmpty, "args");
+                throw new ArgumentNullException(Resources.CannotBeNullOrEmpty, nameof(args));
```
Evangelink, line 46: The message and param name arguments are inverted. I would vote for removing the message as it is incorrect (talks also about being empty) and is "useless" as it's equivalent to the default msg of `ArgumentNullException`.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
+                foreach (Delegate handler in invocationList)
                 {
+                    Stopwatch stopwatch = Stopwatch.StartNew();
+
                     try
                     {
                         handler.DynamicInvoke(sender, args);
+                        if (EqtTrace.IsVerboseEnabled)
+                        {
+                            stopwatch = Stopwatch.StartNew();
```
Evangelink, line 67: I might be missing something but this re-declaration seems off.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
+                        }
                     }
-                    catch (TargetInvocationException e)
+                    catch (TargetInvocationException exception)
                     {
                         if (EqtTrace.IsErrorEnabled)
                         {
                             EqtTrace.Error(
-                                "{0}: Exception occurred while calling handler of type {1} for {2}: {3}",
+                                "MulticastDelegateUtilities.SafeInvoke: {0}: Invoking callback {1}/{2} for {3}.{4}, failed after {5} ms with: {6}",
```
Evangelink, line 82: Nit: Could we add the end of the sentence period?

```suggestion
                                "MulticastDelegateUtilities.SafeInvoke: {0}: Invoking callback {1}/{2} for {3}.{4}, failed after {5} ms with: {6}.",
```

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
```diff
@@ -63,6 +63,13 @@ public Task Send(string data)
...
+            // actually happening, and doing what it suggests would potentially lose messages.
+            // For example in the case where we start testhost process, send it version, and
+            // it responds, we then replace the handler with a new one, and there is quite a long time
+            // (tens of milliseconds) when there is no handler present, which would pump the message
+            // and dump it.
```
MarcoRossignoli, line 71: it will be consumed by poll after right?

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
             {
-                throw new ArgumentException(Resources.CannotBeNullOrEmpty, traceDisplayName);
+                throw new ArgumentException(Resources.CannotBeNullOrEmpty, nameof(traceDisplayName));
             }
 
             if (delegates != null)
             {
-                foreach (Delegate handler in delegates.GetInvocationList())
+                var invocationList = delegates.GetInvocationList();
+                var i = 0;
```
MarcoRossignoli, line 57: nit: for?

MarcoRossignoli:
I think it's a good addition thx @nohwnd, it should help a lot during the investigations.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
MarcoRossignoli: no a great gain I think we invoke with https://docs.microsoft.com/en-us/dotnet/api/system.delegate.dynamicinvoke?view=net-6.0 so we'll have anyway some boxing if needed.

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
MarcoRossignoli: I think that the idea is not to start a timer but measure the time, and it's done with stopWatch.ElapsedMilliseconds(implicit stop/read)

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
nohwnd: Yes, please do it 😏

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
nohwnd: I will keep it as is.

@Evangelink  Technically you are right, this is adding few ticks between the time the WaitOne returns, and we capture the value. But in reality it does not matter, we are looking at the number in milliseconds and don't really care if 1.00004 ms or 1.00013 ms is the correct measurement. So this just saves one line of code. Stopwatch also has no overhead when we don't stop it, because it is not doing anything until you ask it for time, so doing it like this is fine.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd: The new method would have the same signature, and would change the meaning of the traceDisplayName slightly, because it would now capture the caller, instead of the provided value. You could still provided value, so on compile time the signature would stay compatible, but the assembly would not, because in order to be useful, parameter with [CallerMemberName] attribute needs to have default value (`= null`). This changes the public api, because it is not a backward compatible code change.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd: Good call.

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
nohwnd: Not sure I get what you are asking. From how I understand how this works is that we poll, and invoke notification (NotifyDataAvailable) when there are data. 

Then in the notification handler (in the code here), we read the stream to get the message data, and send it to receivers by calling MessageReceived. 

If we poll, and read even where there are no subscribers to MessageReceived we just lose that message. 

Am I wrong?

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd, line 41:
```suggestion
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd:
```suggestion
                throw new ArgumentNullException(nameof(args));
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
             if (args == null)
             {
-                throw new ArgumentNullException(Resources.CannotBeNullOrEmpty, "args");
+                throw new ArgumentNullException(Resources.CannotBeNullOrEmpty, nameof(args));
             }
 
             if (string.IsNullOrWhiteSpace(traceDisplayName))
             {
-                throw new ArgumentException(Resources.CannotBeNullOrEmpty, traceDisplayName);
+                throw new ArgumentException(Resources.CannotBeNullOrEmpty, nameof(traceDisplayName));
```
nohwnd, line 51:
```suggestion
                throw new ArgumentException(nameof(traceDisplayName));
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd, line 67:
```suggestion
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
```diff
@@ -22,40 +22,93 @@ public static class MulticastDelegateUtilities
...
             }
 
             if (delegates != null)
             {
-                foreach (Delegate handler in delegates.GetInvocationList())
+                var invocationList = delegates.GetInvocationList();
+                var i = 0;
+                foreach (Delegate handler in invocationList)
                 {
+                    Stopwatch stopwatch = Stopwatch.StartNew();
```
nohwnd, line 60:
```suggestion
                    var stopwatch = Stopwatch.StartNew();
```

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd: Nope you are right. Was initializing on the top, but needed to use that also for the error in the end.

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
Evangelink: The time diff wasn't the main concern.

> Stopwatch also has no overhead when we don't stop it, because it is not doing anything until you ask it for time, so doing it like this is fine.

This was, and I thought were would be some overhead of having timers running in the background for some time but as you say the overhead is neglectable, fine to keep as-is.

nohwnd: Just to be clear. There are no timers inside of stopwatch. It asks for the current timestamp when started, and then asks for another when you call ElapsedTime or Stop. It reports the difference between those two. 

Apart from the call to get timestamp, and having one more object created, there is no permanent overhead of stopwatch running in background.

src/Microsoft.TestPlatform.CoreUtilities/Utilities/MulticastDelegateUtilities.cs
nohwnd: I did not want to introduce off by one error and have the loop run forever. This code is uglier, but the worse that can happen is that the log is slightly incorrect.

C# please give me `foreach (var h, var i in invocationList) { }`  :)

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
MarcoRossignoli: Nope it's true(like I understood at least), that "consumed by poll" was a bad way to say what you wrote.

nohwnd: @Evangelink  for awareness on how this works.

test/Microsoft.TestPlatform.Client.UnitTests/Execution/TestRunRequestTests.cs
```diff
@@ -170,7 +170,14 @@ public void OnTestSessionTimeoutShouldLogMessage()
...
 
-        _testRunRequest.TestRunMessage += (object sender, TestRunMessageEventArgs e) => handleLogMessageCalled = true;
+        _mockDataSerializer
+            .Setup(s => s.SerializePayload(It.IsAny<string>(), It.IsAny<Object>()))
+            .Returns("non-empty rawMessage");
+
+        _testRunRequest.TestRunMessage += (object sender, TestRunMessageEventArgs e) =>
+            {
+                handleLogMessageCalled = true;
+            };
```
Evangelink, line 180:
```suggestion
        _testRunRequest.TestRunMessage += (object sender, TestRunMessageEventArgs e) => handleLogMessageCalled = true;
```

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
```diff
@@ -346,7 +346,7 @@ private void AddNewFileTransfer(FileTransferInformation fileTransferInfo, AsyncC
                     }
                 }
             },
-            _cancellationTokenSource.Token);
+                _cancellationTokenSource.Token);
```
Evangelink, line 349: Is this expected? I think the previous formatting was right.

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
```diff
@@ -139,7 +139,11 @@ public async Task AcceptClientAsync()
     /// <returns>True if Client is connected, false otherwise</returns>
     public bool WaitForClientConnection(int clientConnectionTimeout)
     {
-        return _clientConnectedEvent.WaitOne(clientConnectionTimeout);
+        var stopWatch = Stopwatch.StartNew();
+        var result = _clientConnectedEvent.WaitOne(clientConnectionTimeout);
+        EqtTrace.Verbose("SocketCommunicationManager.WaitForClientConnection took: {0} ms, with {1} ms timeout, and finished with {2}", stopWatch.ElapsedMilliseconds, clientConnectionTimeout, result);
```
Evangelink, line 144: Nit:

```suggestion
        EqtTrace.Verbose("SocketCommunicationManager.WaitForClientConnection took: {0} ms, with {1} ms timeout, and finished with {2}.", stopWatch.ElapsedMilliseconds, clientConnectionTimeout, result);
```

src/Microsoft.TestPlatform.CommunicationUtilities/SocketCommunicationManager.cs
```diff
@@ -215,7 +219,11 @@ public async Task SetupClientAsync(IPEndPoint endpoint)
     /// <returns>True, if Server got a connection from client</returns>
     public bool WaitForServerConnection(int connectionTimeout)
     {
-        return _clientConnectionAcceptedEvent.WaitOne(connectionTimeout);
+        var stopWatch = Stopwatch.StartNew();
+        var result = _clientConnectionAcceptedEvent.WaitOne(connectionTimeout);
+        EqtTrace.Verbose("SocketCommunicationManager.WaitForServerConnection took: {0} ms, with {1} ms timeout, and finished with {2}", stopWatch.ElapsedMilliseconds, connectionTimeout, result);
```
Evangelink, line 224: Nit

```suggestion
        EqtTrace.Verbose("SocketCommunicationManager.WaitForServerConnection took: {0} ms, with {1} ms timeout, and finished with {2}.", stopWatch.ElapsedMilliseconds, connectionTimeout, result);
```

Evangelink (approved):
A few minor comments, otherwise LGTM.

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
nohwnd, line 349:
```suggestion
            _cancellationTokenSource.Token);
```

Evangelink (approved):

MarcoRossignoli (approved):

PR #3287, opened 2022-01-28
Recognize AnyCPU in case of fallback to PEReader

MarcoRossignoli:
Starting 7.0 runtime some `AssemblyName` properties are deprecated, for instance `ProcessorArchitecture`. 
We're relying on it to get the architecture for the dll. 
Now it will returns always "None" and we fallback to the PEReader reader, so we need to be able to discover also AnyCPU from it.

I did some manual test with x86/x64/neutral compilation and looks good. 
I ran also failing test on the sdk side with the local build and looks good.

Ref issue https://github.com/dotnet/sdk/pull/23597#issuecomment-1024030574


src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,28 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
...
-                    return Architecture.X64;
                 case Machine.IA64:
                     return Architecture.X64;
                 case Machine.Arm64:
                     return Architecture.ARM64;
                 case Machine.Arm:
                     return Architecture.ARM;
                 case Machine.I386:
-                    return Architecture.X86;
+                    // If requires 32bit or if R2R
```
nohwnd, line 113: It is not clear to me what is the reasoning there (I would also spell out Ready to run at least once in the comment.

MarcoRossignoli: We can distinguish AnyCPU only from the set of `CorFlags.Requires32Bit`, but in case of Ready to Run image that flag is not "updated" and ignored, if so we check it the module is IL only or not if it's IL only it means that is a R2R and we're already in the correct architecture x86. In all other cases the architecture will end inside the correct switch branch

nohwnd:
```suggestion
                    // We can distinguish AnyCPU only from the set of CorFlags.Requires32Bit, but in case of Ready                                         
                    // to Run image that flag is not "updated" and ignored. So we check if the module is IL only or                    
                    // not. If it's IL only it means that is a R2R (Ready to Run) and we're already in the correct architecture x86. In all                    
                    // other cases the architecture will end inside the correct switch branch.
```

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,31 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
...
                     return Architecture.X64;
                 case Machine.Arm64:
                     return Architecture.ARM64;
                 case Machine.Arm:
                     return Architecture.ARM;
                 case Machine.I386:
-                    return Architecture.X86;
+                    // We can distinguish AnyCPU only from the set of CorFlags.Requires32Bit, but in case of Ready                                         
+                    // to Run image that flag is not "updated" and ignored. So we check if the module is IL only or                    
+                    // not. If it's IL only it means that is a R2R (Ready to Run) and we're already in the correct architecture x86. In all
```
MarcoRossignoli, line 115:
```suggestion
                    // not. If it's not IL only it means that is a R2R (Ready to Run) and we're already in the correct architecture x86. In all                    
```

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,31 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
             switch (peReader.PEHeaders.CoffHeader.Machine)
             {
                 case Machine.Amd64:
-                    return Architecture.X64;
                 case Machine.IA64:
```
kasperk81, line 106: itanium (ia64) is also unsupported obsolete architecture (and it's different than x64). you can delete this line.

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,31 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
...
                     return Architecture.X64;
                 case Machine.Arm64:
                     return Architecture.ARM64;
                 case Machine.Arm:
                     return Architecture.ARM;
                 case Machine.I386:
-                    return Architecture.X86;
+                    // We can distinguish AnyCPU only from the set of CorFlags.Requires32Bit, but in case of Ready
```
Evangelink, line 113: I am surprised, it seems that the rule about un-needed ending spaces didn't trigger. Could you remove them manually?

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,31 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
...
+                        return Architecture.X86;
+                    }
+                    else
+                    {
+                        return Architecture.AnyCPU;
+                    }
                 default:
-                    break;
+                    {
+                        EqtTrace.Error($"AssemblyMetadataProvider.GetArchitecture: Unhandled architecture '{peReader.PEHeaders.CoffHeader.Machine}'");
```
Evangelink, line 128: Nit

```suggestion
                        EqtTrace.Error($"AssemblyMetadataProvider.GetArchitecture: Unhandled architecture '{peReader.PEHeaders.CoffHeader.Machine}'.");
```

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -103,17 +103,31 @@ private Architecture GetArchitectureFromAssemblyMetadata(string path)
...
                 case Machine.Arm:
                     return Architecture.ARM;
                 case Machine.I386:
-                    return Architecture.X86;
+                    // We can distinguish AnyCPU only from the set of CorFlags.Requires32Bit, but in case of Ready                                         
+                    // to Run image that flag is not "updated" and ignored. So we check if the module is IL only or                    
+                    // not. If it's not IL only it means that is a R2R (Ready to Run) and we're already in the correct architecture x86. In all                    
+                    // other cases the architecture will end inside the correct switch branch.
+                    var corflags = peReader.PEHeaders.CorHeader.Flags;
+                    if ((corflags & CorFlags.Requires32Bit) != 0 || ((corflags & CorFlags.ILOnly) == 0))
```
Evangelink, line 118: Nit

```suggestion
                    if ((corflags & CorFlags.Requires32Bit) != 0 || (corflags & CorFlags.ILOnly) == 0)
```

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
MarcoRossignoli: Thanks, at the moment it's used also in another place in the same file so I'll keep it here for coherence. 
We'll remove in a follow up PR after some analysis cc: @nohwnd

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
MarcoRossignoli: done thx

Evangelink (approved):

nohwnd (approved):

PR #3293, opened 2022-01-28
[WIP] Enable more rules for the repo

Evangelink:
Enable more rules that would make sense in the repo.


test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestLoggerManagerTests.cs
```diff
@@ -1630,6 +1630,7 @@ private class InvalidLogger
     {
         public static int Counter = 0;
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "REVIEW ME: Is it ok to make this static?")]
```
Evangelink, line 1633: REVIEW ME.

.editorconfig
```diff
@@ -149,8 +187,8 @@ csharp_preferred_modifier_order = public,private,protected,internal,static,exter
 
 # Code-block preferences
 csharp_prefer_braces = true:silent
+# Keep this in sync with the related .net rule: IDE0063
```
nohwnd, line 190:
```suggestion
# Keep this in sync with the related .NET rule: IDE0063
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -105,7 +105,7 @@ public TestPlatform()
         loggerManager.Initialize(discoveryCriteria.RunSettings);
 
         var testHostManager = _testHostProviderManager.GetTestHostManagerByRunConfiguration(discoveryCriteria.RunSettings);
-        ThrowExceptionIfTestHostManagerIsNull(testHostManager, discoveryCriteria.RunSettings);
+        TestPlatform.ThrowExceptionIfTestHostManagerIsNull(testHostManager, discoveryCriteria.RunSettings);
```
nohwnd, line 108: Which rule forces this, and what advantage is it bringing us? Are static methods significantly faster? Will we be forced to mark every method that does not touch class fields as static? Is not touching class fields considered code smell and this rule just makes the smell easier to see?

src/Microsoft.TestPlatform.Common/Utilities/StringExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System;
+
+#if NETFRAMEWORK || WINDOWS_UAP || (NETSTANDARD && !NETSTANDARD2_1_OR_GREATER)
+internal static class StringExtensions
```
nohwnd, line 7: I think we should consider this PR to be added to blame ignore. Make sure you put this in a separate commit, and don't squash.

nohwnd: Are we including this into every project? We could also leverage it to replace `string.IsNullOrWhitespace(someString)` and some of the other annoyingly difficult things we do with strings and other types. If we don't include it everywhere we might want to move this to a shared project and import that here and explicitly to other projects so we avoid adding an extra dll we need to manage.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestLoggerManagerTests.cs
nohwnd: It should be, functionality-wise. Why do you think it would make difference?

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Evangelink: > Which rule forces this, and what advantage is it bringing us?
Are you referring to `static` or the prefix? If that's about the prefix, the code-fix for make member `static` seems to be adding the class before but that's not required for C# so I could simplify all names if we prefer.

> Are static methods significantly faster? 
They are faster but perf is always tricky, it depends on so many things. I just don't see any reason not to benefit from the gain (even if proved small in some context). But I would not worry about breaking public API (I suppressed the warnings for them) only for such small change.

> Will we be forced to mark every method that does not touch class fields as static?
Yes, except if we suppress it. I could keep the rule as suggestion but then it's unlikely to be followed.

> Is not touching class fields considered code smell and this rule just makes the smell easier to see?
I am not sure to understand the question.

nohwnd: I was referring to the static, but I did not realize that you are adding the name of the class, because the rule does that. I thought you don't need that, but I was not sure, and did not bother to try it out. I would prefer it that way, so we only call the method by name, not caring if it is static or not. That way making a method static does not result in many changes, and would be much easier to follow.

src/Microsoft.TestPlatform.Common/Utilities/StringExtensions.cs
Evangelink: About the new project, I guess you are thinking about some shared project rather than a class library project? Could we rename `Microsoft.TestPlatform.Execution.Shared` into `Microsoft.TestPlatform.Shared` and make it referenced by all projects? Or do you prefer a separate project?

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestLoggerManagerTests.cs
Haplois: Please make sure it doesn't break compatibility across, I am still reviewing this but in general even if something is internal, if it crosses AppDomain boundaries we need to tread carefully.

src/Microsoft.TestPlatform.Common/Utilities/StringExtensions.cs
Haplois: @nohwnd Merging without squashing is not possible as per our current repo config.

Evangelink: I can cherry-pick specific commit (or copy files) and create a separate PR when we agree on the scope.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/TestLoggerManagerTests.cs
nohwnd: @Haplois do you have a concrete example where this caused a problem? We can go off on just "don't change things", but then the current state will never change.

src/Microsoft.TestPlatform.Common/Utilities/StringExtensions.cs
nohwnd: @Haplois I thought so, we can change it for this PR or cherry pick the commit as mentioned above.

Evangelink:
@MarcoRossignoli I moved it back to a draft. I am waiting for the input from @Haplois to add some tests that would "guarantee" we don't break any compatibility by having certain non-public members `static`. When this is available, I will resume this PR.

PR #3290, opened 2022-01-28
Reformat projects and script files

Evangelink:

PR #3295, opened 2022-01-30
Remove unused properties and solution files

Evangelink:
Merge after #3290


src/DataCollectors/DumpMinitool.x86/DumpMinitool.x86.csproj
```diff
@@ -8,7 +8,6 @@
   <PropertyGroup>
     <TargetFrameworks>net451</TargetFrameworks>
     <TargetFrameworks Condition=" '$(DotNetBuildFromSource)' == 'true' ">net6.0</TargetFrameworks>
-    <WarningsAsErrors>true</WarningsAsErrors>
```
Evangelink, line 11: It seems this property was used wrongly (or the usage changed over time), according to the [doc](https://docs.microsoft.com/dotnet/csharp/language-reference/compiler-options/errors-warnings#warningsaserrors-and-warningsnotaserrors), it is supposed to accept a list of rule IDs not a bool. I guess the goal was to use `TreatWarningsAsErrors` which is defined globally in a *.props file.

Haplois (approved):

nohwnd (approved):

PR #3296, opened 2022-01-30
Fix race condition inside DataCollectionAttachmentManager

MarcoRossignoli:
* We're analyzing a strange race condition issue, but if we enable diag logs we're hiding real issue for an NRE inside the logs self.

image

* Update `Dictionary` to `ConcurrentDictionary` to make `DataCollectionAttachmentManager` thread safe, concurrent call to the data collector `TestCaseEnded` will eventually fail.

```cs
 [TestMethod]
    public void ParallelAccess()
    {
        _attachmentManager.Initialize(new SessionId(Guid.NewGuid()), Path.GetTempPath(), _messageSink.Object);
        Task.WaitAll(

        Task.Run(() =>
        {
            try
            {
                while (true)
                {
                    var dc = new DataCollectionContext(new SessionId(Guid.NewGuid()));
                    string path = Path.GetTempFileName();
                    File.WriteAllText(path, "test");
                    _attachmentManager.AddAttachment(new FileTransferInformation(dc, path, true), null, new Uri("//test1"), "test1");
                    var res = _attachmentManager.GetAttachments(dc);
                }
            }
            catch (Exception ex)
            {
                Assert.Fail(ex.ToString());
            }
        }),

        Task.Run(() =>
        {
            try
            {
                while (true)
                {
                    var dc = new DataCollectionContext(new SessionId(Guid.NewGuid()));
                    string path = Path.GetTempFileName();
                    File.WriteAllText(path, "test");
                    _attachmentManager.AddAttachment(new FileTransferInformation(dc, path, true), null, new Uri("//test2"), "test2");
                    var res = _attachmentManager.GetAttachments(dc);
                }
            }
            catch (Exception ex)
            {
                Assert.Fail(ex.ToString());
            }
        })
        );
    }
``` 
* Add try/catch on the events to avoid possible unhandled exception that lead to the test host hang.

### DIFF
--- a/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
+++ b/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
@@ -768,10 +768,28 @@ internal class DataCollectionManager : IDataCollectionManager
 
     private void LogAttachments(List<AttachmentSet> attachmentSets)
     {
+        if (attachmentSets is null)
+        {
+            EqtTrace.Error("DataCollectionManager.LogAttachments: Unexpected null attachmentSets.");
+            return;
+        }
+
         foreach (var entry in attachmentSets)
         {
+            if (entry is null)
+            {
+                EqtTrace.Error("DataCollectionManager.LogAttachments: Unexpected null entry inside attachmentSets.");
+                continue;
+            }
+
             foreach (var file in entry.Attachments)
             {
+                if (file is null)
+                {
+                    EqtTrace.Error("DataCollectionManager.LogAttachments: Unexpected null file inside entry attachments.");
+                    continue;
+                }
+
                 EqtTrace.Verbose(
                     "Test Attachment Description: Collector:'{0}'  Uri:'{1}'  Description:'{2}' Uri:'{3}' ",
                     entry.DisplayName,
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
@@ -76,7 +76,7 @@ internal class DataCollectionRequestHandler : IDataCollectionRequestHandler, IDi
             new SocketCommunicationManager(),
             messageSink,
             DataCollectionManager.Create(messageSink, requestData),
-            new DataCollectionTestCaseEventHandler(),
+            new DataCollectionTestCaseEventHandler(messageSink),
             JsonDataSerializer.Instance,
             new FileHelper(),
             requestData)
@@ -162,7 +162,7 @@ internal class DataCollectionRequestHandler : IDataCollectionRequestHandler, IDi
                         communicationManager,
                         messageSink,
                         DataCollectionManager.Create(messageSink, requestData),
-                        new DataCollectionTestCaseEventHandler(),
+                        new DataCollectionTestCaseEventHandler(messageSink),
                         JsonDataSerializer.Instance,
                         new FileHelper(),
                         requestData);
@@ -362,7 +362,7 @@ internal class DataCollectionRequestHandler : IDataCollectionRequestHandler, IDi
                     }
                     catch (Exception e)
                     {
-                        EqtTrace.Error("DataCollectionRequestHandler.HandleBeforeTestRunStart : Error occurred during initialization of TestHost : {0}", e);
+                        EqtTrace.Error("DataCollectionRequestHandler.HandleBeforeTestRunStart : Error occurred during test case events handling: {0}.", e);
                     }
                 },
                 _cancellationTokenSource.Token);
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
@@ -3,14 +3,19 @@
 
 namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.DataCollection;
 
+using System;
+using System.Collections.ObjectModel;
 using System.Net;
 
 using Common.DataCollector;
+
 using Microsoft.VisualStudio.TestPlatform.Common.DataCollector.Interfaces;
 using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Interfaces;
-using ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+using ObjectModel;
 
 /// <summary>
 /// The test case data collection request handler.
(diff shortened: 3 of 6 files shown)


src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
```diff
@@ -768,10 +768,28 @@ private void RemoveDataCollectors(IReadOnlyCollection<DataCollectorInformation>
...
+            EqtTrace.Error("DataCollectionManager.LogAttachments: Unexpected null attachmentSets.");
+            return;
+        }
+
         foreach (var entry in attachmentSets)
         {
+            if (entry is null)
+            {
+                EqtTrace.Error("DataCollectionManager.LogAttachments: Unexpected null entry inside attachmentSets.");
+                continue;
```
Evangelink, line 782: Is it ok to only skip null parts or shall we assume something is wrong and fail? Same remarks for the attachments below.

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -362,7 +362,7 @@ private void HandleBeforeTestRunStart(Message message)
                     }
                     catch (Exception e)
                     {
-                        EqtTrace.Error("DataCollectionRequestHandler.HandleBeforeTestRunStart : Error occurred during initialization of TestHost : {0}", e);
+                        EqtTrace.Error("DataCollectionRequestHandler.HandleBeforeTestRunStart : Error occurred during test case events handling: {0}", e);
```
Evangelink, line 365: Nit:
```suggestion
                        EqtTrace.Error("DataCollectionRequestHandler.HandleBeforeTestRunStart : Error occurred during test case events handling: {0}.", e);
```

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
```diff
@@ -79,7 +85,17 @@ public void ProcessRequests()
...
-                    _dataCollectionManager.TestCaseStarted(testCaseStartEventArgs);
+
+                    try
+                    {
+                        _dataCollectionManager.TestCaseStarted(testCaseStartEventArgs);
+                    }
+                    catch (Exception ex)
+                    {
+                        _messageSink.SendMessage(new DataCollectionMessageEventArgs(TestMessageLevel.Error, $"Error occurred during TestCaseStarted event handling: {ex}"));
+                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests : Error occurred during TestCaseStarted event handling: {ex}");
```
Evangelink, line 96: Nit:

```suggestion
                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests: Error occurred during TestCaseStarted event handling: {ex}");
```

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
```diff
@@ -96,7 +112,19 @@ public void ProcessRequests()
...
+
+                    Collection<AttachmentSet> attachmentSets;
+                    try
+                    {
+                        attachmentSets = _dataCollectionManager.TestCaseEnded(testCaseEndEventArgs);
+                    }
+                    catch (Exception ex)
+                    {
+                        _messageSink.SendMessage(new DataCollectionMessageEventArgs(TestMessageLevel.Error, $"Error occurred during DataCollectionTestEnd event handling: {ex}"));
+                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests : Error occurred during DataCollectionTestEnd event handling: {ex}");
```
Evangelink, line 124:
```suggestion
                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests: Error occurred during DataCollectionTestEnd event handling: {ex}");
```

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
```diff
@@ -79,7 +85,17 @@ public void ProcessRequests()
...
                     var testCaseStartEventArgs = _dataSerializer.DeserializePayload<TestCaseStartEventArgs>(message);
-                    _dataCollectionManager.TestCaseStarted(testCaseStartEventArgs);
+
+                    try
+                    {
+                        _dataCollectionManager.TestCaseStarted(testCaseStartEventArgs);
+                    }
+                    catch (Exception ex)
+                    {
+                        _messageSink.SendMessage(new DataCollectionMessageEventArgs(TestMessageLevel.Error, $"Error occurred during TestCaseStarted event handling: {ex}"));
```
Evangelink, line 95: Probably doesn't matter much but you could extract the `ex.ToString()` in a variable instead of forcing the call twice.

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
```diff
@@ -114,7 +142,15 @@ public void ProcessRequests()
...
 
-                    Close();
+                    try
+                    {
+                        Close();
+                    }
+                    catch (Exception ex)
+                    {
+                        _messageSink.SendMessage(new DataCollectionMessageEventArgs(TestMessageLevel.Error, $"Error occurred during SessionEnd event handling: {ex}"));
+                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests : Error occurred during SessionEnd event handling: {ex}");
```
Evangelink, line 152: Nit
```suggestion
                        EqtTrace.Error($"DataCollectionTestCaseEventHandler.ProcessRequests: Error occurred during SessionEnd event handling: {ex}");
```

test/datacollector.UnitTests/DataCollectionAttachmentManagerTests.cs
```diff
@@ -39,6 +44,47 @@ public void Cleanup()
         File.Delete(Path.Combine(TempDirectoryPath, "filename1.txt"));
     }
 
+    [TestMethod]
+    public void ParallelAccessShouldNotBreak()
+    {
+        string outputDirectory = Path.Combine(TempDirectoryPath, Guid.NewGuid().ToString());
```
Evangelink, line 50: I don't know if the helper is available from this project but that might be better to use it if possible.

test/datacollector.UnitTests/DataCollectionAttachmentManagerTests.cs
```diff
@@ -39,6 +44,47 @@ public void Cleanup()
         File.Delete(Path.Combine(TempDirectoryPath, "filename1.txt"));
     }
 
+    [TestMethod]
+    public void ParallelAccessShouldNotBreak()
```
Evangelink, line 48: That's really cool you found a way to replicate the issue <3

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionManager.cs
MarcoRossignoli: it's a log part...I don't want throw nothing if state is corrupted for some bug and we have enabled diagnostic information.

test/datacollector.UnitTests/DataCollectionAttachmentManagerTests.cs
MarcoRossignoli: It's not available...are unit tests and not integrations we're not inheriting from it unfortunately, maybe we should have something more "base" for everything.

test/datacollector.UnitTests/DataCollectionAttachmentManagerTests.cs
MarcoRossignoli: Rare but it happen :(

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventHandler.cs
MarcoRossignoli: it should never happen(only in case of bug), we can keep as-is for now one less line.

jakubch1 (approved):

PR #3307, opened 2022-01-31
Run dotnet format whitespace

Evangelink:


nohwnd (approved):

.editorconfig
```diff
@@ -53,9 +53,6 @@ indent_size = 2
 indent_size = 4
 tab_width = 4
 
-# New line preferences
```
nohwnd, line 56: Is this a change from default? Please annotate it rather than deleting.

nohwnd (changes requested):

.editorconfig
Evangelink: Looks like my comment was removed when I pushed! I was saying that for some reason this was duplicate with a different value for C#. The entry exists and is set to true for all files at the beginning of the file.

nohwnd (approved):

nohwnd:
Add it to that blame ignore file once it is in main please.

PR #3317, opened 2022-02-02
Fix serialization issue with TestRunSettings

Evangelink:
Use auto-property code-fix changed the readonly property into a private setter property which seems to change the behavior of the xml serializer. I have kept a property only but moved to a readonly property.

cc @AbhitejJohn

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/TestRunSettings.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/TestRunSettings.cs
@@ -12,7 +12,6 @@ using System.Xml;
 /// </summary>
 public abstract class TestRunSettings
 {
-
     #region Constructor
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.ObjectModel/RunSettings/TestRunSettings.cs
```diff
@@ -46,6 +45,7 @@ protected TestRunSettings(string name)
     /// Converter the setting to be an XmlElement.
     /// </summary>
     /// <returns>The Xml element for the run settings provided.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1059:MembersShouldNotExposeCertainConcreteTypes", Justification = "XmlElement is required in the data collector.")]
```
Evangelink, line 48: That's unrelated but I noticed this was removed during the reformatting and should not have been.

nohwnd (approved):

PR #3323, opened 2022-02-03
Suppress assembly architecture, assembly conflict and restore warnings

nohwnd:
Suppress assembly architecture, assembly conflict and restore warnings that we can't avoid. This is preliminary work to enabling Warnings to be just warnings for local development. And then probably warnings as errors in PR, or ideally warnings as errors in real PRs and warnings as warnings in draft PRs.


scripts/build/TestPlatform.Settings.targets
```diff
@@ -68,6 +68,8 @@
     <When Condition="$(TestProject) == 'true'">
       <PropertyGroup>
         <GenerateDocumentationFile>false</GenerateDocumentationFile>
+        <!-- Suppress warnings about testhost being x64 (AMD64)/x86 when imported into AnyCPU (MSIL) test projects. -->
+        <NoWarn>$(NoWarn);MSB3270</NoWarn>
```
Evangelink, line 72: Do we need both this and the `MSBuildWarningsAsMessages` in each project?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/Microsoft.TestPlatform.Extensions.BlameDataCollector.csproj
```diff
@@ -1,6 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <!-- NU1702 suppresses warning about DumpMinitool being .NET Framework reference imported to a .NET project.
+    Unfortunately this can't be supressed at single ProjectReference. Neither completely suppressed by using NoWarn.
```
Evangelink, line 5:
```suggestion
    Unfortunately this can't be suppressed at single ProjectReference. Neither completely suppressed by using NoWarn.
```

playground/TestPlatform.Playground/TestPlatform.Playground.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TestPlatformRoot Condition="$(TestPlatformRoot) == ''">..\..\</TestPlatformRoot>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
+    <!-- MSB3270 Suppress warnings about testhost being x64 (AMD64)/x86 when imported into AnyCPU (MSIL) projects. -->
+    <!-- MSB3276 Suppress warnings about conflicts between different versions of the same dependent assembly -->
+    <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);MSB3270;MSB3276</MSBuildWarningsAsMessages>
```
Evangelink, line 7: I am wondering if it could be better to have these declared in the "global" target with a condition and only add the condition property to the projects (something like `<SuppressTestHostAssemblyWarnings>true</SuppressTestHostAssemblyWarnings>`).

scripts/build/TestPlatform.Settings.targets
nohwnd: We don't suppress this warning in each project, just in some. This is blanket suppression for all TestProjects.

playground/TestPlatform.Playground/TestPlatform.Playground.csproj
nohwnd: Ideally each of those suppressions would be placed directly on the project reference, but that does not work. Probably for the same reason as the NU1702 warning not being suppressed. I did not want to over complicate this, because then it is not very consistent without the NU1702, and understanding what is happening is easy if you look directly at the project. For SuppressTestHostAssemblyWarnings I would have to look at the definition to know what that means. Imho it is good to go as is. But if you really prefer the other solution I can change it.

scripts/build/TestPlatform.Settings.targets
Evangelink: Sorry I wasn't clear. I meant do we need both `NoWarn` and `MSBuildWarningsAsMessages` applied (I see that both are "suppressing" the `MSB3270`).

playground/TestPlatform.Playground/TestPlatform.Playground.csproj
Evangelink: I am fine with current implementation, just wanted to suggest a different path in case that wasn't considered.

scripts/build/TestPlatform.Settings.targets
nohwnd:
```suggestion
        <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);MSB3270</MSBuildWarningsAsMessages>
```

Do I get it correctly that you want me to do this ^^^ ?

playground/TestPlatform.Playground/TestPlatform.Playground.csproj
nohwnd: It wasn't and thank for the comment, it helped me to think about my reasoning further :)

scripts/build/TestPlatform.Settings.targets
Evangelink: Or the other way around depending on which of the 2 property has wider area but yes :)

Evangelink (approved):

PR #3325, opened 2022-02-04
Fix red

nohwnd:
Use Write-Host instead of setting output color and writing output.
Lock when outputting from callbacks, because they happen on thread pool and can run at the same time. Which ends up setting foreground color to red permanently.


scripts/common.lib.ps1
```diff
@@ -233,6 +227,9 @@ function Invoke-Exe {
 
     Write-Verbose "Done. Exit code: $exitCode"
 
+    if ($exitCode -ne 0) {
+
+    }
```
nohwnd, line 232:
```suggestion
```

Haplois (approved):

scripts/common.lib.ps1
```diff
@@ -60,17 +60,11 @@ function Write-Log {
...
-            Write-Output "... $message"
-        }
-    }
-    finally {
-        $Host.UI.RawUI.ForegroundColor = $currentColor
+    
+    if ($message)
+    {
+        $color = if ("Success" -eq $Level) { "Green" } else { "Red" }
+        Write-Host "... $message" -ForegroundColor $color
```
Evangelink, line 67: I am looking for the source but I remember to have read somewhere that it was better to avoid `Write-Host` and always use `Write-Output`/`Write-Information`/`Write-Error` instead.

Evangelink: Nevermind the articles are from 2013, I can't find any new article so that's probably obsolete remark.

nohwnd: That used to be the gospel. But it was never true anyway. But still it was repeated thousands of times, every time you shown someone a script with Write-Host in it :D 

In this case using Write-Host or Write-Output surrounded by setting the output color do in fact have a slightly different meaning and behavior, and we are losing the ability to capture the output of Write-Host just by assigning the result of this script into a variable (but if we tried enough there is new set of recent cmdlets that Write-Host goes through anyway so we now have a way to capture even that). But because we don't plan to do anything like capturing all our output, and because we write to console directly from our process output handlers, this change is absolutely safe to do.

PR #3327, opened 2022-02-04
Fix failfast

nohwnd:
Pass the exit code directly because we no longer rely on LastExitCode for executables. 

Add colors to lines that match " error" and " warning".


scripts/build.ps1
```diff
@@ -55,6 +55,7 @@ Param(
 )
 
 $ErrorActionPreference = 'Stop'
+$ErrorView = 'Normal'
```
nohwnd, line 58: This disables the summary view in Pwsh that tries to summarize the error. We want to see the error in full.

Haplois (approved):

PR #3329, opened 2022-02-04
Enable more rules requiring a polyfill type

Evangelink:


src/Microsoft.TestPlatform.ShimLayer/StringExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: I have named the containing folder `Microsoft.TestPlatform.ShimLayer`, let me know if you prefer a simple and maybe more generic name for such kind of shared file.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -301,11 +301,9 @@ private void ValidateAndAddTriggerBasedProcessDumpParameters(XmlElement collectD
...
                 case XmlAttribute attribute when string.Equals(attribute.Name, Constants.CollectDumpAlwaysKey, StringComparison.OrdinalIgnoreCase):
 
-                    if (string.Equals(attribute.Value, Constants.TrueConfigurationValue, StringComparison.OrdinalIgnoreCase) || string.Equals(attribute.Value, Constants.FalseConfigurationValue, StringComparison.OrdinalIgnoreCase))
-                    {
-                        bool.TryParse(attribute.Value, out _collectDumpAlways);
-                    }
-                    else
+                    if ((!string.Equals(attribute.Value, Constants.TrueConfigurationValue, StringComparison.OrdinalIgnoreCase)
+                        && !string.Equals(attribute.Value, Constants.FalseConfigurationValue, StringComparison.OrdinalIgnoreCase))
+                        || !bool.TryParse(attribute.Value, out _collectDumpAlways))
```
Evangelink, line 306: I have changed a bit the logic to also log the warning about the parameter in case of a failure in the `TryParse` although I don't really expect any problem as we just tested before that the string was equal to `true` or `false`.

PR #3335, opened 2022-02-07
Enable TreatWarningsAsErrors only on CI

Evangelink:
## Description

We raised the severity of some rules from message to warning to ensure they are followed in the team but because of the `TreatWarningsAsErrors` we end up with some "boring" dev cycles where we are forced to cleanup usings etc while we iterate. The idea is to keep warnings locally to not slow us down but still fail in PRs so that we can keep styling consistency.

cc @nohwnd @MarcoRossignoli

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -20,7 +20,9 @@
          serialization. This is also defined in build script. -->
     <AssemblyVersion Condition="'$(AssemblyVersion)' == ''">15.0.0</AssemblyVersion>
 
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <!-- Treat warnings as errors only on CI to prevent painful development process because of unused usings... -->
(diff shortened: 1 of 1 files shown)


scripts/build/TestPlatform.Settings.targets
```diff
@@ -20,7 +20,9 @@
          serialization. This is also defined in build script. -->
     <AssemblyVersion Condition="'$(AssemblyVersion)' == ''">15.0.0</AssemblyVersion>
 
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <!-- Treat warnings as errors only on CI to prevent painful development process because of unused usings... -->
+    <TreatWarningsAsErrors Condition="'$(CIBuild)' == 'true'">true</TreatWarningsAsErrors>
```
Evangelink, line 24: @nohwnd I remember you asked if we could disable in draft PRs but except from using github api I couldn't find something already in place.

MarcoRossignoli (approved):
LGTM at first glance...let's see how much pain it will cause and in case we'll revert.

PR #3340, opened 2022-02-08
Telemetry improvements

cvpoienaru:
Telemetry improvements


MarcoRossignoli:
@cvpoienaru  looks like you've some conflicts

src/Microsoft.TestPlatform.Client/RequestHelper/ITestRequestManager.cs
```diff
@@ -88,6 +88,18 @@ public interface ITestRequestManager : IDisposable
...
         ProtocolConfig protocolConfig);
 
+    /// <summary>
+    /// Stops a test session.
+    /// </summary>
+    /// 
+    /// <param name="testSessionInfo">The test session info.</param>
+    /// <param name="eventsHandler">The events handler.</param>
+    /// <param name="protocolConfig">Protocol related information.</param>
+    public void StopTestSession(
```
Evangelink, line 98: I don't think you wanted to include accessibility.

```suggestion
    void StopTestSession(
```

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
```diff
@@ -117,6 +115,19 @@ public static class TelemetryDataConstants
 
     public static string AttachmentsProcessingState = "VS.AttachmentsProcessing.State";
 
+    // *********************Test Sessions****************************
+    public static string ParallelEnabledDuringStartTestSession = "VS.TestSession.ParallelEnabled";
```
Evangelink, line 119: I know you only copy/pasted pattern in place but could we make the new added "constants" readonly?

```suggestion
    public static readonly string ParallelEnabledDuringStartTestSession = "VS.TestSession.ParallelEnabled";
```

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -48,10 +58,17 @@ public class ProxyTestSessionManager : IProxyTestSessionManager
...
     }
 
     /// <inheritdoc/>
     public virtual bool StartSession(ITestSessionEventsHandler eventsHandler)
+    {
+        return StartSession(eventsHandler, null);
+    }
+
+    /// <inheritdoc/>
+    public virtual bool StartSession(ITestSessionEventsHandler eventsHandler, IRequestData requestData)
```
Evangelink, line 71: Do we expect others to be able to override this implementation?

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -96,21 +116,45 @@ public virtual bool StartSession(ITestSessionEventsHandler eventsHandler)
...
+        // Collecting session metrics.
+        requestData?.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionId,
+            _testSessionInfo.Id);
+        requestData?.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionSpawnedTesthostCount,
+            _proxyContainerList.Count);
+        requestData?.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionTesthostSpawnTimeInSec,
+            stopwatch.Elapsed.TotalSeconds);
```
Evangelink, line 130: [minor] You will do multiple times the null check, it might be better (neglectable) to enclose the adds in a `if`.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -452,13 +451,18 @@ public void ResetOptions()
...
+        // Collect metrics & commands.
+        CollectMetrics(requestData, runConfiguration);
+        LogCommandsTelemetryPoints(requestData);
 
         lock (_syncObject)
         {
             try
             {
-                EqtTrace.Info("TestRequestManager.StartTestRunner: Synchronization context taken.");
+                EqtTrace.Info("TestRequestManager.StartTestSession: Synchronization context taken.");
```
Evangelink, line 465: I know we are pretty strict about backward compatibility. Do we consider that our "logs" are part of the API or not?

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
MarcoRossignoli: I think we can update others as well...looks safe

Evangelink: It will be a change to the public API (part of the file) so although it feels safe it's probably something we should agree upon and do as a separate PR.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -96,21 +116,45 @@ public virtual bool StartSession(ITestSessionEventsHandler eventsHandler)
...
+                TestSessionState.SessionExistsFailure.ToString());
             DisposeProxies();
             return false;
         }
 
+        requestData?.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionState,
+            TestSessionState.Success.ToString());
+
+        _testSessionStopwatch.Start();
```
MarcoRossignoli, line 156: why not include the time of "TestSessionTesthostSpawnTimeInSec" inside the session time?

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
MarcoRossignoli: Nope, we're updating to be able to find in a fast way during analysis. EqtTrace won't be shown to the user.

Evangelink: It's not written in a log or whatever? I mean we are not expecting users to have written something that would parse logs to look for something (or at least if they do, we "don't care" because that's not part of the API, right?)

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
cvpoienaru: I agree the change is needed but I also agree that a separate PR would probably be best. I'd like to keep the changes in this PR as relevant as possible to ease the review process.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
cvpoienaru: I agree, the logs aren't part of the API and I don't think we should treat them as such either going forward. The logs are a tool for us to find issues in TP, users can parse the console output and TRX but we should draw the line at logs (and to be completely honest TRX parsing is causing us a lot of headaches too but it's too late to change that now :D).

This approach (i.e. treating logs as API) would mostly hurt us because we won't be able to change them unless we add fresh logs while the benefit to the user is not immediately apparent. We haven't heard of anyone parsing the logs yet and even if they do I think we shouldn't support that behavior.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: `TestSessionTesthostSpawnTimeInSec` is the actual time we spent on spawning the initial set of testhosts and we'll know this piece of info immediately after we finish the session spinoff. This is a time during which we "boot up" the session so it shouldn't be counted as part of the total session time.

The total session time (i.e. time the session was up and running) is a separate measurement that we'll have info about once someone calls `StopTestSession` on the wrapper and as such ends the session.

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
MarcoRossignoli: oh we're shipping also this stuff public 😢 never mind.

MarcoRossignoli:
@cvpoienaru will you update this PR or our discussion is for follow-up PRs?

cvpoienaru:
> @cvpoienaru will you update this PR or our discussion is for follow-up PRs?

@MarcoRossignoli I'll create a follow up PR for that because this one is getting too large already.

src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -194,8 +194,6 @@ Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.IProxyExecutionManager.In
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.IProxyExecutionManager.IsInitialized.get -> bool
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.IProxyExecutionManager.StartTestRun(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestRunCriteria testRunCriteria, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestRunEventsHandler eventHandler) -> int
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.IProxyTestSessionManager
-Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.IProxyTestSessionManager.StartSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestSessionEventsHandler eventsHandler) -> bool
```
Evangelink, line 197: Are we okay with removing public APIs? I'd say breaking changes should probably be reserved for "major" version bump.

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
Evangelink: Could we still make the newly added constants to be `readonly`? We don't know when/if we will be able to change the others so let's try to make sure the new ones cannot be altered.

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -118,13 +118,10 @@ virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperatio
 virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.UpdateTestProcessStartInfo(Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProcessStartInfo testProcessStartInfo) -> Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProcessStartInfo
 virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager.DequeueProxy(string source, string runSettings) -> Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager
 virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager.EnqueueProxy(int proxyId) -> bool
-virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager.StartSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestSessionEventsHandler eventsHandler) -> bool
```
Evangelink, line 121: Same question about updating public API.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -20,6 +22,13 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine;
 /// </summary>
 public class ProxyTestSessionManager : IProxyTestSessionManager
 {
+    private enum TestSessionState : int
```
Evangelink, line 25: Fine to keep it as-is but `int` is already the default for `enum` so this could be "simplified".

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -20,6 +22,13 @@ namespace Microsoft.VisualStudio.TestPlatform.CrossPlatEngine;
...
+    {
+        Success = 0,
```
Evangelink, line 27: Would it be worth introducing an `unknown` state?

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
Evangelink, line 25: It seems like we are using the enum name and not its value for the telemetry, is there a reason to manually assign values?

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// Gets or sets the metrics.
+    /// </summary>
+    [DataMember]
+    public IDictionary<string, object> Metrics { get; set; } = null;
+
+    /// <summary>
+    /// Gets or sets a value indicating if the session was successfully stopped or not.
+    /// </summary>
+    [DataMember]
+    public bool IsStopped { get; set; } = false;
```
Evangelink, line 48: Could we end up in a "unknow" state that would justify using a nullable bool here?

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
```diff
@@ -272,6 +272,15 @@ public bool StopTestSession()
     /// <inheritdoc/>
     [Obsolete("This API is not final yet and is subject to changes.", false)]
     public bool StopTestSession(ITestSessionEventsHandler eventsHandler)
+    {
+        return StopTestSession(options: null, eventsHandler);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
+    public bool StopTestSession(
```
Evangelink, line 281: Isn't it better to preserve the original signature order? (i.e. keep the event handler arg first).

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
```diff
@@ -462,6 +472,15 @@ public async Task<bool> StopTestSessionAsync()
     /// <inheritdoc/>
     [Obsolete("This API is not final yet and is subject to changes.", false)]
     public async Task<bool> StopTestSessionAsync(ITestSessionEventsHandler eventsHandler)
+    {
+        return await StopTestSessionAsync(options: null, eventsHandler).ConfigureAwait(false);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
+    public async Task<bool> StopTestSessionAsync(
```
Evangelink, line 481: Isn't it better to preserve the original signature order? (i.e. keep the event handler arg first).

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -237,12 +237,26 @@ public void StartSession()
...
+    {
+        return StopTestSession(
+            testSessionInfo,
+            options: null,
+            eventsHandler);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
+    public bool StopTestSession(
```
Evangelink, line 249: Isn't it better to preserve the original signature order? (i.e. keep the event handler arg second and options last).

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -649,12 +663,26 @@ public async Task StartSessionAsync()
...
+    {
+        return await StopTestSessionAsync(
+            testSessionInfo,
+            options: null,
+            eventsHandler).ConfigureAwait(false);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
+    public async Task<bool> StopTestSessionAsync(
```
Evangelink, line 675: Isn't it better to preserve the original signature order? (i.e. keep the event handler arg second and options last).

test/TranslationLayer.UnitTests/TestSessionTests.cs
```diff
@@ -319,6 +320,23 @@ public void StopTestSessionShouldCallConsoleWrapperStopTestSessionWithCorrectArg
         _mockVsTestConsoleWrapper.Verify(
             vtcw => vtcw.StopTestSession(
                 _testSessionInfo,
+                null,
+                mockTestSessionEventsHandler2.Object),
+            Times.Once);
+    }
+
+    [TestMethod]
+    public void StopTestSessionShouldCallConsoleWrapperStopTestSessionWithCorrectArguments3()
```
Evangelink, line 329: [minor] Could we try to find a different name than simply a number suffix? Past experience tells me that it's hard to keep the original intent with numbered "generic" names.

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
MarcoRossignoli: it's ok...very confusing to dev...every time one failed build for newcomers 😄

src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
cvpoienaru: Yes, we should be ok with breaking the API. In this particular case we don't expect anyone other than us to use this interface (i.e. `IProxyTestSessionManager`) as it's not an extension point and it is buried deep in the `vstest.console` code with no current workflow triggering the usage of Test Session. The entry points of the Test Sessions API (found in `VsTestConsoleWrapper.cs` and `TestSession.cs`) are specifically marked as Obsolete because the API is not final. Furthermore we only expect TE in VS to trigger this workflow (and they haven't implement our changes yet) because they are the primary consumers of our translation layer while `dotnet test` doesn't make use of this layer.

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
cvpoienaru: Same answer as above should apply.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
cvpoienaru: We cannot end up in an unknown state, we either close the session successfully or we fail to close it because it's been already closed. I'd rather not introduce a nullable bool if we can avoid it as that would be an additional failure point that users of the `TranslationLayer` would have to guard against.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
Evangelink: Could we add that as a comment?

```suggestion
    public virtual /* for mocking purposes */ bool StartSession(ITestSessionEventsHandler eventsHandler, IRequestData requestData)
```

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: Not really, removing the values. Since the enum is private we can always add them later should we need them.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
cvpoienaru: Actually I did this to be consistent with the signature of the other methods in this class. For example, notice how adding a `TestPlatformOptions` parameter always keeps the events handler last:
```
public void RunTests(
    IEnumerable<TestCase> testCases,
    string runSettings,
    ITestRunEventsHandler testRunEventsHandler)
```
```
public void RunTests(
    IEnumerable<TestCase> testCases,
    string runSettings,
    TestPlatformOptions options,
    ITestRunEventsHandler testRunEventsHandler)
```

On top of this, I'd rather not go through rewiring the logic and fixing the tests if there's no immediate benefit to be gained.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
cvpoienaru: Same answer as above should apply.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
cvpoienaru: Same answer as above should apply.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: To be honest I don't see much value in introducing an `Unknown` state yet, but we'll benefit from it when we'll implement the telemetry revamp proposal (i.e. the one where we create actual classes for telemetry events) so I added it for future proofing. Then we'll need a default value there and that should probably be `Unknown` as you suggested.

test/TranslationLayer.UnitTests/TestSessionTests.cs
cvpoienaru: Done, thanks !

Evangelink (approved):
LGTM!

One small suggestion still open but feel free to ignore it.

Evangelink (approved):

cvpoienaru:
> LGTM!
> 
> One small suggestion still open but feel free to ignore it.

Fixed as discussed offline.

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,3 @@
+virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestSessionPool.KillSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> bool
+virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager.StartSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestSessionEventsHandler eventsHandler, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> bool
+virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager.StopSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> bool
```
MarcoRossignoli, line 3: ditto

src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
```diff
@@ -68,6 +68,11 @@ public TestEngine() : this(TestRuntimeProviderManager.Instance, new ProcessHelpe
         requestData.MetricsCollection.Add(
             TelemetryDataConstants.ParallelEnabledDuringDiscovery,
             parallelLevel > 1 ? "True" : "False");
+        requestData.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionId,
+            discoveryCriteria.TestSessionInfo != null
```
MarcoRossignoli, line 73: nit: `discoveryCriteria.TestSessionInfo?.Id ?? string.Empt`

src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
```diff
@@ -153,6 +158,11 @@ public TestEngine() : this(TestRuntimeProviderManager.Instance, new ProcessHelpe
         requestData.MetricsCollection.Add(
             TelemetryDataConstants.ParallelEnabledDuringExecution,
             parallelLevel > 1 ? "True" : "False");
+        requestData.MetricsCollection.Add(
+            TelemetryDataConstants.TestSessionId,
+            testRunCriteria.TestSessionInfo != null
```
MarcoRossignoli, line 163: ditto

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StartTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+/// Event arguments used to notify the caller about the status of the test session.
+/// </summary>
+[DataContract]
+public class StartTestSessionCompleteEventArgs : EventArgs
+{
+    /// <summary>
+    /// Gets or sets the test session info.
+    /// </summary>
+    [DataMember]
+    public TestSessionInfo TestSessionInfo { get; set; } = null;
```
MarcoRossignoli, line 20:
```suggestion
    public TestSessionInfo TestSessionInfo { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StartTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// <summary>
+    /// Gets or sets the metrics.
+    /// </summary>
+    [DataMember]
+    public IDictionary<string, object> Metrics { get; set; } = null;
```
MarcoRossignoli, line 26:
```suggestion
    public IDictionary<string, object> Metrics { get; set; }
```

ref type are null by default

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+    public StopTestSessionCompleteEventArgs(TestSessionInfo testSessionInfo)
+    {
+        TestSessionInfo = testSessionInfo;
+    }
+
+    /// <summary>
+    /// Gets or sets the test session info.
+    /// </summary>
+    [DataMember]
+    public TestSessionInfo TestSessionInfo { get; set; } = null;
```
MarcoRossignoli, line 36:
```suggestion
    public TestSessionInfo TestSessionInfo { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// Gets or sets the metrics.
+    /// </summary>
+    [DataMember]
+    public IDictionary<string, object> Metrics { get; set; } = null;
```
MarcoRossignoli, line 42:
```suggestion
    public IDictionary<string, object> Metrics { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
MarcoRossignoli, line 48:
```suggestion
    public bool IsStopped { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
```diff
@@ -0,0 +1,49 @@
...
+    { }
+
+    /// <summary>
+    /// Creates an instance of the current class.
+    /// </summary>
+    /// 
+    /// <param name="testSessionInfo">The test session info.</param>
+    public StopTestSessionCompleteEventArgs(TestSessionInfo testSessionInfo)
+    {
+        TestSessionInfo = testSessionInfo;
```
MarcoRossignoli, line 29: null check? do we expect possible null here?

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StartTestSessionAckPayload.cs
```diff
@@ -9,11 +9,12 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.Payloads;
...
+[DataContract]
 public class StartTestSessionAckPayload
 {
     /// <summary>
-    /// Gets or sets the test session info.
+    /// Gets or sets the event args.
     /// </summary>
     [DataMember]
-    public TestSessionInfo TestSessionInfo { get; set; }
+    public StartTestSessionCompleteEventArgs EventArgs { get; set; } = null;
```
MarcoRossignoli, line 19:
```suggestion
    public StartTestSessionCompleteEventArgs EventArgs { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionAckPayload.cs
```diff
@@ -9,17 +9,12 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.Payloads;
...
     /// </summary>
     [DataMember]
-    public TestSessionInfo TestSessionInfo { get; set; }
-
-    /// <summary>
-    /// Gets or sets a value indicating if the session was successfully stopped or not.
-    /// </summary>
-    [DataMember]
-    public bool IsStopped { get; set; }
+    public StopTestSessionCompleteEventArgs EventArgs { get; set; } = null;
```
MarcoRossignoli, line 19:
```suggestion
    public StopTestSessionCompleteEventArgs EventArgs { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionPayload.cs
```diff
@@ -0,0 +1,26 @@
...
+/// layers into design mode.
+/// </summary>
+[DataContract]
+public class StopTestSessionPayload
+{
+    /// <summary>
+    /// Gets or sets the test session info.
+    /// </summary>
+    [DataMember]
+    public TestSessionInfo TestSessionInfo { get; set; } = null;
```
MarcoRossignoli, line 19:
```suggestion
    public TestSessionInfo TestSessionInfo { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionPayload.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// <summary>
+    /// Gets or sets a flag indicating if metrics should be collected.
+    /// </summary>
+    [DataMember]
+    public bool CollectMetrics { get; set; } = false;
```
MarcoRossignoli, line 25:
```suggestion
    public bool CollectMetrics { get; set; }
```

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -4,3 +4,30 @@ Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.DiscoveryCriteria.TestSes
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.DiscoveryRequestPayload.TestSessionInfo.get -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.DiscoveryRequestPayload.TestSessionInfo.set -> void
 Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestPlatform.StartTestSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.StartTestSessionCriteria criteria, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestSessionEventsHandler eventsHandler) -> bool
+Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.StartTestSessionCompleteEventArgs
```
MarcoRossignoli, line 7: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSession.cs
```diff
@@ -195,6 +195,17 @@ public interface ITestSession : IDisposable, ITestSessionAsync
...
 
+    /// <summary>
+    /// Stops the test session.
+    /// </summary>
+    ///
+    /// <param name="options">Test Platform options.</param>
+    /// <param name="eventsHandler">The session event handler.</param>
+    /// 
+    /// <returns>True if the session was successfuly stopped, false otherwise.</returns>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 206: mmm weird Obsolete for a new api? Preview api(in preview package) are subject to change by design, I think it's not needed.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSessionAsync.cs
```diff
@@ -192,6 +192,19 @@ public interface ITestSessionAsync : IDisposable
...
 
+    /// <summary>
+    /// Stops the test session.
+    /// </summary>
+    ///
+    /// <param name="options">Test Platform options.</param>
+    /// <param name="eventsHandler">The session event handler.</param>
+    /// 
+    /// <returns>True if the session was successfuly stopped, false otherwise.</returns>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 203: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
```diff
@@ -85,6 +85,21 @@ public interface IVsTestConsoleWrapper : IVsTestConsoleWrapperAsync
...
+    /// <summary>
+    /// Stops the test session.
+    /// </summary>
+    /// 
+    /// <param name="testSessionInfo">The test session info object.</param>
+    /// <param name="options">Test Platform options.</param>
+    /// <param name="eventsHandler">The session event handler.</param>
+    /// 
+    /// <returns>True if the session was successfuly stopped, false otherwise.</returns>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 97: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapperAsync.cs
```diff
@@ -79,6 +79,19 @@ public interface IVsTestConsoleWrapperAsync
...
         ITestSessionEventsHandler eventsHandler);
 
+    /// <summary>
+    /// Asynchronous equivalent of <see cref="
+    /// IVsTestConsoleWrapper.StopTestSession(
+    ///     TestSessionInfo,
+    ///     TestPlatformOptions,
+    ///     ITestSessionEventsHandler)"/>.
+    /// </summary>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 89: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,2 +1,10 @@
 static Microsoft.VisualStudio.TestPlatform.VsTestConsole.TranslationLayer.Resources.Resources.CannotFindConsoleRunner.get -> string
-static Microsoft.VisualStudio.TestPlatform.VsTestConsole.TranslationLayer.Resources.Resources.ProcessStartWin32Failure.get -> string
\ No newline at end of file
+static Microsoft.VisualStudio.TestPlatform.VsTestConsole.TranslationLayer.Resources.Resources.ProcessStartWin32Failure.get -> string
+Microsoft.TestPlatform.VsTestConsole.TranslationLayer.Interfaces.IVsTestConsoleWrapper.StopTestSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestPlatformOptions options, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestSessionEventsHandler eventsHandler) -> bool
```
MarcoRossignoli, line 3: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
```diff
@@ -272,6 +272,15 @@ public bool StopTestSession()
     /// <inheritdoc/>
     [Obsolete("This API is not final yet and is subject to changes.", false)]
     public bool StopTestSession(ITestSessionEventsHandler eventsHandler)
+    {
+        return StopTestSession(options: null, eventsHandler);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 280: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -237,12 +237,26 @@ public void StartSession()
...
         ITestSessionEventsHandler eventsHandler)
+    {
+        return StopTestSession(
+            testSessionInfo,
+            options: null,
+            eventsHandler);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 248: ditto

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -649,12 +663,26 @@ public async Task StartSessionAsync()
...
         ITestSessionEventsHandler eventsHandler)
+    {
+        return await StopTestSessionAsync(
+            testSessionInfo,
+            options: null,
+            eventsHandler).ConfigureAwait(false);
+    }
+
+    /// <inheritdoc/>
+    [Obsolete("This API is not final yet and is subject to changes.", false)]
```
MarcoRossignoli, line 674: ditto

MarcoRossignoli (approved):
Some comments

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StartTestSessionCompleteEventArgs.cs
cvpoienaru: I know, just wanted to be extra clear.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StopTestSessionCompleteEventArgs.cs
cvpoienaru: No need to null check, if the `testSessionInfo` is null the `TestSessionInfo` property will be null too as it is by default.

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StartTestSessionAckPayload.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionAckPayload.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionPayload.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Payloads/StopTestSessionPayload.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSession.cs
cvpoienaru: As discussed with Artur, we want to be extra clear those APIs shouldn't be used yet. They already went through a few release cycles so they are out there, but this is a disclaimer we won't necessarily stand by them in their current form. When the API takes its final form we will remove the `Obsolete` attribute.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/ITestSessionAsync.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapper.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/Interfaces/IVsTestConsoleWrapperAsync.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/TestSession.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.ObjectModel/Client/Events/StartTestSessionCompleteEventArgs.cs
cvpoienaru: Same

src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
cvpoienaru: Will fix this in a subsequent PR as I don't want to go through the review cycle again.

src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
cvpoienaru: Will fix this in a subsequent PR as I don't want to go through the review cycle again.

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Unshipped.txt
cvpoienaru: Same

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/PublicAPI/PublicAPI.Unshipped.txt
cvpoienaru: Same

PR #3344, opened 2022-02-10
Update version of VS sdk build tools

Evangelink:
## Description

Update version of VS sdk build tools to fix error when building locally.


Evangelink:
Seems like there was something buggy in my `package` folder. After the clean, `build.cmd` works properly without this change locally.

Feel free to close the PR if you prefer to stick with current version of the VS SDK for now.

Haplois (approved):

PR #3345, opened 2022-02-10
Enable more rules

Evangelink:
## Description

Enable more rules that don't require any polyfilling/shim layer.


Evangelink:
@MarcoRossignoli If you want to have a look at these rules too.

MarcoRossignoli (approved):

MarcoRossignoli:
@Evangelink conflicts

PR #3347, opened 2022-02-10
Use fakes to surround vstest core and test it

nohwnd:
Adds a set of fakes (not Microsoft.Fakes), that replicate the basics of functionality that we need complete a test run. This is a first step towards having the ability to write "integration" tests that run in memory, and inject different kinds of failures. If you look at UnitTests1.cs, the code looks terribly complicated right now. But it will clean up when more relationships are clear. Even though there are many Fake implementations, just 3 or 4 are crucial. FakeTestRuntimeProvider, because it is an extension point, and it manages all aspects of setting up and tearing down a testhost. FakeCommunicationChannel, because that provides the way for posting and receiving messages that would normally go via sockets, and it allows me to receive and send messages. 

Only a single test is written here, and it depends on having just a single "testhost" and just a single communication channel to it. But it shows that the idea is possible. 

When moving forward with this I want to keep these rules in mind: 

- Keep automatic behavior to minimum. E.g. ProcessHelper.LaunchProcess should not try to simulate starting any process it is given, but rather find the process object in a list that was injected from test, or throw.
- Keep fake relationships as minimal as possible. The fake code should not need debugging in most cases.
- Record everything. Fakes try to record everything that they did, especially all events that were produced, that are waiting to be processed, and that were processed. So you can stop in a test and look at the state easily. 
- Pass all errors to ErrorAggregator so we can assert on them in the test. 
- Where possible use the real Fake* type rather than some interface, to avoid confusion about which dependencies are real an which are fake. 
- Review where the boundaries are. E.g. AssemblyMetadataProvider is better place to cut off, because that boils down to just few properties, than trying to replicate filesystem, and loading an assembly binary content from virtual file system. 
- Document where boundaries are and find out what is the limitation of this approach. 
-  Not planned: Add ways to test other pieces in isolation, like testhost as a whole, the pieces that we actually cut of (e.g. DotnetTestHostProvider, AssemblyMetadataProvider), and possibly add way to tie two sides together, so we test Sender <-> Handler interaction.
- Avoid breaking changes, and make this possible to merge.


src/Microsoft.TestPlatform.CommunicationUtilities/SocketClient.cs
```diff
@@ -96,17 +96,17 @@ private void OnServerConnected(Task connectAsyncTask)
...
             }
         }
     }
 
-    private void Stop(Exception error)
+    private void StopOnError(Exception error)
     {
         EqtTrace.Info("SocketClient.PrivateStop: Stop communication from server endpoint: {0}, error:{1}", _endPoint, error);
-
+        // TODO: this is here to prevent stack overflow.
```
nohwnd, line 109: remove todo: just keep the comment.
```suggestion
        // This is here to prevent stack overflow.
```

playground/MSTest1/MSTest1.csproj
```diff
@@ -6,7 +6,7 @@
   <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
 
   <PropertyGroup>
-    <TargetFrameworks Condition=" '$(OS)' == 'WINDOWS_NT' ">$(TargetFrameworks);net472</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' == 'WINDOWS_NT' ">$(TargetFrameworks);net472;netcoreapp3.1</TargetFrameworks>
```
nohwnd, line 9: revert

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -152,6 +152,15 @@ protected DataCollectionRequestHandler(IMessageSink messageSink, IRequestData re
...
         ValidateArg.NotNull(messageSink, nameof(messageSink));
+        // TODO: The MessageSink and DataCollectionRequestHandler have circular dependency.
+        // Message sink is injected into this Create method and then into constructor
+        // and into the constructor of DataCollectionRequestHandler. Data collection manager
+        // is then assigned to .Instace (which unlike many other .Instance is not populated
+        // directly in that property, but is created here). And then MessageSink depends on
+        // the .Instance. This is a very complicated way of solving the circular dependency,
+        // and should be replaced by adding a property to Message and assigning it.
+        // .Instance can then be removed.
+
```
nohwnd, line 163:
```suggestion
```

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -29,6 +29,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities;
 public class TestRequestSender : ITestRequestSender
 {
     // Time to wait for test host exit
+    // DONOTMERGE: this was 10s, I made it 1 second, because it makes my tests pass faster when I get in error state
```
nohwnd, line 32:
```suggestion
```

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -29,6 +29,8 @@ namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities;
...
+    // REVIEW: this was 10s, I made it 1 seconds
```
nohwnd, line 33:
```suggestion
```

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -158,6 +174,7 @@ public int InitializeCommunication()
         // Server start returns the listener port
         // return int.Parse(this.communicationServer.Start());
         var endpoint = _communicationEndpoint.Start(_connectionInfo.Endpoint);
+        // TODO: This is forcing us to use ip and port for communication
```
nohwnd, line 177:
```suggestion
        // TODO: This is forcing us to use IP address and port for communication
```

src/Microsoft.TestPlatform.ObjectModel/TestServiceLocator.cs
```diff
@@ -0,0 +1,66 @@
...
+#pragma warning disable RS0037 // Enable tracking of nullability of reference types in the declared API
+
+public static class TestServiceLocator
+{
+    public static Dictionary<string, object> Instances { get; } = new Dictionary<string, object>();
+    public static List<Resolve> Resolves { get; } = new();
+
+    public static void Register<TRegistration>(string name, TRegistration instance) where TRegistration : notnull
+    {
+        Instances.Add(name, instance);
```
Evangelink, line 24: [styling] Any preference between body and `=>` for one line methods/properties? I could turn on some rules.

test/Intent/Extensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public static List<MethodInfo> SkipExcluded(this IEnumerable<MethodInfo> e)
+    {
+        return e.Where(i =>
+        i.Name != nameof(object.ToString)
+        && i.Name != nameof(object.GetType)
+        && i.Name != nameof(object.GetHashCode)
+        && i.Name != nameof(object.Equals)
+        && i.GetCustomAttribute<ExcludeAttribute>() == null).ToList();
```
Evangelink, line 32: Nit:

```suggestion
            i.Name != nameof(object.ToString)
            && i.Name != nameof(object.GetType)
            && i.Name != nameof(object.GetHashCode)
            && i.Name != nameof(object.Equals)
            && i.GetCustomAttribute<ExcludeAttribute>() == null).ToList();
```

test/Intent/Runner.cs
```diff
@@ -0,0 +1,67 @@
...
+                var ts = asm.GetTypes().SkipNonPublic().SkipExcluded();
+                foreach (var t in ts)
+                {
+                    var ms = t.GetMethods().SkipExcluded();
+
+                    // TODO: This chooses the Only tests only for single assembly and single class,
+                    // to support this full we would have to enumerate all classes and methods first,
+                    // it is easy, I just don't need it right now.
+                    var only = ms.Where(m => m.GetCustomAttribute<OnlyAttribute>() != null).ToList();
+                    if (only.Any())
```
Evangelink, line 29: `Any` always causes evaluation but we have a list so the `Count` property is already processed.

```suggestion
                    if (only.Count > 0)
```

test/Intent/Runner.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Intent;
+
+using System.Reflection;
+
+public class Runner
+{
+    public static void Run(IEnumerable<string> path, IRunLogger logger)
```
Evangelink, line 10: Maybe `paths`.

test/Intent/Runner.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        foreach (var p in path)
+        {
+            try
+            {
+                var asm = Assembly.LoadFrom(p);
+                if (asm.IsExcluded())
+                    continue;
+
+                var ts = asm.GetTypes().SkipNonPublic().SkipExcluded();
```
Evangelink, line 20: Maybe `types`.

test/Intent/Runner.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Intent;
+
+using System.Reflection;
+
+public class Runner
+{
+    public static void Run(IEnumerable<string> path, IRunLogger logger)
+    {
+        foreach (var p in path)
```
Evangelink, line 12: Maybe `path` instead of `paths`, or `currentPath`.

test/Intent/Runner.cs
```diff
@@ -0,0 +1,67 @@
...
+            try
+            {
+                var asm = Assembly.LoadFrom(p);
+                if (asm.IsExcluded())
+                    continue;
+
+                var ts = asm.GetTypes().SkipNonPublic().SkipExcluded();
+                foreach (var t in ts)
+                {
+                    var ms = t.GetMethods().SkipExcluded();
```
Evangelink, line 23: Maybe `methods`.

test/vstest.ProgrammerTests/Fakes/EventRecord.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace vstest.ProgrammerTests.Fakes;
+
+internal class EventRecord<T>
```
Evangelink, line 6: For later, we should enable records.

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public FakeAssemblyMetadataProvider(FakeFileHelper fakeFileHelper, FakeErrorAggregator fakeErrorAggregator)
+    {
+        FakeFileHelper = fakeFileHelper;
+        FakeErrorAggregator = fakeErrorAggregator;
+    }
+
+    public Architecture GetArchitecture(string filePath)
+    {
+        var file = (FakeTestDllFile)FakeFileHelper.Files.Single(f => f.Path == filePath);
```
Evangelink, line 25: [minor] Could we manually throw an exception with some helpful message (what needs to be checked or changed) if we find more than one item (or don't find one).

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+        return file.Architecture;
+    }
+
+    public FrameworkName GetFrameWork(string filePath)
+    {
+        var file = (FakeTestDllFile)FakeFileHelper.Files.Single(f => f.Path == filePath);
```
Evangelink, line 31: [minor] Could we manually throw an exception with some helpful message (what needs to be checked or changed) if we find more than one item (or don't find one).

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+        FakeErrorAggregator = fakeErrorAggregator;
+    }
+
+    public Architecture GetArchitecture(string filePath)
+    {
+        var file = (FakeTestDllFile)FakeFileHelper.Files.Single(f => f.Path == filePath);
+        return file.Architecture;
+    }
+
+    public FrameworkName GetFrameWork(string filePath)
```
Evangelink, line 29: I would rename into `GetFramework` to match other namig case.

test/vstest.ProgrammerTests/Fakes/FakeProcessHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+internal class FakeProcessHelper : IProcessHelper
+{
+    // starting from 100 for no particular reason
+    // I want to avoid processId 0 and 1 as they are
+    // "reserved" on Windows (0) and Linux (both 0 and 1)
+    private int _lastProcessId = 100;
+
+    public FakeProcess CurrentProcess { get; }
+    public List<FakeProcess> Processes { get; } = new();
+    public int LastProcessId => _lastProcessId;
```
Evangelink, line 18: Replace by `get; private set;` property?

Evangelink (commented):
I will finish the review later, my brain is dead now.

src/Microsoft.TestPlatform.Common/Hosting/ITestRuntimeProviderManager.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Host;
```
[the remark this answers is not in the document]
nohwnd: header, usings, namespace, seemed most natural to me, and is also what the "move type to file" refactoring uses. But it is not what the current rules mandated. Personally I don't really care which one it will be. But I was fighting a bit to getting the build to pass because of all these issues that are not identified at first, and then in later builds they appear. It is not random, but I also don't feel like I am getting all warnings all the time, as you noticed.

src/Microsoft.TestPlatform.Common/Hosting/ITestRuntimeProviderManager.cs
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Microsoft.VisualStudio.TestPlatform.Common.Hosting;
+internal interface ITestRuntimeProviderManager
+{
+    ITestRuntimeProvider GetTestHostManagerByRunConfiguration(string runConfiguration);
+    ITestRuntimeProvider GetTestHostManagerByUri(string hostUri);
```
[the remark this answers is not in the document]
nohwnd: This is just extracting the current type into an interface. We could use uri, but as you say, not in this PR.

src/Microsoft.TestPlatform.ObjectModel/Client/TestRunCriteria.cs
```diff
@@ -482,7 +482,7 @@ public TestRunCriteria(IEnumerable<TestCase> tests, BaseTestRunCriteria baseTest
             runStatsChangeEventTimeout,
             testHostLauncher)
     {
-        var testCases = tests as IList<TestCase> ?? tests.ToList();
+        var testCases = tests as IList<TestCase> ?? tests?.ToList();
```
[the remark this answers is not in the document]
nohwnd: Not my code, I am just solving the null ref that happens if you don't provide IList. But yes I agree this is not looking great, but it is also in our actual public api, so imho it makes sense to take the most generic type we can.

src/Microsoft.TestPlatform.ObjectModel/TestServiceLocator.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel;
+
+// We don't want this in our shipped code. Build only for debug until I am able to remove it.
+#if DEBUG
```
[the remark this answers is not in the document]
nohwnd: Thanks, I did not think about that. I will keep as is for now.

src/Microsoft.TestPlatform.ObjectModel/TestServiceLocator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+#if !NETSTANDARD1_0
+using System;
+#endif
+using System.Collections.Generic;
+
+#pragma warning disable RS0016 // Add public types and members to the declared API
+#pragma warning disable RS0037 // Enable tracking of nullability of reference types in the declared API
+
+public static class TestServiceLocator
```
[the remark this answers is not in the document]
nohwnd: It could, but I would have to change internals visible to, and this won't be built for our release dll, so I will keep it as is, my goal is to be able to get rid of this anyway, because it is one of the things that prevent me from isolating my tests.

src/Microsoft.TestPlatform.ObjectModel/TestServiceLocator.cs
nohwnd: No strong preference, I would not impose more rules, until we can get the current rule set to work 100% all the time.

test/Intent.Primitives/TestResult.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Intent;
+
+public enum TestResult
+{
+    None = 0,
```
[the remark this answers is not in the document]
nohwnd: 😁it's a habit I am getting rid of.

test/Intent/Extensions.cs
```diff
@@ -0,0 +1,36 @@
...
+public static class Extensions
+{
+    public static bool IsExcluded(this Assembly asm)
+    {
+        return asm.CustomAttributes.Any(a => a.AttributeType == typeof(ExcludeAttribute));
+    }
+
+    public static List<Type> SkipExcluded(this IEnumerable<Type> e)
+    {
+        return e.Where(i => i.GetCustomAttribute<ExcludeAttribute>() == null).ToList();
```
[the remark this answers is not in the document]
nohwnd: This one should probably be just IEnumerable, if we went for performance.

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
nohwnd: I would as well, but this is the name coming from the interface.

src/Microsoft.TestPlatform.Common/Hosting/ITestRuntimeProviderManager.cs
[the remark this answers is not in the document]
Evangelink: > header, usings, namespace, seemed most natural to me

Will do a style change update later then.

> is also what the "move type to file" refactoring uses

On my side, the refactoring seems to be correctly following setup (either inside or outside).

> It is not random, but I also don't feel like I am getting all warnings all the time, as you noticed.

Same problem here.

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
nohwnd: Yes,  and also move this into FakeFileHelper.GetFakeFile.

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
```diff
@@ -101,7 +100,14 @@ public TestRequestSender(ProtocolConfig protocolConfig, ITestRuntimeProvider run
 
         // The connectionInfo here is that of RuntimeProvider, so reverse the role of runner.
         _runtimeProvider = runtimeProvider;
-        _communicationEndpoint = communicationEndPoint;
+
+        // TODO: In various places TestRequest sender is instantiated, and we can't easily inject the factory, so this is last
+        // resort of getting the dependency into the execution flow.
+        _communicationEndpoint = communicationEndPoint
+#if DEBUG
```
Evangelink, line 107: Just to make sure, when in debug we want to have the 3 possible paths, right?

test/Intent/ConsoleLogger.cs
```diff
@@ -0,0 +1,58 @@
...
+        // insert space before every capital letter or number that is after a non-capital letter
+        var spaced = Regex.Replace(noUnderscores, "(?<=[a-z])([A-Z0-9])", " $1");
+        // insert space before every capital leter that is after a number
+        var spaced2 = Regex.Replace(spaced, "(?<=[0-9]|^)([A-Z])", " $1");
+        var newLines = spaced2.Replace("When", $"{Environment.NewLine}     When")
+            .Replace("Then", $"{Environment.NewLine}     Then");
+
+        return newLines.ToLowerInvariant();
+    }
+
```
Evangelink, line 57: Nit

```suggestion
    }
```

test/Intent/Extensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        return e.Where(i =>
+            i.Name != nameof(object.ToString)
+            && i.Name != nameof(object.GetType)
+            && i.Name != nameof(object.GetHashCode)
+            && i.Name != nameof(object.Equals)
+            && i.GetCustomAttribute<ExcludeAttribute>() == null).ToList();
+    }
+
+
+}
```
Evangelink, line 36: Nit

```suggestion
    }
}
```

test/vstest.ProgrammerTests/Fakes/FakeProcessHelper.cs
nohwnd: Would be better to keep as is, but replace make _lastProcessid an Id, and call .Next() on it. Because that encapsulates the  work we do with interlocked.increment.

Evangelink (approved):
Some minor comments from me but otherwise LGTM! Thanks for this awesome improvement!

src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs
nohwnd: Yes, in debug we want to try the service locator before setting the channel.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -278,6 +278,10 @@ public void Abort()
             }
         }
 
+        // REVIEW: Added this, review this change. If we call abort, the event is never set, and we end up waiting for it to complete forever.
+        // I think w
```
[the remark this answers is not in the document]
nohwnd: this was removed entirely.

test/Intent/Runner.cs
nohwnd: done. FWIW probably makes no noticeable difference, especially considering we will have 0-2 methods in that list :)

test/Intent/Runner.cs
nohwnd: done

test/Intent/Runner.cs
nohwnd: done

test/Intent/Runner.cs
nohwnd: done, used path

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
nohwnd: done

test/vstest.ProgrammerTests/Fakes/FakeAssemblyMetadataProvider.cs
nohwnd: done

test/Intent/Runner.cs
nohwnd: done

test/vstest.ProgrammerTests/Fakes/FakeProcessHelper.cs
nohwnd: I removed the last process id property entirely, and used SequentialId.

test/Intent/ConsoleLogger.cs
nohwnd: fixed manually.

PR #3356, opened 2022-02-12
Enable even more rules

Evangelink:
cc @MarcoRossignoli

### DIFF
--- a/src/Microsoft.TestPlatform.AdapterUtilities/TestIdProvider.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/TestIdProvider.cs
@@ -14,9 +14,9 @@ public class TestIdProvider
     internal const int DigestBytes = DigestBits / 8;
 
     private Guid _id = Guid.Empty;
-    private byte[] _hash = null;
+    private byte[] _hash;
     private byte[] _lastBlock = new byte[BlockBytes];
-    private int _position = 0;
+    private int _position;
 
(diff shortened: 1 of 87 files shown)


src/Microsoft.TestPlatform.AdapterUtilities/TestIdProvider.cs
```diff
@@ -14,9 +14,9 @@ public class TestIdProvider
     internal const int DigestBytes = DigestBits / 8;
 
     private Guid _id = Guid.Empty;
-    private byte[] _hash = null;
```
Evangelink, line 17: @MarcoRossignoli There you go!

MarcoRossignoli (approved):

MarcoRossignoli:
we're missing only the null ref annotations...maybe we can start for new files/types and put some effort project by project...it could be a long task.

Evangelink:
I have a branch for it but that's pretty big effort. One solution could be to go with enable nullable and annotate all existing file with the disable so that every new added file follows nullable pattern and we could work on enabling it partially on existing files at a slower pace.

Evangelink:
@nohwnd Is it ok to squash and merge here?

MarcoRossignoli:
> I have a branch for it but that's pretty big effort. One solution could be to go with enable nullable and annotate all existing file with the disable so that every new added file follows nullable pattern and we could work on enabling it partially on existing files at a slower pace.

I think it's good...it's a great improvement anyway.

PR #3366, opened 2022-02-15
Remove regions

Evangelink:
## Description

[As per comment on another PR](https://github.com/microsoft/vstest/pull/3349#discussion_r806056391), I am opening this PR to remove all regions in the code.


MarcoRossignoli (approved):

nohwnd:
@MarcoRossignoli  @Evangelink  I was just going to say that the  regions in tests actually add value, and we should keep them before we replace them with something else. Please revert.

Apart from that I found 1 place where I would like to add a comment, commented on that place above.

Evangelink:
@nohwnd Want me to revert all or just test part?

PR #3371, opened 2022-02-15
Remove architecture validation

### LINKED ISSUE #3354: vstest.console blocks execution on arm64
vstest.console checks incoming settings and validates that the targetArchitecture and OSArchitecture are compatible. Unless the OSArchitecture and targetArchitecture are the same. 

(dotnet test on arm64 device will run as arm64 by default, so it will pass the architecture == osarchitecture) for this case, and for case where we ask for x64 || x86, it will pass the other check architecture == x86 || architecture == x64. 

BUT if we call vstest.console.exe (.net framework) it will run as emulated x64 (and show that as os architecture), required architecture is arm64 (different from os), and that is not x64 || x86, so it throws.

Remove that check.

nohwnd:
## Description
With multiple architectures being added, we are being too strict with this validation. Right now it blocks run of .NET tests on arm64 when vstest.console is emulated to x86. 

## Related issue
Fix #3354

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
+++ b/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
@@ -182,9 +182,6 @@ public class InferRunSettingsHelper
 
         EqtTrace.Verbose("Using effective platform:{0} effective framework:{1}", architecture, framework);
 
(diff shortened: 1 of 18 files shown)


Evangelink (approved):
PR that deletes code are the best!

PR #3377, opened 2022-02-16
Add post processing intergration test

MarcoRossignoli:
Add post processing intergration test

contributes to https://github.com/microsoft/vstest/issues/1811

### DIFF
--- a/test/Microsoft.TestPlatform.AcceptanceTests/PostProcessingTests.cs
+++ b/test/Microsoft.TestPlatform.AcceptanceTests/PostProcessingTests.cs
@@ -1,0 +1,122 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.TestPlatform.AcceptanceTests;
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


test/Microsoft.TestPlatform.AcceptanceTests/PostProcessingTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            Assert.IsTrue(line.StartsWith("SessionEnded_Handler_"));
+            fileContent.Add(line);
+        }
+
+        Assert.AreEqual(5, fileContent.Distinct().Count());
+    }
+
+    private static string GetRunsettingsFilePath(string resultsDir)
+    {
+        var runsettingsPath = Path.Combine(resultsDir, "test_" + Guid.NewGuid() + ".runsettings");
```
Evangelink, line 88: I don't think the guid is required as we have a unique folder name already.

jakubch1 (approved):

test/Microsoft.TestPlatform.AcceptanceTests/PostProcessingTests.cs
MarcoRossignoli: Copy pasted from another tests with runsettings. 
That guid is useful if we'll update test with different runsettings for different projects. I'll keep as-is.

PR #3380, opened 2022-02-16
Test adapter loading strategy

### LINKED ISSUE #2580: Provide a way to change the behavior when picking up default TestAdapters
## Issue
VSTest automatically picks up all the `*TestAdapter.dll` assemblies present with the assembly tested. We don't offer a way to change this yet. A customer reported on a [feedback ticket][1] that this delays their tests for a long time.

## Proposal
Since TestAdapters and DataCollectors can be in the same assembly, specifying a different load order is not feasible. We'll use the same loading order for the assemblies, and provide a way to exclude them.

   [1]: https://developercommunity.visualstudio.com/content/problem/1172102/xunit-discovery-is-slowing-down-tests-even-without.html

[AB#1370910](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1370910)

Haplois:
## Description
The code here is cherry-picked from our 17.1 release.

 - Fixes #2580 

## Work left
 - [x] Fix the bug where we fail to retrieve `TestAdapterLoadingStrategy` from `.runsettings` file.
 - [x] Fix typos in console help text.
 - [ ] Add documentation
 - [ ] Add tests


src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary>
+        /// A strategy not defined, Test Platfrom will load adapters normally. 
+        /// </summary>
+        Default = 0b0000_0000_0000_0000,
+
+        /// <summary>
+        /// Test Plarform will only load adapters specified by /TestAdapterPath (or RunConfiguration.TestAdaptersPaths node). 
+        /// If a specific adapter path is provided, adapter will be loaded; if a directory path is provided adapters directly in that folder will be loaded. 
+        /// If no adapter path is specified, test run will fail.
+        /// This will imply /InIsolation switch and force the tests to be run in an isolated process.
```
Haplois, line 23:
```suggestion
        /// Test Platform will only load adapters specified by /TestAdapterPath (or RunConfiguration.TestAdaptersPaths node). 
        /// If a specific adapter path is provided, adapter will be loaded; if a directory path is provided adapters directly in that folder will be loaded. 
        /// If no adapter path is specified, test run will fail.
        /// This will imply /InIsolation switch and force the tests to be run in an isolated process.
```

src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel
+{
+    /// <summary>
+    /// Represents a loading strategy
+    /// </summary>
+    [Flags]
+    internal enum TestAdapterLoadingStrategy
+    {
+        /// <summary>
+        /// A strategy not defined, Test Platfrom will load adapters normally.
```
Haplois, line 15:
```suggestion
        /// A strategy not defined, Test Platform will load adapters normally. 
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,258 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Haplois, line 1: Switch to file-scoped namespace.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            if (string.IsNullOrWhiteSpace(argument))
+            {
+
+
+                InitializeDefaultStrategy();
+                return;
+            }
+
+            InitializeStrategy(strategy);
```
Haplois, line 172:
```suggestion
            var strategy = TestAdapterLoadingStrategy.Default;

            if (!string.IsNullOrEmpty(argument) && !Enum.TryParse(argument, out strategy))
            {
                throw new CommandLineException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestAdapterLoadingStrategyValueInvalid, argument));
            }

            if (strategy == TestAdapterLoadingStrategy.Recursive) {
                throw new CommandLineException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestAdapterLoadingStrategyValueInvalidRecursive, $"{nameof(TestAdapterLoadingStrategy.Explicit)}, {nameof(TestAdapterLoadingStrategy.NextToSource)}"));
            }

            if (string.IsNullOrWhiteSpace(argument))
            {            
                InitializeDefaultStrategy();
                return;
            }

            InitializeStrategy(strategy);
```

src/vstest.console/Processors/Utilities/HelpContentPriority.cs
```diff
@@ -35,7 +36,7 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine.Processors;
 /// --Port
 ///
 /// Help
-/// -�Help
+/// -�Help
```
Haplois, line 39:
```suggestion
/// --Help
```

src/vstest.console/Resources/Resources.resx
```diff
@@ -560,6 +560,44 @@
...
+
+        - DefaultRuntimeProviders: Load default runtime providers shipped with Test Platform. 
+                                   If this is not specified when "Explicit" option is set, a test host provider need to be specified explicitly.
+
+        - ExtensionsDirectory: Load adapters inside Extensions folder.
+
+        - NextToSource: Load adapters next to source. 
+
+        - Recursive: Recursively search folders when loading adapters. This requires "Explicit" or "NextToSource" to be specified too.
+    </value>
```
Haplois, line 584:
```suggestion
    <value>--TestAdapterLoadingStrategy|/TestAdapterLoadingStrategy:&lt;strategy&gt;
      This affects adapter loading behavior.

      Currently supported behaviors:
        - Explicit: Test Platform will only load adapters specified by /TestAdapterPath (or RunConfiguration.TestAdaptersPaths node). 
                    If a specific adapter path is provided, adapter will be loaded; if a directory path is provided adapters directly in that folder will be loaded, unless Recursive option is also specified.
                    If no adapter path is specified, test run will fail.
                    This will imply /InIsolation switch and force the tests to be run in an isolated process.
        
        - Default: Test Platform will load adapters is if this argument has not been specified. 
                   It will pick up extensions from next to source, provided additional adapter paths and from the default directory.

        - DefaultRuntimeProviders: Load default runtime providers shipped with Test Platform. 
                                   If this is not specified when "Explicit" option is set, a test host provider need to be specified explicitly.

        - ExtensionsDirectory: Load adapters inside Extensions folder.

        - NextToSource: Load adapters next to source. 

        - Recursive: Recursively search folders when loading adapters. This requires "Explicit" or "NextToSource" to be specified too.
    </value>
```

test/vstest.console.UnitTests/Processors/TestAdapterLoadingStrategyArgumentProcessorTests.cs
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Haplois, line 1: Switch to file-scoped namespace

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -177,6 +164,20 @@ public TestPlatform()
 
         return testSessionManager.StartSession(eventsHandler, requestData);
     }
+    private void PopulateExtensions(string runSettings, IEnumerable<string> sources)
```
Evangelink, line 167: [cosmetic]
```suggestion

    private void PopulateExtensions(string runSettings, IEnumerable<string> sources)
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -211,81 +212,54 @@ public void ClearExtensions()
...
-    /// <param name="testRunCriteria">The test run criteria.</param>
-    private void AddExtensionAssembliesFromSource(TestRunCriteria testRunCriteria)
+    /// <param name="sources">The list of sources.</param>
+    private void AddExtensionAssembliesFromSource(IEnumerable<string> sources, TestAdapterLoadingStrategy strategy)
     {
-        IEnumerable<string> sources = testRunCriteria.Sources;
-        if (testRunCriteria.HasSpecificTests)
+        // Skip discovery unless we're using the default behavior, or NextToSource is specified.
+        if (strategy != TestAdapterLoadingStrategy.Default 
+            && (strategy & TestAdapterLoadingStrategy.NextToSource) != TestAdapterLoadingStrategy.NextToSource)
```
Evangelink, line 244: I don't know if you are using this syntax for the perf impact it used to have? I usually find the `HasFlag` function easier to understand.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+            if (strategy == TestAdapterLoadingStrategy.Default
+                || (strategy & TestAdapterLoadingStrategy.ExtensionsDirectory) == TestAdapterLoadingStrategy.ExtensionsDirectory)
+            {
+                defaultExtensionPaths = fileHelper
+                    .EnumerateFiles(extensionsFolder, SearchOption.TopDirectoryOnly, ".dll", ".exe")
+                    .Union(defaultExtensionPaths);
+            }
+        }
+
+        TestPluginCache.Instance.DefaultExtensionPaths = defaultExtensionPaths.Distinct();
```
Evangelink, line 318: Is it really worth having the lazy behavior?

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+    }
+
+    private static SearchOption GetSearchOption(TestAdapterLoadingStrategy strategy, SearchOption defaultStrategyOption) {
+        if (strategy == TestAdapterLoadingStrategy.Default) {
+            return defaultStrategyOption;
+        }
```
Evangelink, line 324: [cosmetic]

```suggestion
    private static SearchOption GetSearchOption(TestAdapterLoadingStrategy strategy, SearchOption defaultStrategyOption) 
    {
        if (strategy == TestAdapterLoadingStrategy.Default) 
        {
            return defaultStrategyOption;
        }
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+
+        var searchOption = SearchOption.TopDirectoryOnly;
+        if ((strategy & TestAdapterLoadingStrategy.Recursive) == TestAdapterLoadingStrategy.Recursive)
+        {
+            searchOption = SearchOption.AllDirectories;
+        }
+
+        return searchOption;
```
Evangelink, line 332: Feel free to ignore this one if you don't like it

```suggestion
        return (strategy & TestAdapterLoadingStrategy.Recursive) == TestAdapterLoadingStrategy.Recursive
            ? SearchOption.AllDirectories
            : SearchOption.TopDirectoryOnly;     
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+
+        // Default behavior is to only accept directories!
+        if (strategy == TestAdapterLoadingStrategy.Default)
+        {
+            return ExpandAdaptersWithDefaultStrategy(adapterPath, fileHelper);
+        }
+
+        var adapters = ExpandAdaptersWithExplicitStrategy(adapterPath, fileHelper, strategy);
+
+        return adapters.Distinct();
```
Evangelink, line 347: Why distinct here but not on line 342 (`TestAdapterLoadingStrategy.Explicit`)?

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+    }
+
+    private static IEnumerable<string> ExpandAdaptersWithExplicitStrategy(string path, IFileHelper fileHelper, TestAdapterLoadingStrategy strategy)
```
Evangelink, line 350: Feel free to ignore but I would rename `path` into `fileOrDirectoryPath` (or `fileOrFolderPath`).

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+                TestPlatformConstants.RunTimeEndsWithPattern);
+
+            return adapterPaths;
+        }
+
+        EqtTrace.Warning(string.Format(CultureInfo.InvariantCulture, "AdapterPath Not Found: {0}", path));
+        return Enumerable.Empty<string>();
+    }
+
+    private static IEnumerable<string> ExpandAdaptersWithDefaultStrategy(string path, IFileHelper fileHelper)
```
Evangelink, line 380: Again, feel free to ignore but I would rename `path` into `directoryPath` (or `folderPath`).

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsUtilities.cs
```diff
@@ -190,4 +190,14 @@ public static IEnumerable<string> GetTestAdaptersPaths(string runSettings)
...
 
+    /// <summary>
+    /// Gets the test adapter loading strategy
+    /// </summary>
+    /// <param name="runSettings">Test run settings</param>
+    /// <returns>Test adapter loading strategy</returns>
+    internal static TestAdapterLoadingStrategy GetLoadingStrategy(string runSettings) {
+        var runConfiguration = XmlRunSettingsUtilities.GetRunConfigurationNode(runSettings);
+        
+        return runConfiguration?.TestAdapterLoadingStrategy ?? TestAdapterLoadingStrategy.Default;
```
MarcoRossignoli, line 201: I see nullcheck here but not in other place where we use `XmlRunSettingsUtilities.GetRunConfigurationNode(runSettings);` is it ok?

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -211,6 +211,12 @@ public bool DisableAppDomain
         }
     }
 
+    /// <summary>
+    /// Gets or sets the test adapter loading strategy.
+    /// </summary>
+    internal TestAdapterLoadingStrategy TestAdapterLoadingStrategy { get; set; }
+
+
```
MarcoRossignoli, line 219:
```suggestion
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+        if ((strategy & TestAdapterLoadingStrategy.Explicit) != TestAdapterLoadingStrategy.Explicit)
+        {
+            return Enumerable.Empty<string>();
         }
+
+        if (fileHelper.Exists(path))
+        {
+            return new[] { path };
+        }
+        else if (fileHelper.DirectoryExists(path))
```
Evangelink, line 361: I would replace all the code from this point to the end of the method with a call to `TestAdapterLoadingStrategy.Explicit` which is doing exactly the same thing.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+    {
+        IEnumerable<string> sources = testRunCriteria.Sources;
+        if (testRunCriteria.HasSpecificTests)
+        {
+            // If the test execution is with a test filter, group them by sources.
+            sources = testRunCriteria.Tests.Select(tc => tc.Source).Distinct();
+        }
+
+        return sources;
     }
```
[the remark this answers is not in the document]
Evangelink: Or even
```suggestion
    => testRunCriteria.HasSpecificTests
        // If the test execution is with a test filter, group them by sources.
        ? sources = testRunCriteria.Tests.Select(tc => tc.Source).Distinct()
      	: testRunCriteria.Sources;
```

src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// Load adapters inside Extensions folder.
+    /// </summary>
+    ExtensionsDirectory  = 0b0000_0000_0000_1000,
+
+    /// <summary>
+    /// Directory wide searches will be recursive, this is required to be with with NextToSource or Explicit.
+    /// </summary>
+    Recursive = 0b0001_0000_0000_0000,
```
MarcoRossignoli, line 45: I don't know if it's ok put `Recursive` as-is if I don't want recursive for some place due to performance reason I cannot...but we can do it in next iteration, this is internals(`NextToSourceRecursive`), to think also if we want `XXXAsc/XXXDesc` because orders matter today.

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsUtilities.cs
```diff
@@ -190,4 +190,14 @@ public static IEnumerable<string> GetTestAdaptersPaths(string runSettings)
         return testAdaptersPaths;
     }
 
+    /// <summary>
+    /// Gets the test adapter loading strategy
+    /// </summary>
+    /// <param name="runSettings">Test run settings</param>
+    /// <returns>Test adapter loading strategy</returns>
+    internal static TestAdapterLoadingStrategy GetLoadingStrategy(string runSettings) {
```
Evangelink, line 198: [cosmetic]

```suggestion
    internal static TestAdapterLoadingStrategy GetLoadingStrategy(string runSettings) 
    {
```

src/vstest.console/CommandLine/Executor.cs
```diff
@@ -222,7 +222,14 @@ private int GetArgumentProcessors(string[] args, out List<IArgumentProcessor> pr
...
         // should always be executed.
         var processorsToAlwaysExecute = processorFactory.GetArgumentProcessorsToAlwaysExecute();
-        processors.AddRange(processorsToAlwaysExecute);
+        foreach (var processor in processorsToAlwaysExecute)
+        {
+            if (processors.Any(i => i.Metadata.Value.CommandName == processor.Metadata.Value.CommandName)) {
+                continue;
+            }
+
+            processors.Add(ArgumentProcessorFactory.WrapLazyProcessorToInitializeOnInstantiation(processor));
```
MarcoRossignoli, line 231: Why do we need this one, can you add a comment?

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -501,8 +503,15 @@ public override XmlElement ToXml()
             root.AppendChild(testAdaptersPaths);
         }
 
+        if(this.TestAdapterLoadingStrategy != TestAdapterLoadingStrategy.Default)
```
Evangelink, line 506: [cosmetic]
```suggestion
        if (this.TestAdapterLoadingStrategy != TestAdapterLoadingStrategy.Default) 
```

src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    Explicit = 0b0000_0000_0000_0001,
+
+    /// <summary>
+    /// Load adapters next to source. 
+    /// </summary>
+    NextToSource = 0b0000_0000_0000_0010,
+
+    /// <summary>
+    /// Default runtime providers inside Extensions folder will be included
```
Evangelink, line 33: Nit:

```suggestion
    /// Default runtime providers inside Extensions folder will be included.
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+    public const string ExplicitStrategy = "Explicit";
+
+    /// <summary>
+    /// Default constructor.
+    /// </summary>
+    /// <param name="options"> The options. </param>
+    /// <param name="testPlatform">The test platform</param>
+    public TestAdapterLoadingStrategyArgumentExecutor(CommandLineOptions options, IRunSettingsProvider runSettingsManager, IOutput output, IFileHelper fileHelper)
+    {
+        Contract.Requires(options != null);
```
MarcoRossignoli, line 146: are we using code contracts yet?

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+    public const string RunSettingsPath = "RunConfiguration.TestAdapterLoadingStrategy";
+    public const string DefaultStrategy = "Default";
+    public const string ExplicitStrategy = "Explicit";
+
+    /// <summary>
+    /// Default constructor.
+    /// </summary>
+    /// <param name="options"> The options. </param>
+    /// <param name="testPlatform">The test platform</param>
+    public TestAdapterLoadingStrategyArgumentExecutor(CommandLineOptions options, IRunSettingsProvider runSettingsManager, IOutput output, IFileHelper fileHelper)
```
MarcoRossignoli, line 144:
```suggestion
    public TestAdapterLoadingStrategyArgumentExecutor(CommandLineOptions options!!, IRunSettingsProvider runSettingsManager!!, IOutput output!!, IFileHelper fileHelper!!)
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+    {
+        Contract.Requires(options != null);
+
+        _commandLineOptions = options;
+        _runSettingsManager = runSettingsManager;
+        _output = output;
+        _fileHelper = fileHelper;
+    }
+
+    #region IArgumentExecutor
```
MarcoRossignoli, line 154: do we need region here?

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+
+    public override HelpContentPriority HelpPriority => HelpContentPriority.TestAdapterLoadingStrategyArgumentProcessorHelpPriority;
+}
+
+/// <summary>
+/// The argument executor.
+/// </summary>
+internal class TestAdapterLoadingStrategyArgumentExecutor : IArgumentExecutor
+{
+    #region Fields
```
Evangelink, line 104: Please remove this region.

MarcoRossignoli (approved):
Some comments

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+    /// </summary>
+    private readonly IOutput _output;
+
+    /// <summary>
+    /// For file related operation
+    /// </summary>
+    private readonly IFileHelper _fileHelper;
+
+
+    private static readonly string[] EmptyStringArray =
```
Evangelink, line 126: I want to add a global helper for that but I have some trouble with my shim layer...

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+#if NET451
+                                                        new string[0];
```
Evangelink, line 128: [cosmetic] Feel free to ignore

```suggestion
        new string[0];
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+#else
+                                                        Array.Empty<string>();
```
Evangelink, line 130: [cosmetic] Feel free to ignore

```suggestion
        Array.Empty<string>();
```

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsUtilities.cs
Haplois: Here, it might be null, in other places we always have one.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+#endif
+
+    #endregion
+
+    public const string RunSettingsPath = "RunConfiguration.TestAdapterLoadingStrategy";
+    public const string DefaultStrategy = "Default";
+    public const string ExplicitStrategy = "Explicit";
```
Evangelink, line 137: Do we want to use `static readonly` instead? It should probably not matter in this case.

src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
Haplois: So, `summary` is not correct here. `Recursive` only works `NextToSource` or `Explicit`. If this is not provider all searches are not recursive by default.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        for (var i = 0; i < testAdapterPaths.Length; i++)
+        {
+            var adapterPath = testAdapterPaths[i];
+            var testAdapterPath = _fileHelper.GetFullPath(Environment.ExpandEnvironmentVariables(adapterPath));
+
+            if (strategy == TestAdapterLoadingStrategy.Default)
+            {
+                if (!_fileHelper.DirectoryExists(testAdapterPath))
```
Evangelink, line 239: Could be merged with previous if.

src/vstest.console/Processors/TestAdapterPathArgumentProcessor.cs
```diff
@@ -101,7 +100,16 @@ internal class TestAdapterPathArgumentExecutor : IArgumentExecutor
     /// <summary>
     /// Separators for multiple paths in argument.
     /// </summary>
-    private readonly char[] _argumentSeparators = new[] { ';' };
+    internal readonly static char[] ArgumentSeparators = new[] { ';' };
+
+    private static readonly string[] EmptyStringArray =
+#if NET451
+                                                        new string[0];
```
Evangelink, line 107: [cosmetic]

```suggestion
        new string[0];
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
Haplois: I like to keep them when we implement an interface, it makes reading the code easier.

src/vstest.console/Processors/TestAdapterPathArgumentProcessor.cs
```diff
@@ -101,7 +100,16 @@ internal class TestAdapterPathArgumentExecutor : IArgumentExecutor
...
+#else
+                                                        Array.Empty<string>();
```
Evangelink, line 109: [cosmetic]

```suggestion
        Array.Empty<string>();
```

src/vstest.console/Processors/TestAdapterPathArgumentProcessor.cs
```diff
@@ -101,7 +100,16 @@ internal class TestAdapterPathArgumentExecutor : IArgumentExecutor
...
+#endif
+
+    public const string RunSettingsPath = "RunConfiguration.TestAdaptersPaths";
```
Evangelink, line 112: Same remark on `const` vs `static readnly`.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
Haplois: I am not sure. I just copied another argument executor :)

src/vstest.console/CommandLine/Executor.cs
Haplois, line 231:
```suggestion
            // We need to initialize the argument executor if it's set to always execute. This ensures it will be initialized with other executors.
            processors.Add(ArgumentProcessorFactory.WrapLazyProcessorToInitializeOnInstantiation(processor));
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -156,9 +141,11 @@ public TestPlatform()
         StartTestSessionCriteria testSessionCriteria!!,
         ITestSessionEventsHandler eventsHandler)
     {
-        AddExtensionAssemblies(testSessionCriteria.RunSettings);
-
         var runConfiguration = XmlRunSettingsUtilities.GetRunConfigurationNode(testSessionCriteria.RunSettings);
+        var strategy = runConfiguration.TestAdapterLoadingStrategy;
```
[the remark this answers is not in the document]
Haplois: `GetRunConfigurationNode` will always return a value. It's never null.

src/Microsoft.TestPlatform.Common/Utilities/RunSettingsUtilities.cs
Haplois: It's not required here too. Fixed.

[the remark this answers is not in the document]
Haplois: It's not required here too, fixed.

src/vstest.console/CommandLine/Executor.cs
```diff
@@ -222,7 +222,14 @@ private int GetArgumentProcessors(string[] args, out List<IArgumentProcessor> pr
         // Examples: processors to enable loggers that are statically configured, and to start logging,
         // should always be executed.
         var processorsToAlwaysExecute = processorFactory.GetArgumentProcessorsToAlwaysExecute();
-        processors.AddRange(processorsToAlwaysExecute);
+        foreach (var processor in processorsToAlwaysExecute)
+        {
+            if (processors.Any(i => i.Metadata.Value.CommandName == processor.Metadata.Value.CommandName)) {
+                continue;
+            }
```
nohwnd, line 229: Why is this needed? We grab **processors to always execute** and exclude some of them from execution.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+    /// </summary>
+    private readonly IFileHelper _fileHelper;
+
+
+    private static readonly string[] EmptyStringArray =
+#if NET451
+                                                        new string[0];
+#else
+                                                        Array.Empty<string>();
+#endif
```
nohwnd, line 131: Use just the old syntax? Or is that preventing build? in that case suppress inline.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
nohwnd, line 137: use nameof?

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
nohwnd, line 146: !! here any anywhere else? (feel free to ignore)

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+using Microsoft.VisualStudio.TestPlatform.Utilities.Helpers.Interfaces;
+
+using CommandLineResources = Microsoft.VisualStudio.TestPlatform.CommandLine.Resources.Resources;
+
+/// <summary>
+/// Allows the user to specify a order of loading custom adapters from.
+/// </summary>
+internal class TestAdapterLoadingStrategyArgumentProcessor : IArgumentProcessor
+{
+    #region Constants
```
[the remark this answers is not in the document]
Haplois: good point.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
Haplois: I hope to do it as well... This was a quick and dirty solution. I am planning refactor all these processors soon. Will do it there.

src/vstest.console/Processors/TestAdapterPathArgumentProcessor.cs
```diff
@@ -101,7 +100,16 @@ internal class TestAdapterPathArgumentExecutor : IArgumentExecutor
     /// <summary>
     /// Separators for multiple paths in argument.
     /// </summary>
-    private readonly char[] _argumentSeparators = new[] { ';' };
+    internal readonly static char[] ArgumentSeparators = new[] { ';' };
+
+    private static readonly string[] EmptyStringArray =
```
nohwnd, line 105: Same here. Just use the old less effective syntax.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+
+    /// <summary>
+    /// The name of the command line argument that the TestAdapterLoadingStrategyArgumentProcessor handles.
+    /// </summary>
+    public const string CommandName = "/TestAdapterLoadingStrategy";
+
+    #endregion
+
+    private Lazy<IArgumentProcessorCapabilities> _metadata;
```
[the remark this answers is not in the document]
Haplois: Correction, I will do it later on, I need to change how this class looks with others.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
Haplois: Any advantages? I like that `conts` are actually compiled into binary as strings. Also, `DefaultStrategy` and `ExplicitStrategy` will be removed before this PR is merged.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Haplois: Old habits die hard :) I'll fix it.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
Haplois: These are relics of an old time. I'll remove it entirely in this PR.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Evangelink: It's ok to keep it you prefer. I think the perf fix was done only in new framework.

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
MarcoRossignoli: remove it

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            throw new CommandLineException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.TestAdapterPathValueRequiredWhenStrategyXIsUsed, ExplicitStrategy));
+        }
+
+        SetStrategy(strategy);
+    }
+
+    private void ForceIsolation()
+    {
+        if (_commandLineOptions.InIsolation)
```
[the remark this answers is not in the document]
MarcoRossignoli: yep, we have the ticket in the backlog...we can prioritize.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -177,6 +164,20 @@ public TestPlatform()
 
         return testSessionManager.StartSession(eventsHandler, requestData);
     }
+    private void PopulateExtensions(string runSettings, IEnumerable<string> sources)
+    {
+        var runConfiguration = XmlRunSettingsUtilities.GetRunConfigurationNode(runSettings);
+        var strategy = runConfiguration.TestAdapterLoadingStrategy;
```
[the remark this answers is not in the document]
Haplois: Removed that check, it's never null.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Haplois: Probably no, but it's the default behavior here, I'll hopefully remove this code path entirely later on.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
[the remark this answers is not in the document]
Haplois: I use VS main, I'll check why it's not crying...

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Haplois: To keep default behavior constant. It's probably OK to add there, but we didn't have it before :)

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -303,21 +277,133 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources)
...
+                TestPlatformConstants.DataCollectorEndsWithPattern,
+                TestPlatformConstants.RunTimeEndsWithPattern);
+    }
+
+    private static IEnumerable<string> GetSources(TestRunCriteria testRunCriteria)
+    {
+        IEnumerable<string> sources = testRunCriteria.Sources;
+        if (testRunCriteria.HasSpecificTests)
+        {
+            // If the test execution is with a test filter, group them by sources.
```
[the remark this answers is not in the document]
Haplois: That's a legacy comment. Removed.

src/Microsoft.TestPlatform.ObjectModel/TestAdapterLoadingStrategy.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Microsoft.VisualStudio.TestPlatform.ObjectModel;
+
+using System;
+
+/// <summary>
+/// Represents a loading strategy
+/// </summary>
+[Flags]
+internal enum TestAdapterLoadingStrategy
```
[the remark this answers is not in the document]
Haplois: This enum is a `middleground` for now. I have some legacy behavior to fix after this PR. When that's fixed, I'll also add support for loading order. I can't really represent the default case with a combination of these other elements. So, if default is specified it doesn't support anything else.

We're not exposing this enum anywhere, it is stored as a comma seperated string. The plan is process that direcly. We won't be having a `None` there. I don't want to have a legacy behavior as too.

Why I called `0` default? I guess I was lazy :)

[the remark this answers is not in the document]
Haplois: As for binaries, It was easier to define the values, I'll conver those back to decimal, if you prefer.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -201,13 +201,13 @@ public void ClearExtensions()
...
 
-    private void ThrowExceptionIfTestHostManagerIsNull(
+    private static void ThrowExceptionIfTestHostManagerIsNull(
         ITestRuntimeProvider testHostManager,
         string settingsXml)
     {
         if (testHostManager == null)
         {
-            EqtTrace.Error("TestPlatform.CreateTestRunRequest: No suitable testHostProvider found for runsettings : {0}", settingsXml);
+            EqtTrace.Error($"{nameof(TestPlatform)}.{nameof(ThrowExceptionIfTestHostManagerIsNull)}: No suitable testHostProvider found for runsettings : {settingsXml}");
```
Evangelink, line 210: Nit.

```suggestion
            EqtTrace.Error($"{nameof(TestPlatform)}.{nameof(ThrowExceptionIfTestHostManagerIsNull)}: No suitable testHostProvider found for runsettings: {settingsXml}");
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -91,15 +90,15 @@ public TestPlatform()
         PopulateExtensions(discoveryCriteria.RunSettings, discoveryCriteria.Sources);
 
         // Initialize loggers.
-        var loggerManager = TestEngine.GetLoggerManager(requestData);
+        ITestLoggerManager loggerManager = TestEngine.GetLoggerManager(requestData);
```
Evangelink, line 93: If we all agree, I could change the `var` rules in the repo to favor explicit type when not obvious.

My opinion: I usually prefer `var` in VS but that's true that `var` in GitHub review is a bit of a pain.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -277,37 +276,42 @@ private void AddExtensionAssembliesFromSource(IEnumerable<string> sources, TestA
...
-        var strategy = runConfiguration.TestAdapterLoadingStrategy;
-
-        var fileHelper = new FileHelper();
-        var defaultExtensionPaths = Enumerable.Empty<string>();
+        // This method needs to run statically before we have any adapter discovery.
+        // TestHostProviderManager get initialized just after this call and it
+        // requires DefaultExtensionPaths to be set to resolve a TestHostProvider.
+        // Since it's static, it forces us to set the adapter paths.
+        //
+        // Otherwise we will always get an "No suitable test runtime provider found for this run." error.
```
Evangelink, line 284: Nit

```suggestion
        // Otherwise we will always get a "No suitable test runtime provider found for this run." error.
```

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Evangelink: My bad! Was referring to the following method (i.e. `ExpandAdaptersWithDefaultStrategy`).

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -211,13 +207,19 @@ private void ForceIsolation()
             return;
         }
 
+        EqtTrace.Warning(
+            $"{nameof(TestAdapterLoadingStrategyArgumentExecutor)}.{nameof(ForceIsolation)}: InIsolation setting is forced when {nameof(TestAdapterLoadingStrategy.Explicit)} strategy is used." +
+            "Test's will run in isolation."
```
Evangelink, line 212: Assuming you meant plural form

```suggestion
            "Tests will run in isolation."
```

src/vstest.console/Processors/TestAdapterLoadingStrategyArgumentProcessor.cs
```diff
@@ -211,13 +207,19 @@ private void ForceIsolation()
...
+        );
         _commandLineOptions.InIsolation = true;
         _runSettingsManager.UpdateRunSettingsNode(InIsolationArgumentExecutor.RunSettingsPath, "true");
     }
 
     private void ValidateTestAdapterPaths(TestAdapterLoadingStrategy strategy)
     {
-        var testAdapterPaths = _commandLineOptions.TestAdapterPath ?? EmptyStringArray;
+#pragma warning disable CA1825 // Avoid zero-length array allocations
```
Evangelink, line 220: Rule is currently setup to be only `Info` so I'd say we could even skip the pragma (this would help finding diags to fix when/if we increase the severity of the rule). 

Note that if you don't change, it' ok too, we can always lookup for pragmas.

Evangelink (approved):
A few extra minor remarks but otherwise LGTM. Might be worth to have a 2nd approval before merging.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Haplois: I love var, but wgen it's not obvious it hard do discern that while reviewing. I would vote for yes.

src/Microsoft.TestPlatform.Client/TestPlatform.cs
Haplois: Ah, that's the legacy behavior. I kept it separate so we wont break it my mistake. I'll add a comment there.

Haplois:
@Evangelink thanks for the review, I'll add the tests and ask for another review.

nohwnd (approved):

PR #3389, opened 2022-02-18
Fix singleton

Evangelink:


src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
```diff
@@ -14,7 +14,7 @@ internal partial class FeatureFlag : IFeatureFlag
 
     private const string Prefix = "VSTEST_FEATURE_";
 
-    public static IFeatureFlag Instance => new FeatureFlag();
+    public static IFeatureFlag Instance { get; } = new FeatureFlag();
```
nohwnd, line 17:
```suggestion
    public static IFeatureFlag Instance { get; } ??= new FeatureFlag();
```

src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
```diff
@@ -14,7 +14,7 @@ internal partial class FeatureFlag : IFeatureFlag
 
     private const string Prefix = "VSTEST_FEATURE_";
```
nohwnd, line 15:
```suggestion
    private const string VSTEST_FEATURE = nameof(VSTEST_FEATURE);
```

nohwnd: And do the same for the rest of the env variables. And join them with `+ "_" +`.

Also the comment says "po*r*st-processing"

test/Microsoft.TestPlatform.CoreUtilities.UnitTests/FeatureFlag/FeatureFlagTests.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.TestPlatform.CoreUtilities.UnitTests.FeatureFlag;
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestPlatform.Utilities;
+
+[TestClass]
+public class FeatureFlagTests
+{
+    [TestMethod]
+    public void SingletonWorksAsExpected()
```
nohwnd, line 13:
```suggestion
    public void SingletonAlwaysReturnsTheSameInstance()
```

MarcoRossignoli (approved):

MarcoRossignoli:
Thx @Evangelink missed that one, feature flag bag is immutable by design so it didn't fail any tests and also maybe singleton is not needed here, but let's keep the style, too sugars in this C# 😄

PR #3392, opened 2022-02-20
nugets should be developmentdependency

SimonCropp:
Currently pulling in `Microsoft.NET.Test.Sdk` pulls in `Microsoft.TestPlatform.TestHost` which pulls in `Newtonsoft.Json` version 9.0.1

note that v9 is 5+ years old and 4 majors behind.

This pollutes intellisense can can cause type/assembly loading issue when most projects are referencing a newer major version of Newtonsoft.Json.

So shouldnt the nuspecs for Microsoft.NET.Test.Sdk (and transitive) all be developmentdependency so they do not pollute the compile/run time of the test project consuming them?

### DIFF
--- a/src/package/nuspec/Microsoft.CodeCoverage.nuspec
+++ b/src/package/nuspec/Microsoft.CodeCoverage.nuspec
@@ -18,7 +18,8 @@
                 url="https://github.com/microsoft/vstest" 
                 branch="$BranchName$" 
                 commit="$CommitId$" />
-                
+    <developmentdependency>true</developmentdependency>
+
     <dependencies>
       <group targetFramework="net45"></group>
       <group targetFramework="netcoreapp1.0"></group>
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.nuspec
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.nuspec
@@ -18,6 +18,7 @@
                 url="https://github.com/microsoft/vstest" 
                 branch="$BranchName$" 
                 commit="$CommitId$" />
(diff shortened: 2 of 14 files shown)


nohwnd:
( From what I see, Test.SDK is rather a development non-dependency, you only want it for runtime :) )

It is how it is because we have a big pile of scenarios to support, and newtonsoft.json is for some reason embedded in our public api that is used for all communication. BUT if you have better solution that ticks all our boxes then I am all for it.


- We use Newtonsoft.Json because our public api (Message) for some reason uses JToken as the public type, rather than making it  a string or string collection. This is a very general problem we have, and I think this api is just overpublished, but we cannot change it at the moment. Without this, we would probably already moved to System.Text.Json.

- We use very old version of Newtonsoft.Json because the dll is output into bin directory, and mixes with your test project dependencies, and we don't want to upgrade your project if you happen to use an older version. If your project uses 13.0.1 (or whatever the latest is now), you will get that.

- We need the dependency to be there when running, because we use it to actually communicate from testhost to vstest.console. In the past (when .net framework only existed) all the testhost dependencies were resolved from runner directory, and there was no nuget you were referencing. For .net "core" the situation is a bit more complicated, because deps.json need to list all dependencies correctly, and especially when you run as self-contained app you need to have the dependencies present in the output folder. Or at least with our current setup, where we don't provide additional deps path. 

- We need both projects that need building, and pre-built assemblies to work.

--- 

It might not look like it from the outside, but we are inch by inch trying to get to a better place where we don't mix dependencies with the tested project. Have the same story for .NET Framework and .NET, and unified story for all the .NET versions we support.

PR #3409, opened 2022-02-23
Update find project pattern for test.sh

Evangelink:
## Description

With the current pattern passing `-p Accept` does not work while `-p Acceptance` does. With the new pattern both syntax work.


nohwnd:
Please also use the `-ifile ` parameter I mentioned to make the file pattern case insensitive.

nohwnd (approved):

PR #3410, opened 2022-02-23
Handle correctly waiting for process exit on Unix systems

### LINKED ISSUE #3375: Process error and exit callbacks are received in incorrect order for Linux
Investigating the issue https://github.com/dotnet/sdk/issues/22311 lead me to discover that we are not handling properly wait for exit in our process helper for Linux. This is causing us to receive the exit callback before having flushed all output/error callbacks.

We are awaiting for some [guidance from runtime team](https://github.com/dotnet/runtime/issues/65377) before submitting the PR to fix it.

Thank you @ArturDorochowicz for reporting this issue!

Evangelink:
## Description
Handle correctly waiting for process exit on Unix systems.

## Related issue
Fixes #3375


test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
```diff
@@ -599,6 +599,11 @@ protected static void ExecuteApplication(string path, string args, out string st
             throw new ArgumentException("Executable path must not be null or whitespace.", nameof(path));
         }
 
+        if (!File.Exists(path))
```
Evangelink, line 602: Without this check we end up with error like `System.ComponentModel.Win32Exception: No such file or directory` which isn't really helpful.

test/Microsoft.TestPlatform.AcceptanceTests/ProcessTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System.IO;
+
+using Microsoft.TestPlatform.TestUtilities;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.TestPlatform.AcceptanceTests;
+
+[TestClass]
+public class ProcessTests : AcceptanceTestBase
```
Evangelink, line 12: I couldn't really find in which existing type this would fit best so I created this one, feel free to suggest alternative.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -80,21 +82,38 @@ void InitializeAndStart()
...
+                            var cts = new CancellationTokenSource(500);
+#if NET5_0_OR_GREATER
+                            await p.WaitForExitAsync(cts.Token);
+#else
+                            await Task.Run(() => p.WaitForExit(), cts.Token);
+#endif
+                        }
+                        catch (Exception ex) when (ex is InvalidOperationException or TaskCanceledException)
+                        {
+                        }
```
MarcoRossignoli, line 112: can be useful a log here?

test/Microsoft.TestPlatform.AcceptanceTests/ProcessTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    public void MissingFrameworkTextIsCorrectlyDisplayed(RunnerInfo runnerInfo)
+    {
+        // Arrange
+        SetTestEnvironment(_testEnvironment, runnerInfo);
+        using var tempDir = new TempDirectory();
+        var assemblyPath = GetAssetFullPath("SimpleTestProjectMessedUpTargetFramework.dll", Core21TargetFramework);
+        var runtimeConfigJson = Path.Combine(Path.GetDirectoryName(assemblyPath), "SimpleTestProjectMessedUpTargetFramework.runtimeconfig.json");
+        var fileContent = File.ReadAllText(runtimeConfigJson);
+        var updatedContent = fileContent.Replace("\"version\": \"2.1.0\"", "\"version\": \"42\"");
+        File.WriteAllText(runtimeConfigJson, updatedContent);
```
MarcoRossignoli, line 25: this test is modifying the assets right?

test/Microsoft.TestPlatform.AcceptanceTests/ProcessTests.cs
MarcoRossignoli: `ProcessesInteractionTests`?

MarcoRossignoli (approved):

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
Evangelink: Maybe yes. I will have two separate logs depending on the exception type.

test/Microsoft.TestPlatform.AcceptanceTests/ProcessTests.cs
Evangelink: Yes that's why I created a new test asset. I originally thought that we were doing a local copy of the asset for each test but we don't so I wasn't sure how to best do this.

Given the question, I should probably put a comment in the code to explain what/why.

Thanks for pointing it out!

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
Evangelink: Actually, I can't because this project references no other project (so no access to `EqtTrace`). The signature is public so I can't even take some extra parameter.

test/Microsoft.TestPlatform.AcceptanceTests/ProcessesInteractionTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            // On the contrary to other tests, we need to modify the test asset we are using to replace
+            // the target framework with an invalid framework. This is why we have a specific test asset
+            // that's only meant to be used by this project.
+            // Having an invalid framework is a way to reproduce an issue we had on Unix, where we were
+            // not handling correctly the process exit (causing us to not let time to the process to
+            // flush its output and error streams).See https://github.com/microsoft/vstest/issues/3375
+            var runtimeConfigJson = Path.Combine(Path.GetDirectoryName(assemblyPath), testAssetProjectName + ".runtimeconfig.json");
+            var fileContent = File.ReadAllText(runtimeConfigJson);
+            var updatedContent = fileContent.Replace("\"version\": \"2.1.0\"", "\"version\": \"42\"");
```
nohwnd, line 43: maybe use some version that is lower than when is current, or that we know was not published, I think 2.8.0 was my original proposal? Or 0.0.0?

Evangelink: I went with `0.0.0`, it seems easier to notice this is invalid version.

test/Microsoft.TestPlatform.AcceptanceTests/ProcessTests.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.TestPlatform.AcceptanceTests;
+
+[TestClass]
+public class ProcessTests : AcceptanceTestBase
+{
+    [TestMethod]
+    [NetCoreTargetFrameworkDataSource]
+    public void MissingFrameworkTextIsCorrectlyDisplayed(RunnerInfo runnerInfo)
```
nohwnd:
```suggestion
    public void WhenTestHostProcessExitsBecauseTheTargetedRuntimeIsNoFoundThenTheMessageIsCapturedFromTheErrorOutput(RunnerInfo runnerInfo)
```

Might be also worth adding some more info about the problem we were solving.

nohwnd (approved):

PR #3416, opened 2022-02-25
Made telemetry data constants true constants

cvpoienaru:
Made telemetry data constants true constants


Evangelink (commented):
LGTM if we are ok we introducing a "potential" breaking change. We don't expect people to use these but that's public and not currently readonly.

Also, I don't know if we can end up in a situation were we would benefit from having `static readonly` instead of `const` (I am thinking about some case were we would have different versions loaded at the same time with the string value being different). cc @nohwnd

Evangelink (commented):
LGTM if we are ok we introducing a "potential" breaking change. We don't expect people to use these but that's public and not currently readonly.

Also, I don't know if we can end up in a situation were we would benefit from having `static readonly` instead of `const` (I am thinking about some case were we would have different versions loaded at the same time with the string value being different). cc @nohwnd

MarcoRossignoli:
If we're breaking why not put internal?

nohwnd:
@cvpoienaru Do you have any motivation for this change other than not liking the current syntax? Constants will get compiled into the consuming dll. Readonly static strings won't.

Evangelink:
@nohwnd The problem with current solution is the missing `readonly`. Adding it will cause a breaking change, so it might be a good time to also make the change to `internal` and see if we prefer to go with `const` or `static readonly`.

cvpoienaru:
@nohwnd, @Evangelink pretty much covered the reason for this change. This came up during the telemetry improvements PR review when it was pointed out that it'd be best to make our intention even clearer by making the fields true constants. Whether that means `const` or `static readonly` is debatable, however now I'm leaning towards switching to `internal static readonly` after hearing your arguments. Do we agree on `internal static readonly` then ?

CC: @MarcoRossignoli

nohwnd:
The change to internal should be fine .Common. is shipped with runner and host, and uwp provider also does not load it dynamically. So on binary level it should not break anyone.

```
Microsoft.TestPlatform.17.2.0-dev\tools\net451\Common7\IDE\Extensions\TestPlatform\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.17.2.0-dev\tools\net451\Common7\IDE\Extensions\TestPlatform\TestHost\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.CLI.17.2.0-dev\contentFiles\any\netcoreapp2.1\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.CLI.17.2.0-dev\contentFiles\any\netcoreapp2.1\TestHost\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.Internal.Uwp.17.2.0-dev\lib\netstandard1.3\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.Internal.Uwp.17.2.0-dev\lib\netstandard1.3\Microsoft.VisualStudio.TestPlatform.Common.pdb
Microsoft.TestPlatform.Internal.Uwp.17.2.0-dev\lib\uap10.0\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.Internal.Uwp.17.2.0-dev\lib\uap10.0\Microsoft.VisualStudio.TestPlatform.Common.pdb
Microsoft.TestPlatform.Portable.17.2.0-dev\tools\net451\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.Portable.17.2.0-dev\tools\netcoreapp2.1\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.Portable.17.2.0-dev\tools\netcoreapp2.1\TestHost\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.TestHost.17.2.0-dev\lib\netcoreapp1.0\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.TestHost.17.2.0-dev\lib\netcoreapp2.1\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.TestHost.17.2.0-dev\lib\uap10.0\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.TranslationLayer.17.2.0-dev\lib\net451\Microsoft.VisualStudio.TestPlatform.Common.dll
Microsoft.TestPlatform.TranslationLayer.17.2.0-dev\lib\netstandard2.0\Microsoft.VisualStudio.TestPlatform.Common.dll
```

nohwnd (approved):

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
```diff
@@ -8,60 +8,59 @@ namespace Microsoft.VisualStudio.TestPlatform.Common.Telemetry;
 /// <summary>
 /// The Telemetry data constants.
 /// </summary>
-[System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "Part of the public API.")]
-public static class TelemetryDataConstants
+internal static class TelemetryDataConstants
 {
     // ******************** Execution ***********************
-    public static string ParallelEnabledDuringExecution = "VS.TestRun.ParallelEnabled";
+    internal static readonly string ParallelEnabledDuringExecution = "VS.TestRun.ParallelEnabled";
```
Evangelink, line 14: It was probably fine to keep this level public once you changed the class to be `internal` but that's not worth changing back.

Evangelink (approved):

src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
cvpoienaru: That's a good point. I did it the other way around actually (i.e. first changed the members to `internal` and then figured I could even go for the whole class and make it `internal`), so that's why I ended up with both class and members being `internal`. I'll probably create a follow-up PR and fix this :)

PR #3433, opened 2022-03-03
Do not clear env variables by default

### LINKED ISSUE #3431: Support EnvironmentVariables in ConsoleParameters in .NETStandard
currently the variable is only build for .NETFramework: [link](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fvstest%2Fblob%2Fed5cd101764b4915b474ef55cd6dbd79bbed5f7b%2Fsrc%2FMicrosoft.TestPlatform.VsTestConsole.TranslationLayer%2FConsoleParameters.cs%23L50&data=04%7C01%7CVritant.Bhardwaj%40microsoft.com%7Ca03cdbe8052646ecff5308d9fc5bb779%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637818293123777203%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=N12zQ2xaTLx80MSy%2B3S9piyi1dbWKdyVLshwhmgLJkI%3D&reserved=0)

We will need support in .NetStandard as well for TestWindow

nohwnd:
## Description
Do not clear environment variables by default, and allow setting them on .NET Standard.

## Related issue
Fix #3431


nohwnd:
This is technically a breaking change for VS TestExplorer, but it gives us more sane API.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/PublicAPI/net451/PublicAPI.Shipped.txt
```diff
@@ -1,2 +0,0 @@
-Microsoft.TestPlatform.VsTestConsole.TranslationLayer.ConsoleParameters.EnvironmentVariables.get -> System.Collections.Generic.Dictionary<string, string>
-Microsoft.TestPlatform.VsTestConsole.TranslationLayer.ConsoleParameters.EnvironmentVariables.set -> void
```
nohwnd, line 2: These are moved from the net451 specific api to the universal api file.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
```diff
@@ -40,9 +40,22 @@ public ConsoleParameters(IFileHelper fileHelper)
...
     /// </summary>
-    public Dictionary<string, string> EnvironmentVariables { get; set; }
+    public Dictionary<string, string> EnvironmentVariables { get; set; } = new Dictionary<string, string>();
+
+    /// <summary>
+    /// Clears all environment variables that would be inherited from machine, user and process
+    /// and only sets the entries you provided in <see cref="EnvironmentVariables"/>.
+    /// This allows you to provide an arbitrary set of environment variables, for example when you
+    /// want to skip environment variables that are set in the IDE that starts vstest.console.
+    /// When setting this to true you are responsible for providing a full set of envirionment variables
```
Evangelink, line 53:
```suggestion
    /// When setting this to true you are responsible for providing a full set of environment variables
```

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
```diff
@@ -40,9 +40,22 @@ public ConsoleParameters(IFileHelper fileHelper)
...
+    /// that allow the process to start.
+    /// You most likely want to use <see cref="System.Environment.GetEnvironmentVariables(System.EnvironmentVariableTarget)"/> and combine
+    /// <see cref="System.EnvironmentVariableTarget.Machine"/> and <see cref="System.EnvironmentVariableTarget.User"/> responses.
+    /// </summary>
+    public bool ClearEnvironmentVariables { get; set; }
```
Evangelink, line 58: What about `SkipInheritedEnvironmentVariables`? Or `UseOnlyProvidedEnvironmentVariables`?

Evangelink (approved):

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
nohwnd: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.environment?view=net-6.0 docs

peterwald: @nohwnd The challenge with the "ProcessInfo method" is that in Test Explorer the customer expects us to duplicate the ENV that VS is running in more or less exactly, as if the test host is a child process of VS. So it would require a lot more code to add/modify/delete keys to make the environments match. That is why we originally went with the solution of just setting all the variables.

I think `ClearEnvironmentVariables` is pretty clear, and this looks like a good improvement on the existing API for sure. It won't be hard to fixup Test Explorer to handle this. The only downside to this that I can see is that it could make it difficult to remove specific env variables. But you can always just set them to an empty string, so that might not even be an issue.

nohwnd: @peterwald  you are already providing us with the full set of environment variables that you want to use. We just won't clear them for you by default anymore. So with the ProcessStartInfo like approach, you would need to call Clear() on the EnvironmentVariableDictionary, and then put all your key value pairs into it, e.g. by using .ForEach. It is a tiny bit of wasted performance, but you don't start vstest.console that often.

I created PR this morning on your repo it is <10 lines. Using the ProcessStartInfo approach will be the same.

rouke-broersma: I think we already expected that the vstest process would inherit the environment variables from the main process and that we only needed to add extra environment variables we needed, so I think we won't use `ClearEnvironmentVariables`. It is probably an improvement in our case that environment variables will no longer be cleared by default. 

That said `ClearEnvironmentVariables` seems like an implementation bleed through the API. What the user of the API wants is functionally something like `DisableInheritEnvironmentVariables` or something like that.

rouke-broersma: Actually we use a custom testhostlauncher and don't use ConsoleParameters EnvironmentVariables so I'm not sure this case is relevant to us at all: https://github.com/stryker-mutator/stryker-net/blob/master/src/Stryker.Core/Stryker.Core/TestRunners/VsTest/StrykerVstestHostLauncher.cs#L36

nohwnd: @rouke-broersma IIRC you set dotnet roll-forward in the current process to allow vstest.console to inherit it, and run on any dotnet version available. 

This change would have no impact on you other than being able to set the env variables directly on vstest.console rather than setting them in your process and letting vstest.console inherit them.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -119,15 +119,24 @@ public void StartProcess(ConsoleParameters consoleParameters)
...
+        if (consoleParameters.EnvironmentVariables != null)
+        {
             foreach (var envVariable in consoleParameters.EnvironmentVariables)
             {
                 if (envVariable.Key != null)
                 {
+                    // Not printing the value on purpose, env variables can contain secrets and we don't need to know the values
+                    // most of the time.
+                    EqtTrace.Verbose("VsTestCommandLineWrapper.StartProcess: Setting environment variable: {0}", envVariable.Key);
                     info.EnvironmentVariables.Add(envVariable.Key, envVariable.Value?.ToString());
```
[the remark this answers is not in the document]
nohwnd: Good point, this should use the `[] = ""` api I shown above.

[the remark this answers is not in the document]
nohwnd: Done.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs
nohwnd: I used `InheritEnvironmentVariables` and set it to `true` by default. Seems like the least complicated way to name that option. 

I kept the option approach, to avoid copying the whole environment into the dictionary, only to have VS clear it and copy almost whole environment into the dictionary again.

Anyone else gets the simplest behavior they can. They give us a dictionary of env variables they want set, and we merge them with their current environment. They don't need to be editing an existing dictionary, and possibly get it wrong by using `.Add(key, value)` instead of `[key] = value`.

PR #3448, opened 2022-03-07
Newer approach to reference assemblies on MacOS and Linux

nohwnd:
## Description
Use more modern reference assembly pack for Linux and MacOS build.

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -51,7 +51,7 @@
 
   <!-- Allows build of .NET Framework assemblies on MacOS and Linux https://github.com/dotnet/designs/pull/33#issuecomment-489264196 -->
   <ItemGroup>
-    <PackageReference Condition=" '$(OS)' != 'Windows_NT' " Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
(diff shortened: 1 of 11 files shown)


scripts/build/TestPlatform.Settings.targets
```diff
@@ -49,6 +49,11 @@
   <!-- Package dependency versions -->
   <Import Project="$(MSBuildThisFileDirectory)TestPlatform.Dependencies.props" Condition=" '$(DependencyVersionsImported)' != 'true' "/>
 
+  <!-- Allows build of .NET Framework assemblies on MacOS and Linux https://github.com/dotnet/designs/pull/33#issuecomment-489264196 -->
+  <ItemGroup>
+    <PackageReference Condition=" '$(OS)' != 'Windows_NT' " Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
```
Evangelink, line 54: Any reason not to target `1.0.2`? (see https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies)

nohwnd: Nope, I just did not check what is the latest, because they kept all the other packages we used before stick to 1.0.0.

nohwnd:
```suggestion
    <PackageReference Condition=" '$(OS)' != 'Windows_NT' " Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="All" />
```

Evangelink (approved):

cvpoienaru (approved):

PR #3451, opened 2022-03-08
Fix perf tests build

nohwnd:


nohwnd:
Skipping PR pipeline, passed on internal https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5850566&view=results

PR #3454, opened 2022-03-09
Support reading embedded pdbs

### LINKED ISSUE #1908: Support embedded PDBs in OM
## Description
OM currently supports regular and portable PDBs (see https://github.com/Microsoft/vstest/tree/master/src/Microsoft.TestPlatform.ObjectModel/Navigation). However, as far as I can tell, it doesn't support embedded pdbs.

Not sure if this is the same issue as #1748 but is probably related.

## Steps to reproduce
VS's Test Window uses ISourceInformationProvider to side step PDB based source information lookup in most cases for managed projects. However, ISourceInformationProvider currently does not support multi-target-framework projects - so we end up falling back to PDB based source information for such projects. Now if such project also happens to use embedded PDB then the source information lookup fails and we end up not finding tests corresponding to a particular file.

To reproduce create a test project that is both mtfm as well as uses embedded PDB and try to run tests by right clicking on the test file in the solution explorer in any 16.0 VS build.

## Expected behavior
Tests in the file should be executed (and this works if I switch the repro project to use portable PDB instead of embedded).

## Actual behavior
No tests are found for selected file.

nohwnd:
## Description
Building with `<DebugType>embedded</DebugType>` embeds portable pdb into the dll. This change either takes the pdb that is next to the file, or falls back to trying to read the pdb information from the executable.

## Related issue
Fix #1908

Fixes 1. and 2. from https://github.com/microsoft/vstest/issues/1908#issuecomment-548164376 comment, but not 3. which is a perf improvement). Also keeps the current functionality of looking up the file next to the source in place for backwards compatibility.

If only I saw that comment before I went on to implement this. It spells out exactly what I ended up doing, minus 2 hours spent trying to find the right class and method that can read the embedded pdbs.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
@@ -101,12 +101,21 @@ public class DiaSession : INavigationSession
         // For remote scenario, also look for pdb in current directory, (esp for UWP)
         // The alternate search path should be an input from Adapters, but since it is not so currently adding a HACK
         pdbFilePath = !File.Exists(pdbFilePath) ? Path.Combine(Directory.GetCurrentDirectory(), Path.GetFileName(pdbFilePath)) : pdbFilePath;
-        using var stream = new FileHelper().GetStream(pdbFilePath, FileMode.Open, FileAccess.Read);
-        if (PortablePdbReader.IsPortable(stream))
+
+        if (File.Exists(pdbFilePath))
         {
+            using var stream = new FileHelper().GetStream(pdbFilePath, FileMode.Open, FileAccess.Read);
+            if (PortablePdbReader.IsPortable(stream))
+            {
+                return new PortableSymbolReader();
+            }
+
+            return new FullSymbolReader();
+        }
+        else
+        {
+            // If we cannot find the pdb file, it might be embedded in the dll.
             return new PortableSymbolReader();
         }
-
-        return new FullSymbolReader();
     }
 }
--- a/src/Microsoft.TestPlatform.ObjectModel/Navigation/PortableSymbolReader.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/Navigation/PortableSymbolReader.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
 
 using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
 using Microsoft.VisualStudio.TestPlatform.Utilities.Helpers;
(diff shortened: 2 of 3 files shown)


src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
```diff
@@ -101,12 +101,21 @@ private static ISymbolReader GetSymbolReader(string binaryPath)
         // For remote scenario, also look for pdb in current directory, (esp for UWP)
         // The alternate search path should be an input from Adapters, but since it is not so currently adding a HACK
         pdbFilePath = !File.Exists(pdbFilePath) ? Path.Combine(Directory.GetCurrentDirectory(), Path.GetFileName(pdbFilePath)) : pdbFilePath;
-        using var stream = new FileHelper().GetStream(pdbFilePath, FileMode.Open, FileAccess.Read);
-        if (PortablePdbReader.IsPortable(stream))
+
+        if (File.Exists(pdbFilePath))
         {
+            using var stream = new FileHelper().GetStream(pdbFilePath, FileMode.Open, FileAccess.Read);
```
MarcoRossignoli, line 107: nit: we're using `File` everywhere in this file maybe it's better use also here or create on top and reuse for future testability.

src/Microsoft.TestPlatform.ObjectModel/Navigation/PortableSymbolReader.cs
```diff
@@ -148,4 +153,37 @@ private void PopulateCacheForTypeAndMethodSymbols(string binaryPath)
...
+    /// <summary>
+    /// Reads the pdb data from the dlls itself, either by loading the referenced pdb file, or by reading
+    /// embedded pdb from the dll itself.
+    /// </summary>
+    /// <param name="binaryPath"></param>
+    /// <returns></returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    private static PortablePdbReader CreatePortablePdbReaderFromPEData(string binaryPath)
+    {
+        var peReader = new PEReader(new FileStream(binaryPath, FileMode.Open, FileAccess.Read));
```
MarcoRossignoli, line 166: file stream `using` or `MetadataReaderProvider` needs the open handle?

MarcoRossignoli (approved):

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
nohwnd: This is just the code that was here before. Are you talking about using System.IO.File to create the stream? Or about using FileHelper to improve testability?

MarcoRossignoli: I meant have all `FileHelper` or all `File` usage, but it's not mandatory I don't like the mix.

src/Microsoft.TestPlatform.ObjectModel/Navigation/PortableSymbolReader.cs
nohwnd: Fixed.

src/Microsoft.TestPlatform.ObjectModel/Navigation/DiaSession.cs
nohwnd: File helper is used here to get the stream via platform abstractions. It is more correct to use FileHelper everywhere here, but that is a bigger change than what I am comfortable with without adding tests. And I don't want to give this more time right now.

PR #3461, opened 2022-03-11
Refactor datacollector

MarcoRossignoli:
Refactor datacollector

Manual tested

Before
image

After
image


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -51,6 +50,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
     private const string DotnetTestHostUri = "HostProvider://DotnetTestHost";
     private const string DotnetTestHostFriendlyName = "DotnetTestHost";
     private const string TestAdapterRegexPattern = @"TestAdapter.dll";
+    private const string PROCESSOR_ARCHITECTURE = "PROCESSOR_ARCHITECTURE";
```
nohwnd, line 53:
```suggestion
    private const string PROCESSOR_ARCHITECTURE = nameof(PROCESSOR_ARCHITECTURE);
```

src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/DefaultDataCollectionLauncher.cs
```diff
@@ -59,10 +60,15 @@ public override int LaunchDataCollector(IDictionary<string, string> environmentV
         var currentProcessPath = _processHelper.GetCurrentProcessFileName();
 
         // If current process is dotnet/dotnet.exe and you are here, datacollector.exe is present in TestHost folder.
+        // It's unexpected we have DotnetDataCollectionLauncher for .NET Core and we're not shipping arm64 version inside the SDK
+        // because everything there is AnyCPU and will be "Ready to Run" compiled for the specific architecture shipped.
+        // We'll eventually run emulated x64.
```
nohwnd, line 65: I don't understand this comment. Can you elaborate?

scripts/build.ps1
```diff
@@ -267,7 +267,9 @@ function Publish-Package
     Publish-PackageInternal $settingsMigratorProject $TPB_TargetFramework451 $fullCLRPackage451Dir
 
     Write-Log "Package: Publish src\datacollector\datacollector.csproj"
-    Publish-PackageInternal $dataCollectorProject $TPB_TargetFramework472 $fullCLRPackage451Dir
+    # We build datacollector.arm64.exe before and we ship in the same folder as win7-x64.
```
nohwnd, line 270:
```suggestion
    # We build datacollector.arm64.exe before building datacollector.exe and we put it in the same folder, so they end up shipping together.
```

scripts/build.ps1
```diff
@@ -258,7 +258,7 @@ function Publish-Package
     # Publish vstest.console and datacollector exclusively because *.config/*.deps.json file is not getting publish when we are publishing aforementioned project through dependency.
     Write-Log "Package: Publish src\vstest.console\vstest.console.csproj"
 
-    # We build vstest.console.arm64.exe before and we ship in the same folder as win7-x64 to have the same VSIX packaging to deploy on VS.
+    # We build vstest.console.arm64.exe before and we ship in the same folder as win7-x64.
```
nohwnd, line 261:
```suggestion
        # We build vstest.console.arm64.exe before building vstest.console.exe and we put it in the same folder, so they end up shipping together.
```

src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/DefaultDataCollectionLauncher.cs
MarcoRossignoli: It's not clear to me when we reach this point and if it's possible, shouldn't never have dotnet.exe as current process because in that case we use `DotnetDataCollectionLauncher`

https://github.com/microsoft/vstest/blob/1539d8b6c2a6635d9fae371c707b8fcbac7c1be8/src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/DataCollectionLauncherFactory.cs#L44

So I've updated a bit the comment, we won't ship arm64 collector on the sdk so that file won't be there and the default one will be x64.

I don't know if makes sense update the authoring to have also arm version inside the sdk. What do you think?

MarcoRossignoli: Looks like is used for CodeCoverage, it's not clear, I'll ship for now...but I have to sync with @jakubch1 on it, maybe it's code that we can remove

nohwnd (approved):

PR #3470, opened 2022-03-15
Skip restore on publishing internal package

Evangelink:
## Description
Attempts to make build faster.

## Related issue
Replaces #2494


Evangelink:
Not sure if I am looking at the right build entry but I don't see real benefit. @nohwnd could you confirm?

PR #3473, opened 2022-03-16
Enable nullable on tests

Evangelink:


NullableAttributes.cs
```diff
@@ -0,0 +1,155 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: Added as a loose file that is included by projects needing it. That's probably not the best solution on the long run but until we find the right way to have this shimlayer/polyfill project I want to start with something simple.

test/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector.UnitTests/EventLogDataCollectorTests.cs
```diff
@@ -430,39 +423,35 @@ public void WriteEventLogsShouldFilterTestIfMaxValueExceedsEntries()
...
+    public override event EventHandler<TestHostLaunchedEventArgs>? TestHostLaunched;
+    public override event EventHandler<SessionStartEventArgs>? SessionStart;
+    public override event EventHandler<SessionEndEventArgs>? SessionEnd;
+    public override event EventHandler<TestCaseStartEventArgs>? TestCaseStart;
+    public override event EventHandler<TestCaseEndEventArgs>? TestCaseEnd;
 
     public Delegate[] GetTestHostLaunchedInvocationList()
     {
-        return TestHostLaunched.GetInvocationList();
+        return TestHostLaunched!.GetInvocationList();
```
Evangelink, line 434: For tests I went ahead with assuming that even when we have a potential null access we dismiss the message (with bang) because the test was working without NRE.

This comment applies generally to the PR.

test/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector.UnitTests/EventLogSessionContextTests.cs
```diff
@@ -70,18 +67,18 @@ public void CreateEventLogContainerShouldCreateNegativeEndIndexIfLogEntriesAreEm
...
 }
 
-public class DummyEventLogContainer : IEventLogContainer
+public sealed class DummyEventLogContainer : IEventLogContainer
 {
     public DummyEventLogContainer(bool initialize)
     {
         EventLogEntries = new List<EventLogEntry>(10);
-        EventLog eventLog = new("Application");
+        EventLog = new("Application");
```
Evangelink, line 75: We were creating a local variable but never assigning the property. I assumed this was an error and the tests still don't fail (locally) so I guess this is the right move.

test/Microsoft.TestPlatform.AdapterUtilities.UnitTests/ManagedNameUtilities/FindMethodExtensions.cs
```diff
@@ -54,7 +52,7 @@ internal static MethodInfo FindMethod(this Type type, string signature)
         }
 
         Debug.Assert(candidates.Any() && !candidates.Skip(1).Any());
-        return candidates.SingleOrDefault();
+        return candidates.Single();
```
Evangelink, line 55: Looks like we were using a lot XXXOrDefault methods even when we know we have one element (see debug assert just before) or also doing lots of as cast when we know this won't be null.

test/Microsoft.TestPlatform.AdapterUtilities.UnitTests/TestIdProvider/SHA1ImplTests.cs
```diff
@@ -138,7 +136,7 @@ private void SHA1_TestRepetitionVector(char input, int repetition, string expect
         }
         else
         {
-            expected = expected.Replace(" ", "").ToLowerInvariant();
+            expected = expected!.Replace(" ", "").ToLowerInvariant();
```
Evangelink, line 139: In net48, `string.IsNullOrEmpty` doesn't provide the non-null info to the analyzer. When we will use our shimlayer, we could have our own implementation that would provide this hint to the compiler. In the meantime, I went with bang.

nohwnd:
Is enabling nullable in tests going to help us figure out where we do it wrong in the real code, and how? Because right now, it is a lot of `!` noise.

nohwnd:
Are there no other changes than in tests files? It is impossible to filter that on github. And there are some whitespace changes in `src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/TestSessionPool.cs` for example. So I am not sure about all the impact, and 250 files is a lot to review.

Evangelink:
> Is enabling nullable in tests going to help us figure out where we do it wrong in the real code, and how? 

I had a look at enabling nullable from various parts and it's not always clear when we expect null or not. I have seen that tests helps to see cases where we expect to pass null (when the tests are available obviously). With this merged, if we say something is not expecting nulls but we have a test testing null we would have a warning.

> Because right now, it is a lot of `!` noise.

Lots of tests are making lots of assumption on non-nullability hence all those `!`. They should have been written differently, it seems like a lot of work to "rewrite" the cases. For example, there are lots of `as` cast where there is a direct member access (`.XXX`), this should not have been done. Most calls to `XXXOrDefault` also assume there is a value while we do member access so again assume it's not null so the `XXX` would be better call.

> Are there no other changes than in tests files? It is impossible to filter that on github. And there are some whitespace changes in `src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/TestSessionPool.cs` for example. So I am not sure about all the impact, and 250 files is a lot to review.

There is only `TestSessionPool` that slipped in (will revert). For the review I could easily do PR per test project or number of files to help there.

PR #3499, opened 2022-03-22
Allow attachVS env variables in release build

nohwnd:
Allow lookup of AttachVS tool in Release build, but not ship the tool with the build.


Evangelink (approved):
LGTM - are you ok with merging change to acceptance tests parallelization in this PR?

PR #3504, opened 2022-03-22
Enable nullable on extension and utilities tests

Evangelink:


nohwnd (approved):

StringUtils.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: Adding the file at the root as we did for `NullableAttributes.cs`

PR #3511, opened 2022-03-24
Added telemetry data point for extensions loaded during test discovery/run

cvpoienaru:
Added telemetry data point for extensions loaded during test discovery/run


src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -117,8 +117,8 @@ public List<string> GetExtensionPaths(string endsWithPattern, bool skipDefaultEx
     /// <returns>
     /// The <see cref="Dictionary"/>. of test plugin info.
     /// </returns>
-    public Dictionary<string, TPluginInfo> DiscoverTestExtensions<TPluginInfo, TExtension>(string endsWithPattern)
-        where TPluginInfo : TestPluginInformation
+    public Dictionary<string, TPluginInfo> DiscoverTestExtensions<TPluginInfo, TExtension>(
+        string endsWithPattern) where TPluginInfo : TestPluginInformation
```
Evangelink, line 121: Nit: Feel free to ignore but as you split parameters to new line, I'd be nice to also split the type constraint.

```suggestion
        string endsWithPattern)
        where TPluginInfo : TestPluginInformation
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -309,35 +309,36 @@ internal IEnumerable<string> DefaultExtensionPaths
     /// <returns>
     /// The <see cref="Dictionary"/>.
     /// </returns>
-    internal Dictionary<string, TPluginInfo> GetTestExtensions<TPluginInfo, TExtension>(string extensionAssembly, bool skipCache = false) where TPluginInfo : TestPluginInformation
+    internal Dictionary<string, TPluginInfo> GetTestExtensions<TPluginInfo, TExtension>(
+        string extensionAssembly,
+        bool skipCache = false) where TPluginInfo : TestPluginInformation
```
Evangelink, line 314: Nit: Feel free to ignore but as you split parameters to new line, I'd be nice to also split the type constraint.

```suggestion
        bool skipCache = false) 
        where TPluginInfo : TestPluginInformation
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+    /// Merge two extension maps.
+    /// </summary>
+    /// 
+    /// <param name="extensionMap1">First extension map.</param>
+    /// <param name="extensionMap2">Second extension map.</param>
+    /// 
+    /// <returns>
+    /// A map representing the merger between the two input maps.
+    /// </returns>
+    internal static IDictionary<string, ISet<string>> MergeExtensionMaps(
```
MarcoRossignoli, line 104: Can be reused maybe `MergeMaps` or `Merge` only?

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+        IDictionary<string, ISet<string>> extensionMap1,
+        IDictionary<string, ISet<string>> extensionMap2)
+    {
+        var isExtensionMap1Valid = (extensionMap1 != null && extensionMap1.Count > 0);
+        var isExtensionMap2Valid = (extensionMap2 != null && extensionMap2.Count > 0);
```
Evangelink, line 109: [cosmetic] Parenthesis are not needed.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
```diff
@@ -39,6 +41,10 @@ public ParallelDiscoveryDataAggregator()
     /// </summary>
     public long TotalTests { get; private set; }
 
+    /// <summary>
+    /// A collection of aggregated discovered extensions.
+    /// </summary>
+    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; set; }
```
MarcoRossignoli, line 47:
```suggestion
    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; private set; }
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+
+        // Sanity checks.
+        if (!isExtensionMap1Valid && !isExtensionMap2Valid) return new Dictionary<string, ISet<string>>();
```
Evangelink, line 112: I would prefer not to mix styles (i.e. either parenthesis for 1 line ifs or not). Inside of the foreach you do use parenthesis.

Usually in the code, it's the option with the parenthesis which is favored.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
```diff
@@ -58,6 +60,11 @@ public ParallelRunDataAggregator(string runSettingsXml!!)
 
     public HashSet<string> ExecutorUris => new(_executorUris);
 
+    /// <summary>
+    /// A collection of aggregated discovered extensions.
+    /// </summary>
+    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; set; }
```
MarcoRossignoli, line 66:
```suggestion
    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; private set; }
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
Evangelink, line 104: [minor] "Map" feels javaish to me, I would use dictionary.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -307,6 +375,27 @@ internal TestExtensions GetExtensionsDiscoveredFromAssembly(string extensionAsse
         }
     }
 
+    /// <summary>
+    /// Gets the cached extensions for the current process.
+    /// </summary>
+    /// 
+    /// <returns>A map representing the cached extensions for the current process.</returns>
+    internal IDictionary<string, ISet<string>> GetCachedExtensions()
```
Evangelink, line 383: I don't know what's the team preference but I usually prefer to be as specific as possible on return (even more when "private" api) as it might avoid doing cast or reconstruction of a type later on.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +479,35 @@ internal void InvalidateCache()
...
+
+        foreach (var kvp in extensionMap)
+        {
+            sb.Append(
+                kvp.Value?.Count == 0
+                ? string.Empty
+                : string.Format(
+                    "{0}=[{1}];",
+                    kvp.Key,
+                    string.Join(",", kvp.Value.ToArray())));
```
Evangelink, line 508: You don't need the `ToArray` call as there is an `IEnumerable` overload.

```suggestion
                    string.Join(",", kvp.Value)));
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +479,35 @@ internal void InvalidateCache()
...
+        }
+    }
+
+    private static string SerializeExtensionMap(IDictionary<string, ISet<string>> extensionMap)
+    {
+        StringBuilder sb = new();
+
+        foreach (var kvp in extensionMap)
+        {
+            sb.Append(
```
Evangelink, line 502: I would refactor to add a if condition around and use `AppendFormat`.

src/Microsoft.TestPlatform.CommunicationUtilities/Messages/DiscoveryCompletePayload.cs
```diff
@@ -48,4 +48,9 @@ public class DiscoveryCompletePayload
     /// Gets or sets list of sources which were not discovered at all.
     /// </summary>
     public IList<string> NotDiscoveredSources { get; set; } = new List<string>();
+
+    /// <summary>
+    /// Gets or sets the collection of discovered extensions.
+    /// </summary>
+    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; set; }
```
Evangelink, line 55: Shall we initialize to empty dictionary to avoid nulls?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
```diff
@@ -156,4 +168,9 @@ public void AggregateIsMessageSent(bool isMessageSent)
     /// <returns></returns>
     public List<string> GetSourcesWithStatus(DiscoveryStatus status)
         => DiscoveryManager.GetSourcesWithStatus(status, _sourcesWithDiscoveryStatus);
+
+    private void AggregateDiscoveredExtensions(IDictionary<string, ISet<string>> discoveredExtensions)
```
Evangelink, line 172: IMO there is no need for a separate method here, I would do the init directly.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
```diff
@@ -200,4 +211,8 @@ public void AggregateRunDataMetrics(IDictionary<string, object> metrics)
         }
     }
 
+    private void AggregateDiscoveredExtensions(IDictionary<string, ISet<string>> discoveredExtensions)
```
Evangelink, line 214: Same remark about removing this method.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/DiscoveryCompleteEventArgs.cs
```diff
@@ -53,30 +84,42 @@ public DiscoveryCompleteEventArgs(long totalTests, bool isAborted)
...
     /// Gets the list of sources which were not discovered at all.
     /// </summary>
+    [DataMember]
     public IList<string> NotDiscoveredSources { get; set; }
+
+    /// <summary>
+    /// Gets or sets the collection of discovered extensions.
+    /// </summary>
+    [DataMember]
+    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; set; }
```
Evangelink, line 124: Avoid null by doing init to empty dictionary?

src/Microsoft.TestPlatform.ObjectModel/Client/Events/TestRunCompleteEventArgs.cs
```diff
@@ -106,4 +106,10 @@ public TestRunCompleteEventArgs(ITestRunStatistics stats, bool isCanceled, bool
     /// </summary>
     [DataMember]
     public IDictionary<string, object> Metrics { get; set; }
+
+    /// <summary>
+    /// Gets or sets the collection of discovered extensions.
+    /// </summary>
+    [DataMember]
+    public IDictionary<string, ISet<string>> DiscoveredExtensions { get; set; }
```
Evangelink, line 114: Avoid null by doing init to empty dictionary?

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
nohwnd: Agree. Reading that description comment and the name I have no idea what this is.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+    /// <param name="extensionMap2">Second extension map.</param>
+    /// 
+    /// <returns>
+    /// A map representing the merger between the two input maps.
+    /// </returns>
+    internal static IDictionary<string, ISet<string>> MergeExtensionMaps(
+        IDictionary<string, ISet<string>> extensionMap1,
+        IDictionary<string, ISet<string>> extensionMap2)
+    {
+        var isExtensionMap1Valid = (extensionMap1 != null && extensionMap1.Count > 0);
```
nohwnd, line 108: Please invert the logic and rename to isNullOrEmpty

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+        // Sanity checks.
+        if (!isExtensionMap1Valid && !isExtensionMap2Valid) return new Dictionary<string, ISet<string>>();
+        if (!isExtensionMap1Valid) return extensionMap2;
+        if (!isExtensionMap2Valid) return extensionMap1;
+
+        // We don't use an additional map for merging, instead we use the first map as a
+        // destination for the merging operation.
+        foreach (var kvp in extensionMap2)
+        {
+            // If the "source" set is empty there's no reason to continue.
```
nohwnd, line 120: Says that there is no reason to continue, then it continues.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
nohwnd: I would also avoid prefer getting a new instance of the dictionary with the values merged, rather than getting live reference either to left, right, or new instance. 

The code mixes merging of dictionaries with merging of sets, for probably no measurable impact on performance.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
nohwnd: The logic below would be better in it's own function that merges sets.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,74 @@ public class TestExtensions
...
+        return extensionMap1;
+    }
+
+    /// <summary>
+    /// Write the extension map to the telemetry data.
+    /// </summary>
+    /// 
+    /// <param name="metrics">A collection representing the telemetry data.</param>
+    /// <param name="extensionMap">The input extension map.</param>
+    internal static void WriteExtensionMapToTelemetryData(
```
nohwnd, line 153: Rename to AddExtensionTelemetry?

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
nohwnd: Yeah if this is a Dictionary of hashset let's just use those types.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -307,6 +375,27 @@ internal TestExtensions GetExtensionsDiscoveredFromAssembly(string extensionAsse
...
+    /// <summary>
+    /// Gets the cached extensions for the current process.
+    /// </summary>
+    /// 
+    /// <returns>A map representing the cached extensions for the current process.</returns>
+    internal IDictionary<string, ISet<string>> GetCachedExtensions()
+    {
+        var extensionMap = new Dictionary<string, ISet<string>>();
+
+        // Write all "known" cached extension.
```
nohwnd, line 387: What "known" means here?

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +479,35 @@ internal void InvalidateCache()
         }
     }
 
+    private void AddCachedExtensionToExtensionMap<T>(
+        IDictionary<string, ISet<string>> extensionMap,
+        string extensionType,
+        IList<T> extensions) where T : TestPluginInformation
+    {
+        if (extensions == null) return;
```
nohwnd, line 487: New line please.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +479,35 @@ internal void InvalidateCache()
...
+
+        extensionMap.Add(extensionType, new HashSet<string>());
+        foreach (var extension in extensions)
+        {
+            extensionMap[extensionType].Add(extension.IdentifierData);
```
nohwnd, line 492: Populate the hashset, first then add it. Rather than looking it up from the extension dictionary on every iteration. You can just select IdentifierData from extensions and use hashset object initializer, something like this:

new HashSet<string>{ extensions.Select(e => e.IdentifierData) };

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +479,35 @@ internal void InvalidateCache()
...
+        }
+    }
+
+    private static string SerializeExtensionMap(IDictionary<string, ISet<string>> extensionMap)
```
nohwnd, line 496: Why is this necessary? Is this to avoid approving a lot of different data points in gdpr portal? Can a more native format be used? Like json?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunEventsHandler.cs
```diff
@@ -80,12 +82,21 @@ internal class ParallelRunEventsHandler : ITestRunEventsHandler2
...
-            _requestData.MetricsCollection.Add(TelemetryDataConstants.RunState, _runDataAggregator.IsAborted ? "Aborted" : _runDataAggregator.IsCanceled ? "Canceled" : "Completed");
+            _requestData.MetricsCollection.Add(
+                TelemetryDataConstants.RunState,
+                _runDataAggregator.IsAborted
+                    ? "Aborted"
+                    : _runDataAggregator.IsCanceled ? "Canceled" : "Completed");
+
+            // Add the map containing discovered extensions to the event args. This map contains
+            // only extensions discovered by the testhost processes. Current process extensions
+            // (i.e. vstest.console) are to be added later on.
```
nohwnd, line 93:
```suggestion
            // (i.e. vstest.console) are added later on.
```

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
cvpoienaru: Fixed.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
cvpoienaru: Fixed.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done, thanks.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Fixed, thanks.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
cvpoienaru: Fixed.

src/Microsoft.TestPlatform.CommunicationUtilities/Messages/DiscoveryCompletePayload.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/DiscoveryCompleteEventArgs.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.ObjectModel/Client/Events/TestRunCompleteEventArgs.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done, thanks.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunEventsHandler.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: There's a total of 7 extensions we cache, defined by properties at the top of the file. Pasting them here for convenience:

```
    /// <summary>
    /// Gets or sets test discoverer extensions.
    /// </summary>
    internal Dictionary<string, TestDiscovererPluginInformation> TestDiscoverers { get; set; }

    /// <summary>
    /// Gets or sets test executor extensions.
    /// </summary>
    internal Dictionary<string, TestExecutorPluginInformation> TestExecutors { get; set; }

    /// <summary>
    /// Gets or sets test executor 2 extensions.
    /// </summary>
    internal Dictionary<string, TestExecutorPluginInformation2> TestExecutors2 { get; set; }

    /// <summary>
    /// Gets or sets test setting provider extensions.
    /// </summary>
    internal Dictionary<string, TestSettingsProviderPluginInformation> TestSettingsProviders { get; set; }

    /// <summary>
    /// Gets or sets test logger extensions.
    /// </summary>
    internal Dictionary<string, TestLoggerPluginInformation> TestLoggers { get; set; }

    /// <summary>
    /// Gets or sets test logger extensions.
    /// </summary>
    internal Dictionary<string, TestRuntimePluginInformation> TestHosts { get; set; }

    /// <summary>
    /// Gets or sets data collectors extensions.
    /// </summary>
    internal Dictionary<string, DataCollectorConfig> DataCollectors { get; set; }
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Having to approve a lot of different data points in the GDPR portal is only part of the reason. 

There's also the fact that each different extension type we add telemetry data points for would have to be represented as a separate column in the telemetry table, which causes the invalidation of the previous table schema. In short, we need to recreate and repopulate the telemetry table for each newly added column. Even if that weren't the case, we'd still have to log (possibly) multiple extensions for each extension type which would still require some parsing, so we might as well just cut off the inconvenience of recreating the table and do the parsing we would've done either way, only a little more complex.

Another reason of compacting data like this is also related to the way data collectors telemetry is handled, i.e. all telemetry is being written in the same field in the telemetry table and packed using a custom structure. Granted, we can do better here by giving JSON a try.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Taking all input into consideration, I renamed the method `MergeDictionaries`.

`MergeByKey` gives away implementation details;
`MergeMaps`, both Jakub and Amaury think the term `map` should be replaced by `dictionary` to make the operation more self descriptive;
`Merge`/`Combine` is too generic and implies it merges two `TestExtension` objects together which is not really the case, we merge two dictionaries that happen to contain extensions.

cvpoienaru: Also, the method cannot really be reused since the return type is pretty specific, i.e. `IDictionary<string, ISet<string>>`. We should probably implement a generic merge operation that returns an `IDictionary<K, V>` object, but this seems a little overengineered given the scope.

cvpoienaru: Fixed the return issue. Now we return a new instance representing the result of the merging operation no matter what we do inside the method.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Added more details to the comment. What I meant by "there's no reason to continue" is "there's no reason to continue merging for the current key, so we do a `continue` to skip to the next key" :D

cvpoienaru: FIxed the second part, i.e. moved set merging into separate method.

src/Microsoft.TestPlatform.Client/Discovery/DiscoveryRequest.cs
```diff
@@ -248,6 +250,22 @@ public void HandleDiscoveryComplete(DiscoveryCompleteEventArgs discoveryComplete
                     OnDiscoveredTests.SafeInvoke(this, discoveredTestsEvent, "DiscoveryRequest.DiscoveryComplete");
                 }
 
+                // Add extensions discovered by vstest.console.
+                //
+                // TODO(copoiena): Writing telemetry twice is less than ideal.
+                // We first write telemetry data in the _requestData variable in the ParallelRunEventsHandler
+                // and then we write again here. We should refactor this code and write only once.
```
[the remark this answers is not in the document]
cvpoienaru: The first time, we write the telemetry data in the `_requestData` variable that is used only for logging/debugging purposes (i.e. so that we can check the files created on the local machine and make sure they contain our telemetry keys).

Then we write *here* the actual telemetry that is passed to the TranslationLayer in order to be handled by VS.

There's and identical set of comments for test run, same goes for that operation.

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -309,35 +310,37 @@ internal IEnumerable<string> DefaultExtensionPaths
...
-            {
-                return extensions;
-            }
+        // Check if extensions from this assembly have already been discovered.
+        var extensions = TestExtensions?.GetExtensionsDiscoveredFromAssembly(
+            TestExtensions.GetTestExtensionCache<TPluginInfo>(),
+            extensionAssembly);
 
-            var pluginInfos = GetTestExtensions<TPluginInfo, TExtension>(new List<string>() { extensionAssembly });
+        if (extensions != null && extensions.Count > 0)
```
Evangelink, line 328: [cosmetic] 
```suggestion
        if (extensions?.Count > 0)
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,80 @@ public class TestExtensions
...
+    /// Merge two extension dictionaries.
+    /// </summary>
+    /// 
+    /// <param name="first">First extension dictionary.</param>
+    /// <param name="second">Second extension dictionary.</param>
+    /// 
+    /// <returns>
+    /// A dictionary representing the merger between the two input dictionaries.
+    /// </returns>
+    internal static Dictionary<string, HashSet<string>> MergeDictionaries(
```
Evangelink, line 104: [minor] Now that this is pure function, maybe the name shall be `MakeMergedDictionary` or `CreateMergedDictionary`. Or maybe we should even go with a more functional name `CreateMergedDiscoveredExtensions`. WDYT?

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,80 @@ public class TestExtensions
...
+        {
+            return new Dictionary<string, HashSet<string>>(second);
+        }
+        if (isSecondNullOrEmpty)
+        {
+            return new Dictionary<string, HashSet<string>>(first);
+        }
+
+        // Copy all the keys in the first dictionary into the resulting dictionary.
+        var resultMap = new Dictionary<string, HashSet<string>>(first);
```
Evangelink, line 126: Rename map here :P (maybe `mergedDictionary`, `mergedDic` or simply `merge` or `result`).

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -91,6 +91,80 @@ public class TestExtensions
...
+
+        foreach (var kvp in second)
```
Evangelink, line 128: [minor] You could also deconstruct the tuple here to have more meaningful "key" and "value" names.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +485,42 @@ internal void InvalidateCache()
         }
     }
 
+    private void AddCachedExtensionToExtensionMap<T>(
+        Dictionary<string, HashSet<string>> extensionMap,
+        string extensionType,
+        IList<T> extensions) where T : TestPluginInformation
```
Evangelink, line 491: This could be only `IEnumerable<T>` as you only check for `null` and then `Select`.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +485,42 @@ internal void InvalidateCache()
         }
     }
 
+    private void AddCachedExtensionToExtensionMap<T>(
```
Evangelink, line 488: I would also rename this map.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
Evangelink, line 491: [cosmetic]

```suggestion
        IList<T> extensions) 
        where T : TestPluginInformation
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +485,42 @@ internal void InvalidateCache()
...
+            if (kvp.Value?.Count > 0)
+            {
+                sb.AppendFormat("{0}=[{1}];", kvp.Key, string.Join(",", kvp.Value));
+            }
+        }
+
+        return sb.ToString();
+    }
+
+    private static HashSet<string> MergeSets(HashSet<string> firstSet, HashSet<string> secondSet)
```
Evangelink, line 516: Shall we also make this function pure? (i.e. not modify `firstSet`).

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -390,4 +485,42 @@ internal void InvalidateCache()
...
+    {
+        if (extensions == null)
+        {
+            return;
+        }
+
+        extensionMap.Add(extensionType, new HashSet<string>(extensions.Select(e => e.IdentifierData)));
+    }
+
+    private static string SerializeExtensionMap(IDictionary<string, HashSet<string>> extensionMap)
```
Evangelink, line 501: Same remark for `map`.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
```diff
@@ -27,6 +28,7 @@ public ParallelDiscoveryDataAggregator()
         IsAborted = false;
         TotalTests = 0;
         _metricsAggregator = new ConcurrentDictionary<string, object>();
+        DiscoveredExtensions = new Dictionary<string, HashSet<string>>();
```
Evangelink, line 31: [cosmetic] Now that you have "concrete" type, you could simplify declaration here

```suggestion
        DiscoveredExtensions = new();
```

src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: I think we can go with `CreateMergedDictionary`. Going with `CreateMergedDiscoveredExtensions` would throw us back to the discussion where we debated a more generic naming.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Thanks for pointing it out, fixed.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done, made sure there's no more reference to the word "map" :)

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelDiscoveryDataAggregator.cs
cvpoienaru: Done.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: We could in order to be consistent with the way `MergeDictionaries` works, but I didn't do it because I wanted to avoid unnecessary copying. Also the method is private, so there's no concerns regarding misusage.

What are your thoughts ? :)

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: I'd rather keep it as it is since we decided the code is quite generic. We probably shouldn't use more specific names in this case.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
Evangelink: My only concern is that we modify the set of the 1st dictionary (exemple: https://dotnetfiddle.net/?msclkid=3643415baf4711ec9335446118637980)

cvpoienaru: Ah, you're totally right, I missed that we don't do a true deep copy when copying the dictionary. 
Case closed then, copying the hashset it is. Many thanks for the input ! :)

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -385,21 +385,21 @@ internal TestExtensions GetExtensionsDiscoveredFromAssembly(string extensionAsse
...
-        AddCachedExtensionToExtensionMap(extensionMap, "TestDiscoverers", TestDiscoverers?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "TestExecutors", TestExecutors?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "TestExecutors2", TestExecutors2?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "TestSettingsProviders", TestSettingsProviders?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "TestLoggers", TestLoggers?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "TestHosts", TestHosts?.Values.ToList());
-        AddCachedExtensionToExtensionMap(extensionMap, "DataCollectors", DataCollectors?.Values.ToList());
-
-        return extensionMap;
+        AddCachedExtensionToDictionary(extensions, "TestDiscoverers", TestDiscoverers?.Values.ToList());
```
Evangelink, line 394: Sorry I should have called it out in the previous review (feel free to ignore it, the perf impact is most likely neglectable here). The `ToList` call is no longer necessary as you now allow `IEnumerable`.

Evangelink (approved):

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
cvpoienaru: Oh, that makes sense, many thanks for pointing it out. I'll need to push another commit either way since the current build has finished in the meantime but it didn't report the status back to Github so it appears as if it's still running ...

nohwnd (approved):

PR #3531, opened 2022-03-30
Fix smoke and acceptance tests

nohwnd:


test/TestAssets/DiscoveryTestProject/LongDiscoveryTestClass.cs
```diff
@@ -146,7 +146,7 @@ public TestMethodWithDelayAttribute()
         {
             // This will be multiplied by 3 because the framework will internally create this
             // attribute 3 times. And by another 3 because we have 3 slow tests.
-            Thread.Sleep(100);
+            Thread.Sleep(500);
```
Evangelink, line 149: Is it just some "random" value hoping we will wait enough?

Evangelink (approved):

cvpoienaru (approved):

test/TestAssets/DiscoveryTestProject/LongDiscoveryTestClass.cs
nohwnd: Yes!

nohwnd: Sadly 😢

Evangelink: It seems that the original 10s was not always "long" enough.

Evangelink: I guess we are better off ignoring the test for now, WDYT?

nohwnd: Then there is probably a different problem than the timeout not being long enough. Yeah let's ignore.

nohwnd: One last attempt, I added logs. It looks like the cancellation is not working they way it should and it trips us up. The slow assembly will take a long time to report the tests, and then it will finish very quickly, and the second assembly is fast as well. So the delay we have is only adding to the start of the test, but is not keeping us discovering while we cancel as was the intent.


```
Logging diagnostics in C:\Users\jajares\AppData\Local\Temp\vstest\gkQoy\log.txt
Console runner path: C:\p\vstest2\artifacts\Debug\net451\win7-x64\vstest.console.exe
Starting Discovery.
Received test case DiscoveryTestProject3.LongDiscoveryTestClass.LongDiscoveryTestMethod
Cancelling at 4789.6157 ms.
Discovery complete at 4831 ms, with isAborted: True.
Discovery finished.
```

It cancels at 4.7s in, and is done at 100ms after that. My computer is fast so it manages to cancel on time, but CI server is probably not so it manages to discover the second dll before cancelling. 

I switched up the dlls, so the fast one cancels, and then the slow one kicks in.

```
Logging diagnostics in C:\Users\jajares\AppData\Local\Temp\vstest\pyYsO\log.txt
Console runner path: C:\p\vstest2\artifacts\Debug\net451\win7-x64\vstest.console.exe
Starting Discovery.
Received test case SampleUnitTestProject.UnitTest1.PassingTest
Cancelling at 1852.8161 ms.
Received test case SampleUnitTestProject.UnitTest1.FailingTest
Received test case SampleUnitTestProject.UnitTest1.SkippingTest
Discovery complete at 1897 ms, with isAborted: True.
Discovery finished.
```

PR #3537, opened 2022-04-01
Add nullable helpers TPDebug and StringUtils and ban system equivalents

Evangelink:
## Description

I am needing something similar for Parallel discovery and assembly load context so I thought it would be better to create a separate PR for this feature and have the other PRs simpler.

### DIFF
--- a/scripts/build/TestPlatform.Settings.targets
+++ b/scripts/build/TestPlatform.Settings.targets
@@ -53,7 +53,14 @@
   <ItemGroup>
     <PackageReference Condition=" '$(OS)' != 'Windows_NT' " Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="All" />
   </ItemGroup>
-  
+
+  <ItemGroup Condition=" '$(UseBannedApiAnalyzers)' == 'true' ">
+    <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="$(RoslynBannedApiAnalyzersVersion)">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
   <!-- Include PublicApi analyzers into all projects that are in our src directory, this can't use TestProject == true
   because some projects in test and playground are not tests, but we don't want to analyze those for public apis, because
   we don't ship them.
--- a/src/Microsoft.TestPlatform.Common/BannedSymbols.txt
+++ b/src/Microsoft.TestPlatform.Common/BannedSymbols.txt
@@ -1,0 +1,4 @@
+﻿﻿M:System.Diagnostics.TPDebug.Assert(System.Boolean); Use 'TPTPDebug.Assert' instead
+M:System.Diagnostics.TPDebug.Assert(System.Boolean,System.String); Use 'TPTPDebug.Assert' instead
+M:System.String.IsNullOrEmpty(System.String); Use 'StringUtils.IsNullOrEmpty' instead
+M:System.String.IsNullOrWhiteSpace(System.String); Use 'StringUtils.IsNullOrWhiteSpace' instead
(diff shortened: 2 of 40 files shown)


scripts/build/TestPlatform.Settings.targets
```diff
@@ -53,7 +53,14 @@
   <ItemGroup>
     <PackageReference Condition=" '$(OS)' != 'Windows_NT' " Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="All" />
   </ItemGroup>
-  
+
+  <ItemGroup Condition=" '$(UseBannedApiAnalyzers)' == 'true' ">
```
Evangelink, line 57: Any preference between copy-pasting the full package reference block in all projects using the banned api analyzer or this (i.e. adding a simple property)?

src/Microsoft.TestPlatform.Common/BannedSymbols.txt
```diff
@@ -0,0 +1,4 @@
+﻿﻿M:System.Diagnostics.TPDebug.Assert(System.Boolean); Use 'TPTPDebug.Assert' instead
```
Evangelink, line 1:
```suggestion
﻿﻿M:System.Diagnostics.TPDebug.Assert(System.Boolean); Use 'TPDebug.Assert' instead
```

src/Microsoft.TestPlatform.Common/BannedSymbols.txt
```diff
@@ -0,0 +1,4 @@
...
+M:System.Diagnostics.TPDebug.Assert(System.Boolean,System.String); Use 'TPTPDebug.Assert' instead
```
Evangelink, line 2:
```suggestion
M:System.Diagnostics.Debug.Assert(System.Boolean,System.String); Use 'TPDebug.Assert' instead
```

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
```diff
@@ -6,11 +6,13 @@
   <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
   <PropertyGroup>
     <AssemblyName>Microsoft.VisualStudio.TestPlatform.Common</AssemblyName>
-    <TargetFrameworks>netstandard2.0;netstandard1.3;net451</TargetFrameworks>
+    <TargetFrameworks>net6.0;netstandard2.0;netstandard1.3;net451</TargetFrameworks>
```
Evangelink, line 9: We don't want to ship the net6.0, we simply add it to avoid duplicate definition in tests as they target net6.0 but Common didn't target netcoreapp3.0 so was always having the `NullableAttributes`.

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -85,7 +85,7 @@ internal void AddSearchDirectories(IEnumerable<string> directories)
     /// </returns>
     private Assembly OnResolve(object sender, AssemblyResolveEventArgs args)
     {
-        if (string.IsNullOrEmpty(args?.Name))
+        if (StringUtils.IsNullOrEmpty(args?.Name))
```
Evangelink, line 88: Cannot use the extension method syntax here because of the `?.`.

nohwnd:
We should wait with this until we snap for 17.2, at least. @MarcoRossignoli did similar "internal" change and it blew up.

nohwnd (changes requested):
Wait till 17.3 at least.

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
nohwnd:
```suggestion
        if (args is null || args.Name.IsNullOrEmpty())
```

Maybe?

nohwnd: It's probably not much better, and maybe even worse because you need to get the condition right.

Evangelink: I don't know, both are ok to me. We could maybe go with `arg?.Name is { Length: > 0 }` in this case but for people not used to pattern matching that might be too much.

nohwnd: Yeah, way too much. Imho the original static call is okay, and it chains well whe there are more things that can be null in a deeper object structure.

nohwnd: Yeah, way too much. Imho the original static call is okay, and it chains well whe there are more things that can be null in a deeper object structure.

nohwnd (approved):

PR #3538, opened 2022-04-01
Use !! notation instead of ValidateArg.NotNull

Evangelink:


src/Microsoft.TestPlatform.ObjectModel/Utilities/AssemblyHelper.cs
```diff
@@ -32,14 +32,11 @@ public static class AssemblyHelper
...
     /// </summary>
-    public static bool? DoesReferencesAssembly(string source, AssemblyName referenceAssembly)
+    public static bool? DoesReferencesAssembly(string source, AssemblyName referenceAssembly!!)
     {
         try
         {
             ValidateArg.NotNullOrEmpty(source, nameof(source));
-            ValidateArg.NotNull(referenceAssembly, nameof(referenceAssembly));
-
-            Debug.Assert(!string.IsNullOrEmpty(source));
```
Evangelink, line 42: Unneeded assert as we throw if null or empty a few lines before.

Evangelink, line 42: Unneeded assert as we throw if null or empty a few lines before.

nohwnd (approved):

PR #3556, opened 2022-04-08
Do not remove packages folder in ci

Haplois:


nohwnd:
You are reverting to what I did recently. It is okay if you keep the packages unzipped always, not only in ci build. It was useful for me to have them unzipped after build, because then I can easily look into the package and pick and choose dlls from there. 

The only thing I would like to ensure, if this is for signing checks, is that if the folder is deleted then the signing check will fail. Instead of passing.

Haplois:
> @nohwnd
> The only thing I would like to ensure, if this is for signing checks, is that if the folder is deleted then the signing check will fail. Instead of passing.

No, we restore source-build packages there which was built by another pipeline. This removes those packages and manifests. In CI, this folder is already clean and it gets cleared after pipeline is done. So I just disabled this in CI.

nohwnd:
Please comment that in the code, that is not obvious.

nohwnd (approved):

PR #3559, opened 2022-04-08
Fixed test extensions telemetry formatting

cvpoienaru:
Fixed test extensions telemetry formatting

### DIFF
--- a/THIRD-PARTY-NOTICES.txt
+++ b/THIRD-PARTY-NOTICES.txt
@@ -1,5 +1,5 @@
-vstest uses third-party libraries or other resources that may be
-distributed under licenses different than the vstest software.
+VSTest uses third-party libraries or other resources that may be
+distributed under licenses different than the VSTest software.
 
 In the event that we accidentally failed to list a required notice, please
 bring it to our attention by posting an issue.
--- a/THIRD-PARTY-NOTICES.txt
+++ b/THIRD-PARTY-NOTICES.txt
@@ -1,0 +1,34 @@
+vstest uses third-party libraries or other resources that may be
+distributed under licenses different than the vstest software.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention by posting an issue.
+
+The attached notices are provided for information only.
+
+License notice for SimpleJSON
+-------------------------------------
+https://github.com/Bunny83/SimpleJSON
+
+The MIT License (MIT)
+
+Copyright (c) 2012-2017 Markus Göbel (Bunny83)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
@@ -12,6 +12,8 @@ using Microsoft.VisualStudio.TestPlatform.Common.Telemetry;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
 
+using Newtonsoft.Json;
+
 #nullable disable
 
 namespace Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.Utilities;
@@ -162,7 +164,7 @@ public class TestExtensions
     {
         metrics.Add(
             TelemetryDataConstants.DiscoveredExtensions,
-            SerializeExtensionDictionary(extensions));
+            JsonConvert.SerializeObject(extensions));
     }
 
     /// <summary>
@@ -497,21 +499,6 @@ public class TestExtensions
         }
 
         extensionDict.Add(extensionType, new HashSet<string>(extensions.Select(e => e.IdentifierData)));
-    }
-
-    private static string SerializeExtensionDictionary(IDictionary<string, HashSet<string>> extensions)
-    {
-        StringBuilder sb = new();
-
-        foreach (var kvp in extensions)
-        {
-            if (kvp.Value?.Count > 0)
-            {
-                sb.AppendFormat("{0}=[{1}];", kvp.Key, string.Join(",", kvp.Value));
-            }
-        }
-
-        return sb.ToString();
     }
 
     private static HashSet<string> MergeSets(HashSet<string> firstSet, HashSet<string> secondSet)
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
@@ -124,9 +124,8 @@ public class TestExtensions
         }
 
         // Copy all the keys in the first dictionary into the resulting dictionary.
-        var result = new Dictionary<string, HashSet<string>>(
-            first.Where(kvp => (kvp.Value != null && kvp.Value.Count > 0))
-            .ToDictionary(kvp => kvp.Key, kvp => kvp.Value));
+        var result = first.Where(kvp => (kvp.Value != null && kvp.Value.Count > 0))
+            .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
 
         foreach (var kvp in second)
         {
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
@@ -5,14 +5,13 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Text;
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
```diff
@@ -34,6 +34,9 @@
     <ProjectReference Include="..\Microsoft.TestPlatform.CoreUtilities\Microsoft.TestPlatform.CoreUtilities.csproj">
       <FromP2P>true</FromP2P>
     </ProjectReference>
+    <PackageReference Include="Newtonsoft.Json">
+      <Version>$(JsonNetVersion)</Version>
```
nohwnd, line 38: Did you analyze the impact of this? Why is it safe?

nohwnd (changes requested):

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
cvpoienaru: @nohwnd the only other project that depends on `Newtonsoft.Json` is `Microsoft.TestPlatform.CommunicationUtilities`, so I checked that both `Microsoft.TestPlatform.CommunicationUtilities` and `Microsoft.VisualStudio.TestPlatform.Common` are packaged together in the following packages (so the dependency requirements were already in place in each `.nuspec`):

- `Microsoft.TestPlatform.TranslationLayer`
- `Microsoft.TestPlatform.Internal.Uwp`
- `Microsoft.TestPlatform`
- `Microsoft.TestPlatform.Portable`

Because of this, I reconsidered the previous position where I didn't want to take a dependency on `Newtonsoft.Json`.

Other than that, I did manual validation for the metrics format and here's a sample of how it looks:

```
{
   "TestSettingsProviders":[
      "WebTestRunConfiguration",
      "LegacySettings",
      "MSTest"
   ],
   "TestHosts":[
      "HostProvider://DefaultTestHost",
      "HostProvider://DotnetTestHost"
   ]
}
```

Do you have any particular checks in mind ?

nohwnd: This is not problematic for projects that already depend on the library you are adding, as long as they both depend on the same version. 

This is problematic for projects that don't depend on it. Did you check if all the projects that depend on common already depend on newtonsoft.json and ship it?

THIRD-PARTY-NOTICES.txt
```diff
@@ -0,0 +1,34 @@
+vstest uses third-party libraries or other resources that may be
```
Evangelink, line 1: Nit: 
```suggestion
VSTest uses third-party libraries or other resources that may be
```

or
```suggestion
Visual Studio Test Platform uses third-party libraries or other resources that may be
```

Evangelink (commented):
You will have to disable the various analyzers triggering on SimpleJSON.cs or fix the diagnostics.

THIRD-PARTY-NOTICES.txt
cvpoienaru: Fixed, thanks. Decided to go with VSTest :)

cvpoienaru:
> You will have to disable the various analyzers triggering on SimpleJSON.cs or fix the diagnostics.

@Evangelink thanks, I actually fixed a whole lot of analyzer issues in `SimpleJSON` before but somehow missed 3 of them :D Fixed them as well now (actually suppressed two and fixed one).

I was going to point out before asking for review that there are a bunch of styling fixes in the `SimpleJSON.cs` file in order to make it compliant with our style standards.

src/Microsoft.TestPlatform.Common/Microsoft.TestPlatform.Common.csproj
cvpoienaru: Moved to `SimpleJSON`, resolving the conversation.

Evangelink (approved):
LGMT

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestExtensions.cs
```diff
@@ -124,7 +124,9 @@ public class TestExtensions
         }
 
         // Copy all the keys in the first dictionary into the resulting dictionary.
-        var result = new Dictionary<string, HashSet<string>>(first);
+        var result = new Dictionary<string, HashSet<string>>(
```
Evangelink, line 127: Is it still required to do `new Dictionary...`?

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
```diff
@@ -239,4 +241,110 @@ public void GetExtensionsDiscoveredFromAssemblyShouldReturnTestDiscoveresAndLogg
...
+        // Merging two empty dictionaries should result in an empty dictionary.
+        Assert.IsTrue(merged.Count == 0);
+
+        // Make sure the method is "pure" and returns a new reference.
+        Assert.IsFalse(ReferenceEquals(merged, first));
+        Assert.IsFalse(ReferenceEquals(merged, second));
+    }
+
+    [TestMethod]
+    public void MergedDictionaryOfOneEmptyAndOneNotEmptyDictionaryShouldBeNotEmpty()
```
nohwnd, line 261: should be not empty is not really what you want it to do.  ShouldContainAllTheItemsOfTheNonEmptyDictionary is more like it. 

also use Fluent assertions and should be equivalent to simplify the check.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
```diff
@@ -239,4 +241,110 @@ public void GetExtensionsDiscoveredFromAssemblyShouldReturnTestDiscoveresAndLogg
...
+        // Make sure the method stays "pure" and returns a new reference regardless of the input.
+        Assert.IsFalse(ReferenceEquals(merged1, first));
+        Assert.IsFalse(ReferenceEquals(merged1, second));
+        Assert.IsFalse(ReferenceEquals(merged2, first));
+        Assert.IsFalse(ReferenceEquals(merged2, second));
+        Assert.IsFalse(ReferenceEquals(merged1, merged2));
+    }
+
+    [TestMethod]
+    public void MergedDictionaryShouldBeSuccessful()
```
nohwnd, line 288: BeSuccessful says nothing about what the desired result is. 

Also use fluent assertions to do the equality / equivalency check.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
```diff
@@ -239,4 +241,110 @@ public void GetExtensionsDiscoveredFromAssemblyShouldReturnTestDiscoveresAndLogg
...
+        Assert.IsTrue(merged.Count == expected.Count);
+        Assert.IsFalse(merged.Where(kvp => !expected.ContainsKey(kvp.Key)).Any());
+        Assert.IsFalse(expected.Where(kvp => !merged.ContainsKey(kvp.Key)).Any());
+
+        // Make sure the hashsets for each key are equal.
+        Assert.IsFalse(merged.Where(kvp => !kvp.Value.SequenceEqual(expected[kvp.Key])).Any());
+    }
+
+    [TestMethod]
+    public void AddExtensionTelemetryShouldBeSuccessful()
```
nohwnd, line 330: Same remark for the name, being successful says nothing about the actual intent.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
```diff
@@ -239,4 +241,110 @@ public void GetExtensionsDiscoveredFromAssemblyShouldReturnTestDiscoveresAndLogg
...
+        var telemetryData = new Dictionary<string, object>();
+        var extensions = new Dictionary<string, HashSet<string>>
+        {
+            { "key1", new HashSet<string>(new List<string>() { "ext1", "ext2", "ext3", "ext4", "ext5" }) },
+            { "key4", new HashSet<string>(new List<string>() { "ext1" }) },
+            { "key5", new HashSet<string>(new List<string>() { "ext1", "ext2" }) }
+        };
+
+        var expectedTelemetry =
+            "{\"key1\":[\"ext1\",\"ext2\",\"ext3\",\"ext4\",\"ext5\"],"
```
Evangelink, line 341: I think we should be able to use the raw string feature which would increase readability here.

Evangelink (approved):

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
cvpoienaru: Done.

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
```diff
@@ -239,4 +242,110 @@ public void GetExtensionsDiscoveredFromAssemblyShouldReturnTestDiscoveresAndLogg
...
+        var telemetryData = new Dictionary<string, object>();
+        var extensions = new Dictionary<string, HashSet<string>>
+        {
+            { "key1", new HashSet<string>(new List<string>() { "ext1", "ext2", "ext3", "ext4", "ext5" }) },
+            { "key4", new HashSet<string>(new List<string>() { "ext1" }) },
+            { "key5", new HashSet<string>(new List<string>() { "ext1", "ext2" }) }
+        };
+
+        var expectedTelemetry =
+            @"{""key1"":[""ext1"",""ext2"",""ext3"",""ext4"",""ext5""],"
```
Evangelink, line 342: Was thinking about https://docs.microsoft.com/dotnet/csharp/whats-new/csharp-11#raw-string-literals so something but current solution is better than all these backslashes :)

nohwnd (approved):

test/Microsoft.TestPlatform.Common.UnitTests/ExtensionFramework/Utilities/TestExtensionsTests.cs
cvpoienaru: So I've been experimenting with this new raw string functionality but I got stuck at removing line breaks from the multiline string. Since the JSON-formatted string we get by converting the extension dictionary to JSON has no additional spaces and no line brakes, that causes my assert to break. I can do a `Replace` but that seems overly complicated for a test and goes against a more "readable" string in my opinion, so it'll probably be best to stick with the current form of the string unless there's a simpler solution to remove line breaks that I may have missed.

For reference this is the way my raw string looks like right now:
```
var expectedTelemetry = """
    {"key1":["ext1","ext2","ext3","ext4","ext5"], <--LINE BREAK HERE
    "key4":["ext1"], <--LINE BREAK HERE
    "key5":["ext1","ext2"]}
    """;
```

PR #3560, opened 2022-04-08
Add baseline to perf tests

nohwnd:
## Description
Adds test adapter called Perfy that has almost no overhead to the perf tests. This will help us separate TP and adapter overhead. TEST_COUNT env variable is set for the adapter which generates that many test cases, or test results. 

This is work is common to all adapters, and is reasonably cheap, any overhead that is measured in the run, can be attributed to TestPlatform. In our results we can then see how much time it took for Perfy to run 1, 100, 1000, and 10k tests, and compare that with other adapters, which will show e.g. 1.1s for Perfy and 1.5s for NUnit, meaning that 1.1s is TestPlatform overhead, and 400ms is NUnit overhead.


test/TestAssets/performance/Perfy.TestAdapter/Perfy.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    [ExtensionUri(Id)]
+    [DefaultExecutorUri(Id)]
+    public class Perfy : ITestDiscoverer, ITestExecutor
+    {
+        static Perfy()
+        {
+            // No meaning to the number, it is just easy to find when it breaks. Better than returning 1000 or 0 which are both
+            // less suspicious. It could throw, but that is bad for interactive debugging.
+            Count = int.TryParse(Environment.GetEnvironmentVariable("TEST_COUNT") ?? "356", out var count) ? count : 356;
```
cvpoienaru, line 27: nit: Since we have no preference for the number, can we still use something like a power of 2, or maybe a prime number ?

cvpoienaru: Also, can we make "TEST_COUNT" a constant ?

test/TestAssets/performance/Perfy.TestAdapter/Perfy.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Data;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace PerfyPassing
```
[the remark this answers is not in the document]
nohwnd: Won't work. Our test assets don't use preview language features, yet.

test/TestAssets/performance/Perfy.TestAdapter/Perfy.cs
nohwnd: Why power of 2 or prime number? :)

Why constant and where it would go? I don't want to put it into PerfTestBase, and there are 4 tests that use it in 2 different classes. If you get it wrong, the test will fail. 

In the code above it is used in a single place in single file, so making it a constant only adds mental overhead.

test/TestAssets/performance/Perfy.TestAdapter/Perfy.cs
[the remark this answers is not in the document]
nohwnd: I know it won't work because I did it and it did not work. Same for Uri initializer. And I don't want to just add the lang feature tag into this project when the other projects don't use it.

MarcoRossignoli (approved):

test/TestAssets/performance/Perfy.TestAdapter/Perfy.cs
nohwnd: I merged it to start measuring. Let me know if you feel strongly about this, and I can make another PR for it.

PR #3583, opened 2022-04-19
Fix hang dump type none

### LINKED ISSUE #15555: Dotnet test `--blame-hang-dump-type` does not seem to have effect.
When I run `--blame-hang-dump-type none`, the Microsoft docs specify:

_"It should be full, mini, or none. When none is specified, test host is terminated on timeout, but no dump is collected."_ 

## My Command
```
dotnet test .\Faultify.Benchmark.NUnit\Faultify.Benchmark.NUnit.csproj --blame-hang-timeout 2000 --blame-hang-dump-type none

OR

dotnet test .\Faultify.Benchmark.XUnit\Faultify.Benchmark.XUnit.csproj --blame-hang-timeout 2000 --blame-hang-dump-type none
```

## Result
When the test timesout it generates:
image

```
The active test run was aborted. Reason: Test host process crashed
Data collector 'Blame' message: The specified inactivity time of 2 seconds has elapsed. Collecting hang dumps from testhost and its child processes.
Data collector 'Blame' message: Dumping 33252 - testhost.

**Attachments:**
  XXX\bin\Debug\netcoreapp3.1\test-duplication-3\TestResults\583a5b3b-981c-414d-92dc-316891a31e19\Sequence_d4e9a65520c549d7a5381b7db58ff94a.xml
  XXX\bin\Debug\netcoreapp3.1\test-duplication-3\TestResults\583a5b3b-981c-414d-92dc-316891a31e19\testhost_33252_20210129T085444_hangdump.dmp
```
## Version
.NET SDK (5.0.101)
Microsoft.NET.Test.Sdk 16.9.0-preview-20210127-04
Projects are dotnet core 3.1

NUnit 3.13.0
XUnit3TestAdapter 4.0.0-beta.1

XUnit 2.4.1
xunit.runner.visualstudio 2.4.3

## Replicate
The following zip contains both a NUnit and XUnit project that contains a timeout test. When running the commands in the `README.txt` you will see it generates a `TestResults` while the argument `--blame-hang-dump-type none` is provided.

[Replication.zip](https://github.com/dotnet/sdk/files/5892536/Replication.zip)

## Question
How can I disable this dump, am I doing something wrong here?

nohwnd:
## Description
Fix hang dump type None which was planned and documented to just kill testhost but was never properly implemented.

## Related issue
Fix https://github.com/dotnet/sdk/issues/15555

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -40,14 +40,18 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
     private XmlElement _configurationElement;
     private int _testStartCount;
     private int _testEndCount;
-    private bool _collectProcessDumpOnTrigger;
-    private bool _collectProcessDumpOnTestHostHang;
+    private bool _collectProcessDumpOnCrash;
+    private bool _collectProcessDumpOnHang;
     private bool _collectDumpAlways;
-    private bool _processFullDumpEnabled;
-    private bool _inactivityTimerAlreadyFired;
     private string _attachmentGuid;
+
+    private CrashDumpType _crashDumpType;
+    private HangDumpType? _hangDumpType;
+
+    private bool _inactivityTimerAlreadyFired;
     private IInactivityTimer _inactivityTimer;
     private TimeSpan _inactivityTimespan = TimeSpan.FromMinutes(DefaultInactivityTimeInMinutes);
+
     private int _testHostProcessId;
     private string _targetFramework;
     private readonly List<KeyValuePair<string, string>> _environmentVariables = new();
@@ -131,11 +135,11 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
         if (_configurationElement != null)
         {
             var collectDumpNode = _configurationElement[Constants.DumpModeKey];
-            _collectProcessDumpOnTrigger = collectDumpNode != null;
+            _collectProcessDumpOnCrash = collectDumpNode != null;
 
-            if (_collectProcessDumpOnTrigger)
+            if (_collectProcessDumpOnCrash)
             {
-                ValidateAndAddTriggerBasedProcessDumpParameters(collectDumpNode);
+                ValidateAndAddCrashProcessDumpParameters(collectDumpNode);
 
                 // enabling dumps on MacOS needs to be done explicitly https://github.com/dotnet/runtime/pull/40105
                 _environmentVariables.Add(new KeyValuePair<string, string>("COMPlus_DbgEnableElfDumpOnMacOS", "1"));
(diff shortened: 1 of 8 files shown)


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -131,11 +135,11 @@ public BlameCollector()
         if (_configurationElement != null)
         {
             var collectDumpNode = _configurationElement[Constants.DumpModeKey];
-            _collectProcessDumpOnTrigger = collectDumpNode != null;
+            _collectProcessDumpOnCrash = collectDumpNode != null;
```
Evangelink, line 138: Maybe that's outside of the scope of this PR but do we need this bool field? I feel like we could just do the null check here on line 140 and use `_crashDumpType` later on. 

I think the same remark applies for the `_collectProcessDumpOnHang` field.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -215,63 +219,67 @@ private void CollectDumpAndAbortTesthost()
...
+                catch (Exception ex)
+                {
+                    _logger.LogError(_context.SessionDataCollectionContext, $"Blame: Collecting hang dump failed with error.", ex);
+                }
             }
-            catch (Exception ex)
+            else
             {
-                _logger.LogError(_context.SessionDataCollectionContext, $"Blame: Collecting hang dump failed with error.", ex);
+                EqtTrace.Info("BlameCollector.CollectDumpAndAbortTesthost: Custom path to dump directory was provided via VSTEST_DUMP_PATH. Skipping attachment upload, the caller is responsible for collecting and uploading the dumps themselves.");
```
Evangelink, line 280: [minor][pedantic] Not entirely sure here but shouldn't it be `itself/himself/herself` instead of `themselves` as we have one caller?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/Constants.cs
```diff
@@ -122,6 +122,11 @@ internal static class Constants
     /// </summary>
     public const string MiniConfigurationValue = "Mini";
 
+    /// <summary>
+    /// Configuration value for none
+    /// </summary>
+    public const string NoneConfigurationValue = "None";
```
Evangelink, line 128: Is it used?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: I like having a bool field that enables the whole functionality more than leaning on nullability on a field that selects which dump type should be used. Feels unnecessary to mix two different concepts (enabling it and dump type) into one field, when fields are free. It is also easier to see that we really meant to check if crash dump is enabled even though we are currently in hang dump related method: 

image


e.g. here where we need to detach procdump before killing testhost.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: It could be, but because we are not sure about the gender and we don't need to imply one, we can use a gender-neutral pronoun "they" which is valid in singular form: https://en.wikipedia.org/wiki/Singular_they

Evangelink: Oh thanks didn't know about this.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
Evangelink: Fine by me.

Evangelink (approved):
Just one remark about whether or not the newly added constant is used somewhere.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/Constants.cs
nohwnd: Nope, I changed the whole logic around parsing params to use enums and removed it.

test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/BlameCollectorTests.cs
```diff
@@ -586,7 +587,7 @@ public void TriggerTestHostLaunchedHandlerShouldLogWarningForWrongDumpType()
     public void TriggerTestHostLaunchedHandlerShouldLogWarningForNonBooleanCollectAlwaysValue()
     {
         var dumpConfig = GetDumpConfigurationElement();
-        var dumpTypeAttribute = dumpConfig.OwnerDocument.CreateAttribute("DumpType");
+        var dumpTypeAttribute = dumpConfig.OwnerDocument.CreateAttribute("CollectAlways");
```
nohwnd, line 590: This test was testing another thing that what it said in the name.

Evangelink (approved):

PR #3588, opened 2022-04-20
Condition out BannedApiAnalyzers usage in source-build

MichaelSimons:
BannedApiAnalyzers is not part of source-build and therefore cannot be referenced without introducing prebuilts.  Other repos have conditioned out the reference for source-build.  This is viewed as acceptable as the "normal" product builds will run the analyzers and surface any issues.

cc @lbussell, @crummel


MichaelSimons:
cc @Haplois

Haplois (approved):

MichaelSimons:
@Haplois, Thanks for reviewing and merging.  Will the changes from main flow into dotnet 7.0 (e.g. main) or do I need to backport this PR to another branch?

PR #3589, opened 2022-04-21
Adapter time without TP overhead

nohwnd:
This is a proof of concept to report time that is actually spent executing adapter code, and not executing test platform code. Every callback is wrapped in timer and the total is subtracted from the total. It works fine with non-parallel execution but probably won't work with parallel execution.

### DIFF
--- a/playground/TestPlatform.Playground/Program.cs
+++ b/playground/TestPlatform.Playground/Program.cs
@@ -42,7 +42,7 @@ internal class Program
         var consoleOptions = new ConsoleParameters
         {
             LogFilePath = Path.Combine(here, "logs", "log.txt"),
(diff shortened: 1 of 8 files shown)


nohwnd:
overhead

To illustrate what I mean. Normally the execution of adapter and TP callback to report the test would be interspersed, but to show the point better I've split it to generate test results, and then report them in bulk. 

The green part is representing the actual work that adapter does to run tests. Plus a relatively tiny overhead of creating the objects that TP understands `TestCase` and `TestResult`. In real adapter this would mean lookup tests by reflection (for example), run them, and convert the results. This example overhead does nothing of that, it just generates the results.

The red part represents TP callback to report the result, this should be accounted as TP overhead, and not as time spent in adapter as it is now.

PR #3608, opened 2022-04-26
Better deserialization and serialization performance

nohwnd:
## Description
Improve performance of deserialization and serialization. 
For serialization we save a lot of time by not serializing into JToken and then to string. 

And for deserialization the biggest impact is coming from not deserializing payloads into JTokens, and not deserializing them every time we inspect the header of the message.


src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -81,7 +135,106 @@ public T DeserializePayload<T>(Message message)
...
+    {
+        // PERF: This can be also done slightly better using ReadOnlySpan<char> but we don't have that available by default in .NET Framework
+        // and the speed improvement does not warrant additional dependency. This is already taking just few ms for 10k messages.
+        version = 0;
+        messageType = null;
+
+        try
+        {
+            // {"Version":6,"MessageType":"TestExecution.GetTestRunnerProcessStartInfoForRunAll","Payload":{
+            if (rawMessage[2] != 'V')
```
MarcoRossignoli, line 175: out of range check?

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -81,7 +135,106 @@ public T DeserializePayload<T>(Message message)
...
+                return Deserialize<T>(payloadSerializer, Deserialize<VersionedMessage>(rawMessage).Payload);
+            }
+        }
+        else
+        {
+            return Deserialize<T>(payloadSerializer, message.Payload);
+        }
+    }
+
+    private bool FastHeaderParse(string rawMessage, out int version, out string messageType)
```
MarcoRossignoli, line 165: General not mandatory info, if you start to check from the highest possible index all lower access will be optimized without range check by jitter.

https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgMwAEOZAwmQN5FmMXmwJkCyAFKwNoC6ZCAEo6DJuNZkATgFMUAVwA2bALxkADAG4x4xjt2yFysgGo1EHtj7bCupoaVszgnrmv7xD484vr3tuzIPJkwAdmk5RxtxAF99fVIySXZsbhgEfkERegC7SS9VDWjA4MYC03MeP2K7cp9Xf0CIoydKqxrdUopwgo64whigA==

nohwnd: I don't know the highest index until I find the last message type `"`, but I don't know where to start looking for it, until after I know how long the version number is. 

Or is there some trick you can think about?

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -81,7 +135,106 @@ public T DeserializePayload<T>(Message message)
...
+    }
+
+    /// <summary>
+    ///  Try getting substring until a given delimiter, but don't search more characters than maxSearchLength.
+    /// </summary>
+    private bool TryGetSubstringUntilDelimiter(string rawMessage, int start, char character, int maxSearchLength, out string substring, out int delimiterIndex)
+    {
+        var length = rawMessage.Length;
+        var searchEnd = start + maxSearchLength;
+        for (int i = start; i < length && i <= searchEnd; i++)
```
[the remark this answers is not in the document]
nohwnd: Right, I was looking exactly for this but did not find it, and I was surprised that it is not there :D Not sure if it would be any faster, and if I get the count wrong it will throw. v2 maybe? :)

NuGet.config
```diff
@@ -20,6 +20,7 @@
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="vs-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vs-impl/nuget/v3/index.json" />
     <add key="vssdk" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/vssdk/nuget/v3/index.json" />
+    <add key="mstest" value="C:\p\testfx\artifacts\Debug\MSTestPackages" />
```
Evangelink, line 23: Probably something to revert.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -21,8 +25,12 @@ public class JsonDataSerializer : IDataSerializer
 {
     private static JsonDataSerializer s_instance;
 
+    private static bool DisableFastJson { get; } = FeatureFlag.Instance.IsSet(FeatureFlag.DISABLE_FASTER_JSON_SERIALIZATION);
```
Evangelink, line 25: Any reason to make it a property rather than a field?

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -21,8 +25,12 @@ public class JsonDataSerializer : IDataSerializer
...
+
     private static JsonSerializer s_payloadSerializer; // payload serializer for version <= 1
     private static JsonSerializer s_payloadSerializer2; // payload serializer for version >= 2
+    private static JsonSerializerSettings s_jsonSettings7; // serializer settings for faster json
```
Evangelink, line 29: What about `s_fastJsonSettings`? Instead of a number.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -68,7 +94,35 @@ private JsonDataSerializer()
     /// <returns>A <see cref="Message"/> instance.</returns>
     public Message DeserializeMessage(string rawMessage)
     {
-        return Deserialize<VersionedMessage>(s_serializer, rawMessage);
+        if (!DisableFastJson)
```
Evangelink, line 94: [cosmetic] I would revert the if and remove the else part to reduce nesting.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -81,7 +135,106 @@ public T DeserializePayload<T>(Message message)
     {
         var versionedMessage = message as VersionedMessage;
         var payloadSerializer = GetPayloadSerializer(versionedMessage?.Version);
-        return Deserialize<T>(payloadSerializer, message.Payload);
+
+        if (!DisableFastJson)
+        {
+            var messageWithRawMessage = message as MessageWithRawMessage;
+            var rawMessage = messageWithRawMessage.RawMessage;
```
Evangelink, line 139: Potential NRE, you did a `as` cast and here you do a member access.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -81,7 +135,106 @@ public T DeserializePayload<T>(Message message)
     {
         var versionedMessage = message as VersionedMessage;
         var payloadSerializer = GetPayloadSerializer(versionedMessage?.Version);
-        return Deserialize<T>(payloadSerializer, message.Payload);
+
+        if (!DisableFastJson)
```
Evangelink, line 136: [cosmetic] Same suggestion about inverting the if and remove the else to reduce nesting.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
Evangelink: If you only check for size 3 you will need to do a similar check for `10` and `11` later on so I would check at least this size.

nohwnd: I am checking for minimal valid message size.

src/Microsoft.TestPlatform.ObjectModel/TestProperty/CustomKeyValueConverter.cs
```diff
@@ -38,11 +46,12 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+            // PERF: The values returned here can possibly be cached, but the benefits are very small speed wise,
+            // and it is unclear how many distinct objects we get, and how much memory this would consume. I was seeing around 100ms improvement on 10k tests.
+
             using var stream = new MemoryStream(Encoding.Unicode.GetBytes(data));
             // Converting Json data to array of KeyValuePairs with duplicate keys.
-            var serializer = new System.Runtime.Serialization.Json.DataContractJsonSerializer(typeof(TraitObject[]));
-            var listOfTraitObjects = serializer.ReadObject(stream) as TraitObject[];
-
+            var listOfTraitObjects = _serializer.ReadObject(stream) as TraitObject[];
             return listOfTraitObjects.Select(i => new KeyValuePair<string, string>(i.Key, i.Value)).ToArray();
```
Evangelink, line 55: Potential NRE, you are doing a `as` cast.

NuGet.config
nohwnd: done

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -36,13 +44,31 @@ private JsonDataSerializer()
...
+            ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
         };
 
+        s_jsonSettings = jsonSettings;
+
         s_serializer = JsonSerializer.Create();
         s_payloadSerializer = JsonSerializer.Create(jsonSettings);
         s_payloadSerializer2 = JsonSerializer.Create(jsonSettings);
 
+        var jsonSettings7 = new JsonSerializerSettings
```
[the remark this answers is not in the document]
nohwnd: nope.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
nohwnd: Inverted

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -68,7 +94,35 @@ private JsonDataSerializer()
     /// <returns>A <see cref="Message"/> instance.</returns>
     public Message DeserializeMessage(string rawMessage)
     {
-        return Deserialize<VersionedMessage>(s_serializer, rawMessage);
+        if (!DisableFastJson)
+        {
+            // PERF: Try grabbing the version and message type from the string directly, we are pretty certain how the message is serialized
+            // when the format does not match all we do is that we check if 6th character in the message is 'V'
+            if (!FastHeaderParse(rawMessage, out int v, out string m))
```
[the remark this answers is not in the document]
nohwnd: done

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
nohwnd: Inverted

src/Microsoft.TestPlatform.ObjectModel/TestProperty/CustomKeyValueConverter.cs
```diff
@@ -18,6 +19,13 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel;
 /// </summary>
 internal class CustomKeyValueConverter : TypeConverter
 {
+    private readonly DataContractJsonSerializer _serializer;
+
+    public CustomKeyValueConverter()
+    {
+        _serializer = new DataContractJsonSerializer(typeof(TraitObject[]));
```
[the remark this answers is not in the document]
nohwnd: done

src/Microsoft.TestPlatform.ObjectModel/TestProperty/CustomKeyValueConverter.cs
nohwnd: Fixed

Evangelink (approved):

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -36,13 +41,29 @@ private JsonDataSerializer()
             DateParseHandling = DateParseHandling.DateTimeOffset,
             DateTimeZoneHandling = DateTimeZoneHandling.Utc,
             TypeNameHandling = TypeNameHandling.None,
-            ReferenceLoopHandling = ReferenceLoopHandling.Ignore
+            ReferenceLoopHandling = ReferenceLoopHandling.Ignore,
```
MarcoRossignoli, line 44:
```suggestion
            ReferenceLoopHandling = ReferenceLoopHandling.Ignore
```

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
MarcoRossignoli: I mean you check for 2 and after for 10...so if 10 is there 2 is there...you can do it until the "known" length. 
But I think it's too much and risky maybe to skip a pair of test. 

Take it as an interesting info, it's ok as-is.

MarcoRossignoli: Oh I've missed the check for the malformed messages...so I think that now it's optimized until the 31 index access.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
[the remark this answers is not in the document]
MarcoRossignoli: > Not sure if it would be any faster

It's faster because it's vectorized https://github.com/dotnet/runtime/blob/181bf3122278f9b6ae1cd755e2ab99dfe1706604/src/libraries/System.Private.CoreLib/src/System/SpanHelpers.Char.cs#L518-L794 but I think that it's ok as-is here we don't have numbers to justify atm.

MarcoRossignoli (approved):

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
Evangelink: It's pretty convenient to keep the last comma so that when you add a new entry you don't have to bother adding it. But I am fine if we remove it too.

nohwnd: I am keeping it :)

MarcoRossignoli: It's ok to me(my 2c, never liked it...looks like a typo and it's also at the end of a list so anyway you need to add another element and a comma to keep same idea so I don't see a great benefit, looks weird for younger that comes from standard programming books.)

raghavthind2005 (approved):

PR #3610, opened 2022-04-27
Refined runsettings matching criteria for test sessions

cvpoienaru:
Refined runsettings matching criteria for test sessions

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
@@ -45,7 +45,22 @@ public class ProxyTestSessionManager : IProxyTestSessionManager
     private readonly IList<ProxyOperationManagerContainer> _proxyContainerList;
     private readonly IDictionary<string, int> _proxyMap;
     private readonly Stopwatch _testSessionStopwatch;
-    private IDictionary<string, string> _runSettingsEnvironmentVariables;
+    private IDictionary<string, string> _testSessionEnvironmentVariables = new Dictionary<string, string>();
+
+    private IDictionary<string, string> TestSessionEnvironmentVariables
+    {
+        get
+        {
+            if (_testSessionEnvironmentVariables.Count == 0)
+            {
+                _testSessionEnvironmentVariables = InferRunSettingsHelper.GetEnvironmentVariables(
+                        _testSessionCriteria.RunSettings)
+                    ?? _testSessionEnvironmentVariables;
+            }
+
+            return _testSessionEnvironmentVariables;
+        }
+    }
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ProxyTestSessionManager"/> class.
@@ -65,7 +80,6 @@ public class ProxyTestSessionManager : IProxyTestSessionManager
 
         _proxyContainerList = new List<ProxyOperationManagerContainer>();
         _proxyMap = new Dictionary<string, int>();
-        _runSettingsEnvironmentVariables = new Dictionary<string, string>();
         _testSessionStopwatch = new Stopwatch();
     }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -382,6 +384,35 @@ private void DisposeProxies()
             _proxyMap.Clear();
         }
     }
+
+    private bool CheckRunSettingsAreCompatible(string requestRunSettings)
```
drognanar, line 388: perhaps this should be converted to a void and called ProcessRunSettings? and throw if the runsettings are invalid?
might be slightly cleaner than having a bool that underneath mutates the classes fields

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -382,6 +384,35 @@ private void DisposeProxies()
...
+        if (requestEnvironmentVariables != null
+            && _runSettingsEnvironmentVariables != null
+            && (requestEnvironmentVariables.Count != _runSettingsEnvironmentVariables.Count
+                || requestEnvironmentVariables.Except(_runSettingsEnvironmentVariables).Any()))
+        {
+            return false;
+        }
+
+        // Data collection is not supported for test sessions yet.
+        if (XmlRunSettingsUtilities.IsDataCollectionEnabled(requestRunSettings))
```
drognanar, line 409: it would be good to figure out if this check can be relaxed and if so how

cvpoienaru: The check can be relaxed and I was planning to do it, but it will require more work and that means adopting Test Sessions API would be blocked for longer. This check was only added to reflect the state of the art, i.e. we cannot have data collectors when starting the test session and because of that we cannot have them when matching the testhosts with test discovery/run requests. But again, this is only temporary.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
cvpoienaru: Oh, but we don't actually mutate the class fields. The only thing that can make it look like we are is this piece of code, but this is only for caching purposes so that we don't have to reparse the test session run settings (which never change) each time we have to check if the new runsettings match the old ones:
```
if (_runSettingsEnvironmentVariables.Count == 0)
{
    _runSettingsEnvironmentVariables = InferRunSettingsHelper.GetEnvironmentVariables(
            _testSessionCriteria.RunSettings)
        ?? _runSettingsEnvironmentVariables;
}
```

Also, I don't feel like `ProcessRunSettings` is an accurate name. All this function does is to literally check if the test session run settings match the new run settings so a testhost from the current session may be reused, hence coming up with the current naming.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
```diff
@@ -382,6 +384,35 @@ private void DisposeProxies()
             _proxyMap.Clear();
         }
     }
+
+    private bool CheckRunSettingsAreCompatible(string requestRunSettings)
+    {
+        if (_runSettingsEnvironmentVariables.Count == 0)
+        {
+            _runSettingsEnvironmentVariables = InferRunSettingsHelper.GetEnvironmentVariables(
```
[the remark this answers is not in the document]
drognanar: does that mean that if the session is created with valid runsettings, but a discovery request with invalid runsettings, we'll default to session's runsettings? it might be more clear if an error would be thrown then

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
drognanar: but the cached value is what's being mutated. thus calling CheckRunSettingsAreCompatible() twice with the same runsettings would return false and then return true. usually for bool returning methods I'd expect that this is a query and so it should return the same result if called twice

cvpoienaru: That's not the case, `CheckRunSettingsAreCompatible()` will always return the same value given the same input. I wrote a test in order to confirm and also have this as a check against future changes.

That being said, I did some refactoring and hid the caching part behind a property to make the `CheckRunSettingsAreCompatible()` method easier to follow.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/ProxyTestSessionManager.cs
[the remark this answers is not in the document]
cvpoienaru: Depends on what you mean by invalid runsettings. Invalid in that they're not compatible when matched against the original test session runsettings ? If that's the case we don't default to the original test session runsettings but this method will return false and one layer above, in `DequeProxy` we throw an `InvalidOperationException`.

drognanar (approved):

PR #3622, opened 2022-05-04
Use System.Text.Json for deserialization

### LINKED ISSUE #2316: Can Newtonsoft.Json get updated to 11.x or 12.x?

nohwnd:
## Description
Uses System.Text.Json for deserialization

## Related issue
Fix https://github.com/microsoft/vstest/issues/2316


nohwnd:
This does the basics of upgrading to System.Text.Json, but there are some features missing in that library: 

- [ ] Serializing and deserializing many of the basic types, like DateTimeOffset, Type, or Exception. 
- [ ] Respecting our annotations on data types using *DataMember attributes, to avoid forcing ObjectModel to tie to any serialization library.  Which is not supported right now, and there are no plans for it: https://github.com/dotnet/runtime/issues/29975

Issue #3595, opened 2022-04-22
Does vstest support running in-process?


ViktorHofer:
This is a question therefore I removed the default issue template.

As I haven't been following what changed in vstest in the last 9 months I was wondering if VSTest already supports running in-proc. Repositories that are lower in the stack like dotnet/runtime want to avoid dependencies on the networking stack, i.e. communication over Sockets. The last time I spoke with @nohwnd, VSTest heavily depended on Sockets communication and there was not a way to force an in-process execution.

Running in-proc should also enable testing single file applications. AFAIK not being able to run in-proc is the remaining blocker for us in dotnet/runtime to adopt VSTest fully and avoid the unsupported xunit.console test runner.

nohwnd:
It is still the same, we don't want to endorse in process execution, instead we should endorse configuring the transport between processes, to have more options than sockets and tcpip. Problem is, we don't have much motivation other than runtime to do it though :/

agocke:
I've been looking at this for a bit -- I think the best way to support this might be a custom test host process. Main problem right now seems to be that this code path isn't very well tested. In fact, I don't think it's ever worked.

https://github.com/microsoft/vstest/blob/c8f881a518377cb723e41f13dbb7a02d2f2712fb/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L314

That line goes looking for `testhost.dll` if `testHostExeFound` is false (which it always is if `useCustomDotnetHostPath` is true), and `testHostPath` is empty, which it always is as well.

I'll see how far I can get.

Eventual goal is that tests are compiled into an exe, and then the hosting layer itself is also compiled into that exe.

agocke:
Never mind, I misunderstood the docs here. I thought "host" meant a custom exe that could run tests -- instead it looks like it means a dotnet core host. 

I think, ideally, the concept would be generalized to be completely language agnostic, and simply communicate over the RPC framework.

MarcoRossignoli:
I think there're a lot of benefits in moving towards that direction. 
We should discuss a bit more on it.

agocke:
Happy to help out if necessary. Also, I think the biggest advantage is not just in testing for `dotnet/runtime` -- the new version of xunit, 3.0, doesn't support DLL-based tests at all. It uses exes exclusively. That likely means it will not be compatible with vstest until this feature is implemented.

nohwnd:
[...]

PR #3623, opened 2022-05-04
Allow transports other than sockets

nohwnd:
## Description
Add options to allow additional transports other than sockets if we decide to add one.

## Related issue
Related #3595


nohwnd:
Just a draft now, to figure out what needs to be done to even make this customizable.

PR #3643, opened 2022-05-16
Fix logging setup when running in remote

nohwnd:
## Description
In remote scenario we can be passed path surrounded with quotes from parameters, and we should ignore those extra quotes. We should also create the full path for log, because in local run vstest.console handles that for us, and so testhost always has the full path, but in remote run the folder is not there and setting up logs fails, and nothing is logged


src/Microsoft.TestPlatform.CoreUtilities/Tracing/EqtTrace.cs
```diff
@@ -161,6 +161,9 @@ public static bool InitializeVerboseTrace(string customLogFile)
     /// <returns>Trace initialized flag.</returns>
     public static bool InitializeTrace(string customLogFile, PlatformTraceLevel traceLevel)
     {
+        // Remove extra quotes if we get them passed on the parameter,
+        // System.IO.File does not ingore them when checking the file existence.
```
Evangelink, line 165:
```suggestion
        // System.IO.File does not ignore them when checking the file existence.
```

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
```diff
@@ -369,6 +369,12 @@ private static bool EnsureTraceIsInitialized()
                     runnerLogFileName = LogFile;
                 }
 
+                var runnlerLogFileInfo = new FileInfo(runnerLogFileName);
+                if (!Directory.Exists(runnlerLogFileInfo.DirectoryName))
+                {
+                    Directory.CreateDirectory(runnlerLogFileInfo.DirectoryName);
+                }
```
Evangelink, line 376:
```suggestion
                var runnerLogFileInfo = new FileInfo(runnerLogFileName);
                if (!Directory.Exists(runnerLogFileInfo.DirectoryName))
                {
                    Directory.CreateDirectory(runnerLogFileInfo.DirectoryName);
                }
```

Evangelink (approved):

src/Microsoft.TestPlatform.CoreUtilities/Tracing/EqtTrace.cs
nohwnd: fixed

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
nohwnd: fixed

src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/PlatformEqtTrace.cs
```diff
@@ -367,6 +367,12 @@ private static bool EnsureTraceIsInitialized()
                     runnerLogFileName = LogFile;
                 }
 
+                var runnerLogFileInfo = new FileInfo(runnerLogFileName);
+                if (!Directory.Exists(runnerLogFileInfo.DirectoryName))
```
Evangelink, line 371: Nit: You could do `runnerLogFileInfo.Directory.Exists` instead.

Evangelink (approved):
Approved with 1 nit.

PR #3661, opened 2022-05-20
Run tests with Server GC enabled & concurrent GC disabled.

JoshKeegan:
## Description
Maximises throughput over latency, reducing runtime for larger test packs or tests that will create a lot of objects on the heap.

## Related issue
See discussion in #1967 
There was also an older PR for these changes that got closed due to inactivity, that also contains some relevant discussion. See #1987

### DIFF
--- a/src/testhost.arm64/app.config
+++ b/src/testhost.arm64/app.config
@@ -4,6 +4,9 @@
     <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.0" />
   </startup>
   <runtime>
+    <gcServer enabled="true" />
+    <gcConcurrent enabled="false" />
+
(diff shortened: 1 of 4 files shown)


Evangelink (approved):

nohwnd (approved):

MarcoRossignoli:
@nohwnd @Evangelink I think we should revert this one and add an option inside the runsetting for the host mode.

Evangelink:
Hi @cjprecord, the change was actually reverted. We didn't do any runsettings option to enable it so we are back to previous state and so you shouldn't be experiencing any issue if you are using the stable release.

MarcoRossignoli:
@cjprecord the revert will be included inside 17.4.3 release.

PR #3671, opened 2022-05-24
Enable nullable on AttachVS

Evangelink:

### DIFF
--- a/src/AttachVS/AttachVs.cs
+++ b/src/AttachVS/AttachVs.cs
@@ -10,8 +10,6 @@ using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
 using System.Threading;
 
-#nullable disable
-
 namespace Microsoft.TestPlatform.AttachVS;
 
 internal class DebuggerUtility
@@ -26,7 +24,7 @@ internal class DebuggerUtility
                 return false;
             }
             var process = Process.GetProcessById(pid.Value);
-            Trace($"Starting with pid '{pid}({process?.ProcessName})', and vsPid '{vsPid}'");
+            Trace($"Starting with pid '{pid}({process.ProcessName})', and vsPid '{vsPid}'");
             Trace($"Using pid: {pid} to get parent VS.");
             var vs = GetVsFromPid(Process.GetProcessById(vsPid ?? process.Id));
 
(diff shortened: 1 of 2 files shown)


src/AttachVS/AttachVs.cs
```diff
@@ -215,18 +218,18 @@ private static bool IsVsOrNull(Process process)
         return isVs;
     }
 
-    private static bool IsCorrectParent(Process currentProcess, Process parent)
+    private static bool IsCorrectParent(Process currentProcess, Process? parent)
```
Evangelink, line 221: We could also ensure that `parent` is not nullable and handle potential nulls on calling sites. Any preference?

src/AttachVS/AttachVs.cs
```diff
@@ -38,20 +36,25 @@ internal static bool AttachVSToProcess(int? pid, int? vsPid)
...
+                            return null;
+                        }
+                    })
+                    .Where(p => p != null && !p.HasExited)
+                    .OrderBy(p => p!.StartTime)
+                    .ToList();
 
-                var firstVs = processes.FirstOrDefault();
-                Trace($"Found VS {firstVs.Process.Id}");
+                var firstVs = processes.First();
```
Evangelink, line 56: We were using `FirstOrDefault` but then doing member access. So it's unclear to me whether or not we expect to have an empty list.

nohwnd: I will have to look at the logic again. Plan was originally to: pickup the "current" process from env variable, or use the current. Then for that  pickup parent VS, and if not found fallback to first VS started. And if not found then fail.

src/AttachVS/AttachVs.cs
```diff
@@ -215,61 +223,51 @@ private static bool IsVsOrNull(Process process)
...
+                return p;
             }
             catch
             {
-                // throws when parent no longer runs
+                return -1;
             }
         }
-
-        return IsCorrectParent(process, parent) ? parent : null;
```
nohwnd, line 269: Why did this logic have to move into the body of the other method directly? Separating it into a named method seemed clearer.

src/AttachVS/AttachVs.cs
```diff
@@ -215,61 +223,51 @@ private static bool IsVsOrNull(Process process)
...
-            id = p;
-        }
-        catch
-        {
-            id = -1;
+                Trace($"Process {parent.ProcessName} ({parent.Id}) is not a valid parent because it started after the current process.");
+            }
+            catch
+            {
+                // throws when parent no longer runs
```
nohwnd, line 246: This catch is now not as narrow scoped as before, it can throw for more reasons I think.

Evangelink: All the `IsCorrect` was inside a global try/catch and the call to `GetProcessById` was also inside a global try/catch so I have simply merged the 2.

src/AttachVS/AttachVs.cs
Evangelink: Seems more complex to have all the logic split but I will revert.

Separated looks like:

image

Merged:

image

nohwnd (approved):

PR #3674, opened 2022-05-24
Enable nullable for EventLogCollector

Evangelink:


src/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector/EventLogDataCollector.cs
```diff
@@ -149,12 +147,13 @@ internal EventLogDataCollector(IFileHelper fileHelper)
...
     /// <param name="dataCollectionEnvironmentContext">Provides contextual information about the agent environment</param>
+    [MemberNotNull(nameof(_events), nameof(_dataSink), nameof(_logger))]
     public override void Initialize(
-        XmlElement configurationElement,
+        XmlElement? configurationElement,
         DataCollectionEvents events!!,
         DataCollectionSink dataSink!!,
         DataCollectionLogger logger!!,
-        DataCollectionEnvironmentContext dataCollectionEnvironmentContext)
+        DataCollectionEnvironmentContext dataCollectionEnvironmentContext!!)
```
Evangelink, line 156: This is a change of behavior but we weren't doing null check so if `Initialize` is called with null for this arg we would fail with NRE.

src/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector/EventLogDataCollector.cs
```diff
@@ -149,12 +147,13 @@ internal EventLogDataCollector(IFileHelper fileHelper)
     /// Used by the data collector to send warnings, errors, or other messages
     /// </param>
     /// <param name="dataCollectionEnvironmentContext">Provides contextual information about the agent environment</param>
+    [MemberNotNull(nameof(_events), nameof(_dataSink), nameof(_logger))]
     public override void Initialize(
-        XmlElement configurationElement,
+        XmlElement? configurationElement,
```
Evangelink, line 152: Functionally we should accept only non-null but that would be a change of behavior so I decided to allow null.

src/DataCollectors/Microsoft.TestPlatform.Extensions.EventLogCollector/EventLogDataCollector.cs
```diff
@@ -267,6 +266,7 @@ internal string WriteEventLogs(List<EventLogEntry> eventLogEntries, int maxLogEn
         // Write the event log file
         FileTransferInformation fileTransferInformation =
             new(dataCollectionContext, eventLogPath, true, _fileHelper);
+        TPDebug.Assert(_dataSink != null, "Initialize should have been called.");
```
Evangelink, line 269: We cannot use the initialize pattern we were talking about because this method is called from within the `Initialize` method.

MarcoRossignoli (approved):

PR #3676, opened 2022-05-25
Fixed telemetry data sharing exception

cvpoienaru:


src/Microsoft.TestPlatform.CrossPlatEngine/Execution/ExecutionManager.cs
```diff
@@ -64,6 +64,9 @@ protected ExecutionManager(ITestPlatformEventSource testPlatformEventSource, IRe
     /// <param name="pathToAdditionalExtensions"> The path to additional extensions. </param>
     public void Initialize(IEnumerable<string> pathToAdditionalExtensions, ITestMessageEventHandler testMessageEventsHandler)
     {
+        // Clear the request data metrics left over from a potential previous run.
+        _requestData.MetricsCollection.Metrics.Clear();
```
Evangelink, line 68: On discovery side you are null checking `MetricsCollection` and `Metrics`, shall you do it here too?

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -102,7 +102,7 @@ Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestExtensionManager.UseAddi
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.GetDiscoveryManager() -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.TesthostProtocol.IDiscoveryManager
 Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.GetExecutionManager() -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.TesthostProtocol.IExecutionManager
-Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.TestHostManagerFactory(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> void
+Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestHostManagerFactory.TestHostManagerFactory(bool telemetryOptedIn) -> void
```
Evangelink, line 105: Are we ok with this breaking change on a minor release?

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -120,7 +120,7 @@ virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionMana
...
 virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestSessionPool.AddSession(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.ProxyTestSessionManager proxyManager) -> bool
 virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestSessionPool.ReturnProxy(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, int proxyId) -> bool
-virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestSessionPool.TryTakeProxy(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, string source, string runSettings) -> Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager
+virtual Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.TestSessionPool.TryTakeProxy(Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.TestSessionInfo testSessionInfo, string source, string runSettings, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.IRequestData requestData) -> Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager
```
Evangelink, line 123: Same question about breaking change here.

src/testhost.x86/DefaultEngineInvoker.cs
```diff
@@ -155,29 +155,15 @@ public void Invoke(IDictionary<string, string?> argsDictionary)
...
-        {
-            MetricsCollection =
-                telemetryOptedIn
-                    ? new MetricsCollection()
-                    : new NoOpMetricsCollection(),
-            IsTelemetryOptedIn = telemetryOptedIn
-        };
-        return requestData;
+        return !string.IsNullOrWhiteSpace(telemetryStatus)
+            && telemetryStatus.Equals("true", StringComparison.Ordinal);
```
Evangelink, line 166: You can simplify by doing simple string check with either the const on the left side or by using `string.Equals`.

```suggestion
        return string.Equals("true", telemetryStatus, StringComparison.Ordinal);
```

Evangelink (commented):
Do we have any test ensuring we won't regress on the cleaning of the metrics in both `Initialize` methods?

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/ExecutionManager.cs
cvpoienaru: Fixed, thanks !

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
cvpoienaru: I would say so. No one other than us uses this type so the risks are minimal here.

Unfortunately there's no good way of fixing this without an API change since the request data is bound to a testhost during its entire lifespan due to a design choice that didn't take into account the possibility of a testhost living past its expected lifetime (i.e. fulfilling a test discovery/run request).

src/Microsoft.TestPlatform.CrossPlatEngine/PublicAPI/PublicAPI.Shipped.txt
cvpoienaru: We should be ok with this breaking change too. The only consumer of `TestSessionPool.TryTakeProxy` is `vstest.console` and its use is buried quite deep. The risks should be minimal here too.

src/testhost.x86/DefaultEngineInvoker.cs
cvpoienaru: Done, thanks for the suggestion !

cvpoienaru:
> Do we have any test ensuring we won't regress on the cleaning of the metrics in both `Initialize` methods?

Sorry, missed this comment. We don't have any such tests yet but I am working on adding them.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/TestSessionPool.cs
```diff
@@ -147,7 +149,13 @@ public virtual bool KillSession(TestSessionInfo testSessionInfo, IRequestData re
         try
         {
             // Deque an actual proxy to do work.
-            return sessionManager.DequeueProxy(source, runSettings);
+            var proxy = sessionManager.DequeueProxy(source, runSettings);
+
+            // Make sure we use the per-request request data instead of the request data used when
+            // creating the test session.
```
nohwnd, line 155: Please make this sure why need to make sure we use them here.

nohwnd, line 155: Please make this sure why need to make sure we use them here.

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/ExecutionManager.cs
```diff
@@ -40,7 +40,7 @@ public class ExecutionManager : IExecutionManager
     /// <summary>
     /// Initializes a new instance of the <see cref="ExecutionManager"/> class.
     /// </summary>
-    public ExecutionManager(IRequestData requestData) : this(TestPlatformEventSource.Instance, requestData)
+    public ExecutionManager(IRequestData requestData!!) : this(TestPlatformEventSource.Instance, requestData)
```
Evangelink, line 43: The protected ctor doesn't throw on null `requestData`, that seems weird to require non-null for public but allow null for ones inheriting.

src/Microsoft.TestPlatform.CrossPlatEngine/TestSession/TestSessionPool.cs
cvpoienaru: Done, thanks !

src/Microsoft.TestPlatform.CrossPlatEngine/Execution/ExecutionManager.cs
cvpoienaru: That's a nice catch, thanks. I will make the protected one require non-null as well but I will let the public one like this. I know it's not strictly necessary since the constructors are chained but I want to future proof in case we'll ever remove the chaining.

nohwnd (approved):

Issue #3647, opened 2022-05-18
Improve performance of parallel discovery


shyamnamboodiripad:
Currently sequential discovery of tests with MaxCpuCount = 1 is faster than parallel discovery (MaxCpuCount = 0) for test assemblies that only contain a few (100) tests. This is because -

1. Sequential discovery reuses the same testhost process to discover all supplied test sources. Parallel discovery on the other hand creates a new testhost process for every supplied test source. The overhead of process startup, extension loading etc. is significantly greater than the time required to discover a few (100) tests in this case. Sharing / reusing each teshost processes to discover more than one test source over the lifetime of the discovery operation should improve performance significantly.

1. The testshost processes for parallel discovery are started on a staggered basis and the actual number of testhosts running at any point is always (significantly) smaller than the specified MaxCpuCount. Starting the max allowed number of testhost processes up front should help to better amortize the startup time overhead for individual processes.

shyamnamboodiripad:
FYI @Evangelink @nohwnd Logging this so that I have something I can link to in my AzDO side changes.

Evangelink:
Is it worth to consider having vstest process having a higher priority compared to testhosts so that we ensure it's less of a bottleneck?

shyamnamboodiripad:
> This is only true for .NET Framework (we still can't repro what you said you saw 😁)

My bad @nohwnd. I went back and confirmed that the solution I was trying against (MSTest_100K) was indeed targeting .NET Framework 😢 I was probably confused because the projects in question were new SDK-style and incorrectly assumed they were also targeting .NET Core. Apologies for the confusion.

Its a pity .NET Framework and .NET Core differ in behavior here. Does .NET Framework also isolate each source being discovered in its own AppDomain?

> Is it worth to consider having vstest process having a higher priority compared to testhosts so that we ensure it's less of a bottleneck?

@Evangelink as we discussed before I agree it would be great to use low priority by default for testhosts. MSBuild has a [similar setting](https://stackoverflow.com/questions/63214007/how-can-i-make-msbuild-use-less-cpu-resources) for build. All that does is to [set the process priority to `BelowNormal`](
https://github.com/dotnet/msbuild/blob/ac305ec92b09385c3e1f656ba47b8389468b4299/src/MSBuild/XMake.cs#L2093).
```
Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.BelowNormal;
```

[...]

PR #3682, opened 2022-05-26
Allow sharing testhosts for non-paralell runs on a Shared testhost

nohwnd:
When runtime provider indicates that it is Shared (.NET Framework testhost), batch up the sources that are provided and 
run the on the same testhost process.

For any non-shared, or any parallel run, prefer splitting the workloads per source and putting them on separate processes. 

This is in-line with how TP 16.6.1 works for non-paralell discovery and execution and ensures that big runs on shared host are able to share the testhost for all dlls.

## Related issue
Fixes part of #3647


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -182,21 +185,45 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
+                    var workload = new ProviderSpecificWorkload<DiscoveryCriteria>(updatedCriteria, testhostProviderInfo);
+                    workloads.Add(workload);
+                }
+            }
         }
 
         return workloads;
 
-        static DiscoveryCriteria NewDiscoveryCriteriaFromSourceAndSettings(string source, DiscoveryCriteria discoveryCriteria, string runsettingsXml)
+        static DiscoveryCriteria NewDiscoveryCriteriaFromSourceAndSettings(string[] sources, DiscoveryCriteria discoveryCriteria, string runsettingsXml)
```
Evangelink, line 223: This could be `IEnumerable<string>` instead of array (this is the "level" expected by `DiscoveryCriteria`.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -182,21 +185,45 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
+                var runsettings = testhostProviderInfo.RunSettings;
+                var sourcesToDiscover = group.Select(w => w.Work).ToArray();
+                var updatedCriteria = NewDiscoveryCriteriaFromSourceAndSettings(sourcesToDiscover, discoveryCriteria, runsettings);
+                var workload = new ProviderSpecificWorkload<DiscoveryCriteria>(updatedCriteria, testhostProviderInfo);
+                workloads.Add(workload);
+            }
+            else
+            {
+                // Create one workload for each source
+                foreach (var w in group.ToList())
```
Evangelink, line 210: I don't think the `ToList` is adding much value and forces a double iteration.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -182,21 +185,45 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
-            var testHostProviderInfo = sourceToTestHostProviderMap[source];
-            var runsettingsXml = testHostProviderInfo.RunSettings;
-            var updatedDiscoveryCriteria = new ProviderSpecificWorkload<DiscoveryCriteria>(NewDiscoveryCriteriaFromSourceAndSettings(source, discoveryCriteria, runsettingsXml), testHostProviderInfo);
-            workloads.Add(updatedDiscoveryCriteria);
+            var testhostProviderInfo = group.Key;
+            // If the run is not parallel and the host is shared, put all testcases on single testhost.
+            if (!_isParallel && testhostProviderInfo.Shared)
+            {
+                var runsettings = testhostProviderInfo.RunSettings;
+                var sourcesToDiscover = group.Select(w => w.Work).ToArray();
```
Evangelink, line 202: If you change `NewDiscoveryCriteriaFromSourceAndSettings` to take `IEnumerable<string>` you won't need the `ToArray` call here.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -182,21 +185,45 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
         List<ProviderSpecificWorkload<DiscoveryCriteria>> workloads = new();
-        foreach (var source in discoveryCriteria.Sources)
+        foreach (var group in providerGroups)
         {
-            var testHostProviderInfo = sourceToTestHostProviderMap[source];
-            var runsettingsXml = testHostProviderInfo.RunSettings;
-            var updatedDiscoveryCriteria = new ProviderSpecificWorkload<DiscoveryCriteria>(NewDiscoveryCriteriaFromSourceAndSettings(source, discoveryCriteria, runsettingsXml), testHostProviderInfo);
-            workloads.Add(updatedDiscoveryCriteria);
+            var testhostProviderInfo = group.Key;
+            // If the run is not parallel and the host is shared, put all testcases on single testhost.
```
Evangelink, line 198: Might be worth to explain briefly why we don't think it adds value to share for parallel.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -182,21 +185,45 @@ public bool HandlePartialDiscoveryComplete(IProxyDiscoveryManager proxyDiscovery
...
+            else
+            {
+                // Create one workload for each source
+                foreach (var w in group.ToList())
+                {
+                    var runsettings = testhostProviderInfo.RunSettings;
+                    var sourcesToDiscover = new[] { w.Work };
+                    var updatedCriteria = NewDiscoveryCriteriaFromSourceAndSettings(sourcesToDiscover, discoveryCriteria, runsettings);
+                    var workload = new ProviderSpecificWorkload<DiscoveryCriteria>(updatedCriteria, testhostProviderInfo);
+                    workloads.Add(workload);
```
Evangelink, line 216: Outside for the sources to discover the logic of the loop is exactly the same as in previous if. I think it might be worth to update the local static function to either handle `ProviderSpecificWorkload` creation and locally do `workloads.Add(workload)` or also do this from the function.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -210,36 +212,81 @@ private List<ProviderSpecificWorkload<TestRunCriteria>> SplitToWorkloads(TestRun
...
             {
-                var testHostProviderInfo = sourceToTestHostProviderMap[group.Key];
-                var runsettings = testHostProviderInfo.RunSettings;
-                // ToList because it is easier to see what is going on when debugging.
-                var testCases = group.ToList();
-                var updatedCriteria = CreateTestRunCriteriaFromTestCasesAndSettings(testCases, testRunCriteria, runsettings);
-                var workload = new ProviderSpecificWorkload<TestRunCriteria>(updatedCriteria, testHostProviderInfo);
-                workloads.Add(workload);
+                var testhostProviderInfo = group.Key;
+                // If the run is not parallel and the host is shared, put all testcases on single testhost.
```
Evangelink, line 235: Most of my remarks for discovery can also be applied here.

Evangelink (commented):
Could we try to add tests to avoid regressions around the new logic?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: But I want to see them when I debug. not somewhere x layers deeper, just to avoid calling toArray on collection of <10 items. Is there any perf improvement you can show me? :)

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
```diff
@@ -210,36 +212,81 @@ private List<ProviderSpecificWorkload<TestRunCriteria>> SplitToWorkloads(TestRun
         // is shared we try to find a workload that is appropriate for it. If we don't find any work that the
         // running testhost can do. Or if the testhost already exited (possibly because of crash), we start another one
         // and give it the next workload.
+
+
+        // List all the sources (dlls) we have and group them by their provider, so we can run either multiple sources
+        // on a single instance of the appropriate testhost (for non-paralell, shared),
+        // or each source in its own instance of the provider (for non-parallel, non-shared, and both parallel shared, and parallel non-shared).
```
nohwnd, line 219:
```suggestion
        // List all the sources (dlls) we have and group them by their provider, so we can run either multiple sources
        // on a single instance of the appropriate testhost (for non-paralell, shared),
        // or each source in its own instance of the provider (for non-parallel, non-shared, and both parallel shared, and parallel non-shared).
        //
        // We prefer running more work in parallel when we can, because doing work on 10 testhosts in
        // parallel is faster than doing the same work on 1 testhost. Even when the sources to discover
        // contain just a few tests.
```

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: Or teach me :)

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
Evangelink: You can call `ToList`/`ToArray` or even simply expand when debugging. I don't have any perf but the global reasoning of "it's only adding a small amount" everywhere in the code might cause problems in the long run.

Your call to fix or close :)

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: done

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: Changed it anyway. If that makes you happy :D

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: removed.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
nohwnd: I end up grouping the sources into a collection of arrays. 

It re-uses the logic because we then run on collection of string arrays. It it still not very elagant, and you will hate me for that type. But it is clearer than using IEnumerable<IEnumerable<string>> 😁

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
nohwnd: added.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyExecutionManager.cs
nohwnd: also applied here.

nohwnd:
Tests are in place, they just need to be reverted to reflect that we reverted to this new-old behavior. We will see them fail here imho. If not it will be very strange.

nohwnd:
@shyamnamboodiripad this is the regression fix.

Evangelink (approved):

PR #3684, opened 2022-05-26
Perform file roll only occasionally when file is locked

### LINKED ISSUE #3413: Rolling file trace listener slows down significantly when the file is in use
## Description

When file is in use and we reach the size limit of the file (e.g. 20 MB), then SafeMove will try to move the file (2 times) on every write to the log, and will fail to do it because the file is in use. This slows the execution significantly because the filesystem is touched and 2 exceptions are thrown for ever write to the log.

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/RollingFileTraceListener.cs#L283

Run with diagnostic enabled. Have a 20MB log file, open it in notepad or some other process that keeps it in use. In SafeMove you should see the first call to Move fail, and the second call in finally also fail.

## Expected behavior

Attempting to move the file should be given up, or use exponentiona back-off.

## Actual behavior

Execution becomes really slow.

## Diagnostic logs

They don't matter the operation does not log into log.

nohwnd:
## Description
When the log file becomes bigger than the roll limit and is opened in editor it becomes locked, and we try to move it twice on every message. This avoids rolling it for 1 minute since we last tried and failed.

## Related issue
Fix #3413

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/RollingFileTraceListener.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/common/Tracing/RollingFileTraceListener.cs
@@ -128,6 +128,9 @@ public class RollingFileTraceListener : TextWriterTraceListener
         /// this listener.
         /// </summary>
         private TallyKeepingFileStreamWriter _managedWriter;
+        private bool _lastRollFailed;
+        private readonly Stopwatch _sinceLastRoll = Stopwatch.StartNew();
+        private readonly long _failedRollTimeout = TimeSpan.FromMinutes(1).Ticks;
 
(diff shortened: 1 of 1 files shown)


Evangelink (commented):
What happens to the traces performed in between the rolling?

Could we use something mockable instead of `Stopwatch` and add some tests for this behavior?

nohwnd:
Nothing happens to them, they will continue being written into the file, same as they did before. We just don't attempt to move the file on every line we write, and bring performance to halt.

Feel free to be the other person in your "we" and add as many unit tests as you want 😁

Evangelink (approved):

PR #3688, opened 2022-05-30
Enable nullable on adapter utilities

Evangelink:

### DIFF
--- a/src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.MethodBase.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.MethodBase.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Reflection;
 
-#nullable disable
-
 namespace Microsoft.TestPlatform.AdapterUtilities.Helpers;
 
 internal static partial class ReflectionHelpers
@@ -33,7 +31,7 @@ internal static partial class ReflectionHelpers
 #endif
     }
 
-    internal static Type GetReflectedType(MethodBase method)
+    internal static Type? GetReflectedType(MethodBase method)
     {
 #if !NETSTANDARD1_0 && !NETSTANDARD1_3 && !WINDOWS_UWP
         return method.ReflectedType;
--- a/src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.Type.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.Type.cs
@@ -4,8 +4,6 @@
 using System;
 using System.Reflection;
 
-#nullable disable
-
 namespace Microsoft.TestPlatform.AdapterUtilities.Helpers;
 
 internal static partial class ReflectionHelpers
@@ -19,7 +17,7 @@ internal static partial class ReflectionHelpers
 #endif
     }
 
-    internal static MethodBase GetDeclaringMethod(Type type)
+    internal static MethodBase? GetDeclaringMethod(Type type)
     {
 #if !NETSTANDARD1_0 && !NETSTANDARD1_3 && !WINDOWS_UWP
         return type.DeclaringMethod;
--- a/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
@@ -7,8 +7,6 @@ using System;
 using System.Runtime.Serialization;
 #endif
 
-#nullable disable
-
 namespace Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities;
 
 #if !NETSTANDARD1_0 && !WINDOWS_UWP
@@ -20,7 +18,7 @@ public class InvalidManagedNameException :
     , ISerializable
 #endif
 {
-    public InvalidManagedNameException(string message) : base(message) { }
+    public InvalidManagedNameException(string? message) : base(message) { }
 
 #if !NETSTANDARD1_0 && !WINDOWS_UWP
     protected InvalidManagedNameException(SerializationInfo info, StreamingContext context) : base(info, context) { }
--- a/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
@@ -18,7 +18,7 @@ public class InvalidManagedNameException :
     , ISerializable
 #endif
 {
-    public InvalidManagedNameException(string? message) : base(message) { }
+    public InvalidManagedNameException(string message) : base(message) { }
 
 #if !NETSTANDARD1_0 && !WINDOWS_UWP
     protected InvalidManagedNameException(SerializationInfo info, StreamingContext context) : base(info, context) { }
--- a/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
@@ -80,15 +80,19 @@ public static partial class ManagedNameHelper
     /// More information about <paramref name="managedTypeName"/> and <paramref name="managedMethodName"/> can be found in
     /// <see href="https://github.com/microsoft/vstest-docs/blob/main/RFCs/0017-Managed-TestCase-Properties.md">the RFC</see>.
     /// </remarks>
-    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[]? hierarchyValues)
+    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[] hierarchyValues)
     {
         if (!ReflectionHelpers.IsMethod(method))
         {
+            // TODO: @Haplois, exception expects a message and not a param name.
             throw new NotSupportedException(nameof(method));
         }
 
-        var semanticType = ReflectionHelpers.GetReflectedType(method);
-        if (semanticType is not null && ReflectionHelpers.IsGenericType(semanticType))
+        var semanticType = ReflectionHelpers.GetReflectedType(method)
+            // TODO: @Haplois, exception expects a message and not a param name.
+            ?? throw new NotSupportedException(nameof(method));
+
+        if (ReflectionHelpers.IsGenericType(semanticType))
         {
             // The type might have some of its generic parameters specified, so make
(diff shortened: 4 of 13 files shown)


src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.MethodBase.cs
```diff
@@ -33,7 +31,7 @@ internal static bool IsMethod(MethodBase method)
 #endif
     }
 
-    internal static Type GetReflectedType(MethodBase method)
+    internal static Type? GetReflectedType(MethodBase method)
```
Evangelink, line 34: Because we return `as Type`.

src/Microsoft.TestPlatform.AdapterUtilities/Helpers/ReflectionHelpers.Type.cs
```diff
@@ -19,7 +17,7 @@ internal static bool IsGenericType(Type type)
 #endif
     }
 
-    internal static MethodBase GetDeclaringMethod(Type type)
+    internal static MethodBase? GetDeclaringMethod(Type type)
```
Evangelink, line 20: Property `DeclaringMethod` on `Type` is nullable.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -142,10 +140,13 @@ public static void GetManagedName(MethodBase method!!, out string managedTypeNam
 
         managedTypeName = typeBuilder.ToString();
         managedMethodName = methodBuilder.ToString();
-        hierarchyValues = new[] {
+        hierarchyValues = hierarchyPos != null
```
Evangelink, line 143: It's not clear what we want to do when `hierarchyPos` is null so I went with returning null for `hierarchyValues` but maybe we could throw?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -260,16 +256,17 @@ bool Filter(MemberInfo mbr, object param)
...
+        return (MethodInfo?)methods.SingleOrDefault();
     }
 
-    private static int[] AppendTypeString(StringBuilder b, Type type, bool closedType)
+    private static int[]? AppendTypeString(StringBuilder b, Type? type, bool closedType)
     {
-        int[] hierarchies = null;
+        int[]? hierarchies = null;
+
+        if (type is null)
```
Evangelink, line 266: Not clear what to do when `type` is null.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
```diff
@@ -20,7 +18,7 @@ public class InvalidManagedNameException :
     , ISerializable
 #endif
 {
-    public InvalidManagedNameException(string message) : base(message) { }
+    public InvalidManagedNameException(string? message) : base(message) { }
```
Haplois, line 21: I think the `message` is never null, and it shouldn't be. If we are throwing this, we should report to user what's wrong.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Haplois: We should return `null` in that case. This method might be called recursively until we reach the end of type tree. We stop when it's null.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Haplois: No, that can never happen. It always has a value. `AppendTypeString` only returns null if provided type is null, here it's not.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -82,15 +80,15 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
...
     /// </remarks>
-    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[] hierarchyValues)
+    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[]? hierarchyValues)
     {
         if (!ReflectionHelpers.IsMethod(method))
         {
             throw new NotSupportedException(nameof(method));
         }
 
         var semanticType = ReflectionHelpers.GetReflectedType(method);
```
Haplois, line 90: This shouldn't ever be `null`. We can throw if it is.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameParser.cs
```diff
@@ -66,7 +64,7 @@ public static void ParseManagedTypeName(string managedTypeName, out string names
     /// <exception cref="InvalidManagedNameException">
     /// Thrown if <paramref name="managedMethodName"/> contains spaces, incomplete, or the arity isn't numeric.
     /// </exception>
-    public static void ParseManagedMethodName(string managedMethodName, out string methodName, out int arity, out string[] parameterTypes)
+    public static void ParseManagedMethodName(string managedMethodName, out string methodName, out int arity, out string[]? parameterTypes)
```
Haplois, line 67: `parameterTypes` can't ever be `null`, if it doesn't exactly have the elements we expect we throw.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -82,15 +80,15 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
     /// More information about <paramref name="managedTypeName"/> and <paramref name="managedMethodName"/> can be found in
     /// <see href="https://github.com/microsoft/vstest-docs/blob/main/RFCs/0017-Managed-TestCase-Properties.md">the RFC</see>.
     /// </remarks>
-    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[] hierarchyValues)
+    public static void GetManagedName(MethodBase method!!, out string managedTypeName, out string managedMethodName, out string[]? hierarchyValues)
```
Haplois, line 83: This can't be null, we throw if don't return as per the RFC.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -502,62 +509,4 @@ private static string GetTypeString(Type type, bool closedType)
         AppendTypeString(builder, type, closedType);
         return builder.ToString();
     }
-
-#if NET20
```
Haplois, line 506: This library by-design supports older frameworks too, we want everyone to be able to use it. Please don't remove this support. I know we don't target these, but if people copy-and-paste this code to their projects, we want it to work. (This point can be discussed though, I don't know if we want to do that :))

src/Microsoft.TestPlatform.AdapterUtilities/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -1,27 +1,28 @@
...
-static Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.ManagedNameHelper.GetMethod(System.Reflection.Assembly assembly, string managedTypeName, string managedMethodName) -> System.Reflection.MethodBase
-static Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.ManagedNameParser.ParseManagedMethodName(string managedMethodName, out string methodName, out int arity, out string[] parameterTypes) -> void
-static Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.ManagedNameParser.ParseManagedTypeName(string managedTypeName, out string namespaceName, out string typeName) -> void
\ No newline at end of file
+static Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.ManagedNameHelper.GetManagedName(System.Reflection.MethodBase! method, out string! managedTypeName, out string! managedMethodName) -> void
+static Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.ManagedNameHelper.GetManagedName(System.Reflection.MethodBase! method, out string! managedTypeName, out string! managedMethodName, out string![]? hierarchyValues) -> void
```
Haplois, line 25: this will always be an array of string with a length of `Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.TotalLevelCount`. It can't be null; one or more items can be.

Haplois (changes requested):
Please see my comments about nullability.

src/Microsoft.TestPlatform.AdapterUtilities/Microsoft.TestPlatform.AdapterUtilities.csproj
```diff
@@ -5,7 +5,7 @@
   <Import Project="$(TestPlatformRoot)scripts/build/TestPlatform.Settings.targets" />
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;net45;net6.0</TargetFrameworks>
```
Evangelink, line 8:
```suggestion
    <TargetFrameworks>netstandard1.0;netstandard2.0;net45</TargetFrameworks>
```

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/InvalidManagedNameException.cs
Evangelink: All basic .net exceptions allow null message that's why I am allowing null too.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Evangelink: Minimal version supported by MS is net462 so it should be safe to remove code targeting older versions.

Haplois (approved):

nohwnd (approved):

Evangelink:
TODOs will be addressed by @Haplois on a different PR he is currently working on.

PR #3693, opened 2022-05-30
Added telemetry for in-process runs

cvpoienaru:
Added telemetry for in-process runs

### DIFF
--- a/src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
+++ b/src/Microsoft.TestPlatform.Common/Telemetry/TelemetryDataConstants.cs
@@ -13,6 +13,8 @@ internal static class TelemetryDataConstants
     // ******************** General ***********************
     public static readonly string DiscoveredExtensions = "VS.TestPlatform.DiscoveredExtensions";
 
+    public static readonly string InProcessRequest = "VS.TestPlatform.InProcessRequest";
+
     // ******************** Execution ***********************
     public static readonly string ParallelEnabledDuringExecution = "VS.TestRun.ParallelEnabled";
 
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
@@ -99,6 +99,9 @@ public class TestEngine : ITestEngine
             }
             var testHostManagerInfo = testHostManagers[0];
 
+            // Collect telemetry for in-process runs.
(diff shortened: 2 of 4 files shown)


nohwnd:
What is this trying to achieve? We collect telemetry only under VS, and there we never run in-process, because we are in design mode.

cvpoienaru:
> What is this trying to achieve? We collect telemetry only under VS, and there we never run in-process, because we are in design mode.

You're right in that we don't collect telemetry for VS scenarios (or even for dotnet ones since there's an explicit check for that too), but there might be other places where `vstest.console` is run as standalone (i.e. not in server mode) and could be able to report telemetry back provided such a mechanism exists. The main scenario I have in mind (but not limited to) is AzDo, but in order to see if/how common is that use case we have to have such telemetry in place.

The change overall is benign but can uncover use cases that we are not aware of right now.

nohwnd:
Can you rather inspect the code of the consumer than adding a datapoint and hoping to see it somewhere? It also removes the check for telemetry opted in. Which is pretty serious problem if we collect telemetry where the user did not opt in.

nohwnd (changes requested):
Always check if telemetry was opted in.

cvpoienaru:
> Can you rather inspect the code of the consumer than adding a datapoint and hoping to see it somewhere? It also removes the check for telemetry opted in. Which is pretty serious problem if we collect telemetry where the user did not opt in.

I agree that telemetry collection without explicit consent is a very serious problem, however that's not the case here because of the way the `RequestData` is built. In case telemetry is not opted in, we default to a `NoOpMetricsCollection` that basically does a no-op on all telemetry `Add`s, basically ignoring them. I find this approach much better than cluttering the code with `if`s to check if telemetry is opted in and that's actually the case today (i.e. this was the only place in code where we explicitly checked if telemetry was opted in, every other telemetry call exploits the mechanism described below).

```
private IRequestData GetRequestData(ProtocolConfig protocolConfig)
{
    return new RequestData
    {
        ProtocolConfig = protocolConfig,
        MetricsCollection =
            _telemetryOptedIn || IsTelemetryOptedIn()
                ? (IMetricsCollection)new MetricsCollection()
                : new NoOpMetricsCollection(),
        IsTelemetryOptedIn = _telemetryOptedIn || IsTelemetryOptedIn()
    };
}
```

nohwnd:
What about the extra work that is done just to be thrown away, can that be avoided be returning early?

MarcoRossignoli:
Going to close this one for stale state, feel free to reopen if want to complete.

PR #3696, opened 2022-05-30
Keep handler versions on edge

nohwnd:
## Description
Replace the public interfaces we re-use in way too many places with internal interfaces that we can improve, and then on edges translate data. 

- [ ] Discovery
- [ ] Replace all primitive messages with messages that carry objects, so they are expandable.

## Related issue


nohwnd:
This can be abandoned; I did most of the changes in #3701

PR #3700, opened 2022-05-31
Enable nullable on TP utilities

Evangelink:


src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -584,7 +584,7 @@ private static void ValidateRunConfiguration(XPathNavigator runSettingsNavigator
         XmlUtilities.AppendOrModifyChild(xmlDocument, TargetFrameworkNodePath, TargetFrameworkNodeName, effectiveFramework.ToString());
     }
 
-    public static bool TryGetPlatformXml(XPathNavigator runSettingsNavigator, out string platformXml)
+    public static bool TryGetPlatformXml(XPathNavigator runSettingsNavigator, out string? platformXml)
```
Evangelink, line 587: If platform is null we return true, that seems weird to me. Usually in TryXXX methods you return true when you have something valid.

nohwnd (approved):

PR #3703, opened 2022-05-31
Enable nullables on remaining tests

Evangelink:


Evangelink:
Dropped and recreated as smaller chunk because the CI error makes no sense.

PR #3704, opened 2022-05-31
Fixed bugs in ManagedMethod parsing, and updated hierarchies.

### LINKED ISSUE #2926: Exception: Sequence contains more than one element
## Description

In version 2.2.3 there is a regression (compared to 2.1.2) for this code:

```c#
[TestClass]
public sealed class MyTestClass : MyTestClassBase
{
    [TestMethod]
    [TestCategory("MyTests")]
    public new void TestSomething()
    {
        base.TestSomething();
    }
}

public abstract class MyTestClassBase
{
    protected void TestSomething()
    {
        // Do something
    }
}
```

Executing this via:

```
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" "MyTests.dll" /TestCaseFilter:"TestCategory=MyTests"
```

Leads to this error (I had to manually translate it because the error is in German on my machine):

```
Exception in call of Executor "executor://mstestadapter/v2": The sequence contains more than one element.
```

This does not happen in version 2.1.2.

## Steps to reproduce

1. Create a modern-SDK test project
1. Select `net48` as target framework
1. Add code above
1. Reference NuGet packages like this:

```
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.1" />
    <PackageReference Include="MSTest.TestAdapter" Version="2.2.3" />
    <PackageReference Include="MSTest.TestFramework" Version="2.2.3" />
```

Run the tests via `vstest.console` (like shown above).

## Expected behavior

There should be no exception.

## Actual behavior

The exception above is thrown.

## Environment

* Windows 10
* Visual Studio 2019 16.7.12
* vstest.console 16.7.0

[AB#1335565](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1335565)

Haplois:
Our FQN API is finalized, this PR makes it compliant with Visual Studio.

 - [x] Hierarchy order reversed to make it easily expandable. Currently, we only support 4 levels.
 - [x] Hierarchy API revised to return all 4 levels which are: `Test group`, `class name`, `namespace`, and `container`.
 - [x] `Test group` by default is managed method name without adapters.
 - [x] Fixes #2926

Hierarchies are for display purposes; we need to change how we generate those:
 - [x] Switch to using closed types instead of open. `TestMethod<string>` instead of ```TestMethod`1```.
 - [x] Append method parameters to `Test group` if any, do not append `()` if not.
 - [x] `Namespace` can be `string.Empty` in case of no namespace. Make sure that works as expected, and we're able to resolve when provided.

# Breaking changes
 - Hierarchy generation changed, RFC to be updated to reflect this. (https://github.com/microsoft/vstest-docs/pull/261)
 - Although this breaks compatibility with older versions of `Hierarchy` since those versions were officially released and never used by anyone - support will start with the first version we release. It will be bound to TP version, and adapter version. Check is done inside Test Explorer.


CODEOWNERS
```diff
@@ -3,6 +3,11 @@
 PublicAPI.Shipped.txt @nohwnd @MarcoRossignoli
 PublicAPI.Unshipped.txt @nohwnd @MarcoRossignoli
 
+# Changes here might break our contracts with other adapters, and possibly
+# Visual Studio. 
+/src/Microsoft.TestPlatform.AdapterUtilities/ @haplois
```
Evangelink, line 8: Shall we have a backup in case you are not around @Haplois?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -119,11 +118,11 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
         var methodBuilder = new StringBuilder();
 
         // Namespace and Type Name (with arity designation)
+        // hierarchyPos contains [startIndexOfNamespace, endIndexOfNameSpace, endIndexOfTypeName]
         var hierarchyPos = AppendTypeString(typeBuilder, semanticType, closedType: false);
-        if (hierarchyPos is null || hierarchyPos.Length != HierarchyConstants.Levels.TotalLevelCount)
+        if (hierarchyPos is null || hierarchyPos.Length != 3)
```
Evangelink, line 123: Could we avoid magic numbers? At least add some comment to explain how you get this number.

nohwnd:
Can you add more info to the PR description please?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -265,7 +266,17 @@ bool Filter(MemberInfo mbr, object? param)
...
+        if (methods.Length == 1)
+        {
+            method = (MethodInfo?)methods[0];
+        }
+        else if (methods.Length > 1)
+        {
+            method = (MethodInfo?)methods.SingleOrDefault(i => i.DeclaringType == type);
+        }
+
+        return method;
```
Evangelink, line 279: Nit: Switch pattern would be easier to read. I am suggesting something without having tested it so there might be some small fixes to do:

```suggestion
        return methods.Length switch        
        {
            1 => (MethodInfo?)methods[0],
            > 1 =>(MethodInfo?)methods.SingleOrDefault(i => i.DeclaringType == type),
            _ => null,
        };
```

src/Microsoft.TestPlatform.AdapterUtilities/TestIdProvider.cs
```diff
@@ -31,8 +31,27 @@ public void AppendString(string str)
         {
             throw new InvalidOperationException();
         }
+        if (str == null)
```
Evangelink, line 34: Method signature should be updated to allow null: `string? str`.

src/Microsoft.TestPlatform.AdapterUtilities/HierarchyConstants.cs
```diff
@@ -26,16 +26,42 @@ public static class Levels
         /// <summary>
         /// Total length of Hierarchy array.
         /// </summary>
-        public const int TotalLevelCount = 2;
+        /// <remarks>
+        /// Currently the order for this is usually:
+        /// <c>Managed Method Name</c>, <c>ClassName</c>, <c>Namespace</c>, <c>Assembly Display Name</c>.
+        /// </remarks>
+        public const int TotalLevelCount = 4;
```
Evangelink, line 33: Is it ok to merge these breaking changes now?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -86,8 +86,7 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
 
         if (!ReflectionHelpers.IsMethod(method))
         {
-            // TODO: @Haplois, exception expects a message and not a param name.
-            throw new NotSupportedException(nameof(method));
+            throw new ArgumentException(Resources.Resources.ErrorMethodExpectedAsAnArgument, nameof(method));
```
Evangelink, line 89: Could we add test for this case?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Haplois: See line 121 :)

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -119,11 +118,11 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
...
 
         // Namespace and Type Name (with arity designation)
+        // hierarchyPos contains [startIndexOfNamespace, endIndexOfNameSpace, endIndexOfTypeName]
         var hierarchyPos = AppendTypeString(typeBuilder, semanticType, closedType: false);
-        if (hierarchyPos is null || hierarchyPos.Length != HierarchyConstants.Levels.TotalLevelCount)
+        if (hierarchyPos is null || hierarchyPos.Length != 3)
         {
-            // TODO: @Haplois, exception expects a message and not a param name.
-            throw new NotSupportedException(nameof(method));
+            throw new ArgumentException(Resources.Resources.ErrorMethodExpectedAsAnArgument, nameof(method));
```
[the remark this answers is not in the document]
Haplois: Nope, I don't know how to create that state, as long as we get a method that will never happen. If we don't get a method, then we err out earlier.

src/Microsoft.TestPlatform.AdapterUtilities/HierarchyConstants.cs
Haplois: Yes, our contract with Visual Studio and adapters is changing.

src/Microsoft.TestPlatform.AdapterUtilities/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -1,9 +1,11 @@
 #nullable enable
 const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.HierarchyLabel = "Hierarchy" -> string!
 const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.HierarchyPropertyId = "TestCase.Hierarchy" -> string!
-const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.ClassIndex = 1 -> int
-const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.NamespaceIndex = 0 -> int
-const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.TotalLevelCount = 2 -> int
+const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.ContainerIndex = 0 -> int
+const Microsoft.TestPlatform.AdapterUtilities.HierarchyConstants.Levels.NamespaceIndex = 1 -> int
```
nohwnd, line 5: So how does this work when adapter utilities on testhost side and on IDE side are different? How do we keep compatibility with MSTest that is currently using this, like 2.2.9?

src/Microsoft.TestPlatform.AdapterUtilities/Resources/Resources.Designer.cs
```diff
@@ -1,157 +1,163 @@
...
-        {
-            get
-            {
+        internal static string ErrorTypeNotFound {
+            get {
                 return ResourceManager.GetString("ErrorTypeNotFound", resourceCulture);
             }
         }
-
+
```
nohwnd, line 135: Can the generator no emit these whitespace empty lines?

nohwnd (commented):
My main concern is about breaking backwards compatibility. What is the consequences of that?

src/Microsoft.TestPlatform.AdapterUtilities/Resources/Resources.Designer.cs
Haplois: Idk, I'll check.

src/Microsoft.TestPlatform.AdapterUtilities/PublicAPI/PublicAPI.Shipped.txt
Haplois: Yes, we will only enable this on next release. (That part is not baked in TE yet.) Also if an adapter returns something not supported we will fallback to current behavior.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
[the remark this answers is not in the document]
Evangelink: Is it worth transforming this into some `InvalidOperationException` then? It usually represents some invalid state (which seems to match this case).

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -81,20 +81,50 @@ public static void GetManagedName(MethodBase method, out string managedTypeName,
...
+    /// The hierarchy values.
+    /// </returns>
+    public static string[] GetManagedHierarchy(MethodBase method)
+    {
+        GetManagedNameAndHierarchy(method, true, out _, out _, out var hierarchyValues);
+
+        return hierarchyValues;
+    }
+
+    private static void GetManagedNameAndHierarchy(MethodBase method, bool useClosedTypes, out string managedTypeName, out string managedMethodName, out string[] hierarchyValues)
```
Evangelink, line 114: It's not entirely clear to me what is a "closed type", could you please either explain here or add comment on one of the methods?

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Evangelink: It seems that test for this case is missing.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
Haplois: Yes. 'List<string>' is a closed type; and 'List<T>' is open.

Evangelink (commented):
One last comment.

PS: It'd be better to have a separate PR for the assemblyinfo changes as they do not relate to this change.

src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameHelper.Reflection.cs
```diff
@@ -352,7 +399,9 @@ private static void AppendNamespace(StringBuilder b, string? namespaceString)
         }
     }
 
-    private static void AppendMethodString(StringBuilder methodBuilder, string name, int methodArity)
+#pragma warning disable IDE0060 // Remove unused parameter
+    private static void AppendMethodString(StringBuilder methodBuilder, string name, int methodArity, bool closedType)
```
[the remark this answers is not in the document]
Haplois: I broke those in this PR - reverted them back. But for some reason it is shown as a change. It isn't :)

nohwnd (approved):
Approving, because I am promised we can iterate on the design. And hopefully make the values separate properties that will be output only when they have non-default values. And same for the FQN properties, because keeping them custom (in property bag) adds a lot of overhead.

PR #3714, opened 2022-06-02
Update framework detection logic to not rely on throwing/catching NRE

Evangelink:
## Description

Update framework detection logic to not rely on throwing/catching NRE. When running some tests locally in Debug we had `Debug.Assert` failures, in CI/release the behavior was the same as before because assert are ignored.

### DIFF
--- a/src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
+++ b/src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
@@ -32,7 +32,7 @@ internal class AssemblyMetadataProvider : IAssemblyMetadataProvider
     }
 
     /// <inheritdoc />
-    public FrameworkName GetFrameWork(string filePath)
+    public FrameworkName GetFrameworkName(string filePath)
     {
         FrameworkName frameworkName = new(Framework.DefaultFramework.Name);
         try
@@ -42,10 +42,10 @@ internal class AssemblyMetadataProvider : IAssemblyMetadataProvider
         }
         catch (Exception ex)
         {
-            EqtTrace.Warning("AssemblyMetadataProvider.GetFrameWork: failed to determine TargetFrameworkVersion exception: {0} for assembly: {1}", ex, filePath);
+            EqtTrace.Warning("AssemblyMetadataProvider.GetFrameworkName: failed to determine TargetFrameworkVersion exception: {0} for assembly: {1}", ex, filePath);
         }
 
-        EqtTrace.Info("AssemblyMetadataProvider.GetFrameWork: Determined framework:'{0}' for source: '{1}'", frameworkName, filePath);
+        EqtTrace.Info("AssemblyMetadataProvider.GetFrameworkName: Determined framework:'{0}' for source: '{1}'", frameworkName, filePath);
 
         return frameworkName;
     }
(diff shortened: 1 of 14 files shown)


src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -32,7 +32,7 @@ internal AssemblyMetadataProvider(IFileHelper fileHelper)
     }
 
     /// <inheritdoc />
-    public FrameworkName GetFrameWork(string filePath)
+    public FrameworkName GetFrameworkName(string filePath)
```
Evangelink, line 35: Somehow unrelated but I got confused by the API name compared to its return type when fixing the auto-detection algorithm so I also fixed the name.

src/vstest.console/CommandLine/InferHelper.cs
```diff
@@ -133,35 +133,24 @@ public Architecture AutoDetectArchitecture(IList<string?>? sources, Architecture
...
-            framework = Framework.FromString(finalFx.FullName);
-            if (conflictInFxIdentifier)
-            {
-                // TODO Log to console and client.
-                EqtTrace.Info(
-                    "conflicts in Framework identifier of provided sources(test assemblies), using default framework:{0}",
-                    framework);
-            }
-        }
-        catch (Exception ex)
```
Evangelink, line 154: Try/catch was only catching NRE when `finalFx` is `null` which happens only when all sources provided are `null`. `DetermineFrameworkName` and `Framework.FromString` are having inner try/catch.

src/vstest.console/CommandLine/InferHelper.cs
```diff
@@ -133,35 +133,24 @@ public Architecture AutoDetectArchitecture(IList<string?>? sources, Architecture
     public Framework AutoDetectFramework(IList<string?>? sources, out IDictionary<string, Framework> sourceToFrameworkMap)
     {
         sourceToFrameworkMap = new Dictionary<string, Framework>();
-        Framework framework = Framework.DefaultFramework;
 
         if (sources == null || sources.Count == 0)
```
Evangelink, line 137: I am not filtering `null` sources here because the simplest solution is to actually return `Framework` instead of `FrameworkName` on `DetermineFramework`. This way I can solve the "problem" of instances equality.

src/vstest.console/CommandLine/InferHelper.cs
```diff
@@ -133,35 +133,24 @@ public Architecture AutoDetectArchitecture(IList<string?>? sources, Architecture
...
+                framework);
         }
 
         EqtTrace.Info("Determined framework for all sources: {0}", framework);
-
         return framework;
     }
 
-    private FrameworkName? DetermineFrameworkName(IEnumerable<string?> sources, out IDictionary<string, Framework> sourceToFrameworkMap, out bool conflictInFxIdentifier)
+    private Framework DetermineFramework(IEnumerable<string?> sources, out IDictionary<string, Framework> sourceToFrameworkMap, out bool conflictInFxIdentifier)
```
Evangelink, line 153: Potentially we could move this as local function of `AutoDetetFramework`.

nohwnd (approved):

PR #3717, opened 2022-06-02
[After #3716] Fix debug assert failures on some tests for parallel discovery

Evangelink:
Not sure why we missed these cases but some tests are failing in debug mode because of some `Debug.Assert` conditions.

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
@@ -27,7 +27,7 @@ internal class ParallelProxyDiscoveryManager : IParallelProxyDiscoveryManager
     private readonly ParallelOperationManager<IProxyDiscoveryManager, ITestDiscoveryEventsHandler2, DiscoveryCriteria> _parallelOperationManager;
     private readonly Dictionary<string, TestRuntimeProviderInfo> _sourceToTestHostProviderMap;
     private int _discoveryCompletedClients;
-    private int _availableTestSources = -1;
+    private int _availableTestSources;
     private int _availableWorkloads;
     private bool _skipDefaultAdapters;
     private readonly IRequestData _requestData;
--- a/test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
+++ b/test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
@@ -147,7 +147,7 @@ public class ParallelProxyDiscoveryManagerTests
     }
 
     [TestMethod]
-    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedWithEventHandler()
+    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedBeforeBeingStartedWithEventHandler()
     {
         var parallelDiscoveryManager = new ParallelProxyDiscoveryManager(_mockRequestData.Object, _createMockManager, dataAggregator: new(), parallelLevel: 1, new List<TestRuntimeProviderInfo>());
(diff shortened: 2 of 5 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelProxyDiscoveryManager.cs
```diff
@@ -27,7 +27,7 @@ internal class ParallelProxyDiscoveryManager : IParallelProxyDiscoveryManager
     private readonly ParallelOperationManager<IProxyDiscoveryManager, ITestDiscoveryEventsHandler2, DiscoveryCriteria> _parallelOperationManager;
     private readonly Dictionary<string, TestRuntimeProviderInfo> _sourceToTestHostProviderMap;
     private int _discoveryCompletedClients;
-    private int _availableTestSources = -1;
```
Evangelink, line 30: It brings nothing to initialize with `-1`.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
```diff
@@ -147,7 +147,7 @@ public void DiscoverTestsShouldProcessAllSources()
     }
 
     [TestMethod]
-    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedWithEventHandler()
+    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedBeforeBeingStartedWithEventHandler()
```
Evangelink, line 150: This test calls `Abort` without having called `DiscoverTests` before, that seems illogical but nothing in our code prevents this state so I decided to keep the test for now.

As we were initializing to -1 the number of sources and never calling `DiscoverTests` which resets the value to the actual number of sources (i.e. 0+) we were having assertion failures.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
```diff
@@ -159,7 +159,22 @@ public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedW
...
+
+        parallelDiscoveryManager.DiscoverTests(_discoveryCriteriaWith2Sources, _mockEventHandler.Object);
+        parallelDiscoveryManager.Abort(_mockEventHandler.Object);
+        bool isPartialDiscoveryComplete = parallelDiscoveryManager.HandlePartialDiscoveryComplete(proxyDiscovermanager, 20, new List<TestCase>(), isAborted: false);
+
+        Assert.IsTrue(isPartialDiscoveryComplete);
+    }
+
+    [TestMethod]
+    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedBeforeBeingStarted()
```
Evangelink, line 177: Similar to comment test aborting without starting discovery.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
```diff
@@ -170,13 +185,28 @@ public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAborted(
         Assert.IsTrue(isPartialDiscoveryComplete);
     }
 
+    [TestMethod]
+    public void HandlePartialDiscoveryCompleteShouldReturnTrueIfDiscoveryWasAbortedAfterBeingStarted()
```
Evangelink, line 189: Similar to comment test aborting after starting discovery.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelProxyDiscoveryManagerTests.cs
nohwnd: Cool, I planned to investigate why I am seeing the failure on -1 sources, I thought it is something like this where the initial step is skipped in a unit test. So now I don't have to.

PR #3722, opened 2022-06-07
Get PlatformAbstractions from ObjectModel

drognanar:
Do not pick up the PlatformAbstractions.dll as Content, since they should be included through the ObjectModel package dependency


Evangelink (approved):
LGTM! Is there any ticket/experience linked to this change?

PR #3729, opened 2022-06-08
Changed identifier data for test discoverer plugins

cvpoienaru:
Changed identifier data for test discoverer plugins


Evangelink (commented):
Is there any ticket associated with this change? If so, could you please link it? If not, I think the PR deserves some explanation/description so we can better understand the motivation behind it (especially when we will be looking back).

Also, it would be nice to add some test to avoid regression (it also helps with clarifying intent IMO).

Evangelink:
@cvpoienaru I will close this draft. Please reopen if you plan to finish the work on it.

PR #3740, opened 2022-06-10
Add env var to control host priority

Evangelink:
## Description
Add environment variable to control host process priority. This would help us experiment with different values to find the best combination(s).

## Related issue
Fixes [AB#1545979](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1545979)

### DIFF
--- a/src/Microsoft.TestPlatform.CoreUtilities/Helpers/EnvironmentVariableHelper.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/Helpers/EnvironmentVariableHelper.cs
@@ -18,32 +18,10 @@ internal class EnvironmentVariableHelper : IEnvironmentVariableHelper
         => Environment.GetEnvironmentVariable(variable);
 
     /// <inheritdoc />
-    // TODO: This helper won't be needed when we will stop support for .NET Standard 1.3 and UWP
-    public TEnum GetEnvironmentVariableAsEnum<TEnum>(string variable, TEnum defaultValue = default) where TEnum : Enum
-    => Environment.GetEnvironmentVariable(variable) is string value && !string.IsNullOrEmpty(value)
-        ? (TEnum)Enum.Parse(typeof(TEnum), value)
-        : defaultValue;
-
-    /// <inheritdoc />
-    public T GetEnvironmentVariable<T>(string variable, T defaultValue = default) where T : IConvertible
+    public TEnum GetEnvironmentVariableAsEnum<TEnum>(string variable, TEnum defaultValue = default) where TEnum : struct, Enum
         => Environment.GetEnvironmentVariable(variable) is string value && !string.IsNullOrEmpty(value)
-            ? ConvertToType<T>(value)
+            ? Enum.TryParse<TEnum>(value, out var enumValue) ? enumValue : defaultValue
             : defaultValue;
-
-    private static T ConvertToType<T>(string input)
-    {
-        var targetType = typeof(T);
-        var conversionType = Nullable.GetUnderlyingType(targetType) ?? targetType;
-
-#if !WINDOWS_UWP && !NETSTANDARD1_3
-        if (conversionType.IsEnum)
-        {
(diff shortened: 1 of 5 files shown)


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -469,8 +482,19 @@ private bool LaunchHost(TestProcessStartInfo testHostStartInfo, CancellationToke
             _processHelper.SetExitCallback(processId, ExitCallBack);
         }
 
+        if (_testHostProcess is null)
+        {
+            return false;
+        }
+
+        _testHostProcess.PriorityClass =
+            _environmentVariableHelper.GetEnvironmentVariable("VSTEST_HOST_PRIORITY") is { } processPriorityString
```
Evangelink, line 491: Maybe we should add `INTERNAL` somewhere in the name to make it clear this is not stable API.

nohwnd, line 491: Is there a better syntax than `{ }` to get non-null value? e.g. `is string` or `is not null`?

Maybe it would be easier to understand as

Or _environmentVariableHelper.GetEnvironmentVariableAsEnum<PriorityClass?>("VSTEST_HOST_PRIORITY", defaultValue: null); 

Because that would be useful utility for other places where we want to get value from the env variable and provide default value as well, no matter if null, or actual fallback value like PriorityClass.Normal (which we don't want to do here to avoid touching the priority unless user specifically asked us to set it).

The env variable should also allow

nohwnd (approved):

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
nohwnd: Seems like even more generic GetEnvironmentVariable<T> would be also useful, but is there a universal interface we can use for that like IParseable, or some converter? Are type constraints like: where T is int or is double or is string or is enum possible nowadays?

Evangelink: > Is there a better syntax than `{ }` to get non-null value? e.g. `is string` or `is not null`?

Yes `is string processPriorityString`  would work.

> Seems like even more generic GetEnvironmentVariable would be also useful, but is there a universal interface we can use for that like IParseable, or some converter? Are type constraints like: where T is int or is double or is string or is enum possible nowadays?

I will make some test but I think if `T : IConvertible` should be enough.

src/Microsoft.TestPlatform.CoreUtilities/Helpers/EnvironmentVariableHelper.cs
```diff
@@ -13,8 +13,37 @@ namespace Microsoft.VisualStudio.TestPlatform.CoreUtilities.Helpers;
...
+        var conversionType = Nullable.GetUnderlyingType(targetType) ?? targetType;
+
+#if !WINDOWS_UWP && !NETSTANDARD1_3
+        if (conversionType.IsEnum)
+        {
+            return (T)Enum.Parse(conversionType, input);
+        }
+#endif
+
+        return (T)Convert.ChangeType(input, conversionType);
```
nohwnd, line 45: Won't this throw quite easily?

Evangelink: It's a lot less safe than doing `XXX.TryParse` for sure. Also we are not allowing any format provider. I can add a try/catch but that would still be less good than specific parse call.

Maybe this helper is not a good idea and I should keep only the enum one (replacing `Parse` with `TryParse`).

nohwnd (approved):

nohwnd (approved):

PR #3748, opened 2022-06-13
Experimental feature: enable negative values of MaxCpuCount to match a percentage of number of cores

Evangelink:
Replaces #3747 and fixes [AB#1545983](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1545983)

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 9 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
```diff
@@ -492,15 +499,28 @@ private static int GetDistinctNumberOfSources(TestRunCriteria testRunCriteria)
...
             EqtTrace.Verbose(
                 "TestEngine: Initializing Parallel Execution as MaxCpuCount is set to: {0}",
                 parallelLevelToUse);
 
+            // TODO: EXPERIMENTAL FEATURE - will need to be removed or strengthen/tested.
+            // A negative value is used to indicate that the value should be used as a percentage of the number of cores.
+            if (parallelLevelToUse < 0)
+            {
+                Debug.Assert(parallelLevelToUse > -100, "Parallel level should be above -100 (i.e. 100% of cores)");
+                parallelLevelToUse = (int)Math.Round((double)-parallelLevelToUse * _environment.ProcessorCount / 100.0);
```
nohwnd, line 514:
```suggestion
                // Use the given percent of available cores, while rounding up. Always use at least 1 core.
                Debug.Assert(parallelLevelToUse > -100, "Parallel level should be above -100 (i.e. 100% of cores)");
                parallelLevelToUse = (int)Math.Max(1, Math.Round((double)-parallelLevelToUse * _environment.ProcessorCount / 100.0, MidpointRounding.AwayFromZero));
```

nohwnd: The if below should be removed it this change is used. 

You could also remove that Debug.Assert, because it will only annoy us, and use Math.Min on the result with the count of logical cores, so we never say to use more than what we have.

nohwnd (approved):

PR #3761, opened 2022-06-14
Fix failing assertions on tests

Evangelink:


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -236,7 +236,7 @@ private void CollectDumpAndAbortTesthost()
             {
                 Action<string> logWarning = m => _logger.LogWarning(_context.SessionDataCollectionContext, m);
                 var dumpDirectory = GetDumpDirectory();
-                _processDumpUtility.StartHangBasedProcessDump(_testHostProcessId, dumpDirectory, _hangDumpType == HangDumpType.Full, _targetFramework, logWarning);
+                _processDumpUtility.StartHangBasedProcessDump(_testHostProcessId, dumpDirectory, _hangDumpType == HangDumpType.Full, _targetFramework!, logWarning);
```
Evangelink, line 239: In failing tests `_targetFramework` is null but we pass it to some mock that doesn't care. Real life scenario, the implementation of this method throws on null TFM. I could also keep the assertion and update the tests to not pass null.

This comment also applies to the other assertion below.

nohwnd (approved):

PR #3768, opened 2022-06-16
Support arm64 native CppUnitTestFramework with `dotnet test`

MarcoRossignoli:
Support `dotnet test nativeArm64.dll`
image

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -533,18 +533,6 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
 
         bool SilentlyForceToX64(string sourcePath)
         {
-            // Scenario: dotnet test nativeArm64.dll
-            // If the dll is native and we're not running in process(vstest.console.exe)
-            // the expected target framework is ".NETCoreApp,Version=v1.0".
-            // In this case we don't want to force x64 architecture
-            using var assemblyStream = _fileHelper.GetStream(sourcePath, FileMode.Open, FileAccess.Read);
-            using var peReader = new PEReader(assemblyStream);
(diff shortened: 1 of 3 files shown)


Evangelink (approved):

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -408,7 +408,7 @@ public Task<bool> LaunchTestHostAsync(TestProcessStartInfo testHostStartInfo, Ca
 
             // We silently force x64 only if the target architecture is the default one and is not specified by user
             // through --arch or runsettings or -- RunConfiguration.TargetPlatform=arch
-            bool forceToX64 = SilentlyForceToX64() && _runsettingHelper.IsDefaultTargetArchitecture;
+            bool forceToX64 = SilentlyForceToX64(sourcePath) && _runsettingHelper.IsDefaultTargetArchitecture;
```
nohwnd, line 411: Is this in this order on purpose? SilentlyForceToX64 should be evaluated even if IsDefaultTargetArchitecture is true?

nohwnd (approved):

nohwnd:
@MarcoRossignoli  not sure how to revoke my approval without blocking you from merging. So I just disabled automerge. 

Should the native detection run when we know that the default architecture is set?

MarcoRossignoli:
> Should the native detection run when we know that the default architecture is set?

we should never force if native.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
MarcoRossignoli: > Is this in this order on purpose? SilentlyForceToX64 should be evaluated even if IsDefaultTargetArchitecture is true?

do you meant `false`

PR #3770, opened 2022-06-16
Add default platform option to runsettings

nohwnd:
## Description
Add default platform option that determines the platform to use for AnyCPU sources, without forcing it at sources that are built for specific platform. 

This is equivalent of this dropdown in VS: 

image

Not setting DefaultPlatform is equivalent to the `Auto` option. Which is fallback to the platform of the current process. E.g. x64 for x64 VisualStudio, or arm64 for arm64 VisualStudio.

The difference between TargetPlatform and DefaultPlatform, is that TargetPlatform will force the platform on all sources, and the ones that don't match will be skipped or will fail with error.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -236,7 +236,7 @@ public class RunConfiguration : TestRunSettings
     }
 
     /// <summary>
-    /// Gets or sets the Target platform this run is targeting. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
+    /// Gets or sets the Target platform this run is targeting. Possible values are <see cref="Architecture"/> except for AnyCPU and Default.
     /// </summary>
     public Architecture TargetPlatform
     {
@@ -253,7 +253,7 @@ public class RunConfiguration : TestRunSettings
     }
 
     /// <summary>
-    /// Gets or sets the default platform that will be used for AnyCPU sources, or non-dll sources. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
+    /// Gets or sets the default platform that will be used for AnyCPU sources, or non-dll sources. Possible values are <see cref="Architecture"/> except for AnyCPU and Default.
     /// </summary>
     public Architecture? DefaultPlatform
     {
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -22,6 +22,8 @@ public class RunConfiguration : TestRunSettings
     /// </summary>
     private Architecture _platform;
 
+    private Architecture? _defaultPlatform;
+
     /// <summary>
     /// Maximum number of cores that the engine can use to run tests in parallel
     /// </summary>
@@ -234,7 +236,7 @@ public class RunConfiguration : TestRunSettings
     }
 
     /// <summary>
-    /// Gets or sets the Target platform this run is targeting. Possible values are <c>x86|x64|arm|anycpu</c>.
+    /// Gets or sets the Target platform this run is targeting. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
     /// </summary>
     public Architecture TargetPlatform
     {
@@ -247,6 +249,23 @@ public class RunConfiguration : TestRunSettings
         {
             _platform = value;
             TargetPlatformSet = true;
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets the default platform that will be used for AnyCPU sources, or non-dll sources. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
+    /// </summary>
+    public Architecture? DefaultPlatform
+    {
+        get
+        {
+            return _defaultPlatform;
+        }
+
+        set
+        {
+            _defaultPlatform = value;
+            DefaultPlatformSet = true;
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -234,7 +236,7 @@ public bool DisableParallelization
     }
 
     /// <summary>
-    /// Gets or sets the Target platform this run is targeting. Possible values are <c>x86|x64|arm|anycpu</c>.
+    /// Gets or sets the Target platform this run is targeting. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
```
Evangelink, line 239: Not sure but I would have said `and` instead of `or`.

```suggestion
    /// Gets or sets the Target platform this run is targeting. Possible values are <see cref="Architecture"/> except for AnyCPU and Default.
```

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -250,6 +252,23 @@ public Architecture TargetPlatform
...
+    {
+        get
+        {
+            return _defaultPlatform;
+        }
+
+        set
+        {
+            _defaultPlatform = value;
+            DefaultPlatformSet = true;
```
Evangelink, line 268: Do we expect "null" to be set for the architecture? If not, then I would recommend to make this property an auto-property and remove `DefaultPlatformSet`.

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -250,6 +252,23 @@ public Architecture TargetPlatform
         }
     }
 
+    /// <summary>
+    /// Gets or sets the default platform that will be used for AnyCPU sources, or non-dll sources. Possible values are <see cref="Architecture"/> except for AnyCPU or Default.
+    /// </summary>
+    public Architecture? DefaultPlatform
```
MarcoRossignoli, line 258: Should we name it better like AnyCPU[Concrete/...]Platform?

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -348,6 +367,15 @@ public bool TargetPlatformSet
         private set;
     }
 
+    /// <summary>
+    /// Gets a value indicating whether default platform is set.
+    /// </summary>
+    public bool DefaultPlatformSet
```
Evangelink, line 373: Nit: IsDefaultPlatformSet

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -737,6 +772,34 @@ public static RunConfiguration FromXml(XmlReader reader)
                         runConfiguration.TargetPlatform = archType;
                         break;
 
+                    case nameof(DefaultPlatform):
+                        XmlRunSettingsUtilities.ThrowOnHasAttributes(reader);
+                        Architecture defaultArchType;
+                        string defaultPlatformValue = reader.ReadElementContentAsString();
+                        try
+                        {
+                            defaultArchType = (Architecture)Enum.Parse(typeof(Architecture), defaultPlatformValue, true);
```
Evangelink, line 781: What about using `Enum.TryParse` and avoid the catch?

test/vstest.console.UnitTests/TestPlatformHelpers/TestRequestManagerTests.cs
```diff
@@ -2571,4 +2573,50 @@ private void DiscoverTestsIfThrowsExceptionShouldThrowOut(Exception exception)
 
         _testRequestManager.DiscoverTests(payload, mockDiscoveryEventsRegistrar.Object, _protocolConfig);
     }
+
```
Evangelink, line 2576: Please add a test for the `SettingsException`

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
nohwnd: It is not just for AnyCPU, but for any source that does not determine a platform. best names are imho FallbackPlatform or DefaultPlatform, which is what I used.

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
nohwnd: This is following the same naming pattern as the rest of the file.

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
nohwnd: I am just doing what all the other settings are doing. We do expect the value to be null when the setting is not set. The whole logic of this class around nulls is not very clear, because e.g. _platform always has value, but is not always serialized.

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
nohwnd: Again, same code as above. We are close to final, I am using what was proven to work. We can redesign it, there is experience for it, but just not now :)

test/vstest.console.UnitTests/TestPlatformHelpers/TestRequestManagerTests.cs
nohwnd: Added. UsingInvalidValueForDefaultPlatformSettingThrowsSettingsException

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
nohwnd: Seems equivalent, but I changed it anyway.

Evangelink (approved):

MarcoRossignoli (approved):

PR #3772, opened 2022-06-16
Add background option for testhosts

Evangelink:
Fixes [AB#1555397](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1555397)

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
@@ -531,7 +531,7 @@ public class TestEngine : ITestEngine
                 // In case of a background discovery we want to reduce the number of cores utilized
                 // to leave enough power for other tasks.
                 var runSettingsEnvVariables = InferRunSettingsHelper.GetEnvironmentVariables(runSettings);
-                string? isBackgroundDiscoveryEnabled = null;
+                string isBackgroundDiscoveryEnabled = null;
                 if (runSettingsEnvVariables is not null
                     && runSettingsEnvVariables.TryGetValue("VSTEST_BACKGROUND_DISCOVERY", out isBackgroundDiscoveryEnabled)
                     && isBackgroundDiscoveryEnabled == "1"
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
@@ -522,19 +522,30 @@ public class TestEngine : ITestEngine
                 "TestEngine: Initializing Parallel Execution as MaxCpuCount is set to: {0}",
                 parallelLevelToUse);
 
-            // TODO: EXPERIMENTAL FEATURE - will need to be removed or strengthen/tested.
-            // A negative value is used to indicate that the value should be used as a percentage of the number of cores.
-            if (parallelLevelToUse < 0)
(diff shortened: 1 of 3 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/TestEngine.cs
```diff
@@ -522,19 +522,30 @@ private static int GetDistinctNumberOfSources(TestRunCriteria testRunCriteria)
...
+                    && bool.TryParse(value, out var isBackgroundDiscoveryEnabled)
+                    && isBackgroundDiscoveryEnabled
+                    && userParallelSetting == 0) // If user specifies a CPU count, respect it
+                {
+                    // Dummy logic based on some observations, might need to be tweaked/improved.
+                    parallelLevelToUse = parallelLevelToUse switch
+                    {
+                        1 => 1,
+                        < 8 => (int)Math.Round(parallelLevelToUse / 2.0, MidpointRounding.AwayFromZero),
+                        _ => (int)Math.Round((double)0.75 * parallelLevelToUse / 2.0, MidpointRounding.AwayFromZero),
```
nohwnd, line 545:
```suggestion
                        _ => (int)Math.Round((double)0.75 * parallelLevelToUse, MidpointRounding.AwayFromZero),
```

Evangelink: Doh! Copy paste mistake! Thank you for catching it.

nohwnd (approved):

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -506,20 +507,24 @@ private bool LaunchHost(TestProcessStartInfo testHostStartInfo, CancellationToke
...
-    internal static void SetProcessPriority(Process testHostProcess, IEnvironmentVariableHelper environmentVariableHelper)
+    internal static void AdjustProcessPriorityBasedOnSettings(Process testHostProcess, IDictionary<string, string?>? testHostEnvironmentVariables)
     {
         ProcessPriorityClass testHostPriority = ProcessPriorityClass.BelowNormal;
         try
         {
-            testHostPriority = environmentVariableHelper.GetEnvironmentVariableAsEnum("VSTEST_HOST_INTERNAL_PRIORITY", testHostPriority);
-            testHostProcess.PriorityClass = testHostPriority;
-            EqtTrace.Verbose("Setting test host process priority to {0}", testHostProcess.PriorityClass);
+            if (testHostEnvironmentVariables is not null
```
shyamnamboodiripad, line 521: @Evangelink @nohwnd I am not familiar with the code base - but is DefaultTestHostManager also used when parallel is not in play?

Just want to make sure that it is possible to create low prio test hosts -
a) regardless of whether parallel is turned on and
b) regardless of whether the operation requested is execution versus discovery

Also if the answer to b) above is yes, then it may be better to rename this to something more general like VSTEST_BACKGROUND_OPERATION or VSTEST_BACKGROUND_PRIORITY :)

Evangelink: Yes that code is used even in "sequential". This is used to decide on the level of parallelism (including sequential) to use.

PR #3793, opened 2022-06-21
Enable nullable on ObjectModel

Evangelink:


MarcoRossignoli (approved):

src/Microsoft.TestPlatform.ObjectModel/Adapter/Interfaces/ITestExecutor.cs
```diff
@@ -22,15 +20,15 @@ public interface ITestExecutor
     /// <param name="tests">Tests to be run.</param>
     /// <param name="runContext">Context to use when executing the tests.</param>
     /// <param param name="frameworkHandle">Handle to the framework to record results and to do framework operations.</param>
-    void RunTests(IEnumerable<TestCase> tests, IRunContext runContext, IFrameworkHandle frameworkHandle);
+    void RunTests(IEnumerable<TestCase>? tests, IRunContext? runContext, IFrameworkHandle? frameworkHandle);
```
alexrp, line 23: @Evangelink I just updated to v17.3.0 of Microsoft.TestPlatform.ObjectModel and got nullable errors due to the annotations on these two `RunTests` methods.

Are they actually correct? I've looked at implementations both in this repo and the wild and no one seems to expect any of these to ever be null.

Evangelink: Hi @alexrp, thanks for raising this. I am just coming back from vacation and I will look into it this week.

Evangelink: Looking at mstest, xunit and nunit, it seems that `tests` cannot be `null` but `runContext` can.

PR #3794, opened 2022-06-22
Improve how to retrieve process ID

Evangelink:
## Description
See https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1837

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/PostProcessing/ArtifactProcessingManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/PostProcessing/ArtifactProcessingManager.cs
@@ -65,7 +65,14 @@ internal class ArtifactProcessingManager : IArtifactProcessingManager
         {
             _testSessionCorrelationId = testSessionCorrelationId;
             _processArtifactFolder = Path.Combine(_fileHelper.GetTempPath(), _testSessionCorrelationId);
-            _testSessionProcessArtifactFolder = Path.Combine(_processArtifactFolder, $"{Process.GetCurrentProcess().Id}_{Guid.NewGuid()}");
+#if NET5_0_OR_GREATER
+            var pid = Environment.ProcessId;
+#else
+            int pid;
+            using (var p = Process.GetCurrentProcess())
+                pid = p.Id;
+#endif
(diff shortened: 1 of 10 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/PostProcessing/ArtifactProcessingManager.cs
```diff
@@ -65,7 +65,14 @@ internal class ArtifactProcessingManager : IArtifactProcessingManager
         {
             _testSessionCorrelationId = testSessionCorrelationId;
             _processArtifactFolder = Path.Combine(_fileHelper.GetTempPath(), _testSessionCorrelationId);
-            _testSessionProcessArtifactFolder = Path.Combine(_processArtifactFolder, $"{Process.GetCurrentProcess().Id}_{Guid.NewGuid()}");
+#if NET5_0_OR_GREATER
+            var pid = Environment.ProcessId;
```
MarcoRossignoli, line 69: is it cached?

MarcoRossignoli (approved):

src/Microsoft.TestPlatform.CrossPlatEngine/PostProcessing/ArtifactProcessingManager.cs
Evangelink: System.Diagnostics.Process.GetCurrentProcess().Id is expensive:

It allocates a [Process](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process) instance, usually just to get the Id.
The [Process](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process) instance needs to be disposed, which has a performance impact.
It's easy to forget to call [Dispose()](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.dispose#system-diagnostics-process-dispose) on the [Process](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process) instance.
If nothing else besides Id uses the Process instance, then the linked size grows unnecessarily by increasing the graph of types referenced.
It is somewhat difficult to discover or find this API.
System.Environment.ProcessId avoids all the above.

PR #3806, opened 2022-06-23
Add runtime location to host log

### LINKED ISSUE #3792: No way to determine the runtime used with dotnet test
## Description
Our repository has run into a situation where we have tests failing in one environment but not the other. This happens on occasion but in this specific case it's failing locally but passing in CI. There is some suspicion that it's because the environments are picking different runtimes when invoking `dotnet test`. 

The problem is we have no way of knowing what runtime is being used when `dotnet test` is invoked. It's not listed in the command line, logs, etc ... It's left up to the developer to reason this out and we're failing to do so. 

## Expected behavior
There should be some mechanism that lets us know what runtime actually executed our tests. 

## Actual behavior
There is no such mechanism. The only options available are 

1. Use `%COREHOST_TRACE%`. This produces a ~45,000 line file with multiple host activations that we have to trace through to find the runtime. 
2. Attach debuggers to running processes to see what version got loaded

None of these are viable for CI environments

MarcoRossignoli:
Add runtime location to host log.

closes https://github.com/microsoft/vstest/issues/3792

outcome 
runner
image

host
image

### DIFF
--- a/src/testhost.x86/DefaultEngineInvoker.cs
+++ b/src/testhost.x86/DefaultEngineInvoker.cs
@@ -108,7 +108,7 @@ internal class DefaultEngineInvoker :
             string objectTypeLocation = typeof(object).Assembly.Location;
             if (!string.IsNullOrEmpty(objectTypeLocation))
             {
-                EqtTrace.Verbose($"Runtime location: {Path.GetDirectoryName(typeof(object).Assembly.Location)}");
+                EqtTrace.Verbose($"Runtime location: {Path.GetDirectoryName(objectTypeLocation)}");
             }
 #endif
         }
--- a/src/testhost.x86/DefaultEngineInvoker.cs
+++ b/src/testhost.x86/DefaultEngineInvoker.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Net;
 using System.Reflection;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


src/vstest.console/Processors/EnableDiagArgumentProcessor.cs
```diff
@@ -138,6 +138,13 @@ public void Initialize(string? argument)
         // Write version to the log here, because that is the
         // first place where we know if we log or not.
         EqtTrace.Verbose($"Version: {Product.Version} Current process architecture: {_processHelper.GetCurrentProcessArchitecture()}");
+        // https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=net-6.0#remarks
+        // In .NET 5 and later versions, for bundled assemblies, the value returned is an empty string.
+        string objectTypeLocation = typeof(object).Assembly.Location;
+        if (!StringUtils.IsNullOrEmpty(objectTypeLocation))
```
Evangelink, line 144: Nit

```suggestion
        if (!objectTypeLocation.IsNullOrEmpty())
```

src/testhost.x86/DefaultEngineInvoker.cs
```diff
@@ -100,7 +101,16 @@ public void Invoke(IDictionary<string, string?> argsDictionary)
                 .GetTypeInfo()
                 .Assembly
                 .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
-            EqtTrace.Verbose($"Version: {version}");
+            EqtTrace.Verbose($"Version: {version} Current process architecture: {_processHelper.GetCurrentProcessArchitecture()}");
+#if NETCOREAPP2_0_OR_GREATER || NETFRAMEWORK
+            // https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.location?view=net-6.0#remarks
+            // In .NET 5 and later versions, for bundled assemblies, the value returned is an empty string.
+            string objectTypeLocation = typeof(object).Assembly.Location;
+            if (!string.IsNullOrEmpty(objectTypeLocation))
```
Evangelink, line 109: Can you use the helper here?

```suggestion
            if (!objectTypeLocation.IsNullOrEmpty())
```

Evangelink (approved):
Let's see if this provides enough info.

Evangelink (approved):

PR #3815, opened 2022-06-27
Upgrade to Newtonsoft.Json 13.0.1

nohwnd:
Upgrade to Newtonsoft.Json 13.0.1.


Haplois (approved):

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
```diff
@@ -96,7 +96,7 @@ public Message DeserializeMessage(string rawMessage)
             // PERF: This is slow, we deserialize the message, and the payload into JToken just to get the header. We then
             // deserialize the data from the JToken, but that is twice as expensive as deserializing the whole object directly into the final object type.
             // We need this for backward compatibility though.
-            return Deserialize<VersionedMessage>(rawMessage);
+            return Deserialize<VersionedMessage>(rawMessage)!;
```
Evangelink, line 99: Are we sure to always have a value here? I see that below you are doing a null check on the deserialized data.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestObjectConverter.cs
```diff
@@ -55,10 +55,10 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         // key value pairs.
         foreach (var property in properties)
         {
-            var testProperty = property["Key"].ToObject<TestProperty>(serializer);
+            var testProperty = property["Key"]!.ToObject<TestProperty>(serializer)!;
 
             // Let the null values be passed in as null data
-            var token = property["Value"];
+            var token = property["Value"]!;
```
Evangelink, line 61: Same question here (i.e. above you are checking for null token and here not )

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentProcessorAppDomain.cs
```diff
@@ -157,7 +157,9 @@ public async Task<ICollection<AttachmentSet>> ProcessAttachmentSetsAsync(XmlElem
         cancellationToken.Register(() => _wrapper.CancelProcessAttachment());
         _processAttachmentSetsLogger = logger;
         _progressReporter = progressReporter;
+#pragma warning disable CS8603 // Possible null reference return.
```
Evangelink, line 160: Can you add an explanation of what's wrong here?

src/testhost.arm64/app.config
```diff
@@ -46,6 +46,11 @@
         <assemblyIdentity name="Microsoft.VisualStudio.QualityTools.Tips.UnitTest.ObjectModel" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
         <bindingRedirect oldVersion="10.0.0.0-16.0.0.0" newVersion="17.0.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <!-- Microsoft.Extensions.DependencyModel depends on version 9.0.0, and newer versions of that nuget don't support all the frameworks we need anymore. -->
+        <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
+        <bindingRedirect oldVersion="9.0.0.0-13.0.0.0" newVersion="13.0.0.0" />
```
Evangelink, line 52: I see that the package is `13.0.1`, is the dll still in `13.0.0.0`?

Evangelink (commented):
I cannot see coverage but I am unsure all the new ifs added are covered.

samples/Microsoft.TestPlatform.Protocol/Microsoft.TestPlatform.Protocol.csproj
```diff
@@ -11,6 +11,6 @@
     <!-- / CVE-2017-11770 -->
 
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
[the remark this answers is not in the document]
nohwnd: No it is in samples. Those should be hopefully standalone.

scripts/build.ps1
```diff
@@ -641,11 +641,11 @@ function Publish-Package {
     }
 
     # Copy dependency of Microsoft.TestPlatform.TestHostRuntimeProvider
-    $newtonsoft = Join-Path $env:TP_PACKAGES_DIR "newtonsoft.json\9.0.1\lib\net45\Newtonsoft.Json.dll"
+    $newtonsoft = Join-Path $env:TP_PACKAGES_DIR "newtonsoft.json\13.0.1\lib\net45\Newtonsoft.Json.dll"
```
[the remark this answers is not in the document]
nohwnd: yes.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
Evangelink: When unsure, I usually prefer a temp var and a `Debug.Assert` as it would fail and give me some context info instead of potentially failing somewhere else and not realizing because of all our try/catch.

Maybe we could just add a todo comment above so we can try to check null cases later?

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestObjectConverter.cs
```diff
@@ -55,10 +55,10 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         // key value pairs.
         foreach (var property in properties)
         {
-            var testProperty = property["Key"].ToObject<TestProperty>(serializer);
+            var testProperty = property["Key"]!.ToObject<TestProperty>(serializer)!;
```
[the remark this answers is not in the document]
nohwnd: nope. Fixed.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestObjectConverter.cs
nohwnd: nope. Fixed.

src/testhost.arm64/app.config
nohwnd: Yup, file version is the same.

src/Microsoft.TestPlatform.CrossPlatEngine/AttachmentsProcessing/DataCollectorAttachmentProcessorAppDomain.cs
nohwnd: Deserialization of the result might be null. Not sure why I did not see that before.

scripts/build.sh
```diff
@@ -436,7 +436,7 @@ function publish_package()
         done
         #*************************************************************************************************************#
 
-        newtonsoft=$TP_PACKAGES_DIR/newtonsoft.json/9.0.1/lib/netstandard1.0/Newtonsoft.Json.dll
+        newtonsoft=$TP_PACKAGES_DIR/newtonsoft.json/13.0.1/lib/netstandard1.0/Newtonsoft.Json.dll
```
Haplois, line 439: Since we're reading the version from Versions.props in build.ps1, can we do it here as well? This file has samples.

scripts/build/TestPlatform.Dependencies.props
```diff
@@ -57,7 +57,7 @@
     <!-- This version also needs to be updated in src\package\nuspec\TestPlatform.ObjectModel.nuspec -->
     <NuGetFrameworksVersion>5.11.0</NuGetFrameworksVersion>
     <ILAsmPackageVersion>5.0.0</ILAsmPackageVersion>
-    <JsonNetVersion>9.0.1</JsonNetVersion>
+    <JsonNetVersion>13.0.1</JsonNetVersion>
```
Haplois, line 60: Is there a reason to have a separate property here not use the on in Versions.props?

nohwnd: I did not add it as you can see here. It was probably you who did not migrate it 😛

scripts/build.sh
nohwnd: done.

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.cs
nohwnd: It worked until now, so imho not really a big problem.

MarcoRossignoli (approved):
Hoping that all that `return null` are not breaking something(if for some reason we're accidentally relying on some try/catch for NRE) 🤞

scripts/build.sh
Evangelink: You said done but I still see the version used here.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestObjectConverter.cs
```diff
@@ -55,12 +55,17 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
         // key value pairs.
         foreach (var property in properties)
         {
-            var testProperty = property["Key"].ToObject<TestProperty>(serializer);
+            var testProperty = property["Key"]?.ToObject<TestProperty>(serializer);
```
Evangelink, line 58: On `TestCaseConverter` you are doing `property?["Key"]` do we need the same here?

test/TestAssets/NewtonSoftDependency/NewtonSoftDependency.csproj
```diff
@@ -26,7 +26,7 @@
   <ItemGroup>
     <PackageReference Include="MSTest.TestAdapter" Version="$(MSTestAdapterVersion)" />
     <PackageReference Include="MSTest.TestFramework" Version="$(MSTestAdapterVersion)" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
Evangelink, line 29: Could you use the msbuild variable here?

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestObjectConverter.cs
nohwnd: I think doing that confuses the compiler to think that property can be null afterwards, even though we continue when testProperty is null. In any case we are probably not serializing null pair in the dictionary, so this should be safe. It also worked till now.

test/TestAssets/NewtonSoftDependency/NewtonSoftDependency.csproj
nohwnd: I think the intent here was to use a version that is newer than what we have in TP. But currently there is no such stable version. Added comment.

Evangelink (approved):

PR #3831, opened 2022-06-29
Bump to 17.4.0

MarcoRossignoli:
Bump to 17.4.0

completes https://github.com/microsoft/vstest/pull/3818

### DIFF
--- a/scripts/build/TestPlatform.Dependencies.props
+++ b/scripts/build/TestPlatform.Dependencies.props
@@ -11,7 +11,7 @@
     <!-- this version also needs to be "statically" readable because the test fixture will inspect this file for the version
     and because during the test `dotnet test` will run and re-build some of the test projects and at that time the version
     from a build parameter would not be available, so I am writing this version from the build.ps1 script to keep it in sync -->
-    <NETTestSdkVersion>17.3.0-dev</NETTestSdkVersion>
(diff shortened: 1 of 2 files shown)


Evangelink (approved):
@haplois Can we share the version somehow between the 2 parts? Or is it possible to create a script that validates the versions produced so we avoid doing a partial bump as we did?

scripts/build/TestPlatform.Dependencies.props
```diff
@@ -11,7 +11,7 @@
     <!-- this version also needs to be "statically" readable because the test fixture will inspect this file for the version
     and because during the test `dotnet test` will run and re-build some of the test projects and at that time the version
     from a build parameter would not be available, so I am writing this version from the build.ps1 script to keep it in sync -->
-    <NETTestSdkVersion>17.4.0-dev</NETTestSdkVersion>
+    <NETTestSdkVersion>17.3.0-dev</NETTestSdkVersion>
```
Evangelink, line 14: Might be worth to add a comment why this can't be latest version.

Haplois:
> @Haplois Can we share the version somehow between the 2 parts? Or is it possible to create a script that validates the versions produced so we avoid doing a partial bump as we did?

I will update our build script to build and update the version accordingly like we do in TestFx when we provide external versions.

PR #3832, opened 2022-06-29
Enable more rules on test projects

Evangelink:


test/Microsoft.TestPlatform.AcceptanceTests/AppDomainTests.cs
```diff
@@ -20,7 +20,7 @@ public class AppDomainTests : AcceptanceTestBase
 {
     [TestMethod]
     [TestCategory("Windows-Review")]
-    [NetFullTargetFrameworkDataSource]
+    [NetFullTargetFrameworkDataSourceAttribute]
```
nohwnd, line 23: Why we should spell out `Attribute` on this attribute? TestMethod is also an attribute and does not spell it out.

test/TestAssets/performance/MSTest1000Passing/MSTests.cs
```diff
@@ -5453,5 +5452,5 @@ public void Test_50_20()
         #region FailingTests
         #endregion
     }
- 
+
```
nohwnd, line 5455: was this fixed by fixing the generator and re-generating the file?

nohwnd (approved):

PR #3852, opened 2022-07-15
Enforce use of correct dispose pattern

Evangelink:


src/Microsoft.TestPlatform.CoreUtilities/Utilities/JobQueue.cs
```diff
@@ -197,7 +197,14 @@ public void Flush()
...
     public void Dispose()
     {
-        if (_isDisposed)
+        Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    protected virtual void Dispose(bool disposing)
+    {
+        if (_isDisposed
```
nohwnd, line 206: This is the only class that checks _isDisposed, why the other classes don't use/check that?

src/Microsoft.TestPlatform.CrossPlatEngine/EventHandlers/TestRequestHandler.cs
```diff
@@ -155,8 +155,17 @@ public void ProcessRequests(ITestHostManagerFactory testHostManagerFactory)
     /// <inheritdoc />
     public void Dispose()
     {
-        _communicationEndPoint?.Stop();
-        _channel?.Dispose();
+        Dispose(true);
+        GC.SuppressFinalize(this);
```
nohwnd, line 159: Isn't this useful only for classes that actually have a finalizer? Or is it done as a precaution because we don't mark the class as sealed, and we don't know if an inheritor will implement a finalizer?

Evangelink: The second part :)
It's usually "better" to start with limited scope for a class (i.e. `internal sealed`) and open the scope if needed. Obviously that depends on what kind of code you write (for a framework you usually tend to open by default and close on some specific types).

src/Microsoft.TestPlatform.CoreUtilities/Utilities/JobQueue.cs
Evangelink: I haven't changed the actual implementation of the `Dispose`, but only applied the pattern. Preventing multiple dispose is not always needed.

MarcoRossignoli (approved):

src/Microsoft.TestPlatform.CoreUtilities/Utilities/JobQueue.cs
nohwnd: >   Preventing multiple dispose is not always needed. 

Same as suppressing finalization 😁 Yet you still enforce it.

nohwnd (approved):

src/Microsoft.TestPlatform.CoreUtilities/Utilities/JobQueue.cs
Evangelink: I have addressed the comment as there were some merge conflicts.

PR #3646, opened 2022-05-17
Follow .NET lifecycle: update codebase to net462

Evangelink:
## Description

Follow Microsoft .NET Framework component lifecycle policy by dropping support for .NET Framework older than 4.6.2.

## Related issue

Replaces #3517 and fixes [AB#1491257](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1491257)

PR #3856, opened 2022-07-15
Cleanups post move to net462

Evangelink:
Follow-up from #3646 (see https://github.com/microsoft/vstest/pull/3646#discussion_r922174075)


src/Microsoft.TestPlatform.ObjectModel/TraitCollection.cs
```diff
@@ -19,13 +19,9 @@ public class TraitCollection : IEnumerable<Trait>
     internal const string TraitPropertyId = "TestObject.Traits";
     private static readonly TestProperty TraitsProperty = TestProperty.Register(
         TraitPropertyId,
-#if !NET462
         // TODO: Fix this with proper resourcing for UWP and Win 8.1 Apps
         // Trying to access resources will throw "MissingManifestResourceException" percolated as "TypeInitialization" exception
         "Traits",
-#else
```
nohwnd, line 26: I think I got it backwards. This should be used: Resources.Resources.TestCasePropertyTraitsLabel,

Because the condition is a compiler directive, so net451 used Resources.Resources.TestCasePropertyTraitsLabel, and net462 should to, and net(coreapp) should probably too, even thought it used the harcoded value till now. I hpe the resource is just set to Traits :D

Evangelink: The resource is set to `Traits` for English.

Why don't we simply use the hardcoded string for `WINDOWS_UWP` (and whatever the condition is for Win8.1)?

nohwnd: is it different for other languages? wtf? I hope we are not using that as an identifier. (would have been broken for netcoreapp so I think we don't)

Evangelink: Yes it's being translated.

Evangelink: I still don't understand the original condition. Basically for .NET 451 we use resx and for everything else we use hardcoded value. BUT comment says we want to use hardcoded for UWP and Win8.1. So we should have WINDOWS_UAP for "Traits" and everything else using resx. Am I missing something?

nohwnd: I don't think you are missing something, but there are 2 things: 
- we probably won't build for windows_uap anymore, because we will only update winUI runner which is netstandard2.0, so adding that condition is unnecessary.
- because the condition was not updated when netcoreapp was introduced, netcoreapp was also using the hardcoded string, so changing the condtion is potentially a breaking change, and it would be interesting to have more context on how it is actually used.

Evangelink: > * we probably won't build for windows_uap anymore, because we will only update winUI runner which is netstandard2.0, so adding that condition is unnecessary.

It seems we are still building it:
https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj#L9-L11

> * because the condition was not updated when netcoreapp was introduced, netcoreapp was also using the hardcoded string, so changing the condtion is potentially a breaking change, and it would be interesting to have more context on how it is actually used.

That's possible. So using the hardcoded string always would be the more alike to what was there before. I will add a TODO and a task to fix loading from a resource so we keep track of it.

nohwnd: - We build it, but no-one is consuming. AFAIK. UWP runtime provider is not updated with those dll, because it needs to stick to supporting older versions of UWP than what we are able to support.
- Hopefully there is another ID that is used, and this string is used only as a description, there are imho 2 names sent for the property, you should be able to see it in json messages on any TestRunChangedEventArgs (test run update).

MarcoRossignoli (approved):

PR #3858, opened 2022-07-15
Allow DotNetHostPath to contain env vars

Therzok:
This is useful for controlling the paths to the dotnet host path when you're running `dotnet test src/tests/TestProjectA.csproj` to avoid duplicating the runsettings.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -891,7 +891,6 @@ public class RunConfiguration : TestRunSettings
 
                     case "DotNetHostPath":
                         XmlRunSettingsUtilities.ThrowOnHasAttributes(reader);
-                        
                         string? dotnetHostPath = reader.ReadElementContentAsString();
 
 #if !NETSTANDARD1_0
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -891,7 +891,14 @@ public class RunConfiguration : TestRunSettings
 
                     case "DotNetHostPath":
                         XmlRunSettingsUtilities.ThrowOnHasAttributes(reader);
-                        runConfiguration.DotnetHostPath = reader.ReadElementContentAsString();
+                        
+                        string? dotnetHostPath = reader.ReadElementContentAsString();
+
+#if !NETSTANDARD1_0
+                        dotnetHostPath = Environment.ExpandEnvironmentVariables(dotnetHostPath);
+#endif
+
+                        runConfiguration.DotnetHostPath = dotnetHostPath;
                         break;
                     case "TreatNoTestsAsError":
                         XmlRunSettingsUtilities.ThrowOnHasAttributes(reader);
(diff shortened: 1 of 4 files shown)


src/Microsoft.TestPlatform.Utilities/ClientUtilities.cs
```diff
@@ -43,6 +44,12 @@ public static void FixRelativePathsInRunSettings(XmlDocument xmlDocument, string
         {
             FixNodeFilePath(resultsDirectoryNode, root);
         }
+
+        var dotnetHostPathNode = xmlDocument.SelectSingleNode(DotnetHostPathXPath);
```
Therzok, line 48: Tests are failing because SelectSingleNode is likely case sensitive?

src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
```diff
@@ -891,7 +891,13 @@ public static RunConfiguration FromXml(XmlReader reader)
 
                     case "DotNetHostPath":
```
Therzok, line 892: I think there's a bigger problem at hand here. This uses upper-case `N`, which means that the XML is only valid when it has upper-case N in the setting.

But it sets lower-case N here.

src/Microsoft.TestPlatform.Utilities/ClientUtilities.cs
Evangelink: No test is failing on CI. Is there any change you would have forgot to push?

Evangelink:
@Therzok Could you add some tests (as mentioned by @nohwnd)?

Therzok:
Yes, sorry, I completely forgot about the PR :D

Evangelink:
No worries and thanks for the remaining work.

src/Microsoft.TestPlatform.Utilities/ClientUtilities.cs
Therzok: Check previous CI runs, the problem was with the casing. If I used `<RunConfiguration><DotnetHostPath>` instead of `DotNetHostPath` it would break.

Therzok:
I've modified an existing test to handle this scenario.

src/Microsoft.TestPlatform.Utilities/ClientUtilities.cs
Evangelink: Thanks! We are planning on improving the way we read/write values in runsettings so I guess this could be handled here.

Evangelink (approved):
LGTM! @nohwnd For a second review.

nohwnd:
Probably an edge case, but would you expect environment variables that are defined in the runsettings (or from commandline via `-e`) to be also usable and expandable here?

Therzok:
Didn't even know that's possible. I just followed the logic that applied to `ResultsDirectory`, and made it also work on `DotNetHostPath` :D

The way I wanted to integrate it is via `msbuild` setting the environment variable (some common props file) from a test project, so it could support different nesting inside a solution directory layout.

```xml
<!-- this is imported via some shared props -->
<DotNetHostPath>build/bin/App</DotNetHostPath>
```

```
src
- Implementation
- - Implementation.Tests
- Tests
```

Currently, with the relative path, if you use the solution as an argument to `dotnet test` it works, but when you pass in a project, it breaks, as it tries to probe into directories relative to the project.

nohwnd (approved):

nohwnd:
Let's leave that suggestion for later then.

PR #3860, opened 2022-07-18
CA1001: Types that own disposable fields should be disposable

Evangelink:


src/vstest.console/Internal/ProgressIndicator.cs
```diff
@@ -115,4 +116,10 @@ private void Timer_Elapsed(object sender, System.Timers.ElapsedEventArgs e)
             }
         }
     }
+
+    public void Dispose()
+    {
+        _timer?.Dispose();
+        _timer = null;
```
MarcoRossignoli, line 123: not needed

MarcoRossignoli, line 123:
```suggestion
```

MarcoRossignoli (approved):

src/vstest.console/Internal/ProgressIndicator.cs
Evangelink: Arg another 3h of waiting for CI.... Thanks for the suggestion!

PR #3873, opened 2022-07-20
Fix fakes

nohwnd:
## Description
Pass the detected frameworks together with sources to Fakes. 

Fix [AB#1557193](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1557193)

### DIFF
--- a/src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
+++ b/src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
@@ -250,7 +250,7 @@ static Microsoft.VisualStudio.TestPlatform.Common.SettingsProvider.SettingsProvi
 static Microsoft.VisualStudio.TestPlatform.Common.SettingsProvider.SettingsProviderExtensionManager.LoadAndInitializeAllExtensions(bool shouldThrowOnError) -> void
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.CancellationTokenExtensions.ThrowTestPlatformExceptionIfCancellationRequested(this System.Threading.CancellationToken token) -> void
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.ExceptionUtilities.GetExceptionMessage(System.Exception? exception) -> string!
-static Microsoft.VisualStudio.TestPlatform.Common.Utilities.FakesUtilities.GenerateFakesSettingsForRunConfiguration(string![]! sources, string! runSettingsXml) -> string!
+static Microsoft.VisualStudio.TestPlatform.Common.Utilities.FakesUtilities.GenerateFakesSettingsForRunConfiguration(System.Collections.Generic.IDictionary<string!, Microsoft.VisualStudio.TestPlatform.ObjectModel.Framework!>! sourceToFrameworkMap, string! runSettingsXml) -> string!
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.RunSettingsUtilities.CreateAndInitializeRunSettings(string? settingsXml) -> Microsoft.VisualStudio.TestPlatform.Common.RunSettings?
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.RunSettingsUtilities.GetMaxCpuCount(Microsoft.VisualStudio.TestPlatform.ObjectModel.RunConfiguration? runConfiguration) -> int
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.RunSettingsUtilities.GetMaxCpuCount(string? settingXml) -> int
--- a/src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
+++ b/src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Xml;
 using System.Xml.XPath;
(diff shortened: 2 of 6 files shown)


src/Microsoft.TestPlatform.Common/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -250,7 +250,7 @@ static Microsoft.VisualStudio.TestPlatform.Common.SettingsProvider.SettingsProvi
 static Microsoft.VisualStudio.TestPlatform.Common.SettingsProvider.SettingsProviderExtensionManager.LoadAndInitializeAllExtensions(bool shouldThrowOnError) -> void
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.CancellationTokenExtensions.ThrowTestPlatformExceptionIfCancellationRequested(this System.Threading.CancellationToken token) -> void
 static Microsoft.VisualStudio.TestPlatform.Common.Utilities.ExceptionUtilities.GetExceptionMessage(System.Exception? exception) -> string!
-static Microsoft.VisualStudio.TestPlatform.Common.Utilities.FakesUtilities.GenerateFakesSettingsForRunConfiguration(string![]! sources, string! runSettingsXml) -> string!
+static Microsoft.VisualStudio.TestPlatform.Common.Utilities.FakesUtilities.GenerateFakesSettingsForRunConfiguration(System.Collections.Generic.IDictionary<string!, Microsoft.VisualStudio.TestPlatform.ObjectModel.Framework!>! sourceToFrameworkMap, string! runSettingsXml) -> string!
```
Evangelink, line 253: 💭 Just some open question: Could we make some internals visible to for Test Window if we except them to be the only consumers rather than having something public? It's always weird to me to see us doing "breaking change" in the middle of a version.

test/vstest.console.UnitTests/CommandLine/GenerateFakesUtilitiesTests.cs
```diff
@@ -37,7 +40,7 @@ public void FakesShouldNotBeGeneratedIfDisableAutoFakesSetToTrue()
         CommandLineOptions.Instance.DisableAutoFakes = true;
```
Evangelink, line 40: Could we add a test for when disable is false?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -142,32 +145,24 @@ internal static void InsertOrReplaceFakesDataCollectorNode(XmlDocument runSettin
         XmlRunSettingsUtilities.InsertDataCollectorsNode(runSettings.CreateNavigator()!, settings);
     }
 
-    private static IDictionary<string, FrameworkVersion> CreateDictionary(IEnumerable<string> sources, FrameworkVersion framework)
+    private static IDictionary<string, FrameworkVersion> ConvertToFrameworkVersion(IDictionary<string, Framework> sourceToFrameworkMap)
```
Evangelink, line 148: What about `ConvertToSourceToFrameworkVersion`?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
```diff
@@ -34,9 +35,9 @@ public static class FakesUtilities
     /// <param name="sources">test sources</param>
     /// <param name="runSettingsXml">runsettings</param>
     /// <returns>updated runsettings for fakes</returns>
-    public static string GenerateFakesSettingsForRunConfiguration(string[] sources, string runSettingsXml)
+    public static string GenerateFakesSettingsForRunConfiguration(IDictionary<string, Framework> sourceToFrameworkMap, string runSettingsXml)
```
Evangelink, line 38: What about simply `sourceToFramework`? I am not a big fan of "map" suffix in C# (feels Java). If you really need the suffix, I would vote for `Dictionary`.

test/vstest.console.UnitTests/CommandLine/GenerateFakesUtilitiesTests.cs
nohwnd: @vritant24 could you? :)

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
nohwnd: +Map?

src/Microsoft.TestPlatform.Common/Utilities/FakesUtilities.cs
nohwnd: too late, I already used in in many other places, and honestly, reading sourceToFramework, does not give me the same understanding as reading sourceToFrameworkMap, or sourceToFrameworkDictionary

Evangelink: Not too late to not introduce more `map` instead of `dictionary`, PR isn't yet merged :)

test/vstest.console.UnitTests/CommandLine/GenerateFakesUtilitiesTests.cs
vritant24: Yup. I'm on it :)

vritant24: Should I just push it to this branch?

vritant24: Ah I now remember why we can't have that test. The code under test dynamically loads an assembly through reflection that isn't present in the test env. So testing a scenario where things should work won't be possible unless:
1. We introduce a nuget package that ships the assembly needed and reference it with the test project
2. We fake the static method `FakesUtilities.GenerateFakesSettingsForRunConfiguration` to not call into its logic and instead do something else.
3. We let it be.
Thoughts?

MarcoRossignoli:
Should we close? Looks stale

Evangelink:
@nohwnd Is this PR meant to be completed and merged or shall we delete it?

PR #3879, opened 2022-07-21
Make Microsoft.NET.Test.Sdk package transitive

### LINKED ISSUE #3227: Microsoft.NET.Test.Sdk package is not transitive (request for moving content from `build` to `buildTransitive`)
> Even when we use `PrivateAssets="contentfiles;analyzers"` it works only for one hierarchy layer (`TestSDKNotTransitive_WithPrivateAssets` project `B` works correctly but `C` does not )

[TestSDKNotTransitive.zip](https://github.com/microsoft/vstest/files/7762106/TestSDKNotTransitive.zip)
[TestSDKNotTransitive_WithPrivateAssets.zip](https://github.com/microsoft/vstest/files/7762107/TestSDKNotTransitive_WithPrivateAssets.zip)

Failed to run as a self-contained app.
 > - The application was run as a self-contained app because 'C:\Users\nedav\source\repos\TestSDKNotTransitive\B\bin\Debug\net5.0\testhost.runtimeconfig.json' was not found.
>  - If this should be a framework-dependent app, add the 'C:\Users\nedav\source\repos\TestSDKNotTransitive\B\bin\Debug\net5.0\testhost.runtimeconfig.json' file and specify the appropriate framework.

Evangelink:
Fixes #3227

I have tested locally with the project linked in the issue that wasn't working before and is now working. 

I have also checked that the testhost and other relevant files are correctly copied to the output.

Hopefully there is no unanticipated bad behaviour.

### DIFF
--- a/src/package/nuspec/Microsoft.NET.Test.Sdk.nuspec
+++ b/src/package/nuspec/Microsoft.NET.Test.Sdk.nuspec
@@ -44,15 +44,15 @@
   <files>
     <file src="Icon.png" target="" />
     <file src="licenses\LICENSE_NET.txt" target="" />
-    <file src="netcoreapp\*" target="build\netcoreapp1.0\" />
-    <file src="netcoreapp\*" target="build\netcoreapp2.1\" />
-    <file src="netfx\*" target="build\net462\" />
+    <file src="netcoreapp\*" target="buildTransitive\netcoreapp1.0\" />
+    <file src="netcoreapp\*" target="buildTransitive\netcoreapp2.1\" />
+    <file src="netfx\*" target="buildTransitive\net462\" />
 
     <file src="Microsoft.NET.Test.Sdk.props" target="buildMultiTargeting\" />
-    <file src="Microsoft.NET.Test.Sdk.props" target="build\netcoreapp1.0\" />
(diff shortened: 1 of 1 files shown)


Haplois (changes requested):
Please check compatibility with 16.11 version of VS too.

Evangelink:
It works well
image

Haplois (approved):

jonpryor:
This change causes #4098; for certain projects, it means that simply updating the Microsoft.NET.Test.Sdk can cause the project to stop building.

Those "certain projects" are projects which provide their own `Main()` method and reference Microsoft.NET.Test.Sdk indirectly via transitive dependencies.

Should this be considered a "breaking" change, insofar as updating the Microsoft.NET.Test.Sdk package can *cause* build failures?

What is the best way to resolve this?  Document this change somewhere?  (NuGet.org doesn't have a link for "release notes", so I'm not sure how anyone would *find* any such documentation, but it's a least a plausible answer?)

Make inclusion of e.g. `Microsoft.NET.Test.Sdk.Program.cs` *conditional* on `'$(OutputType)' != 'Exe'`?

PR #3883, opened 2022-07-21
Use net7.0 TargetFramework during source-build

### LINKED ISSUE #3846: Update projects to build net7.0 TFM
Source build requires projects to build with the current TFM or a supported netstandard TFM. vstest currently targets net6.0. Ideally this would be updated to net7.0. If that is not possible, the TFM could be conditioned for source-build but ideally the regular and source build would be identical.

cc @Haplois 

[AB#1568715](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1568715)

lbussell:
Fixes https://github.com/microsoft/vstest/issues/3846. There are still some build failures here but I'm working on this. @Haplois if you have any input on what to change I would appreciate it.


Haplois:
Closing in favor of #3944.

PR #3896, opened 2022-07-25
Set discovery batch size to 1000

nohwnd:
## Description
Measures show that 1000 for discovery size makes it run faster because there is less contention when reporting the test results in parallel. This is because we report through a single channel and so there are less times threads run into each other and need to wait to send back their results.

We also don't want to push the results into a queue because then we have no backpressure.

There no risk that this value will break stuff, if we fail discovering tests we throw the whole dll results away anyway, no matter if we missed 1 tests or 1000 tests from the results.

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
+++ b/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
@@ -24,6 +24,7 @@ namespace Microsoft.VisualStudio.TestPlatform.Utilities;
 public class InferRunSettingsHelper
 {
     private const string DesignModeNodeName = "DesignMode";
+    private const string BatchSizeNodeName = "BatchSize";
     private const string CollectSourceInformationNodeName = "CollectSourceInformation";
     private const string RunSettingsNodeName = "RunSettings";
     private const string RunConfigurationNodeName = "RunConfiguration";
@@ -33,6 +34,7 @@ public class InferRunSettingsHelper
     private const string TargetDevice = "TargetDevice";
 
     private const string DesignModeNodePath = @"/RunSettings/RunConfiguration/DesignMode";
+    private const string BatchSizeNodePath = @"/RunSettings/RunConfiguration/BatchSize";
     private const string CollectSourceInformationNodePath = @"/RunSettings/RunConfiguration/CollectSourceInformation";
     private const string RunConfigurationNodePath = @"/RunSettings/RunConfiguration";
     private const string TargetPlatformNodePath = @"/RunSettings/RunConfiguration/TargetPlatform";
@@ -202,6 +204,16 @@ public class InferRunSettingsHelper
     public static void UpdateDesignMode(XmlDocument runSettingsDocument, bool designModeValue)
     {
         AddNodeIfNotPresent(runSettingsDocument, DesignModeNodePath, DesignModeNodeName, designModeValue);
+    }
+
+    /// <summary>
+    /// Updates the <c>RunConfiguration.BatchSize</c> value for a run settings. Doesn't do anything if the value is already set.
+    /// </summary>
+    /// <param name="runSettingsDocument">Document for runsettings xml</param>
+    /// <param name="batchSizeValue">Value to set</param>
+    public static void UpdateBatchSize(XmlDocument runSettingsDocument, long batchSizeValue)
+    {
+        AddNodeIfNotPresent(runSettingsDocument, BatchSizeNodePath, BatchSizeNodeName, batchSizeValue);
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -204,6 +206,16 @@ public static void UpdateDesignMode(XmlDocument runSettingsDocument, bool design
         AddNodeIfNotPresent(runSettingsDocument, DesignModeNodePath, DesignModeNodeName, designModeValue);
     }
 
+    /// <summary>
+    /// Updates the <c>RunConfiguration.BatchSize</c> value for a run settings. Doesn't do anything if the value is already set.
+    /// </summary>
+    /// <param name="runSettingsDocument">Document for runsettings xml</param>
+    /// <param name="batchSizeValue">Value to set</param>
+    public static void UpdateBatchSize(XmlDocument runSettingsDocument, long batchSizeValue)
```
Evangelink, line 214: Seems more like `SetBatchSize` or `AddBatchSize` because it doesn't update the value (it does nothing if already set).

Evangelink, line 214: It could be nice to return a bool to say if the value was actually set or not.

src/vstest.console/CommandLine/CommandLineOptions.cs
```diff
@@ -24,10 +24,15 @@ namespace Microsoft.VisualStudio.TestPlatform.CommandLine;
 internal class CommandLineOptions
 {
     /// <summary>
-    /// The default batch size.
+    /// The default batch size for Run.
     /// </summary>
     public const long DefaultBatchSize = 10;
```
Evangelink, line 29: Maybe you could rename it to `DefaultRunBatchSize` for consistency.

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -174,13 +174,16 @@ public void ResetOptions()
...
+                isDiscovery: true,
                 out string updatedRunsettings,
                 out IDictionary<string, Architecture> sourceToArchitectureMap,
                 out IDictionary<string, Framework> sourceToFrameworkMap))
         {
             runsettings = updatedRunsettings;
         }
 
+
+
```
Evangelink, line 186: Cosmetic nit:
```suggestion
```

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -909,6 +917,27 @@ private bool UpdateDesignMode(XmlDocument document, RunConfiguration runConfigur
...
 
+    private bool AddOrUpdateBatchSize(XmlDocument document, RunConfiguration runConfiguration, bool isDiscovery)
+    {
+        // On run keep it as is to fall back to the current default value (which is 10 right now).
+        if (!isDiscovery)
+        {
+            // We did not update runnsettings.
+            return false;
+        }
+
```
Evangelink, line 928: Cosmetic nit:

```suggestion
```

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -909,6 +917,27 @@ private bool UpdateDesignMode(XmlDocument document, RunConfiguration runConfigur
         return updateRequired;
     }
 
+    private bool AddOrUpdateBatchSize(XmlDocument document, RunConfiguration runConfiguration, bool isDiscovery)
```
Evangelink, line 920: Can we test the various paths of this method?

src/Microsoft.TestPlatform.Utilities/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -38,3 +38,5 @@ static Microsoft.VisualStudio.TestPlatform.Utilities.MSTestSettingsUtilities.Imp
 static Microsoft.VisualStudio.TestPlatform.Utilities.MSTestSettingsUtilities.IsLegacyTestSettingsFile(string? settingsFile) -> bool
 static Microsoft.VisualStudio.TestPlatform.Utilities.ParallelRunSettingsUtilities.UpdateRunSettingsWithParallelSettingIfNotConfigured(System.Xml.XPath.XPathNavigator! navigator) -> void
 static Microsoft.VisualStudio.TestPlatform.Utilities.StringExtensions.Tokenize(this string? input, char separator, char escape) -> System.Collections.Generic.IEnumerable<string!>!
+static Microsoft.VisualStudio.TestPlatform.Utilities.InferRunSettingsHelper.UpdateBatchSize(System.Xml.XmlDocument! runSettingsDocument, long batchSizeValue) -> void
```
shyamnamboodiripad, line 41: Just curious: Shouldn't the contents of this file be sorted to ensure fewer merge conflicts?

nohwnd: How does that ensure less merge conflicts?

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
Evangelink: Leaving as-is as it's following pattern defined by other methods.

Evangelink: Following the same pattern as other methods on this class so ignoring this remark.

Evangelink (approved):

PR #3904, opened 2022-07-27
Fix verification of signing

Evangelink:


Haplois (approved):
Let's hope this is the last fix.

PR #3919, opened 2022-08-02
VS: move Solution Items folder out of scripts

Evangelink:
Use a more standard structure to reflect how it's done on disk and also how it's done in other projects.


nohwnd:
Why? :)

Evangelink:
> Why? :)

Why not :)

On a more serious note, this is to be more uniform with how it's done on disk and also how it's done on other projects.

nohwnd (approved):

PR #3923, opened 2022-08-03
Console logger splits path using directory and alt directory separators

### LINKED ISSUE #3376: Console logger has a hardcoded path separator for the source
## Description
`vstest` console logger has a hardcoded path separator for the source: https://github.com/microsoft/vstest/blob/f67ad493f406b20de05ec066d8b027c0d7f2a837/src/vstest.console/Internal/ConsoleLogger.cs#L744

## Steps to reproduce
1. Checkout https://github.com/MarinAtanasov/AppBrix
2. Run `./Test.ps -Build`
3. Look at the output in the console

## Expected behavior
On Linux, Windows and MacOS, the output should be the same:
```
Passed!  - Failed:     0, Passed:    20, Skipped:     0, Total:    20, Duration: 77 ms - AppBrix.Tests.dll (net6.0)
```

## Actual behavior
On Linux and MacOS, the output contains the full path to the test assembly:
```
Passed!  - Failed:     0, Passed:    20, Skipped:     0, Total:    20, Duration: 100 ms - /home/runner/work/AppBrix/AppBrix/Tests/AppBrix.Tests/bin/Release/net6.0/AppBrix.Tests.dll (net6.0)
```
The issue will probably be reproduced if Windows is set to use a different path separator.

## Diagnostic logs
Not necessary.

## Environment
I am running this [GitHub action](https://github.com/MarinAtanasov/AppBrix/blob/0.28.0/.github/workflows/build.yml), which is using `ubuntu-latest`, `windows-latest` and `macOS-latest` with `dotnet-version: 6.0.x`

Evangelink:
## Description

Console logger splits path using directory and alt directory separators. I decided to use both separators because I have no idea if we are ensured to receive backslashes for windows paths.

## Related issue

Fixes #3376

### DIFF
--- a/src/vstest.console/Internal/ConsoleLogger.cs
+++ b/src/vstest.console/Internal/ConsoleLogger.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using System.Text;
 
@@ -160,7 +161,6 @@ internal class ConsoleLogger : ITestLoggerWithParameters
     /// tracking counts per source for the minimal and quiet output.
     /// </summary>
     private ConcurrentDictionary<Guid, MinimalTestResult>? LeafTestResults { get; set; }
-
 
     #region ITestLoggerWithParameters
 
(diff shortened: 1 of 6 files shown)


test/Microsoft.TestPlatform.TestUtilities/TestUtils.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.TestPlatform.TestUtilities;
+
+public static class TestUtils
```
nohwnd, line 8: This name seems soo generic that this class will become another "IntegrationTestBase" with thousands of lines of unrelated code. 

Maybe rename it to OSUtils?

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -765,7 +765,7 @@ private void TestRunCompleteHandler(object? sender, TestRunCompleteEventArgs e)
                     : $"({_targetFramework})";
 
                 var duration = GetFormattedDurationString(sourceSummary.Duration);
-                var sourceName = sd.Key.Split('\\').Last();
+                var sourceName = sd.Key.Split(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar).Last();
```
nohwnd, line 768: Could `Path.GetFileName` be used here? It handles both normal and alt separator char. But does not handle UNC paths well. So maybe this is a lower risk option.

It still seems like a one-off solution of something that should be in a "library" and used everywhere. That is different discussion though.

nohwnd (approved):

src/vstest.console/Internal/ConsoleLogger.cs
Evangelink: > But does not handle UNC paths well

Do you know specific cases that are not well handled? Because replacing the code with `Path.GetFileName` seems pretty sensible to me.

PR #3932, opened 2022-08-09
Support test discovery in sources that are directories

shyamnamboodiripad:
At the moment, test platform only supports discovery of tests present inside test sources that are files. With the change in this PR, it can additionally support discovery of tests present inside test sources that are directories. This helps test adapters for certain languages like python which prefer to discover tests at the granularity of a directory as opposed to individual files.

Python adapter currently supplies test containers that are individual .py files to the VS test explorer. However, when these files are sent to the python test discoverer, the discoverer ignores the individual files and always jumps up to the containing project's directory to discover tests present inside all .py files present under this directory.

Python adapter has some reasons for this behavior (tickets linked below contain more context). However, there doesn't seem to be any significant reason why test platform can't support test discovery at a directory level granularity.

This PR relaxes the requirement that supplied test containers (sources) must always be files. It now also allows test containers that are directories. It also introduces a new `[DirectoryBasedDiscoverer]` attribute that can be applied on `ITestDiscoverer`s that only care to process sources that are directories (similar to the existing `[FileExtension]` attribute that can be applied on `ITestDiscoverer`s that only care to process sources that are files with specific extensions).

Related tickets:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1586970
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1560000

### DIFF
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestDiscovererPluginInformation.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestDiscovererPluginInformation.cs
@@ -29,6 +29,7 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
             FileExtensions = GetFileExtensions(testDiscovererType);
             DefaultExecutorUri = GetDefaultExecutorUri(testDiscovererType);
             AssemblyType = GetAssemblyType(testDiscovererType);
+            IsDirectoryBased = GetIsDirectoryBased(testDiscovererType);
         }
     }
 
@@ -39,7 +40,7 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
     {
         get
         {
-            return new object?[] { FileExtensions, DefaultExecutorUri, AssemblyType };
+            return new object?[] { FileExtensions, DefaultExecutorUri, AssemblyType, IsDirectoryBased };
         }
     }
 
@@ -71,15 +72,25 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
     }
 
     /// <summary>
-    /// Helper to get file extensions from the FileExtensionAttribute on the discover plugin.
+    /// <c>true</c> if the discoverer plugin is decorated with <see cref="DirectoryBasedTestDiscovererAttribute"/>,
+    /// <c>false</c> otherwise.
     /// </summary>
-    /// <param name="testDicovererType">Data type of the test discoverer</param>
+    public bool IsDirectoryBased
+    {
+        get;
+        private set;
+    }
+
+    /// <summary>
+    /// Helper to get file extensions from the <see cref="FileExtensionAttribute"/>S on the discover plugin.
+    /// </summary>
+    /// <param name="testDiscovererType">Data type of the test discoverer</param>
     /// <returns>List of file extensions</returns>
-    private static List<string> GetFileExtensions(Type testDicovererType)
+    private static List<string> GetFileExtensions(Type testDiscovererType)
     {
         var fileExtensions = new List<string>();
 
-        var attributes = testDicovererType.GetTypeInfo().GetCustomAttributes(typeof(FileExtensionAttribute), false).ToArray();
+        var attributes = testDiscovererType.GetTypeInfo().GetCustomAttributes(typeof(FileExtensionAttribute), inherit: false).ToArray();
         if (attributes != null && attributes.Length > 0)
         {
             foreach (var attribute in attributes)
@@ -96,15 +107,15 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
     }
 
     /// <summary>
-    /// Returns the value of default executor Uri on this type. 'Null' if not present.
+    /// Returns the value of default executor Uri on this type. <c>null</c> if not present.
     /// </summary>
     /// <param name="testDiscovererType"> The test discoverer Type. </param>
     /// <returns> The default executor URI. </returns>
     private static string GetDefaultExecutorUri(Type testDiscovererType)
     {
-        string result = string.Empty;
+        var result = string.Empty;
 
-        object[] attributes = testDiscovererType.GetTypeInfo().GetCustomAttributes(typeof(DefaultExecutorUriAttribute), false).ToArray();
+        var attributes = testDiscovererType.GetTypeInfo().GetCustomAttributes(typeof(DefaultExecutorUriAttribute), inherit: false).ToArray();
         if (attributes != null && attributes.Length > 0)
         {
             DefaultExecutorUriAttribute executorUriAttribute = (DefaultExecutorUriAttribute)attributes[0];
@@ -119,7 +130,7 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
     }
 
     /// <summary>
-    /// Helper to get the supported assembly type from the CategoryAttribute on the discover plugin.
+    /// Helper to get the supported assembly type from the <see cref="CategoryAttribute"/> on the discover plugin.
     /// </summary>
     /// <param name="testDiscovererType"> The test discoverer Type. </param>
     /// <returns> Supported assembly type. </returns>
@@ -133,5 +144,16 @@ internal class TestDiscovererPluginInformation : TestPluginInformation
         // Get assembly type from category.
         Enum.TryParse(category, true, out AssemblyType assemblyType);
         return assemblyType;
+    }
+
+    /// <summary>
+    /// Returns <c>true</c> if the discoverer plugin is decorated with
+    /// <see cref="DirectoryBasedTestDiscovererAttribute"/>, <c>false</c> otherwise.
+    /// </summary>
+    /// <param name="testDiscovererType">Data type of the test discoverer</param>
+    private static bool GetIsDirectoryBased(Type testDiscovererType)
+    {
+        var attribute = testDiscovererType.GetTypeInfo().GetCustomAttribute(typeof(DirectoryBasedTestDiscovererAttribute), inherit: false);
+        return attribute is DirectoryBasedTestDiscovererAttribute;
     }
 }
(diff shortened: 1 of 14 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscoveryManager.cs
```diff
@@ -255,7 +255,7 @@ internal static HashSet<string> GetValidSources(IEnumerable<string>? sources, IM
             // It is possible that runtime provider sent relative source path for remote scenario.
             string src = !Path.IsPathRooted(source) ? Path.Combine(Directory.GetCurrentDirectory(), source) : source;
 
-            if (!File.Exists(src))
+            if (!File.Exists(src) && !Directory.Exists(src))
```
shyamnamboodiripad, line 258: Note: This is the crux of the change. The rest of the PR is basically plumbing for the new `[DirectoryBasedTestDiscoverer]` attribute.

src/Microsoft.TestPlatform.ObjectModel/DirectoryBasedDiscovererAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+/// This attribute is applied to <see cref="ITestDiscoverer"/>s. It indicates the test discoverer discovers tests
+/// present inside a directory (as opposed to the <see cref="FileExtensionAttribute"/> which indicates that the
+/// discoverer discovers tests present in files with a specified extension).
+/// </summary>
+/// <remarks>
+/// If neither this attribute nor the <see cref="FileExtensionAttribute"/> is provided on the test discoverer,
+/// it will be called for all relevant test files and directories.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public sealed class DirectoryBasedTestDiscovererAttribute : Attribute
```
AbhitejJohn, line 18: We probably already thought of this but what is our recommendation for adapter authors that use this and are worried about compat with older VS/vstest versions that might not have this attribute defined in the object model?

shyamnamboodiripad: That's a good point @AbhitejJohn. This would not an issue for the python adapter specifically since it ships as part of VS (VSIX-based).

Otherwise, my assumption was that the attribute will be ignored by older versions of TP (which don't know about its existence and therefore won't look for it / try to reflection load it). I will try and test this out.

shyamnamboodiripad: Adapter authors also have the choice to leave out this attribute entirely. In that case, the adapter would be invoked for all sources (not just directories), and it would be up to the adapter to filter out directories in the supplied list...

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscovererEnumerator.cs
```diff
@@ -371,32 +371,62 @@ private static void SetAdapterLoggingSettings(IMessageLogger messageLogger, IRun
         var result = new Dictionary<LazyExtension<ITestDiscoverer, ITestDiscovererCapabilities>, IEnumerable<string>>();
         var sourcesForWhichNoDiscovererIsAvailable = new List<string>(sources);
 
+        sources = sources.Distinct().ToList();
```
MarcoRossignoli, line 374:
```suggestion
        sources = sources.Distinct();
```

MarcoRossignoli: I think it's not needed IEnumerable should be accepted by `Where` etc...

src/Microsoft.TestPlatform.ObjectModel/DirectoryBasedDiscovererAttribute.cs
MarcoRossignoli: @nohwnd does it work with all the "above" part related to inferring the host?

MarcoRossignoli:
@nohwnd does it work with all the "above" part related to inferring the host?

src/Microsoft.TestPlatform.ObjectModel/DirectoryBasedDiscovererAttribute.cs
nohwnd: The compatibility is probably the same as for any other adapter, once you use this latest version of ObjectModel and use any new feature from it, you won't be able to load into older version of VS for .NET Framework tests. (I am writing a proposal to mitigate this.).

nohwnd: Luckily the python adapter ships with VS (as shyam said above, so the version is fixed for .NET Framework, and can be updated for .NET Core by updating the version of Test.SDK.

src/Microsoft.TestPlatform.Common/ExtensionFramework/Utilities/TestDiscovererPluginInformation.cs
```diff
@@ -71,15 +72,25 @@ public AssemblyType AssemblyType
...
     /// </summary>
-    /// <param name="testDicovererType">Data type of the test discoverer</param>
+    public bool IsDirectoryBased
+    {
+        get;
+        private set;
+    }
+
+    /// <summary>
+    /// Helper to get file extensions from the <see cref="FileExtensionAttribute"/>S on the discover plugin.
```
nohwnd, line 85:
```suggestion
    /// Helper to get file extensions from the <see cref="FileExtensionAttribute"/> on the discover plugin.
```

nohwnd:
Thinking about this a bit more:

The reason this capability just (accidentally) works, is because we provide a default runtime environment for non-dll sources. The hidden assumption here is that we will start a testhost.exe to run the adapter, which will then do the work on their own or delegate to additional executable (e.g. xmlTests can be handled in process, but .py tests need to be delegated to python). And we also assume that the provided file type represents only one type of runtime, or that the adapter can detect the correct runtime and start it itself (e.g. .py can be started with python2.exe or python3.exe based on metadata in the file, or if someone would want to implement this for usage with dlls, they would have to multiplex to .NET, .NET Framework, UWP, WinUI and Native).


I guess this brings us back to discussion I think we had in the past: We use the sources to indirectly determine which TestRuntime provider should be used. In runner we check TargetFramework, and if not specified, we auto-detect TargetFramework by looking at the dll metadata. TargetFramework is then used by TestRuntime providers to determine if they can run the workload or not. 

So while it is not blocking for this PR, it would be nice to determine a way to provide additional metadata on the source level, for us to skip some of those checks. Especially in cases where VS possibly has all the info already e.g. what dll type this is, what dlls the project references, and if there are any well known adapters among them.

src/Microsoft.TestPlatform.CrossPlatEngine/Discovery/DiscovererEnumerator.cs
shyamnamboodiripad: Ah the reason was to ensure that the two where queries below operate on a realized collection with distinct elements (so that the `Distinct` call does not need to be repeated when below queries operate on `sources`). Similar reason for the queries in the below variables too...

MarcoRossignoli: ah ok so allocation vs computation...it's ok.

shyamnamboodiripad:
@nohwnd Totally agree that it would be great if source could be an open-ended concept. This could open up other interesting scenarios where the source used need not be present on disk at all (for example, it could be a URL to some online resource which serves the appropriate tests and which the adapter knows to interpret).

I don't think it is bad to have a reasonable default when a target runtime is not specified. But it would certainly be nice if there was a way to provide additional metadata about target runtimes. In cases where adapter knows it will be the same runtime for all sources (like python), perhaps the adapter could specify this declaratively via an attribute. For other cases (especially dll-based) VS could provide info about source-level target runtimes via new API / runsettings. Iirc, VS already provides the TargetFramework, does it not?

Detecting precise adapter references for a project could be done - but this may be a tricky change as it can impact backwards compatibility. Currently VS detects and passes all adapters referenced in the solution regardless of the set of adapters actually referenced by the tests being executed. But changing this would also prevent some redundant sniffs in the adapters - so this would also be nice to optimize down the line. Perhaps VS can have a special mode where projects opt into the new behavior where it would only pass relevant adapters to TP...

src/Microsoft.TestPlatform.ObjectModel/DirectoryBasedDiscovererAttribute.cs
AbhitejJohn: Sounds good, the only other thing I can think of is graceful failure for adapters that do use new functionality but on older versions that do not support it. At the moment, I think they wouldn't even load failing to discover or execute any tests but perhaps @nohwnd's proposal would discuss what our recommendation there would be.

nohwnd (approved):

nithinbandaru1:
Does this even work? @nohwnd

By the way it makes sense that there can be a Python Test Adaptor that works with VS Test as it used JSON based socket protocol. Probably that is part of python SDK that comes with visual studio? Good to know.

When I give directory instead of DLL path, I still get error.

Command:

> D:\prototype> "C:\Program Files\Microsoft Visual Studio\2022\Professional\Common7\IDE\Extensions\TestPlatform\vstest.console.exe" Smpl.Calc.Lib.UT.NuTest\bin\Debug\net8.0\

Error:

> Microsoft (R) Test Execution Command Line Tool Version 17.10.0 (x64)
> Copyright (c) Microsoft Corporation.  All rights reserved.
>
>The test source file "D:\prototype\Smpl.Calc.Lib.UT.NuTest\bin\Debug\net8.0\" provided was not found.

nohwnd:
Yes the python adapter can be installed with VS.

nohwnd:
@nithinbandaru1 I just had a deeper look on this as I am writing some docs. The experience works via the programmatic api (TranslationLayer), but does not work in commandline as you observed. There the directory fails validation. 

It is used by PythonPyTestAdapter (one of the test adapters for python). 

Adapter path needs to be specified in runsettings if the adapter is not built in, because otherwise it cannot be found "next" to sources.

PR #3946, opened 2022-08-19
Fix stack overflow in FilterExpression.ValidForProperties

### LINKED ISSUE #3905: Test process crashes with stack overflow in FilterExpression.ValidForProperties
On the mac machines used by helix, with a long test case filter in an RSP file the `vstest.console.dll` test process will crash with a stack overflow in recursive code.  See below for stack trace.

The same rsp file (and test filter) works on windows.

This was consistently reproducible for me on arcade helix OSX machines.  I don't have another mac to test on.
```
Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.15
 OS Platform: Darwin
 RID:         osx.10.15-x64
```

1.  Build roslyn on the branch https://github.com/dibarbet/roslyn/tree/partition_tests 
2.  Move to the artifacts/bin folder.
3.  Invoke vstest.console.dll on the rsp file directly via `../../.dotnet/dotnet exec ../../.dotnet/sdk/7.0.100-preview.6.22352.1/vstest.console.dll @mac_rsp.rsp`  See attached file.
4.  Expect tests all tests to run (failure isn't necessarily a problem).

RSP file is here (you'll have to rename to .rsp, github wouldn't let me upload without changing the extension)
[mac_rsp.txt](https://github.com/microsoft/vstest/files/9204482/mac_rsp.txt)

All tests specified in the filter in the rsp file should run.

On mac, the test process crashes with
```
Data collector 'Blame' message: All tests finished running, Sequence file will not be generated.
Blame: Attaching crash dump utility to process dotnet (47440).
[createdump] Gathering state for process 47440 
[createdump] Crashing thread 0009c4af signal 00000006
[createdump] Writing full dump to file /tmp/helix/working/B0BD09B7/t/54124976-53fb-4331-9ba0-5b858a159ae4/dotnet_47440_1658882986_crashdump.dmp
[createdump] Written 2474930520 bytes (604231 pages) to core file
[createdump] Dump successfully written
The active test run was aborted. Reason: Test host process crashed : Stack overflow.
Repeat 6491 times:
--------------------------------
   at Microsoft.VisualStudio.TestPlatform.Common.Filtering.FilterExpression.ValidForProperties(System.Collections.Generic.IEnumerable`1<System.String>, System.Func`2<System.String,Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProperty>)
--------------------------------
   at Microsoft.VisualStudio.TestPlatform.Common.Filtering.FilterExpressionWrapper.ValidForProperties(System.Collections.Generic.IEnumerable`1<System.String>, System.Func`2<System.String,Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProperty>)
   at Microsoft.VisualStudio.TestPlatform.Common.Filtering.TestCaseFilterExpression.ValidForProperties(System.Collections.Generic.IEnumerable`1<System.String>, System.Func`2<System.String,Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProperty>)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Discovery.DiscoveryContext.GetTestCaseFilter(System.Collections.Generic.IEnumerable`1<System.String>, System.Func`2<System.String,Microsoft.VisualStudio.TestPlatform.ObjectModel.TestProperty>)
   at Xunit.Runner.VisualStudio.TestCaseFilter.GetTestCaseFilterExpression(Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IRunContext, LoggerHelper, System.String, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.ITestCaseFilterExpression ByRef)
   at Xunit.Runner.VisualStudio.TestCaseFilter..ctor(Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IRunContext, LoggerHelper, System.String, System.Collections.Generic.HashSet`1<System.String>)
   at Xunit.Runner.VisualStudio.VsTestRunner.RunTestsInAssembly(Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IRunContext, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IFrameworkHandle, LoggerHelper, Xunit.Runner.VisualStudio.TestPlatformContext, Xunit.Runner.VisualStudio.RunSettings, Xunit.IMessageSinkWithTypes, Xunit.Runner.VisualStudio.AssemblyRunInfo)
   at Xunit.Runner.VisualStudio.VsTestRunner+<>c__DisplayClass20_0.<RunTests>b__4(Xunit.Runner.VisualStudio.AssemblyRunInfo)
   at System.Collections.Generic.List`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].ForEach(System.Action`1<System.__Canon>)
   at Xunit.Runner.VisualStudio.VsTestRunner.RunTests(Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IRunContext, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IFrameworkHandle, LoggerHelper, Xunit.Runner.VisualStudio.TestPlatformContext, Xunit.Runner.VisualStudio.RunSettings, System.Func`1<System.Collections.Generic.List`1<Xunit.Runner.VisualStudio.AssemblyRunInfo>>)
   at Xunit.Runner.VisualStudio.VsTestRunner.Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.ITestExecutor.RunTests(System.Collections.Generic.IEnumerable`1<System.String>, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IRunContext, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IFrameworkHandle)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.RunTestsWithSources.InvokeExecutor(Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.Utilities.LazyExtension`2<Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.ITestExecutor,Microsoft.VisualStudio.TestPlatform.Common.Interfaces.ITestExecutorCapabilities>, System.Tuple`2<System.Uri,System.String>, Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Adapter.RunContext, Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.IFrameworkHandle)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.BaseRunTests.RunTestInternalWithExecutors(System.Collections.Generic.IEnumerable`1<System.Tuple`2<System.Uri,System.String>>, Int64)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.BaseRunTests.RunTestsInternal()
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.BaseRunTests.RunTests()
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.ExecutionManager.StartTestRun(System.Collections.Generic.Dictionary`2<System.String,System.Collections.Generic.IEnumerable`1<System.String>>, System.String, System.String, Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.ClientProtocol.TestExecutionContext, Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine.ITestCaseEventsHandler, Microsoft.VisualStudio.TestPlatform.ObjectModel.Client.ITestRunEventsHandler)
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestHandler+<>c__DisplayClass30_3.<OnMessageReceived>b__3()
   at Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.TestRequestHandler+<>c.<.ctor>b__18_1(System.Action)
   at Microsoft.VisualStudio.TestPlatform.Utilities.JobQueue`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].SafeProcessJob(System.__Canon)
   at Microsoft.VisualStudio.TestPlatform.Utilities.JobQueue`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].BackgroundJobProcessor()
   at Microsoft.VisualStudio.TestPlatform.Utilities.JobQueue`1[[System.__Canon, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].<.ctor>b__12_0()
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
   at System.Threading.Thread.StartCallback()
```

The same rsp file and tests on windows run correctly.

** !!IMPORTANT NOTE!! ** - On the same mac machines I was able to get the tests to pass by removing all tests after `Microsoft.CodeAnalysis.CSharp.UnitTests.GreenNodeTests.TestCastExpressionIdentityRewriter` in the TestCaseFilter.  It isn't exact, but roughly more than that will cause a crash.  Removing the same amount from the front of the TestCaseFilter list also works, so it seems like removing about ~534 tests from the filter prevents the stack overflow.

Wasn't able to get these - can try again if needed.

Helix logs - https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-roslyn-refs-pull-62138-merge-ad2c6705f39a472f8e/Microsoft.CodeAnalysis.CSharp.Symbol.UnitTests_Microsoft.CodeAnalysis.CSharp.Syntax.UnitTests_Microsoft.CodeAnalysis.Rebuild.UnitTests_20/1/console.9ed51a23.log?helixlogtype=result

This only reproduced on mac (and was consistent across the mac machines helix uses for tests).  But it may be specific to that model, I do not have another mac to test on
This was consistently reproducible for me on arcade helix OSX machines.  I don't have another mac to test on.

Helix machines:
```
Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.15
 OS Platform: Darwin
 RID:         osx.10.15-x64
```

[AB#1591786](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1591786)

engyebrahim:
## Description
fixed the stack overflow in FilterExpression.ValidForProperties by replacing the recursion by iterative code (stacks and loops)

## issue
 - Fixes #3905


nohwnd:
@engyebrahim  please add description to this PR, and prefix the linked issue by Fix or Fixes if you want to automatically close it. If the PR is still in progress please make it a draft.
image

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
-        bool valid = false;
-        if (_condition != null)
+        FilterExpression? current = this;
+        Stack<FilterExpression> filterStack = new();
+        Stack<string[]?> result = new();
+        do
         {
-            valid = _condition.ValidForProperties(properties, propertyProvider);
-            if (!valid)
+            //Push root's right child and then root to stack then Set root as root's left child
```
nohwnd, line 135:
```suggestion
            // Push root's right child and then root to stack then Set root as root's left child.
```

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +303,49 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+        Stack<bool> result = new();
+        do
         {
-            // & or | operator
-            bool leftResult = _left!.Evaluate(propertyValueProvider);
-            bool rightResult = _right!.Evaluate(propertyValueProvider);
-            filterResult = _areJoinedByAnd ? leftResult && rightResult : leftResult || rightResult;
-        }
-        return filterResult;
+            //Push root's right child and then root to stack then Set root as root's left child
```
nohwnd, line 312:
```suggestion
            // Push root's right child and then root to stack then Set root as root's left child.
```

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +303,49 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+            while (current != null)
+            {
+                if (current._right != null)
+                {
+                    filterStack.Push(current._right);
+                }
+                filterStack.Push(current);
+                current = current._left;
+            }
+            //If the popped item has a right child and the right child is at top of stack, then remove the right child from stack, push the root back and set root as root's right child.
```
nohwnd, line 322:
```suggestion
            // If the popped item has a right child and the right child is at top of stack, then remove the right child from stack, push the root back and set root as root's right child.
```

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +303,49 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+                {
+                    // & or | operator
+                    bool rightResult = current._right != null ? result.Pop() : false;
+                    bool leftResult = current._left != null ? result.Pop() : false;
+                    filterResult = current._areJoinedByAnd ? leftResult && rightResult : leftResult || rightResult;
+                }
+                result.Push(filterResult);
+                current = null;
+            }
+        } while (filterStack.Count() > 0);
```
nohwnd, line 347:
```suggestion
        } while (filterStack.Count > 0);
```

nohwnd:
The filtering logic is now a bit more unclear than it was before. And even before I did not really get fully how it works. Now that you probably have understanding of what is going on, could you please add comments with explanation of how it works to the code? 

Could you also add a test with a very long filter that would otherwise stack overflow please?

nohwnd:
Even though you could theoretically measure how many iterations it would take I think it is safe to assume that 100k entries test filter should be fast to process, and enough to test that we don't stackoverflow.

(measuring how many recursive calls are possible for a simple method: 

```cs
        static int MeasureStack(int count)
        {
            try
            {
                RuntimeHelpers.EnsureSufficientExecutionStack();
                return MeasureStack(++count);
            } catch (InsufficientExecutionStackException)
            {
                return count;
            }
        }
```
)

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
             }
-        }
-        else
-        {
-            invalidProperties = _left!.ValidForProperties(properties, propertyProvider);
-            var invalidRight = _right!.ValidForProperties(properties, propertyProvider);
-            if (null == invalidProperties)
+            //If the popped item has a right child and the right child is at top of stack, then remove the right child from stack, push the root back and set root as root's right child.
+            current = filterStack.Pop();
+            if (filterStack.Count > 0 && current._right == filterStack.First())
```
Evangelink, line 147: Could we use [`Peek`](https://docs.microsoft.com/dotnet/api/system.collections.stack.peek) instead of `First`? It wasn't clear if `First` was starting from top or bottom of the `Stack`.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
+                filterStack.Push(current);
+                current = current._left;
             }
-        }
-        else
-        {
-            invalidProperties = _left!.ValidForProperties(properties, propertyProvider);
-            var invalidRight = _right!.ValidForProperties(properties, propertyProvider);
-            if (null == invalidProperties)
+            //If the popped item has a right child and the right child is at top of stack, then remove the right child from stack, push the root back and set root as root's right child.
```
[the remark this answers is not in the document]
Evangelink: It could be worth to also split line length so that it's easier to read on GitHub.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
+            current = filterStack.Pop();
+            if (filterStack.Count > 0 && current._right == filterStack.First())
             {
-                invalidProperties = invalidRight;
+                filterStack.Pop();
+                filterStack.Push(current);
+                current = current._right;
             }
-            else if (null != invalidRight)
+            else
```
Evangelink, line 153: That's a personal taste so feel free to ignore me. Could we add a `continue;` to the if and remove the else? The loop is quite big, that's not easy to keep track of where "parts" end.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
+                    }
+                }
+                result.Push(invalidProperties);
+                current = null;
             }
-        }
+        } while (filterStack.Count() > 0);
 
-        return invalidProperties;
+        return result.First(); // the result stack will have one element at the end
```
Evangelink, line 182: Could we add an assertion like `TPDebug.Assert(result.Count == 1, "Result stack should have one element at the end.");`. Not sure if we are supposed to have 1+ or only 1.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +303,49 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+            }
+            //If the popped item has a right child and the right child is at top of stack, then remove the right child from stack, push the root back and set root as root's right child.
+            current = filterStack.Pop();
+            if (filterStack.Count > 0 && current._right == filterStack.First())
+            {
+                filterStack.Pop();
+                filterStack.Push(current);
+                current = current._right;
+            }
+            else
```
Evangelink, line 330: This method is exactly the same as the one above except for the content of this if. I am wondering if it's worth having a private method for the code part and just passing a lambda for this part?

@nohwnd WDYT?

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +303,49 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+                    // & or | operator
+                    bool rightResult = current._right != null ? result.Pop() : false;
+                    bool leftResult = current._left != null ? result.Pop() : false;
+                    filterResult = current._areJoinedByAnd ? leftResult && rightResult : leftResult || rightResult;
+                }
+                result.Push(filterResult);
+                current = null;
+            }
+        } while (filterStack.Count() > 0);
+        return result.First();
```
Evangelink, line 348: I guess this needs the same debug assert as in the other method.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -127,30 +127,59 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
...
-                invalidProperties = invalidRight;
+                filterStack.Pop();
+                filterStack.Push(current);
+                current = current._right;
             }
-            else if (null != invalidRight)
+            else
             {
-                invalidProperties = invalidProperties.Concat(invalidRight).ToArray();
+                bool valid = false;
```
Evangelink, line 155: This is used only in the `if` part so I'd suggest to move the declaration close to the use.

Evangelink (commented):
Could you please add a test that was causing the stack overflow before so that we don't regress.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
nohwnd: Sounds good. It seemed like the stack managing logic could be nice to split from the actual filtering.

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
Haplois: I believe that entire comment is unnecessary, and more complicated than the code :)

src/Microsoft.TestPlatform.Common/Filtering/FilterExpression.cs
```diff
@@ -274,19 +312,24 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
...
+             else
+             {
+                 // & or | operator
+                 bool rightResult = current._right != null ? result.Pop() : false;
+                 bool leftResult = current._left != null ? result.Pop() : false;
+                 filterResult = current._areJoinedByAnd ? leftResult && rightResult : leftResult || rightResult;
+             }
+             return filterResult;
+         });
+
```
Evangelink, line 332: Nit: remove blank line + fix indentation for the curly braces.

```suggestion
        return IterateFilterExpression<bool>((current, result) =>
        {
             bool filterResult = false;
             if (null != current._condition)
             {
                 filterResult = current._condition.Evaluate(propertyValueProvider);
             }
             else
             {
                 // & or | operator
                 bool rightResult = current._right != null ? result.Pop() : false;
                 bool leftResult = current._left != null ? result.Pop() : false;
                 filterResult = current._areJoinedByAnd ? leftResult && rightResult : leftResult || rightResult;
             }
             return filterResult;
        });
```

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -90,6 +90,20 @@ public void FastFilterWithSingleEqualsClause()
...
+    public void AvoidingStackOverFlow()
+    {
+        var testCaseFilter = "Test1";
+        for (int i = 0; i < 1e5; i++)
+        {
+            testCaseFilter += "|Test2";
+        }
+        var filterExpressionWrapper = new FilterExpressionWrapper(testCaseFilter);
+
+        filterExpressionWrapper.ValidForProperties(new List<string>() { "FullyQualifiedName" }, null);
```
Evangelink, line 103: This method returns something, please add an assertion.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -90,6 +90,20 @@ public void FastFilterWithSingleEqualsClause()
         Assert.IsFalse(fastFilter.Evaluate(s => "Test2"));
     }
 
+    [TestMethod]
+    public void AvoidingStackOverFlow()
+    {
+        var testCaseFilter = "Test1";
```
Evangelink, line 96: You are not reusing this variable so you could directly hardcode the value inside `FilterExpressionWrapper` ctor.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -90,6 +90,20 @@ public void FastFilterWithSingleEqualsClause()
         Assert.IsFalse(fastFilter.Evaluate(s => "Test2"));
     }
 
+    [TestMethod]
+    public void AvoidingStackOverFlow()
```
Evangelink, line 94: It would be better to split this into 2 tests (one for each method you are testing) and go for a more "functional" name (e.g. `ValidForPropertiesHandlesBigFilteringExpressions`).

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -90,6 +90,20 @@ public void FastFilterWithSingleEqualsClause()
...
+    {
+        var testCaseFilter = "Test1";
+        for (int i = 0; i < 1e5; i++)
+        {
+            testCaseFilter += "|Test2";
+        }
```
Evangelink, line 100: That doesn't matter much for tests but in general you should consider `+=` string concatenation inside a loop as a bug because you will cause a lot of instantiations. In such case, use a `StringBuilder` and when you are done call `ToString` on it to create your string.

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -94,14 +94,19 @@ public void FastFilterWithSingleEqualsClause()
...
         {
             testCaseFilter.Append("|Test2");
         }
 
         var filterExpressionWrapper = new FilterExpressionWrapper(testCaseFilter.ToString());
-        Assert.IsNull(filterExpressionWrapper.ValidForProperties(new List<string>() { "FullyQualifiedName" }, null));
+        string[]? invalidProperties = filterExpressionWrapper.ValidForProperties(new List<string>() { "FullyQualifiedName" }, null);
+
+        Assert.IsNotNull(invalidProperties);
+        Assert.AreEqual(invalidProperties?.Count(), 1);
```
Evangelink, line 108: I believe you don't need the `?.` because it's supposed to be validated as not null on line above.

Also, you have the `Length` property on an array so it's better than calling `Count()`.

```suggestion
        Assert.AreEqual(invalidProperties.Length, 1);
```

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -94,14 +94,19 @@ public void FastFilterWithSingleEqualsClause()
...
+        Assert.AreEqual(invalidProperties![0], "Category");
```
Evangelink, line 109: As said on previous comment, I don't think you need null validation here.

```suggestion
        Assert.AreEqual(invalidProperties[0], "Category");
```

Evangelink (approved):

test/Microsoft.TestPlatform.Common.UnitTests/Filtering/FastFilterTests.cs
```diff
@@ -90,6 +91,37 @@ public void FastFilterWithSingleEqualsClause()
         Assert.IsFalse(fastFilter.Evaluate(s => "Test2"));
     }
 
+    [TestMethod]
+    public void ValidForPropertiesHandlesBigFilteringExpressions()
+    {
+        StringBuilder testCaseFilter = new("Category=Test1");
+
+        for (int i = 0; i < 1e5; i++)  // creating a 100k filter cases string
```
nohwnd, line 99:
```suggestion
        // Create filter with 100k conditions.
        for (int i = 0; i < 100_000; i++)
```

nohwnd: just a nitpick.

Issue #3939, opened 2022-08-11
17.3.0 Fails in CI run


nohwnd:
TestPlatform 17.3.0 was reported to fail in CI run by #3937 and #3938, the issue seems to b

The issue seems to be that after 17.3.0 introduced the ability to run multiple target frameworks at the same time, we are trying to run .NET Standard libraries which in 17.2.0 would have been excluded.

**The workaround** (and best practice) is to filter your test dlls down to what you actually want to run. 

❌ The default VSTest task pattern is *tests*.dll, which includes all assemblies with *test* in name, potentially including resources, and some TestPlatform assemblies

```
- task: VSTest@2
  inputs:
    testSelector: 'testAssemblies'
    testAssemblyVer2: |
      **\*test*.dll
      !**\*TestAdapter.dll
      !**\obj\**
    searchFolder: '$(System.DefaultWorkingDirectory)'
```

✔️ Instead you should use a pattern specific to your test assembly naming convention. Typically `*Tests.dll`, or `*.Tests.dll` name is used for a dll containing tests. In that case it is sufficient to simply pick that dll from the bin folder: 

```
**\bin\**\*Tests.dll
**\bin\**\*Test.dll
```

✔️ You can also amend the original pattern to exclude more dlls, hopefully including only your desired test dll:
```
**\*test*.dll
!**\obj\**
!**\*.resources.dll
!**\*TestAdapter.dll
!**\*Microsoft.*TestPlatform*.dll
!**\*testhost*.dll
!**\testcentric.engine.metadata.dll
```

✔️ Set VSTEST_DISABLE_MULTI_TFM_RUN=1 to revert to the previous behavior (or close to previous).

nohwnd:
A did a bit more investigation on this last week, there are 2 problems, both of which were mitigated before unde the .NET Framework runner (vstest.console.exe) we ship in Microsoft.TestPlatform package, and that is installed by TestPlatformInstaller (or shipped with VS).

This is specific to the .NET Framework runner because the common determined framework was dependent on the target framework of the runner until 17.2. In 17.3 we don't determine a common framework.

1. .NET Standard dlls are included in the run, and previously they would be given to .NET Framework testhost to execute. The host would not find any test and it would just continue. In the 17.3, they are given to .NET Core testhost which fails to run them because they lack runtimeconfig.json or deps.json.
[...]

PR #3949, opened 2022-08-19
Do not match .NET Standard to Dotnet testhost runner

nohwnd:
## Description
.NET Standard is not executable by .NET runner, but in multi TFM we try to execute it which does not work. Instead we should correctly report that there is no runtime provider for it.

Improve error message when testhost crashes to mention the sources that it tried to run.

## Related issue
Related #3939


src/Microsoft.TestPlatform.CrossPlatEngine/Resources/Resources.resx
```diff
@@ -169,7 +169,8 @@
     <value>Logging TestHost Diagnostics in file: {0}</value>
   </data>
   <data name="TestHostExitedWithError" xml:space="preserve">
-    <value>Testhost process exited with error: {0}. Please check the diagnostic logs for more information.</value>
+    <value>Testhost process for source {0} exited with error: {1}. Please check the diagnostic logs for more information.</value>
```
Evangelink, line 172: If it can be a list of sources, could we have a message that reflects it? Currently we would display something like `Testhost process for source a.cs, b.cs exited with error...`. As you say it's pretty rare, maybe we don't need multiple messages.

What about writing `for source(s)`? Maybe we could even add quotes like `for source(s) '{0}'` and in calling sites, instead of doing `string.Join(", ", sources)` you would do `string.Join("', '", sources)`.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/ProxyExecutionManagerTests.cs
```diff
@@ -309,7 +309,7 @@ public void SetupChannelShouldThrowExceptionIfTestHostExitedBeforeConnectionIsEs
         _mockRequestSender.Setup(s => s.WaitForRequestHandlerConnection(It.IsAny<int>(), It.IsAny<CancellationToken>())).Returns(false);
         _mockTestHostManager.Setup(tmh => tmh.LaunchTestHostAsync(It.IsAny<TestProcessStartInfo>(), It.IsAny<CancellationToken>())).Returns(Task.FromResult(true)).Callback(() => _mockTestHostManager.Raise(t => t.HostExited += null, new HostProviderEventArgs("I crashed!")));
 
-        Assert.AreEqual(string.Format(CultureInfo.CurrentCulture, CrossPlatEngineResources.Resources.TestHostExitedWithError, "I crashed!"), Assert.ThrowsException<TestPlatformException>(() => _testExecutionManager.SetupChannel(new List<string> { "source.dll" }, runsettings)).Message);
+        Assert.AreEqual(string.Format(CultureInfo.CurrentCulture, CrossPlatEngineResources.Resources.TestHostExitedWithError, "source.dll", "I crashed!"), Assert.ThrowsException<TestPlatformException>(() => _testExecutionManager.SetupChannel(new List<string> { "source.dll" }, runsettings)).Message);
```
Evangelink, line 312: Please add a test where we have multiple sources.

src/Microsoft.TestPlatform.CrossPlatEngine/Resources/Resources.resx
nohwnd: done

Evangelink (approved):

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/ProxyExecutionManagerTests.cs
nohwnd: done

PR #3950, opened 2022-08-19
Allow grouping/collapsing HTML test results by class

### LINKED ISSUE #3086: Allow grouping/collapsing HTML test results by class
Currently, the test results HTML is grouped by assembly:

image

By clicking the assembly path we can actually collapse the test results for that assembly:
image

It would be really handy if there were also headers per test class which we could click to expand or collapse the test results for that class.

If you guys are open to this change I would be happy to take a stab at making the appropriate HTML template changes. @nohwnd @pavelhorak

engyebrahim:
## Description
Allow grouping/collapsing HTML test results by class
Capture

## Related issue
Fix https://github.com/microsoft/vstest/issues/3086


nohwnd:
FTFY prefix the issue number with Fix or Fixes if you want to auto close it. If you don't please prefix it with something else, like Partial fix for:

Evangelink (commented):
Is it possible to provide the feature without updating the public API?

PR #3958, opened 2022-08-22
[rel/17.3] Do not match .NET Standard to Dotnet testhost runner

nohwnd:
## Description
.NET Standard is not executable by .NET runner, but in multi TFM we try to execute it which does not work. Instead we should correctly report that there is no runtime provider for it.

Improve error message when testhost crashes to mention the sources that it tried to run.

## Related issue
Related #3939 

Backport of #3949


Evangelink (approved):

MarcoRossignoli:
what's the state here? Are we able to merge?

PR #3966, opened 2022-08-25
Bump version to 17.3.1 and cherry-pick loc files

Evangelink:


Haplois (approved):

Evangelink:
We need to wait as we will need to also cherry-pick #3967

Evangelink:
We now have all required translations. We can move forward with this PR and the servicing release processes.

PR #3969, opened 2022-08-25
Remove TargetLatestRuntimePatch properties

Evangelink:
## Description

Looking at history and after discussion in the team, it seems that this was added only to allow `CheckEolTargetFramework` that wasn't supported on older patches.


nohwnd:
Cool. This will also make #3968 work more reliably. 

The reason why using `TargetLatestRuntimePatch` is not desirable, is that this will force the runtimeconfig.json to specify the latest version that the SDK we built with knows about (from the file mentioned in the comment "tools\dotnet\sdk\<version from global json>\Microsoft.NETCoreSdk.BundledVersions.props" instead of using 3.1.0 or newer, no matter which sdk we built with. Forcing the higher version makes our executables less portable, and makes it more difficult to setup the CI machine.

nohwnd (approved):

PR #3974, opened 2022-08-26
Add some polyfill to simplify code across compilations

Evangelink:
## Description

Add some polyfill to simplify code across compilations. This way we can use only one API but correctly target the recommended way for newer .NET.

### DIFF
--- a/src/Microsoft.TestPlatform.CoreUtilities/Extensions/GuidPolyfill.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/Extensions/GuidPolyfill.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System;
+/// <summary>
+/// A polyfill helper for Guid.
+/// </summary>
+internal static class GuidPolyfill
+{
+    public static Guid Parse(string s, IFormatProvider? provider)
+        => Guid.Parse(s
+#if NET7_0_OR_GREATER
+            , System.Globalization.CultureInfo.InvariantCulture
+#endif
+            );
+}
--- a/src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-#if NET7_0_OR_GREATER
 using System.Globalization;
-#endif
 using System.IO;
 using System.Linq;
 using System.Runtime.Serialization;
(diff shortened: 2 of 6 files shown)


src/Microsoft.TestPlatform.CoreUtilities/Extensions/GuidPolyfill.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System;
+/// <summary>
+/// A polyfill helper for Guid.
+/// </summary>
+internal static class GuidPolyfill
```
Evangelink, line 8: @MarcoRossignoli Was suggesting to have only one class called `Polyfill` that would host all polyfills as sub-classes (e.g. `Polyfill.Guid.Parse`. It looks like a good idea but doesn't allow for extension methods (need to be defined at top class-level).

We can:
1. Keep solution implemented in this PR
2. Go with @MarcoRossignoli's suggestion for non-extension methods (not ideal as it would force us to have 2 ways)
3. Same as 2. but we do that also for extension methods (will no longer be extension methods).

nohwnd: I see theoretical way around it but it requires three classes and instances. Seems like PolyfillGuid or GuidPolyfill in separate class is much easier way to go.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -270,11 +268,7 @@ protected override void ProtectedSetPropertyValue(TestProperty property, object?
                 }
                 else if (value is string guidString)
                 {
-#if NET7_0_OR_GREATER
-                    Id = Guid.Parse(guidString, CultureInfo.InvariantCulture);
-#else
-                    Id = Guid.Parse(guidString);
-#endif
+                    Id = GuidPolyfill.Parse(guidString, CultureInfo.InvariantCulture);
```
MarcoRossignoli, line 271: should the used signature been the "older" one, I mean from user perspective looks like the culture is impotant...but it's important only for one version...so I wonder if we should have the one without culture and setup the correct default for the newer one.

src/Microsoft.TestPlatform.CoreUtilities/Extensions/GuidPolyfill.cs
MarcoRossignoli: it's ok as-is to me.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
Evangelink: That's a good suggestion. I don't see us using anything else than `InvariantCulture` so it would simplify calls. My idea here is to match newest API so that code looks like what you would expect if using recent .NET.

Maybe I will keep it as-is.

MarcoRossignoli (approved):

PR #3981, opened 2022-08-30
Enable usage of datacollectors in playground.

nohwnd:
## Description
Enable usage of datacollectors in playground.

Allow non-detailed output in console.


playground/TestPlatform.Playground/Program.cs
```diff
@@ -69,31 +84,34 @@ static void Main()
...
-            if (process.ExitCode != 0)
-            {
-                throw new Exception($"Process failed with {process.ExitCode}");
-            }
-        }
-        finally
-        {
-            try { File.Delete(settingsFile); } catch { }
-        }
+        // Uncomment when providing command line parameters is easier for you
```
Haplois, line 87:
```suggestion
        // Uncomment if providing command line parameters is easier for you
```

Haplois (approved):

playground/TestPlatform.Playground/Program.cs
nohwnd: They are both equivalent?

PR #3985, opened 2022-09-01
Fix variable name in common.lib.ps1

Haplois:
Issue is on line 388: `$process.remove_ErrorDataReceived($errorDataReceived)`.

Variable name was wrong.


fhnaseer (approved):

nohwnd:
> Variable name was wrong. 

What does it mean it was wrong? Is the name just incorrect in relation to the data it holds?

Haplois:
> What does it mean it was wrong

Original name was `$errorDataReceivedEvent`, but we were trying to remove `$errorDataReceived`.

nohwnd:
> but we were trying to remove `$errorDataReceived`. 

Where? Am I blind?

I only see `$errorDataReceivedEvent` used in definition and in use with `_add` and they both have the same name. 


image

Haplois:
image

PR #3986, opened 2022-09-01
Fix common.lib.ps1 use correct dotnet on ARM64 devices.

Haplois:
Build script updated to work on ARM64 devices natively.

### DIFF
--- a/scripts/common.lib.ps1
+++ b/scripts/common.lib.ps1
@@ -105,12 +105,18 @@ function Install-DotNetCli
     $dotnetInstallPath = Join-Path $env:TP_TOOLS_DIR "dotnet"
     New-Item -ItemType directory -Path $dotnetInstallPath -Force | Out-Null
 
+    $x64Architecture = "x64"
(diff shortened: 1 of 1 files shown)


scripts/common.lib.ps1
```diff
@@ -105,12 +105,18 @@ function Install-DotNetCli
     $dotnetInstallPath = Join-Path $env:TP_TOOLS_DIR "dotnet"
     New-Item -ItemType directory -Path $dotnetInstallPath -Force | Out-Null
 
+    $x64Architecture = "x64"
+    if ($env:PROCESSOR_ARCHITECTURE -eq "ARM64")
+    {
+        $x64Architecture = "arm64"
```
Evangelink, line 111: What about x86, shall we have some similar logic?

Haplois: I don't know, I don't have a device to test it.

Evangelink (approved):

PR #3998, opened 2022-09-08
Fixed wording for Github issue template

cvpoienaru:
Fixed wording for Github issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,5 @@
 ## Description
-> **IMPORTANT**: Kindly refrain from reporting issues that are only reproduced as part of a Visual Studio workflow - please report them using Visual Studio's "Send Feedback" option instead. This option can be accessed either from the Help menu OR by directly using [this link](https://developercommunity.visualstudio.com).
->
-> For a defect reproducible from the vstest command line, describe the issue you have observed here.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,23 +1,23 @@
 ## Description
-> **IMPORTANT**: if the defect is reproduced only in a workflow from within the Visual Studio IDE then do not report the issue here - instead, please report it using Visual Studio's "Send Feedback" option that can be accessed from the Help menu OR using this link https://developercommunity.visualstudio.com.
+> **IMPORTANT**: Kindly refrain from reporting issues that are only reproduced as part of a Visual Studio workflow - please report them using Visual Studio's "Send Feedback" option instead. This option can be accessed either from the Help menu OR by directly using [this link](https://developercommunity.visualstudio.com).
```
Evangelink, line 2: Do you know why we recommend not to use GitHub issues that reproduce only in VS?

Evangelink (approved):

.github/ISSUE_TEMPLATE.md
cvpoienaru: I'm not sure, but it may be for some historic reasons that may not be relevant today. This notice has been in the template forever. Let's discuss which course of action we should take offline.

PR #4003, opened 2022-09-09
Fix ManagedNameHelper to support namespaceless methods.

Haplois:


Evangelink (commented):
Could you please add tests so we don't regress.

test/Microsoft.TestPlatform.AdapterUtilities.UnitTests/ManagedNameUtilities/ManagedNameGeneratorTests.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.TestPlatform.AdapterUtilities.ManagedNameUtilities.UnitTests;
+
+[TestClass]
+public class ManagedNameGeneratorTests
```
Evangelink, line 9: I see that this is a new class so I assume we don't have tests for the other paths. Could you please add some tests for namespace (to ensure this is still working). 

Might also be worth adding tests for:
- namespaces with multiple levels (A.B.C)
- Inner class
- Struct
- Record

Haplois: There is. Look at the other classes. This is just for the cases with no namespace. 

We can add structs and records with no namespace though.

Evangelink: Which class are you referring too? I have looked at all classes in https://github.com/microsoft/vstest/tree/main/test/Microsoft.TestPlatform.AdapterUtilities.UnitTests/ManagedNameUtilities and we don't test scenario mentioned above.

Haplois: This has all the tests. It will also test sbd. https://github.com/microsoft/vstest/blob/main/test/Microsoft.TestPlatform.AdapterUtilities.UnitTests/ManagedNameUtilities/ManagedNameRoundTripTests.cs

No record tests though. Adding those.

Evangelink (approved):

PR #4032, opened 2022-09-29
Add postmortem blame mode

MarcoRossignoli:
Add postmortem blame mode

// Install AeDebugger using procdump for all tests

`vstest.console.exe /AeDebugger:”Install;ProcDumpToolDirectoryPath=c:\procdump;DumpDirectoryPath=c:\DumpsDirectory”`

// Run monitoring the blame folder

`vstest.console.exe … /Blame:”MonitorPostmortemDebugger;DumpDirectoryPath=c:\DumpsDirectory”`

// Uninstall AeDebugger

`vstest.console.exe /AeDebugger:”Uninstall;ProcDumpToolDirectoryPath=c:\procdump;”`

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -8,11 +8,16 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading;
 using System.Xml;
 
 using Microsoft.VisualStudio.TestPlatform.Execution;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
+using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
+using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces;
 using Microsoft.VisualStudio.TestPlatform.Utilities;
 using Microsoft.VisualStudio.TestPlatform.Utilities.Helpers;
 using Microsoft.VisualStudio.TestPlatform.Utilities.Helpers.Interfaces;
@@ -37,11 +42,13 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
     private Dictionary<Guid, BlameTestObject>? _testObjectDictionary;
     private readonly IBlameReaderWriter _blameReaderWriter;
     private readonly IFileHelper _fileHelper;
+    private readonly IProcessHelper _processHelper;
     private XmlElement? _configurationElement;
     private int _testStartCount;
     private int _testEndCount;
     private bool _collectProcessDumpOnCrash;
     private bool _collectProcessDumpOnHang;
+    private bool _monitorPostmortemDumpFolder;
     private bool _collectDumpAlways;
     private string? _attachmentGuid;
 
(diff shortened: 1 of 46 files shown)


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -525,6 +558,55 @@ private void SessionEndedHandler(object? sender, SessionEndEventArgs args)
...
+                                EqtTrace.Info($"[MonitorPostmortemDump]Valid pattern start with '{_testHostProcessName}' found for {dumpFileNameFullPath}");
+                                try
+                                {
+                                    var fileTranferInformation = new FileTransferInformation(_context.SessionDataCollectionContext, dumpFileNameFullPath, true);
+                                    EqtTrace.Info($"[MonitorPostmortemDump]Transferring {dumpFileNameFullPath}");
+                                    _dataCollectionSink.SendFileAsync(fileTranferInformation);
+                                }
+                                catch (IOException ex)
+                                {
+                                    // In case of race explained in the comment above we simply log a warning.
```
nohwnd, line 601:
```suggestion
                                    // In case of race condition explained in the comment above we simply log a warning.
```

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
```diff
@@ -251,6 +251,44 @@ public void HangDumpChildProcesses(RunnerInfo runnerInfo)
...
+    [NetCoreTargetFrameworkDataSource]
+    public void BlameDataCollectorAeDebuggerShouldCollectDump(RunnerInfo runnerInfo)
+    {
+        SetTestEnvironment(_testEnvironment, runnerInfo);
+
+        // Install AeDebugger
+        string dumpPath = Path.Combine(TempDirectory.Path, "Dumps");
+        Directory.CreateDirectory(dumpPath);
+
+        ExecuteVsTestConsole($"/AeDebugger:Install;ProcDumpToolDirectoryPath={_procDumpPath};DumpDirectoryPath={dumpPath}",
```
nohwnd, line 266: Could this be done the same way as on blame where we find it in PATH or from env variable?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -8,11 +8,16 @@
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
```
nohwnd, line 11: Is it worth (or needed) to mix this into Blame collector when the parameters used on commandline are not under "blame"?

MarcoRossignoli: The idea is that Blame can control every "postmortem" system used in a agnostic way...it simply check if some dump with the correct name are generated under that folder...it's unrelated to procdump it could work also on WER for instance.

test/Microsoft.TestPlatform.AcceptanceTests/BlameDataCollectorTests.cs
MarcoRossignoli: Why?here we're installing procdump as postmortem debugger...we need all these information and are machine wide, I don't see benefit in use env var. I mean anyway I have to run this code before the real test. 
If you refer to the fact that we could pass parameter using env var, we could but for this first version(I don't know if we're missing something yet) I'd like to be really "simple" in the setup.

nohwnd: I should have put more focus on the PATH part, that makes it easier to install procdump as a step before the test run (e.g. via chocolatey). And using the same procdump_path env variable would make it work automatically with vstest task which installs procdump and populates it.

On CI it is pretty hard to know what exact path you should specify.

MarcoRossignoli: I can follow-up with the PATH support, I mean we can check the `ProcDumpToolDirectoryPath` and fallback to the PATH/env var, make sense.
I need more time for that and I'd like to ship something to let them test it, can we merge this part?

nohwnd (approved):

PR #4046, opened 2022-10-06
Add VSTEST_DISABLE_THREADPOOL_SIZE_INCREASE feature flag

nohwnd:
## Description
Add a feature flag that allows disabling the increase of threadpool size so we can test if it has impact on special runs.


MarcoRossignoli (approved):

MarcoRossignoli:
:shipit:

PR #4077, opened 2022-10-17
Document the current protocol used in TP

nohwnd:
## Description
Document the current protocol used in TP in LSP protocol like fashion, using mermaid rather than images that are used in RFC so we can version it and evolve it.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -1,0 +1,332 @@
+## What is TestPlatform? 
+
+TestPlatform is a collection of libraries and tools that provide test running for Visual Studio Test Explorer, `dotnet test`, Azure DevOps, Omnisharp, Stryker and other tools.
+
+TestPlatform is mainly focused on running C# tests, but it is an extensible platform, that allows additional providers and extensions to be used.
+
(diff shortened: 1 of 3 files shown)


docs/Overview.md
```diff
@@ -0,0 +1,1735 @@
...
+
+Once a version is agreed upon, a higher version should not be used by the lower level components. 
+
+> The above runner should announce 7 to testhost, rather than 9. If testhost announces 10, 7 is used. 
+
+The version is negotiated between the components at the beginning of every workflow using [ProtocolVersion](#protocolversion-request) request.
+
+### Request, Notification and Response Ordering
+
+The server supports processing only a single request at a time. Unless the request is [Cancel](#cancel) or [Abort](#abort) request.
```
Evangelink, line 243: `Cancel` and `Abort` links are broken.

Evangelink (approved):
Small comment about broken links but maybe that's wip.

docs/Overview.md
nohwnd: Yes, how we handle cancellation and abort is missing from the document. Keeping as is for now.

Issue #2506, opened 2020-08-03
Report tests that did not run / hanged as errors in trx


nohwnd:
When test host crashes, or is crashed when hang is detected, report the tests that did not finish as failed / aborted, to make more visible that something happened.

PR #4083, opened 2022-10-21
Add aborted tests to TRX summary

mfkl:
## Description

The current TRX report summary is missing information about various test status, such as aborted, timeout, inconclusive, etc.

Full list is here https://github.com/microsoft/vstest/blob/458e90e8e310b71629953e380ebabcb1546a0ced/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestRunSummary.cs#L105-L116

Public repro here https://github.com/mfkl/trx-missing-attributes-repro

It would be nice to have this information as part of the TRX summary, which currently only reports failed/passed test information.

I've started by adding the aborted test result info to the TRX summary. If the approach seems fine with you, I'll add the other properties so that the TRX summary is completely filled.

To write the test, I had to modify the following public API `src/Microsoft.TestPlatform.ObjectModel/TestOutcome.cs` which does not match the internal API `src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestOutcome.cs`

## Related issue

Contribution towards https://github.com/microsoft/vstest/issues/2506

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestRunSummary.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/TestRunSummary.cs
@@ -91,6 +91,7 @@ internal class TestRunSummary : IXmlTestStore
         int executed,
         int pass,
         int fail,
+        int abort,
         TestOutcome outcome,
         List<RunInfo> runMessages,
         string stdOut,
@@ -102,7 +103,7 @@ internal class TestRunSummary : IXmlTestStore
         _passedTests = pass;
         _failedTests = fail;
         int countForNonExistingResults = 0; // if below values are assigned constants 0, compiler gives warning CS0414
-        _abortedTests = countForNonExistingResults;
+        _abortedTests = abort;
         _errorTests = countForNonExistingResults;
         _timeoutTests = countForNonExistingResults;
         _inconclusiveTests = countForNonExistingResults;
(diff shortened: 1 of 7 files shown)


mfkl:
@microsoft-github-policy-service agree [company="{g-research}"]

mfkl:
@microsoft-github-policy-service agree

mfkl:
Hi, could anyone please take a look and provide their thoughts on this?

MarcoRossignoli:
Thanks @mfkl for the PR, TRX is used in more than one tool and we need to ensure that we're not breaking anybody, at the moment we're busy on higher priority tasks, we'll review and let you know soon as possible but we don't have an ETA at the moment. Thanks for the patience.

mfkl:
Understood, thanks. Will wait for your feedback!

MarcoRossignoli:
We did some check and for the moment we decided to not take this PR for risky edge cases, we're planning to work on a full review of the TRX format and implementation. Thanks for your contribution.

topherhansen:
I think it's really annoying that the TRX says that all of these attributes are zero when it's just making it up. When a sessiontimeout happens (I tell it 300 seconds and it's only run half my tests) it lies and says that the total was some small part of my tests.

<Counters total = "198" executed = "198" passed = "13" failed = "185" error = "0" timeout = "0" aborted = "0" inconclusive = "0" passedButRunAborted = "0" notRunnable = "0" notExecuted = "0" disconnected = "0" warning = "0" completed = "0" inProgress = "0" pending = "0" />

PR #4084, opened 2022-10-21
Fixed CC for in-process console scenarios

cvpoienaru:
In-process vstest.console wrapper didn't implement CC attachment processing methods until now. These methods were implemented based on their counterparts in the out-of-process vstest.console wrapper, as well as by using important bits-and-pieces from the translation layer (which is no longer required since in-proc console removes the need of a translation layer).


src/Microsoft.TestPlatform.Client/AttachmentsProcessing/InProcessTestRunAttachmentsProcessingEventsHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+    public void HandleLogMessage(TestMessageLevel level, string? message)
+    {
+        _oldEventsHandler.HandleLogMessage(level, message);
+    }
+
+    public void HandleProcessedAttachmentsChunk(IEnumerable<AttachmentSet> attachments)
+    {
+        // Not implemented by design, keep in sync with the same named method from
+        // TestRunAttachmentsProcessingEventsHandler.cs.
+        throw new NotImplementedException();
```
Evangelink, line 32: I tend to prefer having some explanation in such kind of exception. If we are really "sure" this is not supposed to be called, I would add some `Debug.Fail` call in front so that we could see failures in our tests if we break this "contract".

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -839,7 +840,47 @@ public Task InitializeExtensionsAsync(IEnumerable<string> pathToAdditionalExtens
...
+                            attachmentProcessingPayload,
+                            new InProcessTestRunAttachmentsProcessingEventsHandler(eventsHandler),
+                            new ProtocolConfig { Version = _highestSupportedVersion }),
+                        CancellationToken.None)
+                    .ConfigureAwait(false);
+            }
+        }
+        catch (Exception ex)
+        {
+            EqtTrace.Error("InProcessVsTestConsoleWrapper.ProcessTestRunAttachmentsAsync: Exception occurred: " + ex ?? "payload is null");
```
Evangelink, line 873: Are you aware of a case where we would have a null exception? If not, I would remove the `?? "payload is null"` part.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -839,7 +840,47 @@ public Task InitializeExtensionsAsync(IEnumerable<string> pathToAdditionalExtens
...
+                await Task.Run(() =>
+                        TestRequestManager?.ProcessTestRunAttachments(
+                            attachmentProcessingPayload,
+                            new InProcessTestRunAttachmentsProcessingEventsHandler(eventsHandler),
+                            new ProtocolConfig { Version = _highestSupportedVersion }),
+                        CancellationToken.None)
+                    .ConfigureAwait(false);
+            }
+        }
+        catch (Exception ex)
```
Evangelink, line 871: I would really like for us to avoid always catching all when possible. I am not sure what is your goal here but if you just want to prevent the errors linked to cancellation, we could catch a narrow set of exceptions.

Evangelink (commented):
Do you have any ticket we could link this PR to?

In term of tests, it would be nice to test:
- case where token is cancelled
- case where exception is thrown
- case where `TestRequestManager` is `null`

src/Microsoft.TestPlatform.Client/AttachmentsProcessing/InProcessTestRunAttachmentsProcessingEventsHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Collections.Generic;
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Client;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace Microsoft.VisualStudio.TestPlatform.Client;
+
+internal class InProcessTestRunAttachmentsProcessingEventsHandler : ITestRunAttachmentsProcessingEventsHandler
```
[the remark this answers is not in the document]
cvpoienaru: Yep, that is exactly the case, the class is supposed to act as a wrapper to re-route some of the calls to the original event handler. This paradigm is needed for discovery/run as well as for test session start/stop calls because there's some additional logic that runs with some of the calls, but it's not really needed here in retrospect. Removing the wrapper class altogether, thanks for pointing this out !

src/Microsoft.TestPlatform.Client/AttachmentsProcessing/InProcessTestRunAttachmentsProcessingEventsHandler.cs
cvpoienaru: The comment is no longer relevant since I removed the wrapper class altogether, but answering the question for posterity: we are indeed sure this is not supposed to be called, because as far as I can tell, the original attachment processing logic doesn't use this method at all. It may have been added to be used later, but it's not been the case to date.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Hmm, not really, this logic has been copied from the TestRequestManager.cs implementation, but it's not really relevant here. Removing, thanks !

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Catching all possible exceptions here is by design and it's more of a safety net for defensive programming. All methods in the in-process wrapper follow this convention. With the out-of-process implementation we had before, no TP exceptions would have been able to cross the process boundary to TW via exception propagation. Instead, errors would be propagated via special methods on each of the event handlers. With the in-process implementation this is no longer true and TP exceptions could be propagated to TW code which may not be ready to handle them and nasty crashes could occur. Instead, I prefer to propagate the errors via the same old mechanism used by the out-of-process wrapper (i.e. special purpose methods on event handlers). This ensures both the safety and interchangeability of the two wrappers (in-proc vs out-of-proc), as well as preserving the old tried-and-tested error propagation mechanism.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -839,7 +840,47 @@ public Task InitializeExtensionsAsync(IEnumerable<string> pathToAdditionalExtens
...
+                // Awaiting the attachment processing task here. The implementation of the
+                // underlying operation guarantees the event handler is called when processing
+                // is complete, so when awaiting ends, the results have already been passed to
+                // the caller via the event handler. No need for further synchronization.
+                await Task.Run(() =>
+                        TestRequestManager?.ProcessTestRunAttachments(
+                            attachmentProcessingPayload,
+                            new InProcessTestRunAttachmentsProcessingEventsHandler(eventsHandler),
+                            new ProtocolConfig { Version = _highestSupportedVersion }),
+                        CancellationToken.None)
```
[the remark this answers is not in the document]
cvpoienaru: This is in sync with the original out-of-process implementation of the attachment processing logic where a separate call would set the cancellation token. The implementation of the `ProcessTestRunAttachments` checks the cancellation token before certain steps, and if the cancellation token suddenly got set, the operation as a whole is successfully cancelled. I agree that we can pass the cancellation token to the Task.Run call though, as an optimization, so that we don't start the task at all if the token is set before getting here.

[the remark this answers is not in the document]
cvpoienaru: Backtracking on this. We need to go back to passing `CancellationToken.None` to the `Task.Run` call. We need to make sure the cancel event is correctly propagated to the caller (otherwise the caller could be left hanging), and this only happens in the internal attachment processing logic, so we need to make sure we get there. Cancelling the task before it starts here would mean we need to duplicate that logic here and this will make it hard to maintain.

cvpoienaru:
> Do you have any ticket we could link this PR to?
> 
> In term of tests, it would be nice to test:
> 
> * case where token is cancelled
> * case where exception is thrown
> * case where `TestRequestManager` is `null`

Yes, there's an internal ticket that we can link the PR to, will send it to you offline (and also link the PR).

I agree with the last two test cases, I will write tests for them right away. The first test case though (case where the token is cancelled) is not feasible here, since the token is passed down to the  `TestRequestManager.ProcessTestRunAttachments` and there is the logic that's handling the cancellation (including response propagation). That logic is independently tested in other tests, so we are covered.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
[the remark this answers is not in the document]
Evangelink: Could you please add this comment in the code. This is definitely something that I would consider refactoring if I don't see any explanation. Also, do we have a test (or could we add one) that would ensure that if passing the cancellation token we have some check failing? This would provide belt and braces solution for us not to regress.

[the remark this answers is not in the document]
cvpoienaru: Added a comment as you suggested, thanks ! Unfortunately though, we cannot add a test to make sure we don't pass the wrong cancellation token since Task.Run is not mockable and we cannot verify the arguments it receives.

Evangelink (approved):
Thank you for the nice work @cvpoienaru! :shipit:

PR #4095, opened 2022-10-26
Fix recursive resource lookup

nohwnd:
## Description
Check for recursive resource lookup.

This can happen when we are on non-english locale, and we try to load mscorlib.resources
(or potentially some other resources). This will trigger a new Resolve and call the method
we are currently in. If then some code in this Resolve method (like File.Exists) will again
try to access mscorlib.resources it will end up recursing forever.

So we check if the currently looked up assembly is a resource, and if it is we push it into stack
and then check the stack on additional resource lookup. 

## Related issue
[AB#1654923](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1654923)


Evangelink (commented):
Thanks for the quick fix! Could you please add some tests?

nohwnd:
Pushed, not sure if it will work, I was able to test that it fails locally, but I can't build because of some problems around manifest. So let's try to see if CI will pass.

nohwnd:
@Evangelink works, approve?

Evangelink (approved):

PR #3715, opened 2022-06-02
Fix DOTNET_ROOT env var for .NET 6.0+

Evangelink:
Fixes [AB#1528950](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1528950) and fixes #3586

PR #4112, opened 2022-11-04
Fixed testhost crash for net7

cvpoienaru:
This PR fixes two issues:
- Testhost crashing when using the in-proc wrapper to run net7 tests because dotnet couldn't correctly resolve the net7 runtime. Because we now run in-proc (meaning inside the `TestWindowStoreHost` process), certain environment variables such as `DOTNET_ROOT` and `DOTNET_MULTILEVEL_LOOKUP` are set. We cannot unset them because that would change the behavior of the `TestWindowStoreHost` process, but we certainly need them unset for dotnet runtime resolution to be successful. AppDomains provide no isolation since environment variables are shared at the process level. The simplest and most elegant solution is to set the `VSTEST_WINAPPHOST_DOTNET_ROOT` and this would provide successful resolution. More info about `VSTEST_WINAPPHOST_DOTNET_ROOT` can be found [here](https://github.com/microsoft/vstest/pull/3715).

Error here:
```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Testhost process for source(s) 'C:\Users\<USER>\source\repos\TestProject12\TestProject12\bin\Debug\net7.0\TestProject12.dll' exited with error: You must install or update .NET to run this application.
App: C:\Users\<USER>\source\repos\TestProject12\TestProject12\bin\Debug\net7.0\testhost.exe
Architecture: x64
Framework: 'Microsoft.NETCore.App', version '7.0.0-rc.2.22472.3' (x64)
.NET location: C:\Program Files\Microsoft Visual Studio\2022\Preview\dotnet\runtime
The following frameworks were found:
  6.0.9 at [C:\Program Files\Microsoft Visual Studio\2022\Preview\dotnet\runtime\shared\Microsoft.NETCore.App]
Learn about framework resolution:
https://aka.ms/dotnet/app-launch-failed
To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=7.0.0-rc.2.22472.3&arch=x64&rid=win10-x64
. Please check the diagnostic logs for more information.
```
- An unimplemented handler for raw messages got called on TW side when running in-proc. With out-of-proc, raw messages never reach TW since they aren't dequeued by the translation layer. Duplicating the same logic here, since TW wouldn't even know what to do with raw messages. This is not indicating a malfunction on either TP or TW side, but rather the lack of adherence to a convention. Even if this exception is logged in the output window, the tests are still successfully discovered/run.

Error here:
```
System.NotImplementedException: The method or operation is not implemented.
   at Microsoft.VisualStudio.TestWindow.Host.TestPlatformProvider.TestResponseHandler.HandleRawMessageAsync(String rawMessage)
   at Microsoft.VisualStudio.TestWindow.RemoteAgent.TestAgentBridgeCallback.HandleRawMessageAsync(String rawMessage)
   at Microsoft.VisualStudio.TestWindow.Logging.ILoggerExtensions.<CallWithCatchAsync>d__10.MoveNext()
```


nohwnd:
Please provide more information about the fix and the reasoning the description.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -88,6 +89,8 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
 #endif
         consoleParameters.PortNumber = port;
 
+        UpdateDotnetLookupPath();
```
nohwnd, line 92: This fixes this particular problem, but makes the behavior different between using in-process console and not. When in-process console is not used, we are given a full set of env variables when wrapper is instantiated, and we use those to start vstest.console. Testhost process (and other processes started by vstest.console) then inherits just those env variables. 

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/ConsoleParameters.cs#L42-L55

In-process console should probably do the same, and use those env variables to start additional processes. Otherwise the environment will be different for each usecase.

cvpoienaru: Many thanks for the suggestion ! I've actually looked into this first thing when trying to figure out a way to fix the issue and I agree, it would be ideal if we could fix it like you suggested, but unfortunately it doesn't seem possible (at least from my POV).

The reason is that the in-process wrapper, by definition, runs in process, meaning in the `TestWindowStoreHost` process where there's no isolation we can use like AppDomains since environment variables are set at the process level and this seems to be the most granular option we get. We cannot clear the environment variables inside the in-process wrapper and only set the env variables we get via console parameters because that would mess with the TW process. We get the best of both worlds really by keeping the env vars inherited from the TW and (eventually) overriding them because the env var sets are mostly identical, the only thing being left intact is the env vars exclusive to TW.

Propagating the env var set from console params through TP layers down to testhost and other child processes like data collectors is not trivial and the problem we're trying to solve at this point is more theoretical than a real problem. Because of that I'd rather better use this time someplace else. Moreover, we don't really know what a testhost really is, it can be a separate process but it can be loaded as an extension in vstest.console (as with UWP) and then we get the same issue since the testhost will run as part of the vstest.console process, which in fact will be the same process as `TestWindowStoreHost`. We'd have to convince the UWP folks (and other extension writers on the way) to do this propagation when spawning their own processes if they happen to do this.

All in all, I feel confident we can go with this approach at least for now and fix other issues that may appear on the way rather than undertaking the task of propagating the env vars as outlined above, which is not a bulletproof approach either ...

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -133,6 +136,30 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
...
+        const string dotnetGlobalInstallationDir = "dotnet";
+
+        // Allow users to specify a dotnet private installation path before attempting to use the
+        // global installation path.
+        if (_environmentVariableHelper.GetEnvironmentVariable(dotnetRootOverrideEnvVarName) is not null)
+        {
+            return;
+        }
+
+        var dotnetPathPrefix = _environmentVariableHelper.GetEnvironmentVariable(programFilesPathEnvVarName);
```
Evangelink, line 152: @MarcoRossignoli Can you think of a better way to retrieve the "default" global installation folder?

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -133,6 +136,30 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
 
     internal ITestRequestManager? TestRequestManager { get; set; }
 
+    private void UpdateDotnetLookupPath()
+    {
+        const string dotnetRootOverrideEnvVarName = "VSTEST_WINAPPHOST_DOTNET_ROOT";
```
Evangelink, line 141: This is not enough, you need to check for the x86 one too. See https://github.com/microsoft/vstest/blob/9a5bb2faf130809253b28d7ad2d434d785901025/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L482-L489

Evangelink (commented):
As I said during our talk, this seems to be working but I am quite worried this will end up reverted like the DOTNET_ROOT update I did because we are too far away from understanding all the scenario.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -133,6 +136,30 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
...
+            return;
+        }
+
+        var dotnetPathPrefix = _environmentVariableHelper.GetEnvironmentVariable(programFilesPathEnvVarName);
+        if (dotnetPathPrefix is null)
+        {
+            return;
+        }
+
+        _environmentVariableHelper.SetEnvironmentVariable(
```
MarcoRossignoli, line 158: Looks to me that we're changing this env var for "current" process and so we're affecting not only our host, is it right?
Should it be done before the host start?

src/vstest.console/InProcessVsTestConsoleWrapper.cs
MarcoRossignoli: In windows can be in different way...99.9% I expect to find it there...but it's possible that user will install in different folder.
The "correct code" is here:

https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/Helpers/DotnetHostHelper.cs#L121

src/vstest.console/InProcessVsTestConsoleWrapper.cs
nohwnd: There is existing abstraction over environment, I think you could use that. And make sure it is used everywhere you need it.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Removed this method altogether.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Indeed, we altered the environment of the current process. I've removed this fix and instead propagated the env variables only to the testhost process.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Closing this as we abandoned the idea of patching the dotnet location.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -40,6 +40,13 @@ internal ProcessHelper(IEnvironment environment)
         _environment = environment;
     }
 
+    /// <summary>
+    /// Gets or sets the set of environment variables to be used when spawning a new process.
+    /// Should this set of environment variables be null, the environment variables inherited from
+    /// the parent process will be used.
+    /// </summary>
+    public static IDictionary<string, string?>? ExternalEnvironmentVariables { get; set; }
```
MarcoRossignoli, line 48:
```suggestion
    public static IDictionary<string, string?>? DefaultEnvironmentVariables { get; set; }
```

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -62,7 +65,29 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
         ITestRequestManager? testRequestManager,
         Executor executor,
         ITestPlatformEventSource testPlatformEventSource)
+        : this(
```
nohwnd, line 68: Is there a good reason to add another internal constructor when there already is one?

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -89,9 +114,13 @@ public InProcessVsTestConsoleWrapper(ConsoleParameters consoleParameters)
...
-        // TODO: under VS we use consoleParameters.InheritEnvironmentVariables, we take that
-        // into account when starting a testhost, or clean up in the service host, and use the
-        // desired set, so all children can inherit it.
+        // Running vstest.console in process means we inherit all environment variables from the
+        // process we load the wrapper into. We do not want to alter this environment since that
+        // would mean we may interfer with the way the host process works. However, certain
+        // alterations are desired. The solution is to pass the environment variables we get via
+        // the console parameters directly to the testhost process and make sure that at least the
+        // testhost environment is predictable.
+        ProcessHelper.ExternalEnvironmentVariables = consoleParameters.EnvironmentVariables;
```
nohwnd, line 123: You are not passing the option that says if we should clean up, so if user says we should not (default) and will pass just AAA="bbb", you won't be able to start testhost, because it will get just 1 env variable.

test/vstest.console.UnitTests/InProcessVsTestConsoleWrapperTests.cs
```diff
@@ -114,6 +116,55 @@ public void InProcessWrapperConstructorShouldSetEnvironmentVariablesReceivedAsCo
...
+        var _ = new InProcessVsTestConsoleWrapper(
+            consoleParams,
+            _mockEnvironmentVariableHelper.Object,
+            _mockRequestSender.Object,
+            _mockTestRequestManager.Object,
+            new Executor(_mockOutput.Object, new Mock<ITestPlatformEventSource>().Object, new ProcessHelper(), new PlatformEnvironment()),
+            new Mock<ITestPlatformEventSource>().Object);
+
+        Assert.IsTrue(ProcessHelper.ExternalEnvironmentVariables?.ContainsKey(environmentVariableName));
+        Assert.IsTrue(ProcessHelper.ExternalEnvironmentVariables?[environmentVariableName] == "1");
```
nohwnd, line 136: This should make sure there are also the other env variables, because right now there probably is just 1, but there should be all of them, with AAAA added.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
nohwnd: I like the original name better, because this is not the default.

Evangelink: I also prefer the original name. Maybe we could go with `ProcessEnvironmentVariables` or `BaselineEnvironmentVariables`?

src/Microsoft.TestPlatform.Client/AttachmentsProcessing/InProcessTestRunAttachmentsProcessingEventsHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Microsoft.VisualStudio.TestPlatform.Client;
+
+internal class InProcessTestRunAttachmentsProcessingEventsHandler : ITestRunAttachmentsProcessingEventsHandler
+{
+    private readonly ITestRunAttachmentsProcessingEventsHandler _oldEventsHandler;
+
+    public InProcessTestRunAttachmentsProcessingEventsHandler(
+        ITestRunAttachmentsProcessingEventsHandler oldEventsHandler)
+    {
+        _oldEventsHandler = oldEventsHandler;
```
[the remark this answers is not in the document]
cvpoienaru: Done.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Not really, I was being lazy and I didn't want to change all the tests, but for clarity we should have only 1 internal constructor, I agree. Making the change now 👍

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
```diff
@@ -77,6 +84,23 @@ void InitializeAndStart()
...
+            // replacing them with the desired set of environment variables.
+            if (ExternalEnvironmentVariables is not null)
+            {
+                process.StartInfo.EnvironmentVariables.Clear();
+                foreach (var kvp in ExternalEnvironmentVariables)
+                {
+                    if (kvp.Value is null)
+                    {
+                        continue;
+                    }
```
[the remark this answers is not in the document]
cvpoienaru: We directly pass the environment variable collection we get via `ConsoleParameters`, which has the following type: `Dictionary<string, string?>`, to the `ProcessHelper` property. Note the value in the key value pair can be null. If we don't guard against a null value here, we'll end up setting an environment variable that has null as value, which instead has the effect of deleting the said environment var from the collection at some point. Should that variable not exist in the first place, we can probably expect an exception being thrown. This is just trying to guard against that possibility.

src/vstest.console/InProcessVsTestConsoleWrapper.cs
cvpoienaru: Awesome catch, thanks ! Fixing right away 👍

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
MarcoRossignoli: I have indented this prop like "you inherit from parent and plus you'll always add this default list" so "external" is not super clear to me, as a static "always present" environment variables. But I'm fine with both.

test/vstest.console.UnitTests/InProcessVsTestConsoleWrapperTests.cs
cvpoienaru: Do you mean the final set of environment variables that is used to spawn the testhost should contain, among others, `AAAAA` as well ? Because unfortunately we cannot test that, we don't have access to the process start info due to the way `ProcessHelper` is architectured. I'm just making sure the in-process wrapper correctly sets the console params env variables for the `ProcessHelper` to use, so that we don't regress.

src/Microsoft.TestPlatform.PlatformAbstractions/common/System/ProcessHelper.cs
cvpoienaru: I think I may have partially addressed this naming issue by fixing another comment I got from Jakub. I've added an extra property called `InheritEnvironmentVariables` to be used in tandem with `ExternalEnvironmentVariables`. If `InheritEnvironmentVariables` is set to true, then those env variables are just extra variables to be added to the ones we've already inherited from the parent process, otherwise we first clean up the baseline and add only our external environment variables afterwards.

src/vstest.console/vstest.console.csproj
```diff
@@ -49,7 +50,7 @@
     <ProjectReference Include="..\Microsoft.TestPlatform.ObjectModel\Microsoft.TestPlatform.ObjectModel.csproj" />
     <ProjectReference Include="..\Microsoft.TestPlatform.Client\Microsoft.TestPlatform.Client.csproj" />
     <ProjectReference Include="..\Microsoft.TestPlatform.CommunicationUtilities\Microsoft.TestPlatform.CommunicationUtilities.csproj" />
-    <ProjectReference Include="..\Microsoft.TestPlatform.PlatformAbstractions\Microsoft.TestPlatform.PlatformAbstractions.csproj" />
+    <ProjectReference Include="..\Microsoft.TestPlatform.PlatformAbstractions\Microsoft.TestPlatform.PlatformAbstractions.csproj" Aliases="Abstraction" />
```
Evangelink, line 53: Hack to work around the issue where nullable attributes are coming from both PlatformAbstractions and CoreUtilities. We have tried to solve it in a better way but that was leading to too complex refactoring. Using the alias trick helps to isolate what's coming from this assembly.

Note that I hope to remove such hack by using a source generator for the polyfill (@MarcoRossignoli do you remember the name of the project?)

src/vstest.console/vstest.console.csproj
```diff
@@ -22,6 +22,7 @@
       (e.g. System.Runtime 4.1.2 instead of 4.0.0)
     -->
     <AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
+    <DefineConstants>$(DefineConstants);FIX_NULLABLE_CONFLICTS</DefineConstants>
```
Evangelink, line 25: Using this trick as some of the files that need to be fixed are included from a shared project that would force us to define the alias in all using projects.

Evangelink: @cvpoienaru Re-reading that now, I am thinking that it could be clearer to use `USE_EXTERN_ALIAS` instead, WDYT?

Evangelink (approved):
LGTM. Please wait for approval from @nohwnd and @MarcoRossignoli before merging (I have disabled auto-merge).

test/vstest.console.UnitTests/InProcessVsTestConsoleWrapperTests.cs
nohwnd: The test is not testing if inherit env variables is used correctly. If you pass AAAA=1, then in both cases checking AAAA=1  will be true. But you don't know anything about whether AAAA=1 is now the only env variables in place (inherit=false), or if AAAA=1 was added to the existing set of variables (inherit=true).

src/Microsoft.TestPlatform.Execution.Shared/DebuggerBreakpoint.cs
```diff
@@ -1,14 +1,22 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#if USE_EXTERN_ALIAS
```
nohwnd, line 4: This seems bit unusual, what is the reason to have it and are there other ways?

nohwnd: Answered just above.

nohwnd:
I would prefer to not propagate the inherit flag all the way down to ProcessHelper, instead the code that sets the variables (in wrapper) should check what the preference is and either provide just what it was given (inherit=false), or merge it with the current set of env variables (inherit=true) and set that to ExternalEnvVariables. This way providing external env variables will work the same no matter if we are inheriting or not in terms of updates etc.

src/vstest.console/vstest.console.csproj
MarcoRossignoli: https://github.com/Sergio0694/PolySharp

nohwnd (approved):

test/vstest.console.UnitTests/InProcessVsTestConsoleWrapperTests.cs
cvpoienaru: Fixed, thanks !

MarcoRossignoli (approved):
Pls @cvpoienaru when you have a pair of hours can you add the explanation of the in-process working with this "details" here https://github.com/microsoft/vstest-docs/tree/main/RFCs

And maybe after link in the code here, for future devs, we're moving in different context and this repo doesn't have any information of the other world.

cvpoienaru:
Follow up: https://github.com/microsoft/vstest/issues/4119

PR #4147, opened 2022-11-29
Fix satellite resolution for Microsoft.TestPlatform.Common

MarcoRossignoli:
Fix satellite resolution for Microsoft.TestPlatform.Common

### DIFF
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
@@ -463,6 +463,15 @@ public class TestPluginCache
 
     protected void SetupAssemblyResolver(string? extensionAssembly)
     {
+        // If we don't load the resource for Microsoft.TestPlatform.Common before
+        // to set the assembly resolver we won't be able to use the resources.
+        // This should be the algorith followed during the satellite assembly resolution
+        // https://learn.microsoft.com/en-us/dotnet/core/extensions/package-and-deploy-resources#net-framework-resource-fallback-process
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginCache.cs
```diff
@@ -463,6 +463,15 @@ private static bool TryMergeExtensionPaths(List<string> extensionsList, List<str
 
     protected void SetupAssemblyResolver(string? extensionAssembly)
     {
+        // If we don't load the resource for Microsoft.TestPlatform.Common before
+        // to set the assembly resolver we won't be able to use the resources.
+        // This should be the algorith followed during the satellite assembly resolution
+        // https://learn.microsoft.com/en-us/dotnet/core/extensions/package-and-deploy-resources#net-framework-resource-fallback-process
+        // BUT for some unknown reason the point 10 is not working as explained
+        // satellite resolution should fallback to the NeutralResourcesLanguageAttribute
+        // that we set to en-US but don't and we fail with FileNotFoundException
```
Evangelink, line 472:
```suggestion
        // If we don't load the resource for Microsoft.TestPlatform.Common before
        // to set the assembly resolver we won't be able to use the resources.
        // This should be the algorithm followed during the satellite assembly resolution
        // https://learn.microsoft.com/dotnet/core/extensions/package-and-deploy-resources#net-framework-resource-fallback-process
        // BUT for some unknown reason the point 10 is not working as explained.
        // Satellite resolution should fallback to the NeutralResourcesLanguageAttribute
        // that we set to en-US but don't and we fail with FileNotFoundException.
```

Evangelink (approved):

PR #4152, opened 2022-11-29
Remove specific handling for MSTest in trx logger

Evangelink:
## Description

Remove specific handling for MSTest in trx logger. It's not clear why this specific code was added as it seems that older versions of MSTest don't reach that place.

## Related issue

See #1784 where specific handling for mstest was added. See https://github.com/microsoft/testfx/issues/1149 where issue was detected.


MarcoRossignoli (approved):

MarcoRossignoli:
:shipit:

PR #4221, opened 2022-12-22
Handle ObjectDisposedException

MarcoRossignoli:
Found during some debugging.


Evangelink (approved):
Can't wait to see what we broke :P

cc @pavelhorak

PR #4249, opened 2023-01-06
Remove chutzpah

nohwnd:
## Description

Remove chutzpath from the list of tested frameworks and replace it with our own framework that is chosen for running .js files, and does fake discovery and run to prove that we can delegate run of .js files to dotnet testhost.

## Related issue

- [x] I have ensures that there is a previously discussed and approved issue.


test/TestAssets/NonDll.TestAdapter/TestAdapter.cs
```diff
@@ -0,0 +1,73 @@
...
+using System;
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
+using System.Collections.Generic;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Logging;
+
+namespace NonDll.TestAdapter;
+
+
```
Evangelink, line 13: Nitpick:

```suggestion

```

Evangelink (commented):
The csproj project is seen as "binary" file, I think there is a gitconfig issue. Could you please fix it and readd the project as normal content file?

Otherwise PR lgtm

nohwnd:
The project was encoded as utf-16, we have multiple of those in our solution, not sure how that happened. Fixed the encoding in header and the actual encoding. 

Let's fix the rest of files in separate PR.

Evangelink (approved):

PR #4297, opened 2023-02-10
Add `dotnet vstest` deprecation message

MarcoRossignoli:
image

### DIFF
--- a/test/vstest.console.UnitTests/ExecutorUnitTests.cs
+++ b/test/vstest.console.UnitTests/ExecutorUnitTests.cs
@@ -339,7 +339,9 @@ public class ExecutorUnitTests
     }
 
     [TestMethod]
-    public void ExecutorShouldPrintDotnetVSTestDeprecationMessage()
+    [DataRow("--ShowDeprecateDotnetVSTestMessage")]
+    [DataRow("--showdeprecateDotnetvTestMessage")]
+    public void ExecutorShouldPrintDotnetVSTestDeprecationMessage(string commandLine)
     {
         var mockOutput = new MockOutput();
         Mock<IProcessHelper> processHelper = new();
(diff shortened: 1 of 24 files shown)


nohwnd (approved):

nohwnd:
From technical point of view it looks good, if our PM approves, then let's go.

test/vstest.console.UnitTests/ExecutorUnitTests.cs
```diff
@@ -338,6 +338,25 @@ public void ExecutorShouldPrintRunnerArchitecture()
         Assert.IsFalse(mockOutput.Messages.Any(message => message.Message!.Contains("vstest.console.exe is running in emulated mode")));
     }
 
+    [TestMethod]
+    [DataRow("--ShowDeprecateDotnetVSTestMessage")]
```
Evangelink, line 342: I think you should use `DotNet`.

```suggestion
    [DataRow("--ShowDeprecateDotNetVSTestMessage")]
```

test/vstest.console.UnitTests/ExecutorUnitTests.cs
```diff
@@ -338,6 +338,25 @@ public void ExecutorShouldPrintRunnerArchitecture()
...
+    [DataRow("--showdeprecateDotnetvTestMessage")]
```
Evangelink, line 343: Probably:

```suggestion
    [DataRow("--showDeprecateDotNetVSTestMessage")]
```

test/vstest.console.UnitTests/ExecutorUnitTests.cs
MarcoRossignoli: Here `Dotnet` is the muxer `dotnet`, interesting...I've always used dotnet as single word

test/vstest.console.UnitTests/ExecutorUnitTests.cs
MarcoRossignoli: it's done on purpose to test the ordinal ignore casing.

MarcoRossignoli: never mind me...I'm old.

test/vstest.console.UnitTests/ExecutorUnitTests.cs
nohwnd: What makes you think that? The product name is dotnet or dotNET, with the first one being used way more often imo. So Dotnet is correct. DotNET would also be correct, but dotNet is discouraged.

This guideline says to note the capitalization (and not spell with a space in the middle). 

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/n/net

This brand guide does not mention the spelling we are using `dotnet` explicitly, but uses it throughout the guide: https://github.com/dotnet/brand

nohwnd: https://github.com/dotnet/brand/issues/26 added issue for that.

PR #4299, opened 2023-02-10
[rel/17.3] Update Newtonsoft.Json to 13.0.1

nohwnd:
Cherry pick changes from #3815 into 17.3.


eng/Versions.props
```diff
@@ -2,7 +2,7 @@
 <Project>
   <PropertyGroup>
     <!-- This repo version -->
-    <VersionPrefix>17.3.1</VersionPrefix>
+    <VersionPrefix>17.3.2</VersionPrefix>
```
Evangelink, line 5: Should be 17.3.3 actually, isn't it?

MarcoRossignoli: I think so

nohwnd: Fixed.

Evangelink (approved):

PR #4309, opened 2023-02-16
[rel/17.0] Update Newtonsoft.Json to 13.0.1 (#4309

nohwnd:
Update newtonsoft to 13.0.1


test/Microsoft.TestPlatform.AcceptanceTests/MultitargetingTestHostTests.cs
```diff
@@ -14,6 +14,7 @@ public class MultitargetingTestHostTests : AcceptanceTestBase
     {
         [TestMethod]
         [TestCategory("Windows-Review")]
+        [Ignore("Unreliable")]
```
nohwnd, line 17: Ignore acceptance test for multi targetting, we did not touch that, it relies on an edge case of some api not being available across different versions of .net. It does not depend on newtonsoft and I don't know why it is broken.

Evangelink (approved):

PR #4310, opened 2023-02-16
[rel/17.2] Update Newtonsoft.Json to 13.0.1 (#4310

nohwnd:
Update Newtonsoft.Json to 13.0.1


Evangelink (approved):

Evangelink:
Looks like you will need to update the Linux machine.

nohwnd:
@Evangelink I downgraded and downgraded until I reached a version that finally worked. 😢 Luckily this is only a brownout (a new word to me to be honest), so the images were blocked only yesterday, but the build would have worked today (there is one with ✔️ just above the version updates), so I am merging as is, hopefully we will get nugets before April and won't have to touch this again.

PR #4364, opened 2023-04-04
Don't push logs from Publish, we have none.

nohwnd:
Disable publishing logs from the Publish step. We have none and it makes the public PR pipeline end with warning.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -234,7 +234,7 @@ stages:
           steps:
             # The template job needs a log, otherwise it writes a warning. We can disable log uploading only for
             # the whole stage, which is not what we want to do. So we write an empty file instead.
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -221,39 +221,34 @@ stages:
...
-          ${{ if eq(variables._RunAsInternal, True) }}:
-            name: NetCore1ESPool-Internal
-            demands: ImageOverride -equals 1es-windows-2022
-        strategy:
-          matrix:
-            Release:
-              _BuildConfig: Release
-            # ${{ if eq(variables._RunAsPublic, True) }}:
-            #   Debug:
-            #     _BuildConfig: Debug
```
Evangelink, line 239: For another PR - Keeping the debug build is interesting and would avoid to have all the `Debug.Assert` failures uncaught in CI.

Evangelink (approved):

PR #4378, opened 2023-04-06
Fix hangdump running into crashdump

nohwnd:
## Description

This fixes issue that happens when both hang and crash dumps are enabled, which can be common in CI. The issue results in no crash or hang dump being created. 

When process crashes and procdump starts crashdumping it takes it some time to write all the memory to disk. During that time it is possible that hangdump timer will reach the threshold and initiate hang dump. This will try to cancel procdump, and before this fix we simply sent Ctlr+C to procdump, and immediately went to kill it. Killing attached procdump almost always results in testhost (to which it is attached) crashing with it. This will kill testhost, and you will have no crashdump, nor hangdump.

To solve this, we look at the output that procdump streams to us and try to detect if it is currently dumping a process, and wait for that to finish. When it finishes dumping, or after a given timeout (90 seconds, or "VSTEST_CONNECTION_TIMEOUT"), we cancel and if needed kill procdump. 

Then we progress to hang dumping the process. This is needed because procdump is configurable by environment variables, and it is possible to set dumpcount that is higher than 1, so it is totally possible to create 7 crashdumps, and continue running, and then reach a place where we hang.

Additionally we also write errors from procdump into log.

Testing this in automated tests is quite fragile, so I did not add any tests. But reproducing it locally is quite easy by making crash.dll allocate a bunch of memory (10GB roughly), which makes it write the dump for 10 seconds on my computer, and then fine tune the hang dump timeout to 5-20 seconds to the process has enough time to crash, but not finish crash dumping. 

Modified this test to also create hang dumps after 10 seconds:
```csharp
    [TestMethod]
    [NetCoreRunner("net462;net472;netcoreapp3.1;net5.0")]
    // should make no difference, keeping for easy debug
    // [NetFrameworkRunner("net462;net472;netcoreapp3.1;net5.0")]
    public void CrashDumpOnStackOverflow(RunnerInfo runnerInfo)
    {
        SetTestEnvironment(_testEnvironment, runnerInfo);
        var assemblyPaths = GetAssetFullPath("crash.dll");
        var arguments = PrepareArguments(assemblyPaths, GetTestAdapterPath(), string.Empty, string.Empty, runnerInfo.InIsolationValue);
        arguments = string.Concat(arguments, $" /ResultsDirectory:{TempDirectory.Path}");
        arguments = string.Concat(arguments, $@" /Blame:""CollectHangDump;TestTimeout=10s;CollectDump;DumpType=full"" /diag:S:\t\logs\l1\");

        var env = new Dictionary<string, string?>
        {
            ["PROCDUMP_PATH"] = _procDumpPath
        };

        InvokeVsTest(arguments, env);

        ValidateDump();
    }
```

Allocate some random strings into an array to put stuff in memory:
```cs
// Copyright (c) Microsoft Corporation. All rights reserved.

using Microsoft.VisualStudio.TestTools.UnitTesting;

using System;
using System.Linq;

#pragma warning disable IDE1006 // Naming Styles
namespace timeout
#pragma warning restore IDE1006 // Naming Styles
{
    [TestClass]
    public class UnitTest1
    {
        [TestMethod]
        public void TestMethod1()
        {
            var list = new List<a>();
            list.AddRange(Enumerable.Range(0, 100000).Select(a => new a()));
            // stack overflow
            Span<byte> s = stackalloc byte[int.MaxValue];
        }
    }

    public class a
    {
        public static Random random = new Random();
        public a()
        {
            abc = new string(((char)(byte)random.Next(0, 255)), random.Next(10000, 100_000)); 
        }
        public string abc;
    }
}
```

Output in diag logs, where crashdump is orange and initiates the dump, and hangdump is red, and waits for it to finish: 
image

I also tried it when the hangdump is too fast (e.g. 3s), and it simply collects just hangdump, or when it is too slow (e.g. 100s), and then it collects just crashdump. 

Fix #4351


src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
```diff
@@ -63,25 +68,49 @@ internal ProcDumpDumper(IProcessHelper processHelper, IFileHelper fileHelper, IE
...
+        // [16:06:59] Dump 1 initiated: <path>
+        // [16:07:00] Dump 1 writing: Estimated dump file size is 11034 MB.
+        // [16:07:09] Dump 1 complete: 11034 MB written in 10.1 seconds
+        // We also want to know when we completed writing a dump (and not just set _isCrashDumpInProgress once), because dumpcount larger than 1
+        // can be provided externally and then the first dump would prevent hangdump forever from stopping the process, but the not every dump is crashing the process
+        // so we would run forever.
+        //
+        // Yes the two ifs below depend on the content being in english, and containg those words (which is the case for procdump from 2017 till 2023 at least),
+        // if we get different language it should not break us, we will just cancel more aggressively (unfortunately).
+        if (data != null && data.Contains("initiated"))
```
Evangelink, line 86: Is it possible that we have a "race condition" causing us to check the content too soon?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
```diff
@@ -63,25 +68,49 @@ internal ProcDumpDumper(IProcessHelper processHelper, IFileHelper fileHelper, IE
...
+        if (data != null && data.Contains("complete"))
+        {
+            EqtTrace.Info($"ProcDumpDumper.OutputReceivedCallback: Output received from procdump process contains 'complete' dump is finished, you can cancel procdump if you need.");
+            _isCrashDumpInProgress = false;
+            // Do not reset _crashDumpDetected back to false here. Any detected crash dumping should keep it true, so we don't throw away the dump.
+            _crashDumpTimestamps.Add(DateTime.Now);
+        }
+    };
+
+    internal static Action<object?, string?> ErrorReceivedCallback => (process, data) =>
```
Evangelink, line 101: [cosmetic] Could this be moved with the other fields?

Evangelink, line 101: Do we need to pass the `object?` as we are not using it?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -81,3 +78,6 @@ Microsoft.TestPlatform.Extensions.BlameDataCollector.XmlReaderWriter.ReadTestSeq
...
 Microsoft.TestPlatform.Extensions.BlameDataCollector.XmlReaderWriter.XmlReaderWriter(Microsoft.VisualStudio.TestPlatform.Utilities.Helpers.Interfaces.IFileHelper! fileHelper) -> void
 override Microsoft.TestPlatform.Extensions.BlameDataCollector.BlameCollector.Initialize(System.Xml.XmlElement? configurationElement, Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.DataCollectionEvents! events, Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.DataCollectionSink! dataSink, Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.DataCollectionLogger! logger, Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.DataCollectionEnvironmentContext? environmentContext) -> void
+Microsoft.TestPlatform.Extensions.BlameDataCollector.ICrashDumper.GetDumpFiles(bool processCrashed, System.DateTime? testSessionEndTimestamp) -> System.Collections.Generic.IEnumerable<string!>!
```
Evangelink, line 81: Shall we duplicate APIs instead of doing this binary breaking change?

Evangelink (commented):
Nitpick: I saw you fixed some messages to use `procdump` instead of `proc dump`. We are mixing `procdump` and `ProcDump`, could we use only one or the other (I would favoir latter).

I see quite some code added but no unit test code added. Would it make sense to have some more added? Thinking mainly about `ProcDumpDumper.cs` and `DetachFromTargetProcess`.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -499,6 +503,7 @@ private void EventsTestCaseEnd(object? sender, TestCaseEndEventArgs e)
     /// <param name="args">SessionEndEventArgs</param>
     private void SessionEndedHandler(object? sender, SessionEndEventArgs args)
     {
+        _testSessionEndedTimestamp = DateTime.Now;
```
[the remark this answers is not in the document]
nohwnd: I am comparing local time to local time, so it probably won't matter. Can change.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
```diff
@@ -63,25 +68,49 @@ internal ProcDumpDumper(IProcessHelper processHelper, IFileHelper fileHelper, IE
...
-        // Console.WriteLine(data);
-        // Console.ForegroundColor = ConsoleColor.White;
-        // Log all standard output message of procdump in diag files.
-        // Otherwise they end up coming on console in pipleine.
+    {
         EqtTrace.Info($"ProcDumpDumper.OutputReceivedCallback: Output received from procdump process: {data ?? "<null>"}");
 
+        // This is what procdump writes to the output when it is creating a crash dump. When hangdump triggers while we are writing a crash dump
+        // we probably don't want to cancel, because that crashdump probably has the more interesting info.
+        // [16:06:59] Dump 1 initiated: <path>
```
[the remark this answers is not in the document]
nohwnd: Procdump only works on Windows. There is no lower level of verbosity, only higher.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
```diff
@@ -133,32 +163,84 @@ public void AttachToTargetProcess(int processId, string outputDirectory, DumpTyp
...
             return;
         }
 
+        if (_procDumpProcess.HasExited)
+        {
+            EqtTrace.Info($"ProcDumpDumper.DetachFromTargetProcess: ProcDump already exited, returning.");
+            return;
+        }
+
+        // Process? procDumpCancelProcess = null;
```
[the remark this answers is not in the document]
nohwnd: Yup. Removing.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
```diff
@@ -174,7 +256,41 @@ public IEnumerable<string> GetDumpFiles(bool processCrashed)
...
+            EqtTrace.Verbose($"There was 1 {nameof(_crashDumpTimestamps)} and the test session did not end, keeping all crashdumps.");
+            // When test session did not end and we have 1 crash dump it was a crash for sure.
+            crashDumpDetected = true;
+        }
+        else
+        {
+            EqtTrace.Verbose($"There was 1 {nameof(_crashDumpTimestamps)} and the test session did end. Comparing timestamps to see if we should keep the crashdump.");
+            // When session ended, and we created a dump within 100ms around it, it was most likely an
+            // on process exit dump, and we don't need to keep that, unless the user overrides it (e.g. by specifying CollectAlways).
+            var timestampDiff = Math.Abs((_crashDumpTimestamps.Last() - (DateTime)testSessionEndedTimestamp!).TotalMilliseconds);
```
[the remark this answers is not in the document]
nohwnd: done.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
nohwnd: It has the same shape as OutputReceived. So I would keep it.

Evangelink: Oh yeah my bad!

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/PublicAPI/PublicAPI.Shipped.txt
Evangelink: Can we remove from public API if that's only internal then?

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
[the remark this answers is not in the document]
nohwnd: There is no behavior associated with the time here, and all the methods that take it a decide take it as a parameter, testing the behavior would require no special infrastructure. And we are testing it, because we have all approaches tested (crash dump, collect always crash dump, and collect failfast.

Comparing Now or UtcNow, neither is riskier than the other, in both cases we get the timestamp locally, and get both in the same ways, so it does not matter much if it is offset by timezone or not. But let me change that to UtcNow.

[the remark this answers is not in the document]
nohwnd: It can't be fully reliable anyway because we are racing between procdump seeing that process is exiting (when a breakpoint that it sets is reached), and handling the event that testhost sent to us in eventhandler.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcDumpDumper.cs
nohwnd: If you are talking about data being null, then it just is like that sometimes, I don't think we are missing any content, it cleans up the content before imho and cuts surrounding whitespace which leads to us seeing null.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/PublicAPI/PublicAPI.Shipped.txt
nohwnd: Can I do that later?

Evangelink: Sure, as always, I am just giving some feedback but you are free to ignore them :)

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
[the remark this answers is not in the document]
Evangelink: Resolving this one as you are saying you will use `UtcNow` in the next comment.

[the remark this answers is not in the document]
Evangelink: Ok fine by me - keeping open for the `UtcNow` change if you want to do it, otherwise feel free to close.

Evangelink (approved):

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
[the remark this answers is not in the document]
nohwnd: The unreliability shown in our CI, I think I will have to roll back most of this, and just rely on the test count as it was before to see if we crashed, and only use the output for waiting for procdump, but not for crash detection.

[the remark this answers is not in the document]
nohwnd: reverted that, now I just use the output for detection if we are currently crashdumping, but not for detection if any crash dump was made.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/PublicAPI/PublicAPI.Shipped.txt
nohwnd: ended up reverting all api changes.

PR #4383, opened 2023-04-11
Remove root Localize folder

Evangelink:
## Description

Remove root Localize folder as it should not be used anymore (new pipeline is using `eng/Localize`).


nohwnd (approved):
Ok I guess?

Evangelink:
It was removed with the move to arcade but was readded by a PR I did. It's not used by the new path/pipeline. Only used by old pipeline which should be disabled for main - I guess.

PR #4384, opened 2023-04-11
Fix structure of the eng/Localize folder

Evangelink:
## Description

Based on changes I did on https://github.com/microsoft/testfx, it seems that the subfolder is supposed to be named `loc` and the `LocProject.json` should not exist.


nohwnd (approved):
Ok I guess?

PR #4388, opened 2023-04-12
Exclude old branches from new pipeline trigger (#4388)Co-authored-by: Jakub Jareš <me@jakubjares.com>

Evangelink:
Avoid triggering builds on the dnceng pipeline after we moved to Arcade. Still trigger build for PR, that defintion is in yaml and runs on the same pipeline as the public Arcade build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,8 +5,8 @@ trigger:
     - main
     - rel/*
     exclude:
-    - rel/15*
-    - rel/16*
+    - rel/15.*
+    - rel/16.*
     - rel/17.0
     - rel/17.1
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -1,7 +1,19 @@
...
 trigger:
-- main
-- rel/*
+  branches:
+    include:
+    - main
+    - rel/*
+    exclude:
+    - rel/15*
+    - rel/16*
```
nohwnd, line 9:
```suggestion
    - rel/15.*
    - rel/16.*
```

nohwnd (approved):

Issue #4403, opened 2023-04-18
Packages reported as prebuilts in .NET source-build


MichaelSimons:
The following packages are being reported as prebuilts in source-build and will need to be addressed.

```
    <Usage Id="Microsoft.CodeCoverage.IO" Version="17.7.1-beta.23213.1" File="src/artifacts/obj/Microsoft.TestPlatform.CLI/project.assets.json" IsDirectDependency="true" />
    <Usage Id="Microsoft.Internal.TestPlatform.Remote" Version="17.5.33617.425" File="src/artifacts/obj/Microsoft.TestPlatform.CLI/project.assets.json" IsDirectDependency="true" />
    <Usage Id="System.Private.Uri" Version="4.3.2" File="src/artifacts/obj/Microsoft.TestPlatform.CLI/project.assets.json" />
    <Usage Id="System.Private.Uri" Version="4.3.2" File="src/artifacts/obj/testhost.arm64/project.assets.json" IsDirectDependency="true" />
```

None of these packages (any version) are included in source-build today.  Please refer to the source-build [new dependency documentation](https://github.com/dotnet/source-build/blob/main/Documentation/sourcebuild-in-repos/new-dependencies.md) for guidance on how to handle new dependencies from a source-build perspective.

MichaelSimons:
@nohwnd - is this something you can help with?  These are impacting the preview 4 release for source-build.

MichaelSimons:
Prior to the Arcade refactoring, it looks like `Microsoft.CodeCoverage.IO` and `Microsoft.Internal.TestPlatform.Remote` were part of [external.csproj](https://github.com/dotnet/dotnet/blob/release/8.0.1xx-preview1/src/vstest/src/package/external/external.csproj#L34) and were excluded from source build.  How are these components used are they required from sourcebuild?  If not, then they should be moved to [this section](https://github.com/dotnet/dotnet/blob/main/src/vstest/src/package/Microsoft.TestPlatform.CLI/Microsoft.TestPlatform.CLI.csproj#L63).

jakubch1:
@MichaelSimons @Evangelink Microsoft.CodeCoverage.IO is not needed for source build. It is dynamically loaded in merging of code coverage attachments scenario. I've created this PR to exclude it from source build: https://github.com/microsoft/vstest/pull/4426 
Please check it as I am not source build expert.

MichaelSimons:
[...]

PR #4414, opened 2023-04-24
Addressing System.Private.Uri source-build prebuilt in testhost

MichaelSimons:
Related to https://github.com/microsoft/vstest/issues/4403.

I noticed a discrepancy between [testhost.csproj](https://github.com/microsoft/vstest/blob/main/src/testhost/testhost.csproj#L40), [testhost.x86.csproj](https://github.com/microsoft/vstest/blob/main/src/testhost.x86/testhost.x86.csproj#L40) and [testhost.arm64.csproj](https://github.com/microsoft/vstest/blob/main/src/testhost.arm64/testhost.arm64.csproj#L40).   I'm not sure why testhost.arm64.csproj is the only one to not include the source build condition on System.Private.Uri.  This is causing a prebuilt.

@nohwnd
@Evangelink

### DIFF
--- a/src/testhost.arm64/testhost.arm64.csproj
+++ b/src/testhost.arm64/testhost.arm64.csproj
@@ -37,7 +37,7 @@
     </ProjectReference>
 
     <!-- CVE-2019-0657, CVE-2019-0980 and CVE-2019-0981 mitigation -->
-    <PackageReference Include="System.Private.Uri" Version="$(SystemUriVersion)" />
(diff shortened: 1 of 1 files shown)


Evangelink:
Hey @MichaelSimons,

Thanks for the PR! I will try to see if we can import one project from the others as we are only doing a few properties change between them. This should help with avoiding a fix/change to be ported to the other projects.

In the meantime, I have relaunched CI (flaky test) and will approve this PR.

Evangelink (approved):

PR #4422, opened 2023-04-26
Pass workloads to proxy managers

nohwnd:
## Description

When splitting work into workloads, pass the workload on so we only send the sources that are in that workload instead of all sources that are associated with the give runtime provider (e.g. all net6 sources).

## Related issue

- [x] I have ensured that there is a previously discussed and approved issue.

Fix [AB#1803118](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1803118)


fhnaseer (approved):

Evangelink (approved):
Won't block the PR but would it be hard to add a test to avoid regression?

Evangelink (approved):

PR #4426, opened 2023-04-28
Remove IO from source build

jakubch1:
Relates to #4403


Evangelink (approved):
Fix LGTM based on issue description and build is green so assuming this is all good.

PR #4457, opened 2023-05-16
Fix version of CppUnitTestFramework dlls

Evangelink:
## Related issue

Relates to [AB#1813742](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1813742)


Evangelink:
Converting back to draft as it seems it's not the only/real problem. Investigating more.

jakubch1 (approved):

PR #4477, opened 2023-05-24
Update dependencies from https://github.com/dotnet/arcade build 20230…

nohwnd:
According to https://github.com/dotnet/arcade/issues/13648 we need to update VS to 17.6, is this the way to do it? 

Merging changes from #4470.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -26,9 +26,9 @@
     </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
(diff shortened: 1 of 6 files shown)


nohwnd:
Let's see if this builds, or if some other variable needs updating. Or the VS on disk.

Evangelink (commented):
Thanks for testing this @nohwnd. It's also unclear to me what's the needed change. If we can't figure it out, I will ask on the related arcade ticket.

nohwnd:
waiting for the build agent to have 17.6, then the original PR should be mergeable.

PR #4483, opened 2023-05-25
[rel/17.2] Disable internal build on new pipeline

nohwnd:
Disable build on internal pipeline because anything before 17.6 builds on TestPlatform.CI.Real pipeline.


nohwnd:
Just ubuntu is breaking. That is fine, 18.04 is unsupported.

PR #4485, opened 2023-05-25
Fix arm64 build

nohwnd:
## Description

Setting both `RuntimeIdentifier` and `PlatformTarget` ends up building as `AnyCPU` and selecting the default x86 architecture, irregardles of `RuntimeIdentifier`. Moving the `RuntimeIdentifier` property before the `PlatformTarget` property that checks for presence of it in condition.

Testhost.arm64 is not affected because it does not set `PlatformTarget`.


jakubch1 (approved):

Evangelink (approved):
Feel free to ignore but I was wondering if we could add a test or validation script to ensure that the testhost and vstest.console executables have the correct architectures (as a regression safetynet).

nohwnd:
/azp run

Of course it works locally for me. So is it flaky?

PR #4492, opened 2023-05-29
Update dependencies

### LINKED ISSUE #4418: Several packages are rather old - consider upgrading to newer versions
There are some very old packages used by the build.

`Microsoft.Extensions.FileSystemGlobbing` - version 2.0.0

This is very old version, from 2017. It seems that most other projects in .NET are using version 6.0.0, which is also most downloaded from NuGet.org.

Microsoft.Build.Framework - version 16.0.461
Microsoft.Build.Utilities.Core - version 16.0.461

These versions are also rather old, from 2019.

nohwnd:
## Description

Update packages to latest version, or the latest we can use.

## Related issue

Fix #4418


nohwnd:
DO NOT MERGE, before checking binding redirects.

PR #4496, opened 2023-05-29
Fix version of Microsoft.VisualStudio.Interop.dll

Evangelink:
## Description

We should pick the dll coming from Microsoft.VisualStudio.Interop package and not the one shipped with VSSDK build tools

## Related issue

Fixes [AB#1819789](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1819789)


nohwnd:
Do we have to keep that version in sync with something?

Evangelink:
I have used the version we had before but I don't know the limitation (no other dll is having the same version).

nohwnd (approved):

PR #4503, opened 2023-06-01
Update source-build prebuilt baseline

MichaelSimons:
## Description

This PR contains two changes:

1. Removed two source build prebuilt baseline entries added with https://github.com/microsoft/vstest/pull/4486.  These should not have been added as they would have caused source-build prebuilts.  Rather new reference packages were needed which I added in https://github.com/dotnet/source-build-reference-packages/pull/688 and flowed in with https://github.com/microsoft/vstest/commit/e6ea41a3100d1aaa8ea9bd048c5daf2cc084e6d6.
2. Added some comments and updated the CODEOWNERS to help ensure source-build experts review source-build related changes.  See https://github.com/dotnet/source-build/issues/3435.

## Related issue

Related to https://github.com/dotnet/source-build/issues/3435 and https://github.com/microsoft/vstest/pull/4486

PR #4505, opened 2023-06-01
Fix broken source build path

MichaelSimons:
The CODEOWNERS path for source build specified in https://github.com/microsoft/vstest/pull/4503 was incorrect and I unfortunately didn't catch it.


Evangelink:
Maybe we need to grant the alias some write to the repo:
image

PR #4521, opened 2023-06-06
remove some un-used members

SimonCropp:


Evangelink (changes requested):
It seems that some of the removed entries are actually used through `InternalsVisibleTo`.

Evangelink (approved):

PR #4524, opened 2023-06-06
methodParameterTypes cant be null

SimonCropp:
and i assume the check should be for non empty?

so `methodParameterTypes.Length >= 0` becomes `methodParameterTypes.Length > 0`


Evangelink (commented):
Some tests are failing, see
ManagedNameRoundTripTests.Overloads1
ManagedNameRoundTripTests.Overloads7
ManagedNameRoundTripTests.Overloads8

PR #4512, opened 2023-06-05
[rel/17.6] Downgrade Nuget.Frameworks to 6.5.0

nohwnd:
Downgrade Nuget.Frameworks to 6.5.0, because dotnet SDK ships with nuget version 6.6.0-preview.3 which contains 6.6.0.61, but we upgraded to 6.6.0 which requires 6.6.0.62. During build SDK will replace our version with the newest one to keep us up to date, but this time we got ahead of it and this breaks the insertion. 

https://github.com/dotnet/sdk/runs/13951675455

```
> C:\h\w\B8C009BA\p\d\dotnet.exe msbuild /t:Restore C:\h\w\B8C009BA\t\dotnetSdkTests\th1jboi0.34o\82fd8ac8-1635---0705B68C\VSTestCore.csproj
MSBuild version 17.6.3+07e294721 for .NET
  Determining projects to restore...
  Restored C:\h\w\B8C009BA\t\dotnetSdkTests\th1jboi0.34o\82fd8ac8-1635---0705B68C\VSTestCore.csproj (in 540 ms).
> C:\h\w\B8C009BA\p\d\dotnet.exe test --results-directory C:\h\w\B8C009BA\t\dotnetSdkTests\th1jboi0.34o\82fd8ac8-1635---0705B68C\flag-dir///
  Determining projects to restore...
  All projects are up-to-date for restore.
  VSTestCore -> C:\h\w\B8C009BA\t\dotnetSdkTests\th1jboi0.34o\82fd8ac8-1635---0705B68C\bin\Debug\net7.0\VSTestCore.dll
Test run for C:\h\w\B8C009BA\t\dotnetSdkTests\th1jboi0.34o\82fd8ac8-1635---0705B68C\bin\Debug\net7.0\VSTestCore.dll (.NETCoreApp,Version=v7.0)
Microsoft (R) Test Execution Command Line Tool Version 17.6.1 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

StdErr:
Unhandled exception. System.TypeInitializationException: The type initializer for 'Microsoft.VisualStudio.TestPlatform.ObjectModel.Framework' threw an exception.
 ---> System.IO.FileNotFoundException: Could not load file or assembly 'NuGet.Frameworks, Version=6.6.0.62, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
[...]

PR #4533, opened 2023-06-07
Backport 17.6.2 fixes

nohwnd:
## Description

Cherry-picking changes from rel/17.6 into main to keep them in sync.

https://github.com/microsoft/vstest/pull/4517 - exclude known sources
https://github.com/microsoft/vstest/pull/4528 - exclude .resources. of known sources
https://github.com/microsoft/vstest/pull/4512 - update nuget.frameworks to 6.5.0


nohwnd:
The test was relying on having at least one non-test dll that we now totally filter out, so no additional dlls are now there and the test fails, fixing.

Evangelink (approved):

PR #4545, opened 2023-06-13
Use SemVer1

### LINKED ISSUE #4544: Invalid Preview version breaking VSTS
Getting the exception : Invalid version string for TestPlatformVersion 17.7.0-preview.23280.1+94103c3f3198fe6f1405e1cb8fe486d6cd7132fd

I think the version number is somewhat messing up the VS Installer Task.

nohwnd:
## Description

When using TestPlatformInstaller task we can install a preview version, and deep in DTAAgent the version we provide in ProductVersion on vstest.console.exe is parsed and split into sections. Semver1 is assumed, and maximum of 4 dots in the whole version string is also assumed. Semver2 has more dots (`.`), and the parsing breaks.

## Related issue

Fix #4544

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -2,9 +2,11 @@
 <Project>
   <PropertyGroup Label="Version settings">
     <!-- This repo version -->
-    <!-- Workaround for https://github.com/microsoft/vstest/issues/4544 
-    Use semantic versioning V1 because V2 will produce version like 17.7.0-preview.23280.1+94103c3f and DTAExecutionHost 
-    is trying to parse that version and will consider any version with more than 4 `.` in it as invalid. -->
+    <!-- 
+        Workaround for https://github.com/microsoft/vstest/issues/4544 
+        Use semantic versioning V1 because V2 will produce version like 17.7.0-preview.23280.1+94103c3f and DTAExecutionHost 
+        is trying to parse that version and will consider any version with more than 4 `.` in it as invalid. 
+    -->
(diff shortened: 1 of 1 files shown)


MarcoRossignoli (approved):

eng/Versions.props
```diff
@@ -2,6 +2,10 @@
 <Project>
   <PropertyGroup Label="Version settings">
     <!-- This repo version -->
+    <!-- Workaround for https://github.com/microsoft/vstest/issues/4544 
+    Use semantic versioning V1 because V2 will produce version like 17.7.0-preview.23280.1+94103c3f and DTAExecutionHost 
+    is trying to parse that version and will consider any version with more than 4 `.` in it as invalid. -->
+    <SemanticVersioningV1>true</SemanticVersioningV1>
```
Evangelink, line 8: Nit: can be skipped but I like the "WPF" format .

```suggestion
    <!-- 
        Workaround for https://github.com/microsoft/vstest/issues/4544 
        Use semantic versioning V1 because V2 will produce version like 17.7.0-preview.23280.1+94103c3f and DTAExecutionHost 
        is trying to parse that version and will consider any version with more than 4 `.` in it as invalid. 
    -->
    <SemanticVersioningV1>true</SemanticVersioningV1>
```

nohwnd: Merged, it seems a bit too loud for something we won't be able to change any time soon. But I don't have a strong opinion.

PR #4586, opened 2023-06-30
Remove process suspend

nohwnd:
## Description
Remove suspending process, because it uses undocumented NTDll on Windows.


AbhitejJohn:
Curious, what does removing this logic impact?

AbhitejJohn (approved):

nohwnd:
When you hang dump processes, we used to suspend all of them under the current testhost tree, from the bottom and then dump them, so if for some reason there are two things happening that depend on each other we get them both in the hang more reliably. If we can't use that api, then we won't suspend.

PR #4604, opened 2023-07-14
Add possibility to provide telemetry events by data collectors

jakubch1:
## Description

This PR adds possibility for data collectors to provide its owned telemetry events. Telemetry events are passed back to Test Window and are published to VS.

Example:
image

## Related issue

No collector telemetry in Visual Studio.

### DIFF
--- a/src/Microsoft.TestPlatform.Common/DataCollection/AfterTestRunEndResult.cs
+++ b/src/Microsoft.TestPlatform.Common/DataCollection/AfterTestRunEndResult.cs
@@ -24,6 +24,7 @@ public class AfterTestRunEndResult
         AttachmentSets = null!;
         InvokedDataCollectors = null!;
         Metrics = null!;
+        TelemetryEvents = null!;
     }
 
     /// <summary>
@@ -36,7 +37,7 @@ public class AfterTestRunEndResult
     /// The metrics.
     /// </param>
     public AfterTestRunEndResult(Collection<AttachmentSet> attachmentSets, IDictionary<string, object> metrics)
-        : this(attachmentSets, new Collection<InvokedDataCollector>(), metrics)
+        : this(attachmentSets, new Collection<InvokedDataCollector>(), metrics, new Collection<TelemetryEvent>())
     { }
 
     /// <summary>
@@ -54,10 +55,34 @@ public class AfterTestRunEndResult
     public AfterTestRunEndResult(Collection<AttachmentSet> attachmentSets,
         Collection<InvokedDataCollector>? invokedDataCollectors,
         IDictionary<string, object> metrics)
+        : this(attachmentSets, invokedDataCollectors, metrics, new Collection<TelemetryEvent>())
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AfterTestRunEndResult"/> class.
+    /// </summary>
+    /// <param name="attachmentSets">
+    /// The collection of attachment sets.
+    /// </param>
+    /// <param name="invokedDataCollectors">
+    /// The collection of the DataCollectors invoked during test session
+    /// </param>
+    /// <param name="metrics">
+    /// The metrics.
+    /// </param>
+    /// <param name="telemetryEvents">
+    /// The telemetry events.
+    /// </param>
+    public AfterTestRunEndResult(Collection<AttachmentSet> attachmentSets,
+        Collection<InvokedDataCollector>? invokedDataCollectors,
+        IDictionary<string, object> metrics,
+        Collection<TelemetryEvent>? telemetryEvents)
     {
         AttachmentSets = attachmentSets;
         InvokedDataCollectors = invokedDataCollectors;
         Metrics = metrics;
+        TelemetryEvents = telemetryEvents;
     }
 
     [DataMember]
@@ -68,4 +93,7 @@ public class AfterTestRunEndResult
 
     [DataMember]
     public IDictionary<string, object> Metrics { get; private set; }
+
+    [DataMember]
+    public Collection<TelemetryEvent>? TelemetryEvents { get; private set; }
 }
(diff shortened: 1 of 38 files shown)


src/Microsoft.TestPlatform.Common/DataCollection/AfterTestRunEndResult.cs
```diff
@@ -36,7 +37,7 @@ private AfterTestRunEndResult()
     /// The metrics.
     /// </param>
     public AfterTestRunEndResult(Collection<AttachmentSet> attachmentSets, IDictionary<string, object> metrics)
-        : this(attachmentSets, new Collection<InvokedDataCollector>(), metrics)
+        : this(attachmentSets, new Collection<InvokedDataCollector>(), metrics, new Collection<TelemetryEvent>())
```
nohwnd, line 40: Metrics are already passed here which hold telemetry, can that be reused so we don't have to pass two different objects with telemetry?

src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionRequestHandler.cs
```diff
@@ -410,7 +410,8 @@ private void HandleAfterTestRunEnd(Message message)
             _requestData.MetricsCollection.Add(TelemetryDataConstants.InvokedDataCollectors, string.Join(",", invokedDataCollectorsForMetrics.ToArray()));
         }
 
-        var afterTestRunEndResult = new AfterTestRunEndResult(attachmentsets, invokedDataCollectors, _requestData.MetricsCollection.Metrics);
+        var afterTestRunEndResult = new AfterTestRunEndResult(attachmentsets, invokedDataCollectors,
+            _requestData.MetricsCollection.Metrics, _dataCollectionManager.GetTelemetryEvents());
```
nohwnd, line 414: This should merge the data into metrics if possible, and not add any public apis.

jakubch1: Adding all events into metrics can be theoretically possible but it is not elegant IMO. Currently metrics contains metrics of the execution and is used by Test Explorer to send only 1 telemetry event. By some other components those metrics can be used in different way. To add all telemetry events into metrics we would need to use 1 string and treat value in dictionary as list of tuples:

SPECIAL_KEY -> [ (name, Dictionary<string, object>()),  (name, Dictionary<string, object>()),  (name, Dictionary<string, object>())]

IMO keeping it as separate List is better. I extended public API exactly same as "InvokedDataCollector" so I think this change is fine.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelRunDataAggregatorTests.cs
```diff
@@ -119,34 +119,61 @@ public void AggregateShouldAggregateInvokedCollectorsCorrectly()
...
+        Assert.AreEqual(1, aggregator.TelemetryEvents.Count, "TelemetryEvents List must have data.");
+
+        aggregator.Aggregate(null, null, null, TimeSpan.Zero, false, false, null, null, null, telemetryEvents, null);
+        Assert.AreEqual(2, aggregator.TelemetryEvents.Count, "TelemetryEvents List must have aggregated data.");
+
+        aggregator.Aggregate(null, null, null, TimeSpan.Zero, false, false, null, null, null, telemetryEvents, null);
+        Assert.AreEqual(3, aggregator.TelemetryEvents.Count, "TelemetryEvents List must have aggregated data.");
+
+        Assert.AreEqual(telemetryEvents[0].Name, aggregator.TelemetryEvents[0].Name);
+        Assert.AreEqual(telemetryEvents[0].Name, aggregator.TelemetryEvents[1].Name);
```
cvpoienaru, line 142: Shouldn't the aggregator overwrite the TelemetryEvents[0] for each of the subsequent calls ?

src/vstest.console/InProcessVsTestConsoleWrapper.cs
```diff
@@ -11,6 +11,8 @@
 using System.Threading;
 using System.Threading.Tasks;
 
+using Abstraction::Microsoft.VisualStudio.TestPlatform.PlatformAbstractions;
```
cvpoienaru, line 14: Is this needed ?

src/Microsoft.TestPlatform.Common/DataCollection/AfterTestRunEndResult.cs
```diff
@@ -24,6 +24,7 @@ private AfterTestRunEndResult()
         AttachmentSets = null!;
         InvokedDataCollectors = null!;
         Metrics = null!;
```
cvpoienaru, line 26: Isn't it possible to reuse the metrics dictionary instead of adding an additional argument that we have to propagate everywhere ?

src/Microsoft.TestPlatform.CrossPlatEngine/DataCollection/DataCollectionTestRunEventsHandler.cs
```diff
@@ -115,6 +116,15 @@ public void HandleRawMessage(string rawMessage)
                 }
             }
 
+            _telemetryEvents = dataCollectionResult?.TelemetryEvents;
+            if (_telemetryEvents?.Count > 0)
+            {
+                foreach (var telemetryEvent in _telemetryEvents)
```
cvpoienaru, line 122: Can we use AddRange() ?

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelRunDataAggregatorTests.cs
jakubch1: no, each telemetry event is new event - we do not merge it

PR #4620, opened 2023-07-25
Added notification delay when there are no subscribers

cvpoienaru:
Added notification delay when there are no subscribers


nohwnd:
Does it help with the issue? And do we end up forcing this delay into every run? Because subscribers change at least 1 once IMHO.

nohwnd (changes requested):
^^^

nohwnd:
Replaced by #4735

PR #4629, opened 2023-07-31
Fix timing issue in parallel execution

nohwnd:
## Description

Parallel workers are able to pick up work between setting HasWork and actually associating the workload, which leads to the work being unscheduled. The test run (or test discovery) then never finishes. This was observed on a run with 99 dlls, and can be reproduced locally. To me it was on parallel runs of 12 hosts.

This fix moves HasWork to be the last item to be set under the lock, so the readers that are outside of the lock cannot read it too early before the slot is completely populated.

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelOperationManager.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelOperationManager.cs
@@ -5,10 +5,12 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 using Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.Parallel;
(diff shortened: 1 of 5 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelOperationManager.cs
```diff
@@ -152,6 +165,13 @@ private bool RunWorkInParallel()
                 slot.Work = workload.Work;
 
                 _workloads.Remove(workload);
+
+                EqtTrace.Verbose($"ParallelOperationManager.RunWorkInParallel: Adding 1 workload to slot, remaining workloads {_workloads.Count}.");
+
+                // This must be set last, every loop below looks at this property,
+                // and they can do so from a different thread. So if we mark it as HasWork before actually assigning the work
+                // we can pick up the slot, but it has no associated work yet.
+                slot.HasWork = true;
```
nohwnd, line 174: The fix.

Evangelink (approved):
Linking #4630 here to explain why we approve the change as-is.

PR #4610, opened 2023-07-17
Take System dlls from testhost folder

Evangelink:
## Description

For `TestHostNetFramework` folder, embed dlls from the various net framework folders of the testhost projects rather than from netstandard library package.

## Related issue

Relates to [AB#1843962](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1843962)
Fix #4549

PR #4632, opened 2023-08-02
Backport recent fixes

nohwnd:
Backporting recent simple fixes to have then in 7.0.4xx SDK

- Take System dlls from testhost folder (#4610) to fix running .NET Framework dlls on mono and under VS on Mac
- [Fix timing issue in parallel execution (#4629)


MarcoRossignoli (approved):

cvpoienaru:
We're only interested in inserting this in SDK, not VS, correct ?

PR #4642, opened 2023-08-04
RTM Build Fix

cvpoienaru:
RTM Build Fix

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -97,10 +97,6 @@ variables:
         /p:VisualStudioDropName=$(VisualStudioDropName)
         /p:GenerateSbom=true
 
-  - ${{ if eq(parameters.isRTM, True) }}:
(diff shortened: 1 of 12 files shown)


cvpoienaru:
> What is this supposed to do? It is reverting changes. Can you instead create rel/17.7.0 branch if you need a specific commit to be there?

I'd rather not create a rel/17.7.0 branch. I just want to have the current fix for the RTM build (otherwise I'm not able to produce the build) on top of the last commit we inserted in VS, but I don't want to include your changes too since the nugets will diverge from the VS version.

Don't worry, I'll cherry-pick your commit back on top when I have confirmation the RTM build can be produced.

nohwnd (changes requested):
Please create a rel/17.7.0 branch from the commit you want to build again and build that. Reverting changes here in rel/17.7 will propagate them via darc, and you will have to reapply my changes.

PR #4649, opened 2023-08-08
Set console encoding to UTF8

### LINKED ISSUE #4605: Consider setting Console.OutputEncoding for `dotnet test` and friends
There was a change in Windows Terminal 1.17 that causes .NET applications that output Unicode characters to the console to have them garbled, because of the default output encoding. This does not happen in prior versions of Windows Terminal, nor in classic "conhost.exe" console windows.

xUnit.net uses characters which are corrupted now when run on Windows Terminal through `dotnet test` and friends.

Output from `xunit.console.exe`:

image

Output from `dotnet test` (in Windows Terminal-hosted Powershell):

image

Output from `dotnet test` (in conhost-hosted Powershell):

image

While strictly speaking this isn't a bug in VSTest (or xUnit.net), we opted fixed this in https://github.com/xunit/xunit/commit/c4b6ed0f41bb87aa3a92042bcb4f6bbc4694c12d on our side. I would recommend that you consider doing this as well, so that any output from testing frameworks which is outside of ASCII can be presented properly in Windows Terminal.

[AB#1865057](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1865057)

nohwnd:
Set console encoding to UTF8. 

Fix #4605

### DIFF
--- a/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
@@ -57,6 +57,9 @@ internal partial class FeatureFlag : IFeatureFlag
     // Disable the SerialTestRunDecorator
     public const string DISABLE_SERIALTESTRUN_DECORATOR = VSTEST_ + nameof(DISABLE_SERIALTESTRUN_DECORATOR);
 
+    // Disable setting UTF8 encoding in console.
+    public const string DISABlE_UTF8_CONSOLE_ENCODING = VSTEST_ + nameof(DISABlE_UTF8_CONSOLE_ENCODING);
+
     [Obsolete("Only use this in tests.")]
     internal static void Reset()
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
```diff
@@ -57,6 +57,9 @@ internal partial class FeatureFlag : IFeatureFlag
     // Disable the SerialTestRunDecorator
     public const string DISABLE_SERIALTESTRUN_DECORATOR = VSTEST_ + nameof(DISABLE_SERIALTESTRUN_DECORATOR);
 
+    // Disable setting UTF8 encoding in console.
+    public const string DISABlE_UTF8_CONSOLE_ENCODING = VSTEST_ + nameof(DISABlE_UTF8_CONSOLE_ENCODING);
```
cvpoienaru, line 61: Any reason DISABLE has a lowercase 'l' ?

nohwnd: I wish I could say I am testing your attention, but I have no idea why. My fingers must have typed it that way. I sometimes type 0 instead of o or O, and vice versa. I guess this is another quirk of l L and 1 looking the same and my brain working in visual way.

cvpoienaru (approved):

PR #4651, opened 2023-08-10
Fixed RTM version string

cvpoienaru:
Fixed RTM version string


nohwnd:
This will not fix it. You need to change the version that is in ProductVersion on the file.

PR #4675, opened 2023-08-25
Fix SB ref assembly issue for System.ComponentModel.Composition

mthalman:
This fixes an issue with System.ComponentModel.Composition, as described in https://github.com/dotnet/source-build/issues/3599, that is showing up as a reference assembly in the source-built SDK.

The reason a reference assembly exists in this case is because vstest has a reference to the 7.0.0 version of System.ComponentModel.Composition. When building with source-build, it loads that reference from [SBRP](https://github.com/dotnet/source-build-reference-packages) (which only contains reference assemblies) in order to fulfill compile time references. The problem is that the assembly is also getting included in the output. This should have been detected by poison leak detection [but that doesn't yet handle reference assemblies](https://github.com/dotnet/source-build/issues/2817).

It's not known whether the existence of this ref assembly causes a functional issue. But it is known that the source-built 7.0 SDK doesn't define this as a ref assembly but rather as an implementation assembly. So to maintain parity with 7.0 and avoid potential risk, it's best to ensure this is represented as an implementation assembly in the output.

This removes the special configuration that was done for source-build so that it simply reuses the existing `SystemComponentModelCompositionVersion` property. This property will get overridden by source-build as a result of the inclusion of System.ComponentModel.Composition to Version.Details.xml.

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -3,7 +3,17 @@
 
 <UsageData>
   <IgnorePatterns>
+    <!-- Caused by dependency on System.ComponentModel.Composition.4.5.0. This version is overridden in full source-build. -->
+    <UsagePattern IdentityGlob="Microsoft.NETCore.Platforms/2.0.0" />
     <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*/*" />
     <UsagePattern IdentityGlob="NuGet.Frameworks/6.5.0" />
+    <!-- This version is overridden in full source-build. -->
+    <UsagePattern IdentityGlob="System.ComponentModel.Composition/4.5.0" />
+    <!-- Caused by dependency on System.ComponentModel.Composition.4.5.0. This version is overridden in full source-build. -->
+    <UsagePattern IdentityGlob="System.Security.AccessControl/4.5.0" />
+    <!-- Caused by dependency on System.ComponentModel.Composition.4.5.0. This version is overridden in full source-build. -->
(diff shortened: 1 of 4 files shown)


mthalman:
cc @MichaelSimons for review

MichaelSimons (approved):

Evangelink:
Thanks for the PR @mthalman, would you mind fixing the SB error:

```cmd
5 new pre-builts discovered! Detailed usage report can be found at /__w/1/s/artifacts/source-build/self/prebuilt-report/baseline-comparison.xml.
See https://aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them.
Package IDs are:
Microsoft.NETCore.Platforms.2.0.0
System.ComponentModel.Composition.4.5.0
System.Security.AccessControl.4.5.0
System.Security.Permissions.4.5.0
System.Security.Principal.Windows.4.5.0
```

Evangelink:
TP is used in many different scenarios/contexts and so any package update is always a concern as it could be breaking users. In this particular context, I have no idea what's the impact of changing the composition dll...

mthalman:
Looks like the repo has a dependency on `netcoreapp3.1` TFM which System.ComponentModel.Composition.7.0.0 does not support. So we can't upgrade the default version to 7.0.0. That means I'll need to add System.ComponentModel.Composition.4.5.0 to [SBRP](https://github.com/dotnet/source-build-reference-packages) to resolve the prebuilts.

MichaelSimons (approved):

mthalman:
@Evangelink - Builds are passing now.

Evangelink (approved):

Issue #2488, opened 2020-07-22
Replace Newtonsoft.Json with System.Text.Json


IanKemp:
(This was discussed in #2316 but deserves to be its own issue, as the work necessary to close this issue would be different to the work to close that one.)

I'm tired of typing `JsonSerializer` in my test projects and getting it autocompleted to `Newtonsoft.Json.JsonSerializer` because `Microsoft.Net.Test.Sdk` (via `Microsoft.TestPlatform.TestHost`) references (a very old version) of the now-essentially-deprecated `Newtonsoft.Json`, especially when my greenfields .NET Core app under test has zero references to `Newtonsoft.Json`.

This would also help people who do have explicit dependencies on `Newtonsoft.Json` and are being bitten by #2279.

nohwnd:
We would love to do this, but we need to keep our dependencies compatible with net452 and System.Text.Json is not compatible with that.

0xced:
It's actually much worse, the dependency on `Newtonsoft.Json` is not properly declared in the NuGet package.

I was trying to replace `Newtonsoft.Json` with `System.Text.Json` in [Stryker.NET][1] and the dependency on `Newtonsoft.Json` hit me hard. After some work to replace all JSON handling everywhere in Stryker.NET I was finally able to remove `<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />` from its csproj. But thanks to JetBrains Rider great feature where you can see transitive packages (*Implicitly Installed Packages in Solution*), I saw that `Newtonsoft.Json` was still referenced.

It turns out it was referenced through `Buildalyzer/3.2.2` → `Microsoft.Extensions.DependencyModel/2.1.0` → `Newtonsoft.Json/9.0.1`. So in order to completely get rid of Newtonsoft.Json I explicitly updated `Microsoft.Extensions.DependencyModel` to [version 5.0.0][2] which has zero dependencies for the `net5.0` target framework. I added `<PackageReference Include="Microsoft.Extensions.DependencyModel" Version="5.0.0" />` to Stryker.Core.csproj and `Newtonsoft.Json` was definitely gone.

But then… this happened at runtime:
```
System.IO.FileNotFoundException: Could not load file or assembly 'Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'. The system cannot find the file specified.

[...]

PR #4681, opened 2023-08-29
Upgrade Newtonsoft.Json from 13.0.1 to 13.0.3

mthalman:
.NET repos currently have a mix of Newtonsoft.Json versions that we're trying to get consolidated into the latest version. This is particularly necessary for the .NET's [source-build](https://github.com/dotnet/source-build) which can only reference one version of a dependency.

This takes the commit from https://github.com/microsoft/vstest/pull/4680 since I don't have permission to push directly to the PR's branch.

Related to https://github.com/dotnet/source-build-externals/pull/195

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -14,9 +14,9 @@
       <Sha>5ce78f66d89ea529e459abddb129ab36cb5bd936</Sha>
       <SourceBuild RepoName="diagnostics" ManagedOnly="true" />
     </Dependency>
-    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-externals" Version="8.0.0-alpha.1.23424.1">
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-externals" Version="8.0.0-alpha.1.23429.1">
       <Uri>https://github.com/dotnet/source-build-externals</Uri>
-      <Sha>e3cc6c792114ebdfe6627742d2820dbe1ae5bc47</Sha>
+      <Sha>0603839a51f5e18b89c60a3690aff5e81fa666ca</Sha>
       <SourceBuild RepoName="source-build-externals" ManagedOnly="true" />
     </Dependency>
     <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="8.0.0-alpha.1.23414.1">
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -39,6 +39,7 @@
     <MicrosoftVSUtilitiesInternalVersion>16.3.42</MicrosoftVSUtilitiesInternalVersion>
     <MicrosoftVSSDKBuildToolsVersion>17.4.2116</MicrosoftVSSDKBuildToolsVersion>
     <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
+    <NewtonsoftJsonVersion Condition="'$(DotNetBuildFromSource)' == 'true'">13.0.3</NewtonsoftJsonVersion>
     <RoslynBannedApiAnalyzersVersion>3.3.3</RoslynBannedApiAnalyzersVersion>
     <RoslynPublicApiAnalyzersVersion>3.3.4-beta1.21554.2</RoslynPublicApiAnalyzersVersion>
     <SystemCollectionsImmutableVersion>1.5.0</SystemCollectionsImmutableVersion>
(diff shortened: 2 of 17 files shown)


mthalman:
cc @MichaelSimons for review

MichaelSimons (approved):

mthalman:
cc @Evangelink for review

Evangelink:
Hi @mthalman,

We will have to discuss this change with the team. Because the newtonsoft dll is shipped in the user bin, any update is considered as a breaking change so I am not sure we can move this forward right now.

MichaelSimons:
> > Hi @mthalman,
> > We will have to discuss this change with the team. Because the newtonsoft dll is shipped in the user bin, any update is considered as a breaking change so I am not sure we can move this forward right now.
> 
> If this change isn't accepted, then [SourceBuildPrebuiltBaseline.xml](https://github.com/microsoft/vstest/blob/main/eng/SourceBuildPrebuiltBaseline.xml) will need to be updated to include `Newtonsoft.Json/13.0.1`. That will prevent it from being reported as a prebuilt which is currently causing #4680 to fail.

Alternatively would it be an option to conditionally reference 13.0.3 for source-build?  The benefit of this would be the repo leg would mimic the product source-build.

ViktorHofer:
> We will have to discuss this change with the team. Because the newtonsoft dll is shipped in the user bin, any update is considered as a breaking change so I am not sure we can move this forward right now.

@Evangelink can you please explain why upgrading an assembly to the latest patch version is considered a breaking change? I totally get your point for major/minor version upgrades but patch version increments should be non-breaking.

nohwnd (approved):

nohwnd:
The risk here is that we get included into the test project if the tested library also depends on Newtonsoft.Json, we will update the version of the dependency to a newer one that has more fixes, but the deployed product will not have that fixed dll.

So we should be doing those upgrades cautiously. If this was an update from 13 to 14 I would be more careful. 

How often do you plan to update your consolidated version? Because I would prefer to not always push for the latest version of newtonsoft json, unless there is a vulnerability reported.

ViktorHofer:
Interesting. I would have assumed that with the correct isolation, dependencies shouldn't overlap, i.e. by using AssemblyLoadContext to load the test assembly by the runner. Isn't that possible here? One more side question, why do you still use Newtonsoft.Json instead of System.Text.Json?

nohwnd:
> Interesting. I would have assumed that with the correct isolation, dependencies shouldn't overlap, i.e. by using AssemblyLoadContext to load the test assembly by the runner. 

If the architecture was different this would be true, but for .NET the testhost is installed into the project as a nuget, and it claims dependencies on newtonsoft.json, so you will end up with only 1 copy of the dll in the output folder, the newer one. 

If the testhost was shipped with vstest.console and loaded from a different location (and if assembly load context was used in the first place), then we might be able to isolate it.  But it is not the case, and we don't plan to change it. 

>  One more side question, why do you still use Newtonsoft.Json instead of System.Text.Json?

There are features that are missing from system.text.json (e.g. generic attributes for annotations that don't depend on the library), it is not built into .net framework, and types from Newtonsoft.Json are part of our public communications api, and we don't know what would break if we remove it. 

This is a thread where this question was also asked:  https://github.com/microsoft/vstest/issues/2488#issuecomment-1155072927

nohwnd:
Is it okay if this is merged in to 17.9 and go into net9, instead of doing this change in 17.8, and potentially breaking net8 which is nearing release?

ViktorHofer:
> Is it okay if this is merged in to 17.9 and go into net9, instead of doing this change in 17.8, and potentially breaking net8 which is nearing release?

@MichaelSimons or @mthalman would need to speak to that but I'm pretty sure that this needs to happen as part of .NET 8 because Newtonsoft.Json in https://github.com/dotnet/source-build-externals was updated in the main branch which applies to .NET 8.

mthalman:
We can use the [suggestion](https://github.com/microsoft/vstest/pull/4681#issuecomment-1699439277) from @MichaelSimons to conditionally use 13.0.3 only for source-build and make that change for .NET 8. For .NET 9, we can have it use 13.0.3 for everything.

@nohwnd - Which branches should I be targeting to make these two changes then? It doesn't look like there's a 17.9 branch. And, AFAIK, builds from the main branch here are flowing into .NET 8 SDK.

MichaelSimons (approved):

nohwnd:
@mthalman  yes that is true, I wanted to snap to 17.8, and make main 17.9, but I was told that it is too early for that. So please target main.

mthalman:
> @mthalman yes that is true, I wanted to snap to 17.8, and make main 17.9, but I was told that it is too early for that. So please target main.

Ok. Already targeting main so should be good.

PR #4686, opened 2023-09-04
Set branding to 17.9.0

nohwnd:
## Description

Set branding to 17.9.0.


nohwnd (commented):
DO NOT MERGE YET. There is no VS17.9 channel yet, checking why.

PR #4693, opened 2023-09-22
Remove dependency on Nuget.Frameworks

### LINKED ISSUE #3154: NuGet.Frameworks dependency is causing weird test failures when trying to load MSBuild
## Description
I have a project (https://github.com/dotnet/upgrade-assistant) that loads MSBuild from an available source and lets that run how it will. Tests employing this fail when moving to .NET 6 because the test framework has already loaded an (old) version of the NuGet.Frameworks library.

```
System.IO.FileLoadException: Could not load file or assembly 'NuGet.Frameworks, Version=6.0.0.220, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. Could not find or load a specific file. (0x80131621)
```

I've verified that before the tests start, `NuGet.Frameworks` is already loaded. This worked fine when it was a .NET 5 version of msbuild as the versions aligned. This can be fixed by adding the following line to the test project file:

```xml
<ItemGroup>
  <!-- The test framework brings in an old NuGet.Frameworks that ends up getting used rather than what MSBuild wants -->
  <PackageReference Include="NuGet.Frameworks" Version="6.0.0-rc.258" />
</ItemGroup>
```

I see a PR opened that is still in draft mode to address this (https://github.com/microsoft/vstest/pull/2544). For anyone that is trying to test something that uses MSBuild, they'll probably hit this and end up spending too much time fiddling with things to figure out what is going on.

nohwnd:
Remove dependency on Nuget.Frameworks by copying the relevant code into the repository and changing namespace. The Nuget.Frameworks types were not exposed in our public API directly, but we are relying on the functionality, to parse things like net5.0, net472, or any of similar framework shorthands. We also rely on all the frameworks not being explicitly stated, e.g. net11.0 or any such verison of .NET that does not exist yet.

Replace #2544 

Fix #3154

The risk here is that someone is relying on us shipping Nuget.Frameworks into the bin, and not providing it themselves. 

The newly added public api should not cause issues for the existing loggers because they ship together with vstest.console.


src/Microsoft.TestPlatform.ObjectModel/Framework.cs
```diff
@@ -16,6 +15,675 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel;
...
+            ShortName = "net47",
+        },
+        [".NETFramework,Version=v4.7.1"] = new Framework
+        {
+            Name = ".NETFramework,Version=v4.7.1",
+            FrameworkName = ".NETFramework",
+            Version = "4.7.1.0",
+            ShortName = "net471",
+        },
+        [".NETFramework,Version=v4.7.2"] = new Framework
```
nohwnd, line 120: 4.8? 4.8.1?

eng/verify-nupkgs.ps1
```diff
@@ -16,13 +16,13 @@ function Verify-Nuget-Packages {
     $expectedNumOfFiles = @{
         "Microsoft.CodeCoverage"                      = 59;
         "Microsoft.NET.Test.Sdk"                      = 16;
-        "Microsoft.TestPlatform"                      = 607;
+        "Microsoft.TestPlatform"                      = 605;
```
nohwnd, line 19:
```powershell 
$path = "S:\p\vstest3\artifacts\packages\Debug\Shipping"; (ls $path -re -fo | where { $_.FullName -notlike "*Symbols*" -and $_.Name -eq "Nuget.Frameworks.dll" }  | % fullName) -replace [regex]::escape($path)

\Microsoft.TestPlatform.17.8.0-dev\tools\net462\Common7\IDE\Extensions\TestPlatform\NuGet.Frameworks.dll
\Microsoft.TestPlatform.17.8.0-dev\tools\net462\Common7\IDE\Extensions\TestPlatform\TestHostNet\NuGet.Frameworks.dll

\Microsoft.TestPlatform.CLI.17.8.0-dev\contentFiles\any\netcoreapp3.1\NuGet.Frameworks.dll
\Microsoft.TestPlatform.CLI.17.8.0-dev\contentFiles\any\netcoreapp3.1\TestHostNetFramework\NuGet.Frameworks.dll

\Microsoft.TestPlatform.Portable.17.8.0-dev\tools\net462\NuGet.Frameworks.dll
\Microsoft.TestPlatform.Portable.17.8.0-dev\tools\netcoreapp3.1\NuGet.Frameworks.dll
\Microsoft.TestPlatform.Portable.17.8.0-dev\tools\netcoreapp3.1\TestHostNetFramework\NuGet.Frameworks.dll
```

which is count reduction by 2, 2 and 3 just like here.

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -968,3 +968,6 @@ virtual Microsoft.VisualStudio.TestPlatform.ObjectModel.TestObject.ProtectedSetP
 ~static Microsoft.VisualStudio.TestPlatform.ObjectModel.Resources.CommonResources.SourceIncompatible.get -> string
 Microsoft.VisualStudio.TestPlatform.ObjectModel.DirectoryBasedTestDiscovererAttribute
 Microsoft.VisualStudio.TestPlatform.ObjectModel.DirectoryBasedTestDiscovererAttribute.DirectoryBasedTestDiscovererAttribute() -> void
+Microsoft.VisualStudio.TestPlatform.ObjectModel.Framework.FrameworkName.get -> string!
```
nohwnd, line 971: New public api.

src/vstest.console/Internal/ConsoleLogger.cs
```diff
@@ -224,7 +222,7 @@ public void Initialize(TestLoggerEvents events, Dictionary<string, string?> para
         }
 
         parameters.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out _targetFramework);
-        _targetFramework = !_targetFramework.IsNullOrEmpty() ? NuGetFramework.Parse(_targetFramework).GetShortFolderName() : _targetFramework;
+        _targetFramework = Framework.FromString(_targetFramework)?.ShortName ?? _targetFramework;
```
nohwnd, line 225: This returns null when it gets null.

Evangelink (approved):

Evangelink (approved):

src/Microsoft.TestPlatform.ObjectModel/Framework.cs
```diff
@@ -53,56 +62,71 @@ private Framework()
...
 
-                    break;
+            // e.g. .NETFramework,Version=v3.5
+            name = nugetFramework.DotNetFrameworkName;
+            // e.g. net35
+            try
+            {
+                // .NETPortable4.5 for example, is not a valid framework
+                // and this will throw.
+                shortName = nugetFramework.GetShortFolderName();
```
nohwnd, line 112: This throws in some cases when the framework is invalid, parsing will return it as Unsupported, but for example if we construct Portable 4.5 and don't specify Profile, it will construct it as **supported**, but throw when we ask for short name. We use short name as folder for logs, which we figure out when we have a "valid" runtime identifier, so keeping this null should be okay, we check for it in all the places where we use it, and supply the original value.

PR #4706, opened 2023-10-12
Add MSBuild based VSTestTask

nohwnd:
ℹ️This is a copy of #2702  with some changes that I want to try out, it is not replacement for that PR.

VSTestForwardingTask redirects the console output instead of using MSBuild logs. Renaming it will allow to create a new VSTestTask that uses the MSBuild log infrastructure.## Description

## Related issue


src/vstest.console/Internal/MSBuildLogger.cs
```diff
@@ -0,0 +1,166 @@
...
+                    var testDisplayName = e.Result.DisplayName;
+
+                    if (e.Result.DisplayName.IsNullOrWhiteSpace())
+                    {
+                        testDisplayName = e.Result.TestCase.DisplayName;
+                    }
+
+                    stringBuilder.Append(testDisplayName ?? "<unknown>").Append(": ");
+                    AppendFullError(e.Result, stringBuilder);
+                    Output.Error(false, stringBuilder.ToString());
```
KirillOsenkov, line 134: we could maybe emit Json or XML here, then in the MSBuild task read the whole output and parse it as Json, and emit individual errors based on the structure.

nohwnd:
Merged into the original PR. #2702

PR #4717, opened 2023-10-20
Fixed getting current branch info

cvpoienaru:
Fixed getting current branch info


Evangelink (approved):
Do you need to merge to test or have you been able to manually validate?

cvpoienaru:
@Evangelink, replying here as well. As per our offline chat, merging was partially necessary for testing. Local testing won't work since the branch name is a parameter passed by the pipeline, and on the PR build the branch was named "merge". Not empty anymore, but not what was expected either. However, I looked at the official build running after we merged this change and I can see it performed the check and the branch is indeed called 'main'.

PR #4718, opened 2023-10-21
Fix two source-build poison leaks

NikolaMilosavljevic:
Related to https://github.com/dotnet/source-build/issues/3670

This fixes 2 source-build poison leaks: Microsoft.Extensions.DependencyModel.dll and Microsoft.Extensions.FileSystemGlobbing.dll that come from this repo. This was caused by missing package dependencies, which causes source-build to consume reference packages from SBRP repo.

The fix is a simple addition of explicit package dependencies, for versions used by VSTest repo. Source-build will automatically bump these package versions to latest, built in source-build.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -33,6 +33,16 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>30ab651fcb4354552bd4891619a0bdd81e0ebdbf</Sha>
     </Dependency>
+    <!-- Necessary for source-build. This allows the live version of the package to be used by source-build. -->
+    <Dependency Name="Microsoft.Extensions.DependencyModel" Version="3.0.0">
+      <Uri>https://github.com/dotnet/core-setup</Uri>
+      <Sha>7d57652f33493fa022125b7f63aad0d70c52d810</Sha>
+    </Dependency>
+    <!-- Necessary for source-build. This allows the live version of the package to be used by source-build. -->
+    <Dependency Name="Microsoft.Extensions.FileSystemGlobbing" Version="2.0.0">
+      <Uri>https://github.com/dotnet/core-setup</Uri>
+      <Sha>7d57652f33493fa022125b7f63aad0d70c52d810</Sha>
+    </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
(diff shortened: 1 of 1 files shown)


Evangelink (approved):

eng/Version.Details.xml
```diff
@@ -33,6 +33,16 @@
...
+    <!-- Necessary for source-build. This allows the live version of the package to be used by source-build. -->
+    <Dependency Name="Microsoft.Extensions.DependencyModel" Version="3.0.0">
+      <Uri>https://github.com/dotnet/core-setup</Uri>
+      <Sha>7d57652f33493fa022125b7f63aad0d70c52d810</Sha>
+    </Dependency>
+    <!-- Necessary for source-build. This allows the live version of the package to be used by source-build. -->
+    <Dependency Name="Microsoft.Extensions.FileSystemGlobbing" Version="2.0.0">
+      <Uri>https://github.com/dotnet/core-setup</Uri>
+      <Sha>7d57652f33493fa022125b7f63aad0d70c52d810</Sha>
+    </Dependency>
```
ViktorHofer, line 45: Wow, those versions are quite old. Aren't we afraid that full source build could break when using the live versions (8.0.0)?  Would it make sense to minimize the version diff?

NikolaMilosavljevic: Source-build was using live versions until we switched to dependencies-only flow, few months ago: https://github.com/dotnet/installer/pull/17011 By adding these dependencies, we are essentially going back to live versions.

Granted, this big version difference is worrying, in the long-term.

nohwnd (approved):

mthalman (approved):

MichaelSimons:
Does this comment flow to the latest branches or does it need to be manually ported?

NikolaMilosavljevic:
@Evangelink @nohwnd - internal build is failing on one of the legs: https://dev.azure.com/dnceng/internal/_build/results?buildId=2298870&view=logs&j=22e02971-7ade-5002-875e-eef7feb6e7af&t=ddedce3e-d198-5a81-1ae5-9db245410658&l=217

Who can help investigate and fix this, so that we can get a good build that flows to 8.0 release?

nohwnd:
Looking.

nohwnd:
> Does these changes flow to the latest branches or does it need to be manually ported?

They don't automatically flow to main.

PR #4735, opened 2023-11-03
Fix CPU is consumed by polling frequently when there is no subscriber

nohwnd:
## Description

We disabled pre-started testhosts in 17.6.0 because the waiting host could consume large amounts of CPU when data were waiting on the pipeline and there were no consumers on vstest.console side. This is easy to reproduce when you set   `VSTEST_HOSTPRESTART_COUNT = 1` or more, and enable diagnostic logs. 

What happens is that we run init on the pre-started testhost and after we remove the CheckVersionWithHost subscriber, the diag log will push TestMesage:  Logging TestHost Diagnostics in file... which will be wating on the wire and causing tight loop of notifications to subscriber that is not there. 

In logs this shows up like this, with 0ms since the last poll, because the Poll returns immediately, and the notification as well, causing this tight loop: 
```
TpTrace Verbose: 0 : 59440, 12, 2023/11/03, 12:18:50.769, 655225963354, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: NotifyDataAvailable remoteEndPoint: 127.0.0.1:59895 localEndPoint: 127.0.0.1:59892
TpTrace Verbose: 0 : 59440, 12, 2023/11/03, 12:18:50.769, 655225963520, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: Polling on remoteEndPoint: 127.0.0.1:59895 localEndPoint: 127.0.0.1:59892 after 0 ms
```

Instead of this we add a flag that will wait for the maximum of 1 second for a subscriber to subscribe to data reading, and if that does not happen we repeat the loop. 

In log it shows like this: 

```
TpTrace Verbose: 0 : 59440, 12, 2023/11/03, 12:18:50.769, 655225963354, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: NotifyDataAvailable remoteEndPoint: 127.0.0.1:59895 localEndPoint: 127.0.0.1:59892
TpTrace Verbose: 0 : 59440, 12, 2023/11/03, 12:18:51.769, 655225963520, vstest.console.exe, TcpClientExtensions.MessageLoopAsync: Polling on remoteEndPoint: 127.0.0.1:59895 localEndPoint: 127.0.0.1:59892 after 1000 ms
```

Where we correctly wait for the subscriber to subscribe. We also return immediately when we are being cancelled. 

There are probably other causes for this to happen, but this should solve all of them.

## Related issue

Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1896342
Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1835463

### DIFF
--- a/playground/MSTest1/UnitTest1.cs
+++ b/playground/MSTest1/UnitTest1.cs
@@ -3,7 +3,6 @@
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
-using System.Threading;
 namespace MSTest1;
 
 [TestClass]
@@ -12,6 +11,42 @@ public class UnitTest1
     [TestMethod]
     public void TestMethod1()
     {
-        Thread.Sleep(10_000);
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod2()
+    {
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod3()
+    {
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod4()
+    {
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod5()
+    {
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod6()
+    {
+        // Thread.Sleep(1000);
+    }
+
+    [TestMethod]
+    public void TestMethod7()
+    {
+        // Thread.Sleep(1000);
     }
 }
--- a/playground/MSTest1/UnitTest1.cs
+++ b/playground/MSTest1/UnitTest1.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
+using System.Threading;
 namespace MSTest1;
 
 [TestClass]
@@ -11,42 +12,6 @@ public class UnitTest1
     [TestMethod]
     public void TestMethod1()
     {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod2()
-    {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod3()
-    {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod4()
-    {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod5()
-    {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod6()
-    {
-        // Thread.Sleep(1000);
-    }
-
-    [TestMethod]
-    public void TestMethod7()
-    {
-        // Thread.Sleep(1000);
+        Thread.Sleep(10_000);
     }
 }
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/Communication/ICommunicationChannel.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/Communication/ICommunicationChannel.cs
@@ -2,7 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
+
+using Microsoft.VisualStudio.TestPlatform.Utilities;
 
 namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Interfaces;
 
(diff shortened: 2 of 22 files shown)


playground/MSTest1/UnitTest1.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
```
nohwnd, line 4: revert whole folder.

nohwnd:
@cvpoienaru Please fix the tests, and re-enable the PreStartCount to 2.  🙏🙇

src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/Communication/ICommunicationChannel.cs
```diff
@@ -24,5 +27,66 @@ public interface ICommunicationChannel : IDisposable
...
+            e.SafeInvoke(sender, eventArgs!, traceDisplayName);
+        }
+    }
+
+    public bool WaitForSubscriber(int timeoutMilliseconds, CancellationToken cancellationToken)
+    {
+        var _ = timeoutMilliseconds;
+        var __ = cancellationToken;
+        return Event != null;
+        //return _slim.Wait(timeoutMilliseconds, cancellationToken);
```
nohwnd, line 60: Reverting our test code, which we used to cause the original issue.
```suggestion
        return _slim.Wait(timeoutMilliseconds, cancellationToken);
```

src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/Communication/ICommunicationChannel.cs
```diff
@@ -24,5 +27,63 @@ public interface ICommunicationChannel : IDisposable
...
+internal class UnsubscribeToken : IDisposable
+{
+    private readonly Action _unsubscribeCallback;
+    internal UnsubscribeToken(Action value) => _unsubscribeCallback = value;
+
+    public void Unsubscribe() => _unsubscribeCallback();
+    public void Dispose()
+    {
+        _unsubscribeCallback();
+    }
```
nohwnd, line 88: Unused. 
```suggestion
```

src/Microsoft.TestPlatform.CommunicationUtilities/Interfaces/Communication/ICommunicationChannel.cs
```diff
@@ -24,5 +27,63 @@ public interface ICommunicationChannel : IDisposable
     /// Notification from server/client that data is available.
     /// </summary>
     /// <returns>A <see cref="Task"/> implying async nature of the function.</returns>
-    Task NotifyDataAvailable();
+    Task NotifyDataAvailable(CancellationToken cancellationToken);
+}
+
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable
```
nohwnd, line 33: We would have to change way more code then I like.

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
```diff
@@ -85,14 +85,10 @@ public Task NotifyDataAvailable()
             // Try read data even if no one is listening to the data stream. Some server
             // implementations (like Sockets) depend on the read operation to determine if a
             // connection is closed.
```
nohwnd, line 87:
```suggestion
            // Try read data even if no one is listening to the data stream. Some server
            // implementations (like Sockets) depend on the read operation to determine if a
            // connection is closed.
```

This comment is not accurate, we did not read it even before, and it would cause us to lose messages, or have to cache them until a subscriber appears. Instead of that we now have a backpressure

src/Microsoft.TestPlatform.CommunicationUtilities/LengthPrefixCommunicationChannel.cs
```diff
@@ -85,14 +85,10 @@ public Task NotifyDataAvailable()
...
-            if (MessageReceived != null)
+            if (MessageReceived.WaitForSubscriber(1000, cancellationToken))
             {
                 var data = _reader.ReadString();
-                MessageReceived.SafeInvoke(this, new MessageReceivedEventArgs { Data = data }, "LengthPrefixCommunicationChannel: MessageReceived");
-            }
-            else
-            {
```
nohwnd, line 94: This message is unnecessary, and way too much detail. with prestart it is normal that client is waiting for his turn and this would just add yet another message to the log that would always repeat.

src/Microsoft.TestPlatform.CommunicationUtilities/TcpClientExtensions.cs
```diff
@@ -40,17 +41,30 @@ internal static class TcpClientExtensions
                 socketException);
         }
 
+        // PERF: check if verbose is enabled once, and re-use for all calls in the tight loop below. The check for verbose is shows in perf traces
```
nohwnd, line 44: This shown up as small overhead in VS monitoring, but the fact that it shown up says that the check is way more expensive than I would like for a message that would not be logged 99% of time.

cvpoienaru (approved):

MarcoRossignoli (approved):

PR #4737, opened 2023-11-05
Add platform guard

### LINKED ISSUE #4736: TrxLogger throws on certain platforms
## Description

I'm using the TrxLogger to generate TRX reports on the devices I run the unit tests on (Android, iOS etc), but the TRX logger throws an exception when logging the final testrun complete message.
This is because it for some reason writes to the console, but also tries to set the text color of the console:
https://github.com/microsoft/vstest/blob/9e360a6af7c728903882d3468d3036e173d4af6e/src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs#L127
Setting the foreground color isn't supported on several platforms, as indicated here:
https://learn.microsoft.com/en-us/dotnet/api/system.console.foregroundcolor?view=net-7.0

The call should be guarded with a platform check. It would also be nice to be able to set the output level to `None` so nothing is writting to the console at all, since it doesn't make sense on some platforms.

## Steps to reproduce

Use the TRX logger on iOS or Android. Call `OnTestRunComplete` on the TrxWriter.

## Expected behavior

No PlatformNotSupported exception is thrown

## Actual behavior

`PlatformNotSupported` exception, when TRX Logger hits this:
https://github.com/microsoft/vstest/blob/9e360a6af7c728903882d3468d3036e173d4af6e/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs#L413

Screenshot from trx failling with callstack:
image

## Environment

iOS and Android

dotMorten:
## Description

Allows the use of the TRX Logger on other platforms where test runners are being created.

## Related issue

Fixes #4736.

### DIFF
--- a/src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs
@@ -118,7 +118,8 @@ public static class OutputExtensions
 
 #if NET5_0_OR_GREATER
         if (OperatingSystem.IsAndroid() || OperatingSystem.IsIOS() || OperatingSystem.IsBrowser() || OperatingSystem.IsTvOS())
-            return; // Console color not supported on these platforms.
+            // Console color not supported on these platforms.
+            action.Invoke();
 #endif
 
         var previousForegroundColor = Console.ForegroundColor;
--- a/src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs
+++ b/src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs
@@ -118,8 +118,11 @@ public static class OutputExtensions
 
 #if NET5_0_OR_GREATER
         if (OperatingSystem.IsAndroid() || OperatingSystem.IsIOS() || OperatingSystem.IsBrowser() || OperatingSystem.IsTvOS())
+        {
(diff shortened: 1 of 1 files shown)


src/Microsoft.TestPlatform.CoreUtilities/Output/OutputExtensions.cs
```diff
@@ -116,6 +116,11 @@ private static void SetColorForAction(ConsoleColor foregroundColor, Action actio
             return;
         }
 
+#if NET5_0_OR_GREATER
+        if (OperatingSystem.IsAndroid() || OperatingSystem.IsIOS() || OperatingSystem.IsBrowser() || OperatingSystem.IsTvOS())
+            return; // Console color not supported on these platforms.
```
nohwnd, line 121:
```suggestion
            // Console color not supported on these platforms.
            action.Invoke();
```

nohwnd: It is probably better to still write what we wanted to write and just avoid coloring it.

dotMorten: Oh yeah of course. Doh! (Albeit id prefer class libraries didn't write to a console)

Evangelink (approved):

PR #4744, opened 2023-11-09
Remove javascript in `HtmlLogger`

### LINKED ISSUE #4365: HTML logger use <summary> tag instead of JavaScript in template
## Description

Currently in the HTML logger template there is a JavaScript which helps to expand and collapse the test result.

https://github.com/microsoft/vstest/blob/ac4adfabc6df702c8ee0ce028599960ff7b3399f/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt#L14-L24

## Steps to reproduce

`dotnet test --logger html`

## Expected behavior

Test result page does not use additional JavaScript to enable interactions so that it can be used as a test summary page in GitHub actions.

## Actual behavior

The script tag and its contents are rendered as text in the GitHub actions summary.

## Diagnostic logs

None

## Environment

Build version: current version (code is in main branch)

daveMueller:
## Description

Remove javascript in `HtmlLogger`s xslt template

## Related issue

closes #4365 

- [x ] I have ensured that there is a previously discussed and approved issue.

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
+++ b/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
@@ -135,8 +135,9 @@
     <xsl:for-each select ="tp:TestResultCollection">
       <xsl:variable name="Source" select="tp:Id" />
       <xsl:if test="tp:FailedResultList!=''">
-        <details open>
-          <summary><xsl:value-of select = "tp:Source" /></summary>
+        <details>
+            <xsl:attribute name="open"/>
+            <summary><xsl:value-of select = "tp:Source" /></summary>
           <div class ="inner-row" Id="{concat($Source,'-failedResult')}">
             <xsl:for-each select ="tp:FailedResultList/tp:TestResult"><xsl:call-template name ="TestResult"/></xsl:for-each>
           </div>
--- a/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
+++ b/src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
@@ -11,18 +11,7 @@
         <h1>Test run details</h1>
         <xsl:apply-templates select ="/tp:TestRunDetails"/>
       </body>
-      <script language="javascript">
-        function ToggleClass(id) {
-        var elem = document.getElementById(id);
-        if (elem.style.display == "none") {
-        elem.style.display = "block";
-        }
-        else {
-        elem.style.display = "none";
-        }
-        }
-      </script>
-      <style>
+        <style>
         body { font-family: Calibri, Verdana, Arial, sans-serif; background-color: White; color: Black; }
         h2 {
         margin-top: 15px;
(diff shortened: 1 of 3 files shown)


daveMueller:
Not sure why build is failing. Also I'm having issues testing the `HtmlLogger` when compiling `vstest` locally. Any guidance on that would be helpful to completely verify the changes. Currently I only emulated the use of `Html.xslt`.

nohwnd:
Those fails look unrelated to your change, maybe this is just flakiness. Let me know when you are done with your changes, and I can review and help you fix it the tests. 

Testing html logger is probably easiest if you build via build.cmd, it will create nuget in "artifacts\packages\Debug\Shipping\Microsoft.TestPlatform.17.9.0-dev.nupkg" that you can unpack and run vstest.console from there with --logger:html 


Vstest.console is in the package in this path: "tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe"


so e.g.: & "S:\p\vstest2\artifacts\packages\Debug\Shipping\Microsoft.TestPlatform.17.9.0-dev\tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe" --logger:html s:\t\mstest81\bin\Debug\net8.0\mstest81.dll

nohwnd:
re-run, fails now on acceptance: 

image

You can also test that way. But you must always build.cmd before running the acceptance tests, we pick up all the runners from the produced nugets to ensure what we ship is working. So the workflow is a bit clunky. 

Adding some unit tests for the content produced by the html logger is probably not a bad idea.

daveMueller:
Thanks, I let you know once I adapted the tests.

daveMueller:
I fixed the issue and added an integration test. I can also confirm now that the look and feel of the pure `HTML` report is identical to the `javascript` solution. I think it even improved a bit as the black arrows now indicate that the results can be collapsed. Just let me know if I should improve something.

grafik

nohwnd (approved):

nohwnd:
It's perfect, thank you :)

0xced:
Could it be that this change is responsible for this new _stairs_ effect? 🤔 I'm pretty sure I've never seen that before!

Screenshot 2023-11-17 at 00 21 54

daveMueller:
> Could it be that this change is responsible for this new _stairs_ effect? 🤔 I'm pretty sure I've never seen that before!

Yes you are right, thanks for reporting 🙏 . Sorry I made a mistake in the `xslt` template, I'll fix this tomorrow.

daveMueller:
@0xced I checked this again today and it seems this is not related to the changes in this PR. The issue already exists in [v17.9.0-preview-23531-01](https://github.com/microsoft/vstest/releases/tag/v17.9.0-preview-23531-01) which is before these changes. I analyzed this a bit further and the issue doesn't exist in [v17.7.2](https://github.com/microsoft/vstest/releases/tag/v17.7.2) Could  be that a dependency update caused this because there is definitely no change in the `xslt` template between those two version. 

Continuing the analyzis by inspecting the generated `HTML`, it looks like the transformation is loosing a closing `</div>` from the template...
I suggest to open another issue for this.

0xced:
I have submitted #4754 with full reproduction steps. The "only" difference between the correct and the erroneous report is the SDK version (respectively 6.0.417 and 8.0.100).

nohwnd:
Thanks for the analysis here Dave, it helped a lot. The issue was that with the new writer the settings from xslt were not copied, and it output a self-closing <div /> tag when there was no content in error-info. This is a valid XML syntax but not valid HTML syntax. 

image

Same issue solved on SO: https://stackoverflow.com/questions/71357056/c-sharp-xslt-transformer-uses-self-closing-html-tags-which-are-invalid

src/Microsoft.TestPlatform.Extensions.HtmlLogger/Html.xslt
```diff
@@ -11,18 +11,7 @@
...
-        if (elem.style.display == "none") {
-        elem.style.display = "block";
-        }
-        else {
-        elem.style.display = "none";
-        }
-        }
-      </script>
-      <style>
+        <style>
```
Saibamen, line 14: Fix indentation

nohwnd: This PR is merged, if this bothers you, please start a new PR with the fix.

PR #4856, opened 2024-01-29
Update sourcebuild configuration to build net previous and net current

Evangelink:
## Description

Update sourcebuild configuration so we generate both net previous and net current to help with multi-insertion (today we insert both in net8 and net9). The fix is based on my understand of this comment https://github.com/dotnet/sdk/pull/37983#issuecomment-1894304911 but it's possible we will have to do a few iterations to have it all right.

## Related issue

This should be fixing https://github.com/dotnet/sdk/pull/37983


Directory.Build.props
```diff
@@ -64,6 +64,12 @@
     <!-- Naming is based on dotnet/runtime one -->
     <NetFrameworkMinimum>net462</NetFrameworkMinimum>
     <NetCoreAppMinimum>netcoreapp3.1</NetCoreAppMinimum>
+    <!--
+      Arcade is making things hard to work with. This property is unset when it's value is equal to NetMinimum.
```
Evangelink, line 68:
```suggestion
      Arcade is making things hard to work with. This property is unset when its value is equal to NetMinimum.
```

Directory.Build.targets
```diff
@@ -5,10 +5,8 @@
 
   <!-- Source build properties -->
   <PropertyGroup Condition=" '$(DotNetBuildFromSource)' == 'true' ">
```
MichaelSimons, line 7: I just realized the code I referred you to in Roslyn was not 100% correct.  I think this should be changed to the following:

```
<!--
  CI source build leg: this needs to build the current and previous source build TFM. Both are 
  necessary as the output of this leg is used in other CI source build legs. Those could be 
  targeting NetCurrent or NetPrevious hence we must produce both. 
 -->
<PropertyGroup Condition=" '$(DotNetBuildFromSource)' == 'true'  AND '$(DotNetBuildFromSourceFlavor)' != 'Product'">
  <TargetFrameworks>$(NetPrevious);$(NetCurrent)</TargetFrameworks>
</PropertyGroup>

<!-- 
  Source build the product: this is the all up build of the product which needs only NetCurrent
-->
<PropertyGroup Condition=" '$(DotNetBuildFromSource)' == 'true'  AND '$(DotNetBuildFromSourceFlavor)' == 'Product'">
  <TargetFrameworks>$(NetCurrent)</TargetFrameworks>
</PropertyGroup>
```

MichaelSimons:
Capturing a conversation @Evangelink and I had offline.

The reported prebuilts are implicitly loaded by the SDK.  Since the projects are now targeting NetPrevious, these packages are not available in the source-build graph and thus reported as a prebuilt.  With the suggestion I provided in https://github.com/microsoft/vstest/pull/4856#discussion_r1469735402, it would be acceptable to add these as allowed prebuilts in the SourceBuildPrebuiltBaseline.xml because these are only referenced in CI and not the actual product build.

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -14,5 +14,8 @@
     <!-- Caused by dependency on System.ComponentModel.Composition.4.5.0. This version is overridden in full source-build. -->
     <UsagePattern IdentityGlob="System.Security.Principal.Windows/4.5.0" />
     <UsagePattern IdentityGlob="Microsoft.Diagnostics.NETCore.Client/0.2.452401" />
+    <!-- Caused by multi-targeting source build -->
```
MichaelSimons, line 17:
```suggestion
    <!-- Caused by multi-targeting source build for repo builds/CI. Product source build only targets NetCurrent. -->
```

ViktorHofer (commented):
This change seems to have regressed the source-build product build: https://github.com/dotnet/installer/pull/18457.

test/Intent/Intent.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
ViktorHofer, line 5: nit: This regresses perf as TargetFrameworks kicks-off an inner build from the outer build, even though the project is actually single-targeting. I would revert this change and instead find a better pattern in Directory.Build.props to override the TFM for source-build.

Directory.Build.targets
```diff
@@ -3,12 +3,20 @@
...
-    <TargetFrameworks Condition=" '$(TargetFrameworks)' != '' ">$(NetCurrent)</TargetFrameworks>
-    <!-- Force setting TargetFramework to net7.0 only if it was set -->
-    <TargetFramework Condition=" '$(TargetFramework)' != '' ">$(NetCurrent)</TargetFramework>
+  <!--
+    CI source build leg: this needs to build the current and previous source build TFM. Both are
+    necessary as the output of this leg is used in other CI source build legs. Those could be
+    targeting NetCurrent or NetPrevious hence we must produce both.
+  -->
+  <PropertyGroup Condition=" '$(DotNetBuildFromSource)' == 'true' AND '$(DotNetBuildFromSourceFlavor)' != 'Product' ">
+    <TargetFrameworks>$(NetPrevious);$(NetCurrent)</TargetFrameworks>
```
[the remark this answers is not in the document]
Evangelink: Because we insert in both net8 and net9 and it's causing some insertion. That's the solution recommended by @MichaelSimons.

test/Intent/Intent.csproj
Evangelink: If you have some easy suggestion, I am happy to apply otherwise I fear that's not going to be a priority for me given the other tasks I have on my shoulders at the moment.

Evangelink: I couldn't find any easy trick to unset `TargetFramework` and replace it with `TargetFrameworks` in case of source build.

Directory.Build.targets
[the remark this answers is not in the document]
MichaelSimons: This issue is that the dependent repos (e.g. SDK) target different TFM depending on the version.  Therefore to support the repo level source-builds, this repo has to target multiple TFMs.  This pattern is used in other repos like roslyn.  

Fair feedback regarding documentation/tracability.

PR #4866, opened 2024-01-30
Updating version of Microsoft.VisualStudio.Interop to 17.10

MSLukeWest:
## Description

Updating package version of Microsoft.VisualStudio.Interop to a recent 17.10 build.  This is required to fix a MEF composition issue caused by a new interface in 17.10.

Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1948166


nohwnd:
The sdk package cannot be found, do you know which feed we need to add to resolve it? It needs to be a public feed.

Evangelink (approved):

PR #4872, opened 2024-01-31
Upgrade interop libs

nohwnd:
## Description

Another approach for #4866  

The dll is in the vs sdk, and that is where the VSIX package takes it from anyway. So the explicit interop deps is probably unnecessary.


Evangelink (approved):

nohwnd:
This was done to update the interop lib in the linked PR, but the interop lib is causing issues on build, probably it was in the nuget and is not in there anymore or we don't restore. So closing

PR #4878, opened 2024-02-05
Ensure to send a session complete event

drognanar:
Fixes the issue that if starting sessions fails, vstest console will not receive a completion event, thus will keep waiting on the StartSession call indefinitely.

### DIFF
--- a/src/Microsoft.TestPlatform.Client/TestPlatform.cs
+++ b/src/Microsoft.TestPlatform.Client/TestPlatform.cs
@@ -142,7 +142,6 @@ internal class TestPlatform : ITestPlatform
             // sources tells us we should run in-process (i.e. in vstest.console). Because
             // of this no session will be created because there's no testhost to be launched.
             // Expecting a subsequent call to execute tests with the same set of parameters.
-            eventsHandler.HandleStartTestSessionComplete(new());
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.TestPlatform.Client/TestPlatform.cs
```diff
@@ -142,7 +142,6 @@ public TestPlatform()
             // sources tells us we should run in-process (i.e. in vstest.console). Because
             // of this no session will be created because there's no testhost to be launched.
             // Expecting a subsequent call to execute tests with the same set of parameters.
```
Evangelink, line 144: Stale comment. The new one (on the other file) seems to have a little less information but I am not sure if some part are no longer relevant.

Evangelink (commented):
Is it possible to setup some test to capture this case and avoid regressions?

src/Microsoft.TestPlatform.Client/TestPlatform.cs
drognanar: this looks still relevant. it specifies why the method returns false, whereas the other comment specifies why we need to explicitly send a complete event for failures but not for success

drognanar:
Added a unit test and updated the previous Verify that was moved

src/vstest.console/TestPlatformHelpers/TestRequestManager.cs
```diff
@@ -525,6 +525,10 @@ public void ResetOptions()
                 var testSessionStarted = _testPlatform.StartTestSession(requestData, criteria, eventsHandler, sourceToSourceDetailMap, new NullWarningLogger());
                 if (!testSessionStarted)
                 {
+                    // Note: StartTestSession will invoke the HandleStartTestSessionComplete event
+                    // if the test session is started successfully. If it is not started, we need
+                    // to invoke the event here.
```
nohwnd, line 530: Can the comment also explain why we need to do that?

Evangelink (commented):
From what I can grasp lgtm.

nohwnd (approved):

PR #4888, opened 2024-02-07
specify Win10 + maxversiontested to enable xaml APIs to be used in tests running under testhost.exe

### LINKED ISSUE #4887: Update maxversiontested to Xaml Islands can be used in a VSTest project
## Description

The Windows API DesktopWindowXamlSource requires the process to identify the version of the operating system it was tested on by specifying this in it manifest via [maxversiontested](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests?redirectedfrom=MSDN) (dot net had the [same problem](https://github.com/dotnet/sdk/issues/1899)). Since tests run under testhost.exe, it must specify this to enable tests that use that XAML API to function.

```cpp
    if (CQuirksMode2::QuirkBlockXamlIslands())
    {
        // Xaml Islands are available from 19H1( 10.0.18226.0 ) release onwards.
        ::RoOriginateError(E_UNEXPECTED,
````

## Steps to reproduce

Create a test that ties to construct DesktopWindowXamlSource. 

## Expected behavior

The test can run.

## Actual behavior

the test fails due to a fail fast in Xaml

## Diagnostic logs

I've got stack traces and source for those interesetd.

## Environment

Windows 10 or above. 

## Here is an example manifest that is suppoorted.
```xml
<?xml version='1.0' encoding='utf-8' standalone='yes'?>
<assembly
  manifestVersion='1.0'
  xmlns:asmv3='urn:schemas-microsoft-com:asm.v3'
  xmlns='urn:schemas-microsoft-com:asm.v1'>
    <compatibility xmlns='urn:schemas-microsoft-com:compatibility.v1'>
        <application>
            <maxversiontested Id='10.0.18362.0'/>
            <maxversiontested Id='10.0.19041.0'/>
            <maxversiontested Id='10.0.22000.0'/>
            <supportedOS Id='{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}' />
        </application>
    </compatibility>
    <application xmlns='urn:schemas-microsoft-com:asm.v3'>
        <windowsSettings>
            <dpiAware xmlns='http://schemas.microsoft.com/SMI/2005/WindowsSettings'>true/PM</dpiAware>
            <dpiAwareness xmlns='http://schemas.microsoft.com/SMI/2016/WindowsSettings'>PerMonitorV2, PerMonitor</dpiAwareness>
        </windowsSettings>
    </application>
</assembly>
```

ChrisGuzak:
## Description

Resolves #4887 

## Related issue

### DIFF
--- a/src/testhost/app.manifest
+++ b/src/testhost/app.manifest
@@ -7,6 +7,9 @@
     <application>
       <!-- Windows 10 -->
       <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+      <maxversiontested Id='10.0.18362.0'/>
+      <maxversiontested Id='10.0.19041.0'/>
+      <maxversiontested Id='10.0.22000.0'/>
       <!-- Windows 8.1 -->
       <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
       <!-- Windows Vista -->
@@ -15,11 +18,6 @@
       <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
       <!-- Windows 8 -->
       <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
-      <!-- Windows 10 -->
-      <supportedOS Id='{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}' />
-      <maxversiontested Id='10.0.18362.0'/>
-      <maxversiontested Id='10.0.19041.0'/>
-      <maxversiontested Id='10.0.22000.0'/>
     </application>
   </compatibility>
   <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
--- a/src/testhost/app.manifest
+++ b/src/testhost/app.manifest
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <!-- Adding same as used in TPv1 Refer: /src/vset/Agile/TestPlatform/RocksteadyCLI/app.manifest -->
 <assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
   <!-- The assemblyIdentity node is needed to fix a binding problem on Win2003 (PS72518) -->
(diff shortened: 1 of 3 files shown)


nohwnd:
@MSLukeWest  could you double check this please?

src/testhost/app.manifest
```diff
@@ -7,6 +7,9 @@
     <application>
       <!-- Windows 10 -->
       <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+      <maxversiontested Id='10.0.18362.0'/>
```
MSLukeWest, line 10: Two questions:
1) Have you confirmed that this doesn't prevent tests from running on RS5 (10.0.17763)?
2) Should newer versions be added as well? Like 22621 (22H2) and 22631 (23H2)?

ChrisGuzak: @MSLukeWest I've not tested this on older versions, but I understand this will have no effect on them. and newer versions should not be needed... @asklar can you help me explain why? I recall you (or Gov?) recommended multiple to support running on intermediate versions. I read [this bug](https://task.ms/33672605) but don't understand its implications or if it relates to this quirk or others. In [your sample](https://github.com/asklar/xaml-islands/blob/31c4e0c2e30eca9b90790dd83ed9062cd66bcba5/inc/xaml-islands.manifest#L10) you only specify one value `10.0.18226.0` (the value from the quirk that uses `ENABLED_VERSION_LT`). In my testing that works on testhost.exe.

```xml
          <QUIRK NAME="BlockXamlIslands" CODE_ID="262" ENABLED_VERSION_LT="10.0.18226.0" CONTACT_ALIAS="turekhe" ID="{57064B35-BE7D-4111-A3D0-FF077347F207}">
            <DESCRIPTION>
              Xaml Islands were first shipped in RS5. It was just in preview mode and an incomplete feature.
              However, Islands in 19H1 are more stable and complete. 
              Hence Islands are officially available in 19H1 and onward releases and are blocked on all prior releases.
            </DESCRIPTION>
          </QUIRK>
```

asklar: the original bug I had filed is https://task.ms/33687330
IIUC you have to specify 18362 for xaml islands, and there's another value for segoe ui variable support. Adding a later value doesn't enable the features unlocked by an earlier value which is what makes this maxversiontested flag so confusing. @DHowett might remember this more

ChrisGuzak: specifying `18226` works in my testing.

ChrisGuzak: @MSLukeWest per the email thread I've updated this change to use `10.0.18362.0`, the first public release of the os after `10.0.18226.0`. 

its ready to go so please merge it.

MSLukeWest (approved):

src/testhost/app.manifest
MSLukeWest: Approved, @nohwnd please merge for us

nohwnd (approved):

nohwnd:
Merged, thank you both!

PR #4969, opened 2024-04-09
Fix single quote and space in F# pretty methods

### LINKED ISSUE #4313: Test discovery problem when single quote is used in F# method name
## Description

When specifying a name for a test method with the double backtick syntax, adding a single quote is allowed by the compiler. This, however, causes some issue with the test discovery and it prevents not only the particular test from being discovered, but also all other tests in the test file defined below it.

Running `dotnet test` outputs the following warning:
> [MSTest][Discovery][C:\path\to\solution\SingleQuoteTestName\bin\Debug\net7.0\SingleQuoteTestName.dll] MSTestAdapter failed to discover tests in class 'SingleQuoteTestName.TestClass' of assembly 'C:\path\to\solution\SingleQuoteTestName\bin\Debug\net7.0\SingleQuoteTestName.dll' because Whitespace is not valid in a ManagedName (pos: 19).

Given this warning it is hard to narrow it down immediately to the single quote being the problem. If you have many other tests, it is easy to miss this as the test run will not fail, rather it will just run the tests it managed to discover and succeed if those pass.

In my initial attempts to reproduce this issue, it seemed like having an odd number of single quotes was causing the issue whereas an even number was OK. After some further experiments, I unfortunately couldn't really narrow it down to some fixed pattern for when it happens and when it doesn't happen.

## Steps to reproduce

Simply create a blank F# test project with no other dependencies and paste this in the `Tests.fs` file.

```
namespace SingleQuoteTestName

open System
open Microsoft.VisualStudio.TestTools.UnitTesting

[<TestClass>]
type TestClass () =

    [<TestMethod>]
    member _.``test that doesn't pass`` () =
        Assert.IsTrue(false);

    [<TestMethod>]
    member _.``test that passes`` () =
        Assert.IsTrue(true);
```

Then, run `dotnet test` and observe the output.

## Expected behavior

All tests with names accepted by the compiler are discovered and executed.

## Actual behavior

The single quote in the name of the first test method is preventing it from being discovered as well as all other tests in this file.

## Environment

Replicated this both on .NET 6 (6.0.13) & .NET 7 (7.0.2).

[AB#1819417](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1819417)

nohwnd:
## Description
Whitespace is not valid in method names and the parser was not skipping over escaped single tick `\'`, so the method name from F# ``` ``don't pass`` ``` would be output as ` 'don\'t pass' ` in CIL. And the parser would parse `'don\'` as being quoted, and `t pass` as invalid method part because there is whitespace.

## Related issue

Fix #4313 

### DIFF
--- a/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameParser.cs
+++ b/src/Microsoft.TestPlatform.AdapterUtilities/ManagedNameUtilities/ManagedNameParser.cs
@@ -82,12 +82,21 @@ public class ManagedNameParser
     {
         var i = start;
         var quoted = false;
+        char? previousChar = null;
+        // Consume all characters that are in single quotes as is. Because F# methods wrapped in `` can have any text, like ``method name``.
+        // and will be emitted into CIL  as 'method name'.
+        // Make sure you ignore \', because that is how F# will escape ' if it appears in the method name.
         for (; i < managedMethodName.Length; i++)
         {
-            var c = managedMethodName[i];
-            if (c == '\'' || quoted)
+            if ((i - 1) > 0)
             {
-                quoted = c == '\'' ? !quoted : quoted;
+                previousChar = managedMethodName[i - 1];
+            }
+
+            var c = managedMethodName[i];
+            if ((c == '\'' && previousChar != '\\') || quoted)
+            {
+                quoted = (c == '\'' && previousChar != '\\') ? !quoted : quoted;
                 continue;
             }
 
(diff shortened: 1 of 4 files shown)


nohwnd:
We don't check for any additional characters in to unparse the method name from CIL, so ``` ``one \ two``  ```  we appear as `'one \\ two'` in CIL and its managed name will be `one \\ two`. 

Leaving this for another issue, if someone can confirm this is causing problems, or confirm what are the exact rules that F# uses to escape those names. 

/cc @vzarytovskii

nohwnd:
Moved the issue here: https://github.com/microsoft/vstest/issues/4972 because this PR solves the part where discovery cannot finish. With this fix it produces a managed name, but still not a managed name that can be used to lookup the method by name via reflection, which is a different problem with a bigger scope.

vzarytovskii:
FYI, found the exact grammar for it:
```antlr
token ident =
| ident-text For example, myName1
| `` ( [^'`' '\n' '\r' '\t'] | '`' [^ '`' '\n' '\r' '\t'] )+ ``
```

PR #4999, opened 2024-04-25
chore: Add more details to acquistion section.

voroninp:
It's was not very clear what to do with the NuGet package containing test logger.


nohwnd (approved):

nohwnd:
Thank you.

voroninp:
@nohwnd , do I get it right, that `[FrinedlyName]` attribute [implicitly requires](https://github.com/microsoft/vstest/blob/65d9344b73c9cec20b4f42fc27e6d9ace5ca3c1c/src/Microsoft.TestPlatform.CrossPlatEngine/Client/TestLoggerManager.cs#L168) `[ExtensionUri]`?

If so, I'd also suggest adding this to the documentation.

PR #5018, opened 2024-05-03
Add option to not share .NET Framework testhosts

nohwnd:
Adds runsettings option to disable sharing testhosts:

```xml

<DisableSharedTestHost>true</DisableSharedTestHost>
```

and ENV: 

```
VSTEST_DISABLE_SHARING_NETFRAMEWORK_TESTHOST=1
```

This is useful when solution is using both MSTest v2 and MSTest v3 and would like to run each project in isolated process even when they are not running in parallel, or not disabling appdomains. Otherwise the extension logic will send both extension dlls to the host for initialization and it will fail to find the extension and to run the tests. 

This option only affects .NET Framework test projects that don't enable parallelization, or don't disable app domains. Such projects are sharing host, others are not.

Adds option to skip default adapters (the extensions that are place in Extensions folder in our VisualStudio folder).

```xml
<SkipDefaultAdapters>true</SkipDefaultAdapters>
```

This allows projects to run even if external extensions (such as Chutzpah) add dependencies that are incompatible with the projects. Without this option all extensions from the built in folder are always included in the run.

Fix https://github.com/microsoft/vstest/issues/3475

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
@@ -379,7 +379,7 @@ public class DefaultTestHostManager : ITestRuntimeProvider2
         _targetFramework = runConfiguration.TargetFramework;
         _testHostProcess = null;
 
-        Shared = !runConfiguration.DisableAppDomain;
+        Shared = !runConfiguration.DisableAppDomain && runConfiguration.EnableSharedTestHost;
         _hostExitedEventRaised = false;
 
         IsInitialized = true;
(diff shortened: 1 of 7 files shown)


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -379,7 +379,7 @@ public void Initialize(IMessageLogger? logger, string runsettingsXml)
         _targetFramework = runConfiguration.TargetFramework;
         _testHostProcess = null;
 
-        Shared = !runConfiguration.DisableAppDomain;
+        Shared = !runConfiguration.DisableAppDomain && runConfiguration.EnableSharedTestHost;
```
nohwnd, line 382: This is the meat of the PR.

PR #5081, opened 2024-06-03
Use built in sha1 for id generation

### LINKED ISSUE #2762: Improve performance of ID generation
We have noticed that the ID generation is having an heavy impact on performances of MSTest discovery/run, let's investigate what we can improve and if there are some quick wins let's do them directly.

AB#2050310

nohwnd:
## Description

Use built-in sha1 and slice for id generation.

image
image

Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2050310
Fix https://github.com/microsoft/testfx/issues/2762


MarcoRossignoli (approved):

nohwnd:
@evangelink You are a code owner here, but I am just changing the underlying implementation and no tests were changed, and I also checked that the produced trx from old and new mstest SDK  produces the same guids, so all should be good.

PR #5101, opened 2024-06-15
Bump to macos-12 build image

akoeplinger:
AzDO will remove macos-11 in about two weeks: https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#recent-updates


akoeplinger:
This also needs to be backported to all servicing branches.

nohwnd (approved):

PR #5109, opened 2024-06-18
Add option to ignore tests

nohwnd:
Add option to ignore tests.


nohwnd:
yaml is valid, has no effect here.

PR #5177, opened 2024-08-12
Wait for Discovery to initialize before Cancelling it

nohwnd:
## Description

Because discovery intialization takes time until the populates the _currentDiscoveryRequest, and any call to CancelDiscovery in the meantime would just return, we block until the request is populated, and then cancel. This is workaround for race condition that happens when you issue Discovery and DiscoveryCancel quickly in succession. 

## Related issue

Fix https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1579525


MarcoRossignoli (approved):
I don't know if we can have a test but looks complex, I got the idea

nohwnd:
I tested it via vstest console wrapper so test is possible, but it is quite flaky, so I would rather avoid it.

PR #10622, opened 2024-12-10
Set NetCurrent so that it doesn't roll forward automatically

ViktorHofer:
The VSTest repo can't use a floating NetCurrent value as the NetCurrent TFM is actually hardcoded in a number of places => .nuspec files. Set it to net9.0 explicitly so that the build continues to work with a .NET 10 SDK and Arcade 10 SDK (which is used inside the VMR).

Don't hardcode the value when building from source as in that configuration, the TFM needs to actually target the very latest.

## Description

## Related issue

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -50,6 +50,8 @@
 
   <!-- The TFMs to build and test against. -->
   <PropertyGroup>
+    <!-- Naming is based on dotnet/runtime one -->
+    <NetFrameworkMinimum>net462</NetFrameworkMinimum>
     <!-- When not building from source, hardcode the NetCurrent TFM value as it is hardcoded in a number of places, i.e. nuspec files. -->
     <NetCurrent Condition="'$(DotNetBuildSourceOnly)' != 'true'">net9.0</NetCurrent>
     <NetCoreAppMinimum>netcoreapp3.1</NetCoreAppMinimum>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -50,8 +50,8 @@
 
   <!-- The TFMs to build and test against. -->
   <PropertyGroup>
-    <!-- Naming is based on dotnet/runtime one -->
-    <NetFrameworkMinimum>net462</NetFrameworkMinimum>
+    <!-- When not building from source, hardcode the NetCurrent TFM value as it is hardcoded in a number of places, i.e. nuspec files. -->
+    <NetCurrent Condition="'$(DotNetBuildSourceOnly)' != 'true'">net9.0</NetCurrent>
```
nohwnd, line 54: The change does not look right, did you mean to add the property rather than replacing .net framework?

```suggestion
    <!-- Naming is based on dotnet/runtime one -->
    <NetFrameworkMinimum>net462</NetFrameworkMinimum>
    <!-- When not building from source, hardcode the NetCurrent TFM value as it is hardcoded in a number of places, i.e. nuspec files. -->
    <NetCurrent Condition="'$(DotNetBuildSourceOnly)' != 'true'">net9.0</NetCurrent>
```

ViktorHofer: The NetFrameworkMinimum property is already provided by the Arcade SDK: https://github.com/dotnet/arcade/blob/d5c8bb8726b46b5867961f5d8f56f1b13e72dcb9/src/Microsoft.DotNet.Arcade.Sdk/tools/TargetFrameworkDefaults.props#L27

No need to define it here

Issue #14993, opened 2025-01-29
using globbing pattern doesn't work on windows with forward slashes


DavidVilleneuveAnsys:
On windows, when calling `dotnet test C:/path/to/my/tests/*_Tests.dll` we get the following errors : 

```
Unhandled exception. System.ArgumentOutOfRangeException: length ('-1') must be a non-negative value. (Parameter 'length')
Actual value was -1.
   at System.ArgumentOutOfRangeException.ThrowNegative[T](T value, String paramName)
   at System.ArgumentOutOfRangeException.ThrowIfNegative[T](T value, String paramName)
   at System.String.ThrowSubstringArgumentOutOfRange(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex, Int32 length)
   at vstest.console.Internal.FilePatternParser.SplitFilePatternOnWildCard(String filePattern) in /_/src/vstest.console/Internal/FilePatternParser.cs:line 101
   at vstest.console.Internal.FilePatternParser.GetMatchingFiles(String filePattern) in /_/src/vstest.console/Internal/FilePatternParser.cs:line 75
   at Microsoft.VisualStudio.TestPlatform.CommandLine.CommandLineOptions.AddSource(String source) in /_/src/vstest.console/CommandLine/CommandLineOptions.cs:line 283
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.ArgumentProcessorFactory.<>c__DisplayClass18_0.<WrapLazyProcessorToInitializeOnInstantiation>b__0() in /_/src/vstest.console/Processors/Utilities/ArgumentProcessorFactory.cs:line 280
   at System.Lazy`1.CreateValue()
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.GetArgumentProcessors(String[] args, List`1& processors) in /_/src/vstest.console/CommandLine/Executor.cs:line 283
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.Execute(String[] args) in /_/src/vstest.console/CommandLine/Executor.cs:line 173
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Program.Main(String[] args) in /_/src/vstest.console/Program.cs:line 22
```

This works when using backward slashes. 

I think that since forward slashes work in general when doing other Windows CLI tools, or well, in `dotnet test` when not using globbing.

[...]

PR #15002, opened 2025-02-04
Fix issue #14993 about forward slashes not working on windows when globbing sources

DavidVilleneuveAnsys:
🐛: Using a list of allowed separators (alternate + normal) to search for instead of only using the main separators

## Description

To retrieve the source where we start the globbing pattern, the pattern is first split, then we search for the first Directory separator.

In this search, we only allow for the the principal separator, however, some platforms have alternate separators.

So we use `LastIndexOfAny` instead of `LastIndexOf`

## Related issue

Issue being fixed is here https://github.com/microsoft/vstest/issues/14993 and is flagged as a bug

- [x] I have ensured that there is a previously discussed and approved issue.

### DIFF
--- a/src/vstest.console/Internal/FilePatternParser.cs
+++ b/src/vstest.console/Internal/FilePatternParser.cs
@@ -28,6 +28,7 @@ public class FilePatternParser
     private readonly Matcher _matcher;
     private readonly IFileHelper _fileHelper;
     private readonly char[] _wildCardCharacters = ['*'];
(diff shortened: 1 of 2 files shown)


nohwnd:
making it a draft, waiting for @DavidVilleneuveAnsys to confirm CLA

nohwnd:
Closing without deleting the branch, can't take your code without approving CLA. Please comment below if you'd like to continue on this.

PR #15004, opened 2025-02-07
Fix builds on WinUI and UWP .NET 9 projects

Sergio0694:
## Description

This PR fixes the .targets in the test SDK package to avoid breaking WinUI and UWP .NET 9 apps.

## Related issue

- [X] I have ensured that there is a previously discussed and approved issue.


Evangelink (commented):
@nohwnd We could do a similar trick for WinUI so that if the user refs `Microsoft.NET.Test.Sdk` it's not causing a break of their app. WDYT?

Evangelink (approved):

Evangelink (approved):

Sergio0694:
I've updated the PR to also fix builds for WinUI! We should be good to go now. Thank you! 🙂

Youssef1313 (approved):

Youssef1313:
@nohwnd Any concerns you have here or is it good to merge?

Evangelink (approved):

PR #15054, opened 2025-04-17
Condition property on .NET FX MSBuild

jaredpar:
The `BuildWithNetFrameworkHostedCompiler` property forces a .NET Framework specific compiler package to be used. This property was being unconditionally set which meant it loaded inside of `dotnet build`. In that environment the build task is invalid / unsupported.

In this particular case the task ended up throwing errors trying to connect to the compiler server. That is an expected failure path in the build task and it will fallback silently to csc.exe. That meant this went unnoticed as builds remained functional, just significantly slower.

Related: https://github.com/dotnet/sdk/pull/48557

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -133,8 +133,6 @@
     <DebugType>embedded</DebugType>
     <!-- Default to all packages generating a corresponding symbol package -->
     <IncludeSymbols>true</IncludeSymbols>
-    <!-- Workaround for AD0001 in analyzers with .NET 9. See https://github.com/dotnet/arcade/issues/14311 -->
-    <BuildWithNetFrameworkHostedCompiler Condition="'$(MSBuildRuntimeType)' == 'Full'">true</BuildWithNetFrameworkHostedCompiler>
   </PropertyGroup>
 
   <!-- Sign config -->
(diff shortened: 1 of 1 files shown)


lewing (approved):

rainersigwald (approved):

Directory.Build.props
```diff
@@ -134,7 +134,7 @@
     <!-- Default to all packages generating a corresponding symbol package -->
     <IncludeSymbols>true</IncludeSymbols>
     <!-- Workaround for AD0001 in analyzers with .NET 9. See https://github.com/dotnet/arcade/issues/14311 -->
-    <BuildWithNetFrameworkHostedCompiler>true</BuildWithNetFrameworkHostedCompiler>
+    <BuildWithNetFrameworkHostedCompiler Condition="'$(MSBuildRuntimeType)' == 'Full'">true</BuildWithNetFrameworkHostedCompiler>
```
ViktorHofer, line 137: We shouldn't be setting this property in vstest anymore. That was only necessary for a short period of time during .NET 9. Please also update the comment above. GH doesn't let me do that.

```suggestion
```

ViktorHofer (approved):

nohwnd (approved):

Issue #15069, opened 2025-05-20
Version 17.14.0 fails to run on Net 6


sm-cdecker:
We have a few packages we haven't upgraded yet and our pipeline failed to build and test this morning due to the major upgrade from `17.13.0` to `17.14.0`.
I know Net 6 is end of life and no longer supported, but I didn't expect a major version to break compatibility.

```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Could not find testhost
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Hosting.DotnetTestHostManager.GetTestHostProcessStartInfo(IEnumerable`1 sources, IDictionary`2 environmentVariables, TestRunnerConnectionInfo connectionInfo) in /_/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs:line 410
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.SetupChannel(IEnumerable`1 sources, String runSettings) in /_/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs:line 223
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyDiscoveryManager.InitializeDiscovery(DiscoveryCriteria discoveryCriteria, ITestDiscoveryEventsHandler2 eventHandler, Boolean skipDefaultAdapters) in /_/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyDiscoveryManager.cs:line 149
```

nohwnd:
VSTest is not following semantic versioning, it is following VisualStudio versioning. NET6 tfm is end-of-life for a while, you should be able to stick with microsoft.net.test.sdk 17.13.0 for a (long) while.

Can you tell me more about the crash and what exactly you updated? Which packages or which product?

nohwnd:
I think this is unfortunate "feature" of the package, it also did nothing for years for .net framework 4 (until code coverage was added to it as dependency), but it was still able to restore with it, because it is a common combo of multi tfms, so you probably don't want to condition it.

I am assuming there is no warning because of the buildMulitTargeting, will try.

ForNeVeR:
Folks I think I know what the problem is!

All the havoc is likely caused by the fact that nuget thinks in can use net462-compatible version in .NET 6 (and .NET 7) projects.

[...]

PR #15078, opened 2025-05-28
Make the warning error, and restore previous props so we attempt to run

nohwnd:
## Description

Add targets so we don't fallback on package restore on unsupported tfms, and don't end up restoring wrong tfm by accident. Add props back in so we attempt to run via dotnet test.

## Related issue

Related #15069

### DIFF
--- a/src/package/Microsoft.NET.Test.Sdk/Microsoft.NET.Test.Sdk.nuspec
+++ b/src/package/Microsoft.NET.Test.Sdk/Microsoft.NET.Test.Sdk.nuspec
@@ -31,7 +31,7 @@
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="build\net8.0\" />
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="build\net462\" />
 
-    <!-- Add incompatibility warning, and add the props, in case user decides to ignore it. -->
+    <!-- Add incompatibility error, and add the props, in case user decides to ignore it. -->
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="buildMultiTargeting\netcoreapp2.0\" />
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.targets" target="buildMultiTargeting\netcoreapp2.0\" />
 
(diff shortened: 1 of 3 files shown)


src/package/Microsoft.NET.Test.Sdk/Microsoft.NET.Test.Sdk.nuspec
```diff
@@ -26,13 +31,22 @@
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="build\net8.0\" />
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="build\net462\" />
 
-    <!-- Add incompatibility warning. -->
+    <!-- Add incompatibility warning, and add the props, in case user decides to ignore it. -->
```
nohwnd, line 34:
```suggestion
    <!-- Add incompatibility error, and add the props, in case user decides to ignore it. -->
```

src/package/Microsoft.NET.Test.Sdk/Microsoft.NET.Test.Sdk.nuspec
```diff
@@ -26,13 +31,22 @@
...
 
-    <!-- Add incompatibility warning. -->
+    <!-- Add incompatibility error, and add the props, in case user decides to ignore it. -->
+    <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="buildMultiTargeting\netcoreapp2.0\" />
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.targets" target="buildMultiTargeting\netcoreapp2.0\" />
+
+    <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="build\netcoreapp2.0\" />
     <file src="netstandard2.0\Microsoft.NET.Test.Sdk.targets" target="build\netcoreapp2.0\" />
+
+    <file src="netstandard2.0\Microsoft.NET.Test.Sdk.props" target="buildMultiTargeting\netstandard2.0\" />
```
drognanar, line 41: how come both props and targets need to be added?

drognanar (approved):

PR #15184, opened 2025-07-03
Handle dotnet_root in testhost version aware way

nohwnd:
## Description

Use `VSTEST_DOTNET_ROOT_PATH` and `VSTEST_DOTNET_ROOT_ARCHITECTURE` that will receive the location of dotnet.exe from the dotnet SDK. 

When we find testhost.exe we will set env variables based on the version of testhost, to make sure that old testhost gets dotnet_root and new one gets dotnet_root_<arch> so they can pick up local installation of dotnet.

Related https://github.com/dotnet/sdk/issues/49436

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -494,7 +494,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
 
         args += " " + connectionInfo.ToCommandLineOptions();
 
-        // Create a additional probing path args with Nuget.Client
+        // Create a additional probing path args with Nuget.Clientl
         // args += "--additionalprobingpath xxx"
         // TODO this may be required in ASP.net, requires validation
 
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -513,23 +513,114 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
         // i.e. I've got only private install and no global installation, in this case apphost needs to use env var to locate runtime.
         if (testHostExeFound)
         {
-            string prefix = "VSTEST_WINAPPHOST_";
-            string dotnetRootEnvName = $"{prefix}DOTNET_ROOT(x86)";
-            var dotnetRoot = _environmentVariableHelper.GetEnvironmentVariable(dotnetRootEnvName);
-            if (dotnetRoot is null)
+            // This change needs to happen first on vstest side, and then on dotnet/sdk, so prefer this approach and fallback to the old one.
+            // VSTEST_DOTNET_ROOT v2
+            string? dotnetRootPath = _environmentVariableHelper.GetEnvironmentVariable("VSTEST_DOTNET_ROOT_PATH");
+            if (!StringUtils.IsNullOrWhiteSpace(dotnetRootPath))
             {
-                dotnetRootEnvName = $"{prefix}DOTNET_ROOT";
-                dotnetRoot = _environmentVariableHelper.GetEnvironmentVariable(dotnetRootEnvName);
-            }
+                // This is v2 of the environment variables that we are passing, we are in new dotnet sdk. So also grab the architecture.
+                string? dotnetRootArchitecture = _environmentVariableHelper.GetEnvironmentVariable("VSTEST_DOTNET_ROOT_ARCHITECTURE");
 
-            if (dotnetRoot != null)
-            {
-                EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: Found '{dotnetRootEnvName}' in env variables, value '{dotnetRoot}', forwarding to '{dotnetRootEnvName.Replace(prefix, string.Empty)}'");
-                startInfo.EnvironmentVariables.Add(dotnetRootEnvName.Replace(prefix, string.Empty), dotnetRoot);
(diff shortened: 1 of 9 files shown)


nohwnd:
This was attempted before but the testnost apphost was too old to undestand the architecture specific dotnet_root: https://github.com/microsoft/vstest/pull/3715/files

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -494,7 +494,7 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
 
         args += " " + connectionInfo.ToCommandLineOptions();
 
-        // Create a additional probing path args with Nuget.Client
+        // Create a additional probing path args with Nuget.Clientl
```
nohwnd:
```suggestion
        // Create a additional probing path args with Nuget.Client
```

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/TranslationLayerTests/CustomTestHostTests.cs
```diff
@@ -141,6 +141,7 @@ public void RunTestsWithCustomTestHostLauncherUsesLaunchWhenGivenAnOutdatedITest
     [TestCategory("Windows-Review")]
     [TestCategory("Feature")]
     [RunnerCompatibilityDataSource(AfterFeature = Features.MULTI_TFM)]
+    // [RunnerCompatibilityDataSource("net8.0", "MostDownloaded", "net8.0")]
```
nohwnd:
```suggestion
```

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -1197,7 +1197,7 @@ private void SendMessageAndListenAndReportTestResults(
             EqtTrace.Error("Aborting Test Run Operation: {0}", exception);
             eventHandler.HandleLogMessage(
                 TestMessageLevel.Error,
-                TranslationLayerResources.AbortedTestsRun);
+                TranslationLayerResources.AbortedTestsRun + " " + exception.ToString());
```
Youssef1313, line 1200: Was this change intended for another PR?

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -513,23 +513,115 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+                                startInfo.EnvironmentVariables.Add(dotnetRoot, dotnetRootPath);
+                            }
+                        }
+                    }
+                }
             }
             else
             {
-                EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: Prefix '{prefix}*' not found in env variables");
+                // Fallback, can delete this once the change is in dotnet sdk. because they are always used together.
```
Youssef1313, line 606: Is SDK our only concern here? Don't we insert in Test Explorer as well? Does Test Explorer do anything with `VSTEST_WINAPPHOST_*`?

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -513,23 +513,115 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+                    }
+                    else
+                    {
+                        // Set the architecture specific variable to the environment of the process so it is picked up.
+                        startInfo.EnvironmentVariables.Add(environmentVariableName, dotnetRootPath);
+                    }
+                }
+                else
+                {
+                    // This is an old testhost that built against netcoreapp3.1, it does not understand architecture specific DOTNET_ROOT_<ARCH>, we have to set it more carefully
```
Youssef1313, line 578: When running old testhost (netcoreapp3.1) on 32-bit OS, I think `DOTNET_ROOT(x86)` isn't used. In this case, we should probably set `DOTNET_ROOT` (only for the case of WOW64, not generally if target architecture is x86).

Youssef1313 (approved):

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd: it is only sdk.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
nohwnd: it was failing in the "debugger provides more info test" that you saw in mtp migration, and now it of course does not fail anymore... I'd just keep it for now, we are hiding the exception info there, and I am not sure if that is for good reason.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd: `  string rootVariableName = Environment.Is64BitProcess ? "DOTNET_ROOT" : "DOTNET_ROOT(x86)";`

This is what the code did in dotnet/sdk, so if dotnet.sdk was running as x86 it would set the x86 variable, and the old code here just took that value and replace the prefix. So we are not changing the behavior from what it was before.

so since I am not 100% sure that the DOTNET_ROOT(x86) is not used, I would keep it as it was before. That is keep the code as is in this PR.

nohwnd:
https://github.com/microsoft/vstest/issues/15186

mariam-abdulla (approved):

PR #15211, opened 2025-07-09
Update global.json

nohwnd:
Sign validation fails with SDK being too new for the required msbuild, but testfx should be using the same setup and works, so I try upgrading first. 

https://dnceng.visualstudio.com/internal/_build/results?buildId=2745254&view=logs&j=b11b921d-8982-5bb3-754b-b114d42fd804&t=fb192a8b-e433-5fc8-e2b0-276ab015e7d5

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,9 +1,14 @@
 {
-  "sdk": {
-    "version": "10.0.100-preview.5.25269.23",
-    "rollForward": "minor",
-    "allowPrerelease": false,
-    "architecture": "x64"
+"sdk": {
+    "version": "10.0.100-preview.6.25315.102",
+    "paths": [
+      ".dotnet",
+      "$host$"
+    ],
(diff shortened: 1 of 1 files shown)


global.json
```diff
@@ -30,7 +35,7 @@
       "version": "17.8.0"
     },
     "vswhere": "2.2.7",
-    "dotnet": "10.0.100-preview.5.25269.23"
+    "dotnet": "10.0.100-preview.6.25315.102"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.25325.4"
```
Youssef1313, line 41: I don't know if Arcade version here could be relevant to the issues you have, but maybe worth updating to 10.

nohwnd: https://github.com/microsoft/vstest/pull/15224

PR #15234, opened 2025-07-23
Add Version.Details.props

dkurepa:
https://github.com/dotnet/arcade-services/issues/4998

### DIFF
--- a/eng/Version.Details.props
+++ b/eng/Version.Details.props
@@ -1,0 +1,1 @@
+<!-- will get filled in the next backflow -->


akoeplinger (approved):

Youssef1313 (approved):
Approving while @nohwnd is off.

@dkurepa @akoeplinger I see some repos listed in https://github.com/dotnet/arcade-services/issues/4998#issuecomment-3103573094, but microsoft/testfx isn't amongst them. Any reason why not have similar change in microsoft/testfx as well?

dkurepa:
> Approving while @nohwnd is off.
> 
> @dkurepa @akoeplinger I see some repos listed in [dotnet/arcade-services#4998 (comment)](https://github.com/dotnet/arcade-services/issues/4998#issuecomment-3103573094), but microsoft/testfx isn't amongst them. Any reason why not have similar change in microsoft/testfx as well?

This feature is only required for repos that flow into the VMR, but other repos/branches can opt into this new feature too. We haven't rolled it out yet, but when we do, we'll be sure to let everyone know

PR #15255, opened 2025-08-12
Do half the work in GetManagedName

Youssef1313:
Hopefully this is going to reduce these:

image

image

image

I need to benchmark before/after to double check though.


Evangelink (approved):

Youssef1313:
45% faster, 50% less memory :)

I think there is still much more to optimize in this helper though :)

PR #15266, opened 2025-08-19
Set dotnet_root_<arch> always

nohwnd:
## Description

Always set DOTNET_ROOT_<ARCH> (e.g. DOTNET_ROOT_X64) when running under dotnet test, to make sure that the location of dotnet".exe", propagates to the testhost and child processes when we are on windows, linux and macos. This allows local installations of dotnet to be used to run tests that run child executables. On Windows this is alread in place for a long time when testhost.exe is found, so it can resolve correctly dotnet that is not in program files.

On Linux and MacOS it will be used to run xunit v3 tests.

There is a feature flag to disable this new behavior: `VSTEST_DISABLE_DOTNET_ROOT_ON_NONWINDOWS=1`

## Related issue

Will fix https://github.com/dotnet/sdk/issues/50331 once merged and flown.

Working:

image

Reverting to broken behavior with the feature flag:

image

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -231,6 +231,29 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
         EqtTrace.Verbose($"DotnetTestHostmanager.GetTestHostProcessStartInfo: Platform environment '{_platformEnvironment.Architecture}' target architecture '{_architecture}' framework '{_targetFramework}' OS '{_platformEnvironment.OperatingSystem}'");
 
         var startInfo = new TestProcessStartInfo();
+        startInfo.EnvironmentVariables = environmentVariables ?? new Dictionary<string, string?>(StringComparer.OrdinalIgnoreCase);
+
+        string? dotnetRootPath = _environmentVariableHelper.GetEnvironmentVariable("VSTEST_DOTNET_ROOT_PATH");
+        string? dotnetRootArchitecture = _environmentVariableHelper.GetEnvironmentVariable("VSTEST_DOTNET_ROOT_ARCHITECTURE");
+
+        if (!StringUtilities.IsNullOrWhiteSpace(dotnetRootPath))
+        {
+            if (StringUtils.IsNullOrWhiteSpace(dotnetRootArchitecture))
+            {
+                throw new InvalidOperationException("'VSTEST_DOTNET_ROOT_PATH' and 'VSTEST_DOTNET_ROOT_ARCHITECTURE' must be both always set. If you are seeing this error, this is a bug in dotnet SDK that sets those variables.");
+            }
+
+            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: VSTEST_DOTNET_ROOT_PATH={dotnetRootPath}");
+            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: VSTEST_DOTNET_ROOT_ARCHITECTURE={dotnetRootArchitecture}");
+
+            if (!FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_DISABLE_DOTNET_ROOT_ON_NONWINDOWS))
+            {
+                // Set DOTNET_ROOT_<ARCH> for any run, so it gets propagated to testhost and its child processes, like dotnet run does it. This allows executables that start under testhost to find the path to dotnet
+                // from which we called dotnet test. Before this change we only expected testhost.exe to be in this situation, but with xunit v3 running separate exe under testhost, the need for setting architecture
+                // specific DOTNET_ROOT increases and makes this necessary for users to have good experience.
+                SetDotnetRootForArchitecture(startInfo, dotnetRootPath!, dotnetRootArchitecture);
+            }
+        }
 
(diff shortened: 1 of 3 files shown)


nohwnd:
note for me: 

tested on wsl with rc1 release , patched with the newly built testruntime provider.
cp /mnt/s/p/vstest/artifacts/bin/Microsoft.TestPlatform.TestHostProvider/Debug/netstandard2.0/Microsoft.TestPlatform.TestHostRuntimeProvider.dll dotnet3/sdk/10.0.100-rc.1.25418.110/Extensions/Microsoft.TestPlatform.TestHostRuntimeProvider.dll

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -605,28 +603,6 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
                     }
                 }
             }
-            else
```
nohwnd, line 608: dotnet sdk no longer sets this, it was needed temporarily when we inserted new vstest into sdk, but did not have the respective change there yet.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -560,21 +568,11 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
-                    {
-                        // The variable is already set in the surrounding environment, don't set it, because we want to keep what user provided.
-                    }
-                    else
+                    //
+                    // If the feature flag is set, revert to previous behavior of setting DOTNET_ROOT_<ARCH> only on Windows after we found testhost.exe.
+                    if (FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_DISABLE_DOTNET_ROOT_ON_NONWINDOWS))
                     {
-                        var architectureFromEnv = (Architecture)Enum.Parse(typeof(Architecture), dotnetRootArchitecture, ignoreCase: true);
-                        if (architectureFromEnv == _architecture)
```
nohwnd, line 573: this condition was removed from the new code, we set architecture specific dotnet root, so it should be okay to set always. 

it is possible for us to also set dotnet_root (which is not architecture specific) but we do that more carefully, and in other place.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -506,29 +529,14 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
         // G:\nuget-package-path\microsoft.testplatform.testhost\version\**\testhost.dll
         // G:\tmp\netcore-test\bin\Debug\netcoreapp1.0\netcore-test.dll
         startInfo.Arguments = args;
-        startInfo.EnvironmentVariables = environmentVariables ?? new Dictionary<string, string?>();
```
nohwnd, line 509: moved up, otherwise it was overwriting the env variables we set on the start of the method.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
Youssef1313: > it is possible for us to also set dotnet_root (which is not architecture specific) but we do that more carefully, and in other place.

Do we do that anywhere today already? I would personally avoid completely (well, except for netcoreapp3.1 :)).

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -583,7 +581,7 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
                     // to avoid setting DOTNET_ROOT that points to x64 but is picked up by x86 host.
                     //
                     // Also avoid setting it if we are already getting it from the surrounding environment.
-                    var architectureFromEnv = (Architecture)Enum.Parse(typeof(Architecture), dotnetRootArchitecture, ignoreCase: true);
+                    var architectureFromEnv = (Architecture)Enum.Parse(typeof(Architecture), dotnetRootArchitecture!, ignoreCase: true);
```
Youssef1313, line 584: GH doesn't let me comment on the line I want, but on line 600, we are setting just `DOTNET_ROOT`, which confuses me.

Youssef1313: Ah that's for old testhost, it's fine in that specific case :)

Youssef1313 (approved):
Diff is kinda hard to follow with, but reading the overall logic, LGTM.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd: yes it is only for netcore3.1 based testhost.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd: yes.

drognanar (approved):

PR #15267, opened 2025-08-20
Don't hardcode old vswhere version in global.json

akoeplinger:
arcade provides a newer one now.

### DIFF
--- a/global.json
+++ b/global.json
@@ -29,7 +29,6 @@
     "vs": {
       "version": "17.8.0"
     },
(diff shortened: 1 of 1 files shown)


nohwnd (approved):

nohwnd:
image

This conflict from latest arcade update, I've fixed the conflict by taking the dotnet version, but not taking the vswhere entry. Let's see if that will re-appear on next insertion of arcade.

maybe something more is needed on arcade side?

PR #15279, opened 2025-09-16
Update VSSDK to version with code flow guard

nohwnd:
## Description

Updates to versions of vssdk tools that use control flow guard.

Related https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2550974


nohwnd:
don't merge, we need to use a non-preview version for the VSSDK package.

nohwnd:
IncludeCopyLocalReferencesInVSIXContainer avoids copying resources.dll (that are placed next to the dll) and .xml to be copied into the package. This seems to be artifact of how localization is processed. 

We still copy the resources.dll into the correct respective dirs like `cs`.

Youssef1313 (approved):

PR #15287, opened 2025-10-01
Remove stale copy of S.ComponentModel.Composition from testplatform packages

ViktorHofer:
S.ComponentModel.Composition is no longer used in testplatform packages, so we can remove the stale copy of it. The assembly that was redistributed targets netstandard2.0 and is a PNSE - throws PlatformNotSupportedException. This proves that the assembly never got loaded and is indeed unused in all cases.

This resolves having this stale copy in the SDK layout under i.e. "C:\Program Files\dotnet\sdk\10.0.100-rc.1.25451.107\System.ComponentModel.Composition.dll" which showed up as a difference in https://github.com/dotnet/source-build/issues/5344


ericstj (approved):

ericstj:
Might need to update https://github.com/microsoft/vstest/blob/d9177b1896ddb4d8ff617e3ae6b352486e87f855/eng/verify-nupkgs.ps1#L22-L23

nohwnd (approved):

PR #15289, opened 2025-10-03
Fixing async apis and reducing the number of threads we block

nohwnd:
- [ ] combining sync and async apis to talk with the server does not work. probably some flag for the connection is not set.
e.g. if we do discoveryAsync, and then do RunTests (sync), we get stuck.
- [ ] capture the output but use task factory to get the output and await closing the process with async to ensure the streams are dumped.

### DIFF
--- a/playground/TestPlatform.Playground/Properties/launchSettings.json
+++ b/playground/TestPlatform.Playground/Properties/launchSettings.json
@@ -1,0 +1,11 @@
+{
+  "profiles": {
+    "TestPlatform.Playground": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "DOTNET_ThreadPool_ForceMinWorkerThreads": "1",
+        "DOTNET_ThreadPool_ForceMaxWorkerThreads": "1"
+      }
+    }
+  }
+}
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
@@ -132,10 +132,22 @@ internal class VsTestConsoleRequestSender : ITranslationLayerRequestSender
 
         _processExitCancellationTokenSource = new CancellationTokenSource();
 
-        // todo: report standard error on timeout
-        await Task.WhenAny(_communicationManager.AcceptClientAsync(), Task.Delay(clientConnectionTimeout)).ConfigureAwait(false);
+        _handShakeSuccessful = false;
+        _handShakeComplete.Reset();
 
-        await HandShakeWithVsTestConsoleAsync().ConfigureAwait(false);
+        try
+        {
+            // todo: report standard error on timeout
+            await Task.WhenAny(_communicationManager.AcceptClientAsync(), Task.Delay(clientConnectionTimeout)).ConfigureAwait(false);
+
+            await HandShakeWithVsTestConsoleAsync().ConfigureAwait(false);
+
+            _handShakeSuccessful = true;
+        }
+        finally
+        {
+            _handShakeComplete.Set();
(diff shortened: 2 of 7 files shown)


src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -120,36 +120,34 @@ public bool WaitForRequestHandlerConnection(int clientConnectionTimeout)
...
         EqtTrace.Info($"VsTestConsoleRequestSender.InitializeCommunicationAsync: Started with client connection timeout {clientConnectionTimeout} milliseconds.");
 
         _processExitCancellationTokenSource = new CancellationTokenSource();
+
         _handShakeSuccessful = false;
         _handShakeComplete.Reset();
-        int port = -1;
+
         try
         {
```
nohwnd, line 139: This code is completely broken. We are waiting for console to connect here, but we did not return from this method, so the port was never given to the code that starts the console. So there is no process that would ever call us back.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleRequestSender.cs
```diff
@@ -120,36 +120,34 @@ public bool WaitForRequestHandlerConnection(int clientConnectionTimeout)
...
-            port = _communicationManager.HostServer(new IPEndPoint(IPAddress.Loopback, 0)).Port;
-            var timeoutSource = new CancellationTokenSource(clientConnectionTimeout);
-            await Task.Run(() =>
-                _communicationManager.AcceptClientAsync(), timeoutSource.Token).ConfigureAwait(false);
+            // todo: report standard error on timeout
+            await Task.WhenAny(_communicationManager.AcceptClientAsync(), Task.Delay(clientConnectionTimeout)).ConfigureAwait(false);
 
-            _handShakeSuccessful = await HandShakeWithVsTestConsoleAsync().ConfigureAwait(false);
```
nohwnd, line 139: This code is completely broken. We are waiting for console to connect here, but we did not return from this method, so the port was never given to the code that starts the console. So there is no process that would ever call us back.

playground/TestPlatform.Playground/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "profiles": {
+    "TestPlatform.Playground": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "DOTNET_ThreadPool_ForceMinWorkerThreads": "1",
+        "DOTNET_ThreadPool_ForceMaxWorkerThreads": "1"
```
nohwnd, line 7: forces us to single threadpool thread to make sure we don't get blocked when running, and don't block threadpool threads.

bendasgfyug:
Not fixing stage pull

playground/TestPlatform.Playground/Properties/launchSettings.json
bendasgfyug: Not asylm

bendasgfyug: > forces us to single threadpool thread to make sure we don't get blocked when running, and don't block threadpool threads.

Thread pool is hard to focus

nohwnd: This caused issue in some other place where we observed complete blockage of vstest because the process set those variables. This is playgroud project that was trying to repro it. I think the problem of our code is well understood now, the seemingly async apis for process observation are sync and block threads. Which pushes thread pool to grow, and thread pool grows slowly. New apis are coming in net11, and fixes for the current apis as well.

nohwnd:

PR #15296, opened 2025-10-10
Disable publishing on RTM

nohwnd:
## Description

When we build final build (RTM), we want to be able to do it multiple times. Darc publishing will publish to isolated feed thanks to the DotnetFinalVersionRelease. But we would still push to test-tools and other places. We don't want to do that in case there is problem with the package, and do the upload only after we validated the package is correct.


azure-pipelines-official.yml
```diff
@@ -248,66 +248,67 @@ extends:
...
-              # Publishes setup VSIXes to a drop.
-              # Note: The insertion tool looks for the display name of this task in the logs.
-              - task: 1ES.MicroBuildVstsDrop@1
-                displayName: Upload VSTS Drop
-                inputs:
-                  dropName: $(VisualStudioDropName)
-                  dropFolder: 'artifacts\VSSetup\Release\Insertion'
-                  accessToken: $(_DevDivDropAccessToken)
-                condition: succeeded()
+                # Publishes setup VSIXes to a drop.
```
Youssef1313: We want to skip publishing to nuget feeds but we will want to public build artifacts right? Or am I missing/misunderstanding something?

PR #15298, opened 2025-10-17
[rel/18.0] Disable DynamicNative instrumentation by default

nohwnd:
## Description

Partially fix https://github.com/dotnet/sdk/issues/50950 by disabling the DynamicNative instrumentation by default for all code coverage versions.

This change applies the workaround 1 to all runs under vstest 18: https://github.com/dotnet/sdk/issues/50950#issuecomment-3410388878

```xml
 <CodeCoverage>
    <EnableDynamicNativeInstrumentation>false</EnableDynamicNativeInstrumentation>
</CodeCoverage>
```

This reduces the blast radius of the linked issue. DynamicNative code coverage is not needed for projects that only deal with .NET (managed) components.

If user explicitly enables DynamicNative they will still see the error until a fixed version of Code Coverage is published that loads covrun64 in a way that is compatible with .NET 10.

This change will be effective in VSTest 18, including VisualStudio 18, and dotnet test in .NET 10 SDK 10.0.1xx.

User can globally opt out from this fix by setting `VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING=1`. Or they can opt out per run by using runsettings and setting the value themselves to the desired value.

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
+++ b/src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
@@ -718,4 +718,85 @@ public class InferRunSettingsHelper
         return !sourceFramework.Name.Equals(Framework.DefaultFramework.Name, StringComparison.OrdinalIgnoreCase)
                && !sourceFramework.Name.Equals(targetFramework.Name, StringComparison.OrdinalIgnoreCase);
     }
+
+    public static bool UpdateCollectCoverageSettings(XmlDocument xmlDocument)
+    {
+        var root = xmlDocument.DocumentElement;
+
+        // TODO: is this good way to find the node, can users have different casing, can they have  uri="datacollector://Microsoft/CodeCoverage/2.0" or  assemblyQualifiedName="Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector, Microsoft.VisualStudio.TraceCollector, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"
+        var dataCollectorNodes = root?.SelectNodes("DataCollectionRunSettings/DataCollectors/DataCollector");
+        if (dataCollectorNodes == null)
+        {
+            return false;
+        }
+        foreach (XmlNode dataCollectorNode in dataCollectorNodes)
+        {
+            var dataCollectorFound = false;
+            foreach (XmlAttribute attribute in dataCollectorNode.Attributes!)
+            {
+                if (attribute.Name.Equals("friendlyName", StringComparison.OrdinalIgnoreCase) &&
+                    attribute.Value.Equals("Code Coverage", StringComparison.OrdinalIgnoreCase))
+                {
+                    dataCollectorFound = true;
+                    break;
+                }
+
+                if (attribute.Name.Equals("uri", StringComparison.OrdinalIgnoreCase) &&
+                    attribute.Value.Equals(CodeCoverageCollectorUri, StringComparison.OrdinalIgnoreCase))
+                {
+                    dataCollectorFound = true;
+                    break;
+                }
+
+                if (attribute.Name.Equals("assemblyQualifiedName", StringComparison.OrdinalIgnoreCase) &&
+                    attribute.Value.IndexOf("Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    dataCollectorFound = true;
+                    break;
+                }
+            }
+
+            if (dataCollectorFound)
+            {
+                var coverageCollectorNode = dataCollectorNode;
+                // Code coverage settings are present, we should update them.
+
+                var dynamicNativeInstrumentationNode = coverageCollectorNode.SelectSingleNode("Configuration/CodeCoverage/EnableDynamicNativeInstrumentation");
+
+                if (dynamicNativeInstrumentationNode == null)
+                {
+                    // EnableDynamicNativeInstrumentation is not set explicitly, we should set it. Whole tree might not exist.
+
+                    var currentNode = coverageCollectorNode;
+                    var paths = "Configuration/CodeCoverage/EnableDynamicNativeInstrumentation".Split('/');
+                    foreach (var nodeName in paths)
+                    {
+                        var found = false;
+                        foreach (XmlNode childNode in currentNode.ChildNodes)
+                        {
+                            if (childNode.Name == nodeName)
+                            {
+                                currentNode = childNode;
+                                found = true;
+                                break;
+                            }
+                        }
+
+                        if (!found)
+                        {
+                            var newNode = xmlDocument.CreateElement(nodeName);
+                            currentNode.AppendChild(newNode);
+                            currentNode = newNode;
+                        }
+                    }
+
+                    currentNode.InnerXml = "False";
+
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -718,4 +718,85 @@ private static bool IsFrameworkIncompatible(Framework sourceFramework, Framework
         return !sourceFramework.Name.Equals(Framework.DefaultFramework.Name, StringComparison.OrdinalIgnoreCase)
                && !sourceFramework.Name.Equals(targetFramework.Name, StringComparison.OrdinalIgnoreCase);
     }
+
+    public static bool UpdateCollectCoverageSettings(XmlDocument xmlDocument)
+    {
+        var root = xmlDocument.DocumentElement;
+
+        // TODO: is this good way to find the node, can users have different casing, can they have  uri="datacollector://Microsoft/CodeCoverage/2.0" or  assemblyQualifiedName="Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector, Microsoft.VisualStudio.TraceCollector, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"
```
nohwnd:
```suggestion
```

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -718,4 +718,85 @@ private static bool IsFrameworkIncompatible(Framework sourceFramework, Framework
         return !sourceFramework.Name.Equals(Framework.DefaultFramework.Name, StringComparison.OrdinalIgnoreCase)
                && !sourceFramework.Name.Equals(targetFramework.Name, StringComparison.OrdinalIgnoreCase);
     }
+
+    public static bool UpdateCollectCoverageSettings(XmlDocument xmlDocument)
```
Youssef1313, line 722: Is this intended to be public?

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -718,4 +718,84 @@ private static bool IsFrameworkIncompatible(Framework sourceFramework, Framework
...
+    {
+        var root = xmlDocument.DocumentElement;
```
nohwnd, line 724: There are some helpers that might help doing this, but wherever I looked they either reuired me to re-parse the xml, or I was not sure about some of the quirks in them, so I ended up doing it rather manually, in a way that works. Identifiers in the xml (attributes and node names) are validated and should be case sensitive, so less work for use here.

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
nohwnd: yes, we don't have IVT there.

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -718,4 +718,84 @@ private static bool IsFrameworkIncompatible(Framework sourceFramework, Framework
...
+
+        var dataCollectorNodes = root?.SelectNodes("DataCollectionRunSettings/DataCollectors/DataCollector");
```
Youssef1313, line 726: I thought we didn't want to touch runsettings and rely on the env variable instead?

nohwnd: It was the other way around, setting it in runsettings makes it much easier to see in the logs that the option was specified.

jakubch1 (approved):

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
Youssef1313: But is this user facing? i.e, it would be in future a binary break to remove this?

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
```diff
@@ -718,4 +718,84 @@ private static bool IsFrameworkIncompatible(Framework sourceFramework, Framework
...
+        if (dataCollectorNodes == null)
+        {
+            return false;
+        }
```
Youssef1313, line 730: What does this mean?
If user is not providing runsettings explicitly but is running `dotnet test --collect "Code Coverage"`, what do we get here?

Is the `--collect` transformed automatically and we will get non-null here?

Youssef1313 (approved):
Approving with a nit question.

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
nohwnd: Yes that is exactly how this work, the argument processors will transform the parameters ( and programatic settings) into runsettings, and then we reach the code that patches the runsettings to have all the defaults. This new setting is the default now.


& "S:\p\vstest\artifacts\packages\Debug\Shipping\Microsoft.TestPlatform.18.0.0-dev\tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe" S:\t\profiler-crash\ProfilerCrash\TestProject1\bin\Debug\net48\TestProject1.dll --diag:Hello.txt **--collect:"Code coverage"**

```
TpTrace Information: 0 : 36328, 1, 2025/10/17, 17:22:09.650, 1658468562003, vstest.console.exe, TestRunRequest.ExecuteAsync: Starting run with settings:TestRunCriteria:
   KeepAlive=False,FrequencyOfRunStatsChangeEvent=10,RunStatsChangeEventTimeout=00:00:01.5000000,TestCaseFilter=,TestExecutorLauncher=
   Settingsxml=<RunSettings>
  <RunConfiguration>
    <ResultsDirectory>S:\p\vstest\TestResults</ResultsDirectory>
    <TargetFrameworkVersion>.NETFramework,Version=v4.8</TargetFrameworkVersion>
  </RunConfiguration>
  <DataCollectionRunSettings>
    <DataCollectors>
      <DataCollector friendlyName="Code coverage" enabled="True">
        <Configuration>
          <CodeCoverage>
            <EnableDynamicNativeInstrumentation>False</EnableDynamicNativeInstrumentation>
          </CodeCoverage>
        </Configuration>
      </DataCollector>
    </DataCollectors>
  </DataCollectionRunSettings>
  <LoggerRunSettings>
    <Loggers>
      <Logger friendlyName="Console" uri="logger://microsoft/TestPlatform/ConsoleLogger/v1" assemblyQualifiedName="Microsoft.VisualStudio.TestPlatform.CommandLine.Internal.ConsoleLogger, vstest.console, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" codeBase="S:\p\vstest\artifacts\packages\Debug\Shipping\Microsoft.TestPlatform.18.0.0-dev\tools\net462\Common7\IDE\Extensions\TestPlatform\vstest.console.exe" enabled="True" />
    </Loggers>
  </LoggerRunSettings>
</RunSettings>

```

This correctly works on the patched version, and does fail when I force the old behavior by setting  $env:VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING=1

src/Microsoft.TestPlatform.Utilities/InferRunSettingsHelper.cs
nohwnd: Merging so I can get build on main, since this is green. I wanted to use the helpers in the current class, but ended up using nothing.

PR #1196, opened 2017-10-12
Fixing the issues related to response file.

singhsarab:

PR #15306, opened 2025-10-31
Fix answer file splitting

nohwnd:
## Description
The code was last touched here https://github.com/microsoft/vstest/pull/1196 fixing some indeterminate issues with the parsing. Seeing the examples in tests it handles quotes differently and backslashes.

However the code runs out of bounds (not a problem but throws index out of range unnecessarilly).

And fails to recognize `\"` that is in quoted string.  It detects the `\"`, but the index is not advanced past that `"`, so in the next steps is incorrectly stops being a quoted string and spaces will split the command into more tokens.

This is problematic for providing /Tests: which provides the whole test name including spaces etc, in format like this:

```
/Tests:"UpdateTest.CheckUpdate("C:\\Data\\Config.xml","C:\\Data\\Config","C:\\Data\\User"),UpdateTest.MyTest(\"Lorem Ipsum\")"
mytests.dll
```

The previous logic would split the parameters to be:
```
/Tests:UpdateTest.CheckUpdate("C:\\Data\\Config.xml","C:\\Data\\Config","C:\\Data\\User"),UpdateTest.MyTest(\"Lorem
Ipsum\")
mytests.dll
```

And would fail with `Ipsum\")` test source not found.

New logic correctly splits it to:

```
/Tests:UpdateTest.CheckUpdate("C:\\Data\\Config.xml","C:\\Data\\Config","C:\\Data\\User"),UpdateTest.MyTest(\"Lorem Ipsum\")
mytests.dll
```

I have a much larger file from the issue, but it contains customer data, and the spaces are the only edge case that triggers this problem.

## Related issue

https://developercommunity.visualstudio.com/t/vstest-task-fails-with-Unhandled-Except/10527048

### DIFF
--- a/src/Microsoft.TestPlatform.Utilities/CommandLineUtilities.cs
+++ b/src/Microsoft.TestPlatform.Utilities/CommandLineUtilities.cs
@@ -21,9 +21,9 @@ public static class CommandLineUtilities
 
         try
         {
-            while (true)
+            while (index < args.Length)
             {
-                // skip whitespace
+                // Skip whitespace.
                 while (char.IsWhiteSpace(args[index]))
                 {
                     index++;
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.Utilities/CommandLineUtilities.cs
```diff
@@ -33,49 +33,73 @@ public static bool SplitCommandLineIntoArguments(string args, out string[] argum
...
+
+                                // We processed the escaped \, move to next char.
+                                index++;
+                                continue;
+                            }
+
+                            // If the char after '\' is a '"', output '"' and skip over to the next char.
+                            if (index <= args.Length && args[index] == '"')
                             {
                                 currentArg.Append('"');
```
nohwnd, line 76: Main bug was here imho. We handled the quote, but we don't advance to the next character because the logic is trying to be too clever with the `\` being or not being doubled.

The " is then processed in the next loop, and stops the string from being quoted, even though the quote was escaped.

fhnaseer (approved):

PR #15310, opened 2025-11-05
Update build tools to 17.11.48 to be source buildable

nohwnd:
Updating as suggested here: https://github.com/dotnet/dotnet/pull/3009#issuecomment-3490487545


ViktorHofer:
@nohwnd this needs to get backported into `rel/18.0` to unblock https://github.com/dotnet/dotnet/pull/3236

PR #15382, opened 2026-01-30
Clean generated test projects after build

nohwnd:


nohwnd:
Only affects internal build: https://dnceng.visualstudio.com/internal/_build/results?buildId=2891347&view=results

PR #15392, opened 2026-02-05
Updating System.Collections.Immutable package reference to version 9.0.0

MSLukeWest:
## Description

Starting in VS 18.4, VS and MSBuild now reference this newer version of System.Collections.Immutable.  Now, when the UWP test host runtime provider tries to load MSBuild assemblies it will fail, because the test platform's copy of System.Collections.Immutable doesn't have methods that MSBuild uses:

```
Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Failed to launch testhost with error: System.AggregateException: One or more errors occurred. ---> System.TypeInitializationException: The type initializer for 'Microsoft.Build.Shared.XMakeElements' threw an exception. ---> System.MissingMethodException: Method not found: 'System.Collections.Frozen.FrozenSet`1<!!0> System.Collections.Frozen.FrozenSet.Create(System.Collections.Generic.IEqualityComparer`1<!!0>, System.ReadOnlySpan`1<!!0>)'.
   at Microsoft.Build.Shared.XMakeElements..cctor()
```

## Related issue

Internal DevDiv bug 2692656

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -65,7 +65,7 @@
     <MicrosoftVSSDKBuildToolsVersion>17.14.2119</MicrosoftVSSDKBuildToolsVersion>
     <MicrosoftWin32RegistryVersion>5.0.0</MicrosoftWin32RegistryVersion>
     <NewtonsoftJsonVersion>13.0.3</NewtonsoftJsonVersion>
-    <SystemCollectionsImmutableVersion>9.0.0</SystemCollectionsImmutableVersion>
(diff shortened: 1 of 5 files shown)


MSLukeWest:
> Problem here is with the .net framework testhost we insert into sdk, the dependencies are determined by the packaging project (Microsoft.TestPlatform.CLI), and that project is not updating the version of the System.Collections.Immutable package.
> 
> We can quick fix, but a bigger fix would preferable to align this.

@nohwnd - What would the quick fix you mention be?

nohwnd (approved):

PR #15417, opened 2026-02-24
Revert "Updating System.Collections.Immutable package reference to version 9.0.0"

nohwnd:
Reverts microsoft/vstest#15392


nohwnd:
broke VS, reverting until I fix

nohwnd:
reverting in VS instead to make it less complicated, only vs shipment is broken, dotnet cli and microsoft.testplatform works.

PR #15419, opened 2026-02-24
Run tests against vsix runner

nohwnd:
## Description

Run tests locally against the insertion we put into VS, because it is produced differently than the microsoft.testplatform package, and we want to avoid configuration drift and breaking VS.


nohwnd:
@akoeplinger not sure if this will break source-build / vmr, I know we had problems before when build script was checking vsix (the package we insert into VS) that is not built in all the places where we build.

PR #15442, opened 2026-03-06
Resolve dotnet path from DOTNET_HOST_PATH in VSTestTask

nohwnd:
## Summary

VSTestTask (legacy) was hardcoding `dotnet` as the process filename, assuming it would be found on PATH. This changes it to resolve the dotnet executable using the same strategy as `VSTestTask2.GenerateFullPathToTool()`:

1. Check `DOTNET_HOST_PATH` environment variable
2. Check current directory for dotnet executable
3. Search `PATH`
4. Fall back to bare `dotnet` name

## Validation

- Built with `./build.cmd -pack`
- Ran `RunDotnetTestWithCsproj` acceptance test - 2/2 passed

### DIFF
--- a/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
+++ b/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
@@ -247,6 +247,6 @@ internal static class TestTaskUtils
     internal static string? ResolveDotnetPath()
     {
         var dotnetHostPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
-        return Path.GetFullPath(dotnetHostPath);
+        return string.IsNullOrEmpty(dotnetHostPath) ? null : Path.GetFullPath(dotnetHostPath);
     }
 }
--- a/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
+++ b/src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Runtime.InteropServices;
 
 using Microsoft.Build.Utilities;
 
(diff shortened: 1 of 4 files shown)


nohwnd:
Should be good to go if it passes acceptance tests, we are already doing the same in vstest task 2

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -237,4 +239,50 @@ public static string CreateCommandLineArguments(ITestTask task)
...
         return builder.ToString();
     }
+
+    /// <summary>
+    /// Resolves the full path to the dotnet executable by checking DOTNET_HOST_PATH,
+    /// current directory, and PATH in that order.
+    /// </summary>
+    /// <param name="dotnetExe">The platform-specific dotnet executable name (e.g. "dotnet.exe" or "dotnet").</param>
+    /// <returns>The resolved full path, or <see langword="null"/> if not found.</returns>
+    internal static string? ResolveDotnetPath(string dotnetExe)
```
Youssef1313: Do we really need this overload? Can't all callsites simply use the parameterless overload?

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -237,4 +239,50 @@ public static string CreateCommandLineArguments(ITestTask task)
...
+        foreach (var p in values!.Split(Path.PathSeparator))
+        {
+            var fullPath = Path.Combine(p, dotnetExe);
+            if (File.Exists(fullPath))
+            {
+                return fullPath;
+            }
+        }
+
+        return null;
```
kasperk81: this is redundant. `FileName = TestTaskUtils.ResolveDotnetPath() ?? "dotnet",` already takes care of it, RHS will use dotnet from PATH on both windows and unix
```suggestion
        return null;
```

src/Microsoft.TestPlatform.Build/Tasks/TestTaskUtils.cs
```diff
@@ -237,4 +238,15 @@ public static string CreateCommandLineArguments(ITestTask task)
...
+
+    /// <summary>
+    /// Get path to dotnet from DOTNET_HOST_PATH.
+    /// current directory, and PATH in that order.
+    /// </summary>
+    /// <returns>The resolved full path, or <see langword="null"/> if not found.</returns>
+    internal static string? ResolveDotnetPath()
+    {
+        var dotnetHostPath = Environment.GetEnvironmentVariable("DOTNET_HOST_PATH");
+        return Path.GetFullPath(dotnetHostPath);
```
kasperk81: it will throw `ArgumentNullException` when env var is not set

```suggestion
        return string.IsNullOrEmpty(dotnetHostPath) ? null : Path.GetFullPath(dotnetHostPath);
```

PR #15443, opened 2026-03-06
Move to mstest 4

nohwnd:
## Description

Moved to mstest 4.1, some edits were a bit annoying, with the formatting no longer working for assertions. Also the context moved to be the same directory as deploy, which caused a nice message saying that file is in use, but in reality I was saying 

var destination = samePath;
Copy(samePath, destination)

Fixed unit tests for communication utilities, to no longer reset the timeout to 90s when running in parallel, because then few tests that are unlucky will always run into it and wait 90s for timeout they test. 

Test.cmd before: 09m19s
Test.cmd after: 0m35s

### DIFF
--- a/test/vstest.console.UnitTests/Processors/EnableBlameArgumentProcessorTests.cs
+++ b/test/vstest.console.UnitTests/Processors/EnableBlameArgumentProcessorTests.cs
@@ -211,7 +211,6 @@ public class EnableBlameArgumentProcessorTests
     }
 
     [TestMethod]
-    [ExpectedException(typeof(CommandLineException))]
     public void InitializeShouldThrowIfInvalidParameterFormatIsSpecifiedForCollectDumpOption()
     {
         var invalidString = "CollectDump;sdf=sdg;;as;a=";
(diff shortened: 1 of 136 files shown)


test/vstest.console.UnitTests/Processors/EnableBlameArgumentProcessorTests.cs
```diff
@@ -225,32 +224,7 @@ public void InitializeShouldThrowIfInvalidParameterFormatIsSpecifiedForCollectDu
         _mockEnvronment.Setup(x => x.Architecture)
             .Returns(PlatformArchitecture.X64);
 
-        _executor.Initialize(invalidString);
-        _mockOutput.Verify(x => x.WriteLine(string.Format(CultureInfo.CurrentCulture, CommandLineResources.InvalidBlameArgument, invalidString), OutputLevel.Warning));
```
nohwnd, line 229: removed because the code below did not run in the test, since it throws in the _executor.Initialize method and expects that exception

test/vstest.console.UnitTests/Processors/ShowDeprecateDotnetVStestMessageArgumentProcessorTests.cs
```diff
@@ -13,7 +13,9 @@ public class ShowDeprecateDotnetVStestMessageArgumentProcessorTests
     [TestMethod]
     public void ShowDeprecateDotnetVStestMessageProcessorCommandName()
     {
+#pragma warning disable MSTEST0032 // Assertion condition is always true
         Assert.AreEqual("/ShowDeprecateDotnetVSTestMessage", ShowDeprecateDotnetVStestMessageArgumentProcessor.CommandName);
```
nohwnd, line 17: yeah this is kinda weird, how do I ensure that constant has the given value, maybe you already fixed it in your previews

nohwnd: similar issue in other place with comparing a **long** constant with a `1000` int literal. I get warnings that the assertion is never true, and when I change it to say `1000l` to force the long type, I get warning that the assert is never false.

Evangelink (approved):
You could consider using the metapackage versions for MSTest

PR #15472, opened 2026-03-10
Review simplify compatibility sources, deduplicate tests

### LINKED ISSUE #15454: Deduplicate entries for compatibility tests
Why is there duplicated versions in the table anyway? It helps to make it easier to test previews, plan was to automatically update the table to follow the release cycle of mstest - and other products like xunit), and to offer some flexibility, e.g. when xunit has a most dowloaded version that is different from mstest most downloaded version then we should be able to test them both, but not generate more than needed. 

Currently we don't use compatibility matrix for xunit or nunit at all, but it would be helpful. So simply removing the duplications is not the recommended path forward.

---

When running compatibility tests, and there are some duplicated versions of the dependencies used to test we don't deduplicate the rows that are generated. See how to deduplicate them.

---
Similarly, the compatibility projects are generated at one time in assembly initialize and executed (we are not aware what will run later. And they most likely produce duplicated projects for the entries that have the same versions. See how we resolve the projects from the "LatestPreview" and similar monikers and see if we can reduce the number of projects built by deduplicating them.

```xml
    <!--
      Versions that are used when building projects from TestAssets.sln for compatibility tests. See Invoke-TestAssetsBuild in scripts/build.ps1.
      Exact versions are used to avoid Nuget substituting them by closest match, if we make a typo.
      These versions need to be "statically" readable because we read this file as xml in our build and tests.
    -->
    <!-- <MSTestFrameworkLatestVersion></MSTestFrameworkLatestVersion> is not here, because we don't build MSTest locally, so we don't have access to the latest version. -->
    <MSTestFrameworkLatestPreviewVersion>[3.9.3]</MSTestFrameworkLatestPreviewVersion>
    <MSTestFrameworkLatestStableVersion>[3.9.3]</MSTestFrameworkLatestStableVersion>
    <MSTestFrameworkRecentStableVersion>[3.3.1]</MSTestFrameworkRecentStableVersion>
    <MSTestFrameworkMostDownloadedVersion>[2.2.10]</MSTestFrameworkMostDownloadedVersion>
    <MSTestFrameworkPreviousStableVersion>[2.2.10]</MSTestFrameworkPreviousStableVersion>
    <MSTestFrameworkLegacyStableVersion>[1.4.0]</MSTestFrameworkLegacyStableVersion>
    <!-- Versions that are used to restore previous versions of console, translation layer, and test.sdk for compatibility tests.
    See Invoke-TestAssetsBuild in scripts/build.ps1. Exact versions are used to avoid Nuget substituting them by closest match, if we make a typo.
    These versions need to be "statically" readable because we read this file as xml in our build and tests. -->
    <!-- <VSTestConsoleLatestVersion></VSTestConsoleLatestVersion> is not here, NETTestSdkVersion is used instead, because that is the version of the locally built latest package.  -->
    <VSTestConsoleLatestPreviewVersion>[17.10.0]</VSTestConsoleLatestPreviewVersion>
    <VSTestConsoleLatestStableVersion>[17.10.0]</VSTestConsoleLatestStableVersion>
    <VSTestConsoleRecentStableVersion>[17.9.0]</VSTestConsoleRecentStableVersion>
    <VSTestConsoleMostDownloadedVersion>[17.6.2]</VSTestConsoleMostDownloadedVersion>
    <VSTestConsolePreviousStableVersion>[16.11.0]</VSTestConsolePreviousStableVersion>
    <VSTestConsoleLegacyStableVersion>[15.9.2]</VSTestConsoleLegacyStableVersion>
    <ILAsmPackageVersion>5.0.0</ILAsmPackageVersion>
  </PropertyGroup>
```

Compatibility tests are generating generated test assets for every combination, and they don't reduce the number of projects if there is some duplication in versions. 

See if we can reduce that.

nohwnd:
## Description

Split the compatibility sources to 4 levels, wrapper, runner, testhost and adapter. Review how tests are generated, and reduce the number of tests generated. E.g. mstest source was generating 40 tests for each runner and host, instead of just combination of adapter versions and hosts, which is where adapter runs.

On the other hand RunnerCompatibilitySource generates by default tests for VSIX and inProcess.

## Related issue

Related #15464
Related #15470
Fix #15454

### DIFF
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
@@ -22,8 +22,8 @@ public class ExecutionTests : AcceptanceTestBase
     //TODO: It looks like the first 3 tests would be useful to multiply by all 3 test frameworks, should we make the test even more generic, or duplicate them?
     [TestMethod]
     [TestCategory("Windows-Review")]
-    [MSTestCompatibilityDataSource(InProcess = true)]
-    public void RunMultipleTestAssemblies(RunnerInfo runnerInfo)
+    [MSTestCompatibilityDataSource()]
+    public void RunMultipleTestAssemblies1(RunnerInfo runnerInfo)
     {
         SetTestEnvironment(_testEnvironment, runnerInfo);
 
@@ -32,23 +32,6 @@ public class ExecutionTests : AcceptanceTestBase
         InvokeVsTestForExecution(assemblyPaths, testAdapterPath: null, FrameworkArgValue, string.Empty);
 
         ValidateSummaryStatus(2, 2, 2);
-        ExitCodeEquals(1); // failing tests
-        StdErrHasTestRunFailedMessageButNoOtherError();
-        StdOutHasNoWarnings();
-    }
-
-    [TestMethod]
-    [TestCategory("Windows-Review")]
-    [TestPlatformCompatibilityDataSource]
-    public void RunTestsFromMultipleMSTestAssemblies(RunnerInfo runnerInfo)
-    {
-        SetTestEnvironment(_testEnvironment, runnerInfo);
-
-        var assemblyPaths = BuildMultipleAssemblyPath("MSTestProject1.dll", "MSTestProject2.dll");
-
-        InvokeVsTestForExecution(assemblyPaths, testAdapterPath: null, FrameworkArgValue, string.Empty);
-
-        ValidateSummaryStatus(passed: 2, failed: 2, skipped: 2);
         ExitCodeEquals(1); // failing tests
         StdErrHasTestRunFailedMessageButNoOtherError();
         StdOutHasNoWarnings();
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
@@ -23,7 +23,7 @@ public class ExecutionTests : AcceptanceTestBase
     [TestMethod]
     [TestCategory("Windows-Review")]
     [MSTestCompatibilityDataSource()]
-    public void RunMultipleTestAssemblies1(RunnerInfo runnerInfo)
+    public void RunMultipleTestAssemblies(RunnerInfo runnerInfo)
     {
         SetTestEnvironment(_testEnvironment, runnerInfo);
 
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
@@ -261,6 +261,11 @@ public class CompatibilityRowsBuilder
     private void AddRow(List<RunnerInfo> dataRows, string batch,
         string runnerVersion, string runnerFramework, string hostVersion, string hostFramework, string adapterVersion, string adapter, bool inIsolation)
     {
+        if (adapter != AcceptanceTestBase.MSTEST)
+        {
+            throw new NotSupportedException($"Adapter {adapter} is not supported. Only {AcceptanceTestBase.MSTEST} is supported.");
+        }
+
         RunnerInfo runnerInfo = GetRunnerInfo(batch, runnerFramework, hostFramework, inIsolation);
         runnerInfo.DebugInfo = GetDebugInfo();
         runnerInfo.VSTestConsoleInfo = GetVSTestConsoleInfo(runnerVersion, runnerInfo);
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
@@ -40,7 +40,7 @@ public class CompatibilityRowsBuilder
     }
 
     /// <summary>
-    /// Add run for in-process using the selected .NET Framework runners, and and all selected adapters.
+    /// Add run for in-process using the selected .NET Framework runners, and all selected adapters.
     /// </summary>
     public bool WithInProcess { get; set; }
     // Add runner from VSIX to check the shipment we make into VisualStudio.
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
@@ -261,10 +261,7 @@ public class CompatibilityRowsBuilder
     private void AddRow(List<RunnerInfo> dataRows, string batch,
         string runnerVersion, string runnerFramework, string hostVersion, string hostFramework, string adapterVersion, string adapter, bool inIsolation)
     {
-        if (adapter != AcceptanceTestBase.MSTEST)
-        {
-            throw new NotSupportedException($"Adapter {adapter} is not supported.");
-        }
+
         RunnerInfo runnerInfo = GetRunnerInfo(batch, runnerFramework, hostFramework, inIsolation);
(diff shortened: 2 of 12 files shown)


test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
```diff
@@ -66,23 +63,13 @@ public List<RunnerInfo> CreateData()
     {
         var dataRows = new List<RunnerInfo>();
 
-        if (WithEveryVersionOfRunner)
```
nohwnd, line 69: There were two sources of the data, idea was that TestPlatformCompatibilitySource, has some more powerful api to generate exactly what is needed, but this makes it really hard to understand what it will do, and that source is for experimentation only anyway. So now we simply generate all combinations of the input data (with some special handling for in process  - just because we don't have "unknown" for testhost version.

All in all this is not needed anymore, and was causing bugs for mstest source, which provided all test runner versions, and WithEveryVersionOfRunner = false, which had no effect. So tests were generated for every runner version, which does not make sense.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
```diff
@@ -172,133 +194,77 @@ private static SemanticVersion ParseAndPatchSemanticVersion(string? version)
...
-            AddRow(dataRows, "VSIX", AcceptanceTestBase.LATESTVSIX, AcceptanceTestBase.DEFAULT_RUNNER_NETFX, AcceptanceTestBase.LATEST, hostFramework, AcceptanceTestBase.LATESTSTABLE, inIsolation: false);
-        }
-    }
-
-    private void AddOlderConfigurations(List<RunnerInfo> dataRows)
+    private void AddRows(List<RunnerInfo> dataRows)
     {
-        // Older configurations where the runner, host and adapter version are the same.
-        // We already added the row where all are newest when adding combination with all runners.
         foreach (var runnerVersion in _runnerVersions)
```
nohwnd, line 199: This whole method tries to be less clever. Whatever you specify, it will give you all combinations. Before we had different methods that tried to encode some of the thinking (e.g. when testing testhost, we don't need every version of runner), now instead this is done via the input data - you provide just 1 version of the runner to use, and so you will get 1 combination.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -22,8 +22,8 @@ public class ExecutionTests : AcceptanceTestBase
     //TODO: It looks like the first 3 tests would be useful to multiply by all 3 test frameworks, should we make the test even more generic, or duplicate them?
     [TestMethod]
     [TestCategory("Windows-Review")]
-    [MSTestCompatibilityDataSource(InProcess = true)]
```
nohwnd, line 25: This is now default.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -22,8 +22,8 @@ public class ExecutionTests : AcceptanceTestBase
...
-    public void RunMultipleTestAssemblies(RunnerInfo runnerInfo)
+    [MSTestCompatibilityDataSource()]
+    public void RunMultipleTestAssemblies1(RunnerInfo runnerInfo)
```
nohwnd:
```suggestion
    public void RunMultipleTestAssemblies(RunnerInfo runnerInfo)
```

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
```diff
@@ -37,23 +37,6 @@ public void RunMultipleTestAssemblies(RunnerInfo runnerInfo)
         StdOutHasNoWarnings();
     }
 
-    [TestMethod]
-    [TestCategory("Windows-Review")]
-    [TestPlatformCompatibilityDataSource]
-    public void RunTestsFromMultipleMSTestAssemblies(RunnerInfo runnerInfo)
```
nohwnd, line 43: Same tests as above, just multiplied to 100+ tests which can test combinations of versions that are already released.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/ExecutionTests.cs
nohwnd: Will press commit, when / if the build fails.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
```diff
@@ -145,9 +132,44 @@ public List<RunnerInfo> CreateData()
...
+            }
+        }
+
+        // We added all the version types together (would have been better to have this property as array originally, but that would complicate other code), now deduplicate the names to make it less confusing for user.
+        // Latest, Latest, Latest, LatestPreview -> Latest, LatestPreview
+        foreach (var r in distinctRows.Values)
+        {
+            r.VSTestConsoleInfo!.VersionType = string.Join(", ", r.VSTestConsoleInfo!.VersionType!.Split(',').Distinct());
+            r.TestHostInfo!.VersionType = string.Join(", ", r.TestHostInfo!.VersionType!.Split(',').Distinct());
+            r.AdapterInfo!.VersionType = string.Join(", ", r.AdapterInfo!.VersionType!.Split(',').Distinct());
```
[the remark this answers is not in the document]
nohwnd: No, the original value is just single string, we are re-using property on they type in stupid way in order to avoid changing the types, because the VersionType value is never inspected after this point, and we just use it for making the message more informative.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/CompatibilityRowsBuilder.cs
```diff
@@ -145,9 +132,44 @@ public List<RunnerInfo> CreateData()
...
 
-        // We use ToString to determine which values are unique. Not great solution, but works better than using records.
+        // Figure out the distinct rows, and shrink the values, so if we have multiple rows that use the same versions and same setup, and only differ in how the version types
+        // are called (e.g. Latest and LatestStable have the same version), then we get just single row, with description for both.
+        // like RunMultipleTestAssemblies1 (Row: 0, Matrix, Runner = net10.0, TargetFramework = net462, InIsolation,  vstest.console = 18.6.0-dev [Latest],  Testhost = 18.6.0-dev [Latest],  MSTest = 3.9.3 [LatestPreview, LatestStable])
         var distinctRows = new Dictionary<string, RunnerInfo>();
-        rows.ForEach(r => distinctRows[r.ToString()] = r);
+        foreach (var r in rows)
+        {
+            var key = $"{r.Batch}|{r.RunnerFramework}|{r.VSTestConsoleInfo!.Version}|{r.TargetFramework}|{r.TestHostInfo!.Version}|{r.InIsolationValue}|{r.AdapterInfo!.Name}|{r.AdapterInfo.Version}";
```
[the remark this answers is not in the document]
nohwnd: No, your reasoning is wrong, the debug info is the same for all the rows. Also what a convoluted way to say I am correct in the first 8 sentences.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Extension/WrapperCompatibilityDataSource.cs
```diff
@@ -0,0 +1,73 @@
...
+    public bool DebugTestHost { get; set; }
+    public bool DebugDataCollector { get; set; }
+    public bool DebugStopAtEntrypoint { get; set; }
+    public int JustRow { get; set; } = -1;
+
+    public string? BeforeFeature { get; set; }
+    public string? AfterFeature { get; set; }
+
+    //public string? BeforeTestHostFeature { get; set; }
+    //public string? AfterTestHostFeature { get; set; }
```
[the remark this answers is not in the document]
nohwnd: wanted to revisit why this was commented out, keep for now.

PR #15474, opened 2026-03-10
Cleanup dead TRX code

Youssef1313:


src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/CollectorDataEntry.cs
```diff
@@ -135,8 +115,6 @@ public void Save(XmlElement element, XmlTestStoreParameters? parameters)
 
         XmlPersistence helper = new();
         helper.SaveSimpleField(element, "@agentName", _agentName, null);
-        helper.SaveSimpleField(element, "@agentDisplayName", _agentDisplayName, _agentName);
```
Youssef1313, line 138: What this line was saying is to set agentDisplayName attribute value to `_agentDisplayName` if and only if it's different from `_agentName`.

However, `_agentDisplayName` and `_agentName` were always the same. So, this line was always not doing anything. So simplified this class also by removing `_agentDisplayName` completely.

src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/CollectorDataEntry.cs
```diff
@@ -135,8 +115,6 @@ public void Save(XmlElement element, XmlTestStoreParameters? parameters)
...
-        helper.SaveSimpleField(element, "@isFromRemoteAgent", _isFromRemoteAgent, false);
```
Youssef1313, line 139: This line would only set `isFromRemoteAgent` attribute if and only if it's not `false`. However, it was always false. So removed completely to simplify.

src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/CollectorDataEntry.cs
```diff
@@ -167,14 +145,13 @@ internal void AddAttachment(IDataAttachment attachment)
     /// Clones the instance and attachments, with file paths in file attachments absolute or relative as specified
     /// </summary>
     /// <param name="resultsDirectory">The results directory to use to make paths in the data attachments relative or absolute</param>
-    /// <param name="useAbsolutePaths">True to use absolute paths in this instance, false to use relative paths</param>
     /// <returns>A clone of the instance containing cloned attachments with file paths made absolute or relative</returns>
-    internal CollectorDataEntry Clone(string resultsDirectory, bool useAbsolutePaths)
```
Youssef1313, line 172: This was always called with `useAbsolutePaths: false`.
So, removing the bool parameter and renaming to `CloneWithRelativePath`.

src/Microsoft.TestPlatform.Extensions.TrxLogger/ObjectModel/UriDataAttachment.cs
```diff
@@ -90,18 +89,15 @@ public void Save(XmlElement element, XmlTestStoreParameters? parameters)
     /// Clones the instance and makes the URI in the clone absolute using the specified base directory
     /// </summary>
     /// <param name="baseDirectory">The base directory to use to make the URI absolute</param>
-    /// <param name="useAbsoluteUri">True to use an absolute URI in the clone, false to use a relative URI</param>
     /// <returns>A clone of the instance, with the URI made absolute</returns>
-    internal UriDataAttachment Clone(string baseDirectory, bool useAbsoluteUri)
+    internal UriDataAttachment CloneWithRelativePath(string baseDirectory)
```
Youssef1313, line 93: This was always called with `useAbsoluteUri: false`. So, removed the parameter and renamed to `CloneWithRelativePath`.

nohwnd (approved):

PR #15483, opened 2026-03-12
Allow to use global packages folder

drognanar:
Fixes #15469

### DIFF
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Microsoft.TestPlatform.Acceptance.IntegrationTests.csproj
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Microsoft.TestPlatform.Acceptance.IntegrationTests.csproj
@@ -10,6 +10,8 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>net9.0;net48</TargetFrameworks>
+    <!-- Restore packages to the local test assets cache so integration tests can find them at a known path. -->
+    <RestorePackagesPath>$(RepoRoot)artifacts\.packages\</RestorePackagesPath>
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="TestAssets\**" />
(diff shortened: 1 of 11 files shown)


test/Microsoft.TestPlatform.TestUtilities/IntegrationTestEnvironment.cs
```diff
@@ -29,6 +29,13 @@ public class IntegrationTestEnvironment
     public static string ArtifactsTempDirectory { get; } = Path.Combine(RepoRootDirectory, "artifacts", "tmp", BuildConfiguration);
 
     public static string LocalPackageSource { get; } = Path.Combine(RepoRootDirectory, "artifacts", "packages", BuildConfiguration, "Shipping").TrimEnd(Path.DirectorySeparatorChar);
+
+    /// <summary>
+    /// Gets the local NuGet cache directory used by test assets.
+    /// Test assets are restored into this directory by Build.cs via <c>dotnet restore --packages</c>.
+    /// </summary>
+    public static string TestAssetsNuGetCacheDirectory { get; } = Path.Combine(RepoRootDirectory, ".packages");
```
nohwnd:
```suggestion
    public static string TestAssetsNuGetCacheDirectory { get; } = Path.Combine(RepoRootDirectory, "artifacts", ".packages");
```

I'd move it here to make it less confusing. Tests are responsible for managing that folder.

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/Microsoft.TestPlatform.Acceptance.IntegrationTests.csproj
```diff
@@ -10,6 +10,8 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFrameworks>net9.0;net48</TargetFrameworks>
+    <!-- Restore packages to the local .packages directory so integration tests can find them at a known path. -->
+    <RestorePackagesPath>$(RepoRoot).packages\</RestorePackagesPath>
```
nohwnd: Is this only for the msdia dependency below?

drognanar: pretty much. also reading through how integration tests work.

so far my understanding is that with the global cache:

1. the Build.cs will handle preparing the artifacts/.packages folder for all acceptance tests. there's a method that will remove all -dev packages before the test runs
2. test assets such as CLIProject, MSTestV1UnitTestProject, LegacySettingsUnitTestProject are all using the local .packages cacheDid

it would seem that we can fully make this use the global cache. running all tests to validate that this works

nohwnd: Yeah I think the only "problem" here is figuring out where the global cache is, but maybe we could get rid of all the dependencies on that package instead. I am not if the tests are testing something that an integration test does not test.

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestEnvironment.cs
drognanar: this should now be moved

PR #15512, opened 2026-03-18
Fix thread safety issues

Evangelink:
## Summary

Several data structures shared across concurrent test execution threads were not thread-safe, risking corrupted results and lost updates.

## Changes

- **ParallelRunDataAggregator / DiscoveryDataAggregator**: Replace `TryGetValue` + set pattern with `ConcurrentDictionary.AddOrUpdate` in metrics aggregation to eliminate lost-update races
- **DataCollectionAttachmentManager**: Replace `List<Task>` with `ConcurrentBag<Task>` for `_attachmentTasks`; replace `ContainsKey` + `TryAdd` TOCTOU pattern with `GetOrAdd` for both `AttachmentSets` and `_attachmentTasks`
- **BlameCollector**: Replace `List<Guid>` / `Dictionary<Guid, BlameTestObject>` with `ConcurrentQueue` / `ConcurrentDictionary`; use `Interlocked.Increment` for `_testEndCount`; snapshot collections before passing to `WriteTestSequence`
- **TrxLogger / HtmlLogger**: Use `Interlocked.Increment` for all test counters (`TotalTestCount`, `PassedTestCount`, `FailedTestCount`, `SkippedTests`) to prevent lost updates under concurrent result arrival

### DIFF
--- a/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
+++ b/src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
@@ -51,7 +51,7 @@ internal class DataCollectionAttachmentManager : IDataCollectionAttachmentManage
     /// <summary>
     /// Attachment transfer tasks associated with a given datacollection context.
     /// </summary>
-    private readonly ConcurrentDictionary<DataCollectionContext, List<Task>> _attachmentTasks;
+    private readonly ConcurrentDictionary<DataCollectionContext, ConcurrentBag<Task>> _attachmentTasks;
 
     /// <summary>
     /// Use to cancel attachment transfers if test run is canceled.
@@ -79,7 +79,7 @@ internal class DataCollectionAttachmentManager : IDataCollectionAttachmentManage
     {
         _fileHelper = fileHelper;
         _cancellationTokenSource = new CancellationTokenSource();
-        _attachmentTasks = new ConcurrentDictionary<DataCollectionContext, List<Task>>();
+        _attachmentTasks = new ConcurrentDictionary<DataCollectionContext, ConcurrentBag<Task>>();
         AttachmentSets = new ConcurrentDictionary<DataCollectionContext, ConcurrentDictionary<Uri, AttachmentSet>>();
     }
 
@@ -170,12 +170,8 @@ internal class DataCollectionAttachmentManager : IDataCollectionAttachmentManage
             return;
         }
 
-        if (!AttachmentSets.ContainsKey(fileTransferInfo.Context))
-        {
-            var uriAttachmentSetMap = new ConcurrentDictionary<Uri, AttachmentSet>();
-            AttachmentSets.TryAdd(fileTransferInfo.Context, uriAttachmentSetMap);
-            _attachmentTasks.TryAdd(fileTransferInfo.Context, new List<Task>());
-        }
+        AttachmentSets.GetOrAdd(fileTransferInfo.Context, _ => new ConcurrentDictionary<Uri, AttachmentSet>());
+        _attachmentTasks.GetOrAdd(fileTransferInfo.Context, _ => new ConcurrentBag<Task>());
 
         if (!AttachmentSets[fileTransferInfo.Context].ContainsKey(uri))
         {
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
@@ -144,15 +144,10 @@ internal sealed class DiscoveryDataAggregator
             {
                 var newValue = Convert.ToDouble(metric.Value, CultureInfo.InvariantCulture);
 
-                if (_metricsAggregator.TryGetValue(metric.Key, out object? oldValue))
-                {
-                    double oldDoubleValue = Convert.ToDouble(oldValue, CultureInfo.InvariantCulture);
-                    _metricsAggregator[metric.Key] = newValue + oldDoubleValue;
-                }
-                else
-                {
-                    _metricsAggregator.TryAdd(metric.Key, newValue);
-                }
+                _metricsAggregator.AddOrUpdate(
+                    metric.Key,
+                    newValue,
+                    (_, oldValue) => newValue + Convert.ToDouble(oldValue, CultureInfo.InvariantCulture));
             }
         }
     }
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
@@ -74,23 +74,20 @@ internal class ParallelRunDataAggregator
     {
         var testOutcomeMap = new Dictionary<TestOutcome, long>();
         long totalTests = 0;
-        lock (_dataUpdateSyncObject)
+        if (_testRunStatsList.Count > 0)
         {
-            if (_testRunStatsList.Count > 0)
+            foreach (var runStats in _testRunStatsList)
             {
-                foreach (var runStats in _testRunStatsList)
+                // TODO: we get nullref here if the stats are empty.
+                foreach (var outcome in runStats.Stats!.Keys)
                 {
-                    // TODO: we get nullref here if the stats are empty.
-                    foreach (var outcome in runStats.Stats!.Keys)
+                    if (!testOutcomeMap.ContainsKey(outcome))
                     {
-                        if (!testOutcomeMap.ContainsKey(outcome))
-                        {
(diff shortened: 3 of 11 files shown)


src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
```diff
@@ -170,12 +170,8 @@ public void AddAttachment(FileTransferInformation fileTransferInfo, AsyncComplet
             return;
         }
 
-        if (!AttachmentSets.ContainsKey(fileTransferInfo.Context))
-        {
-            var uriAttachmentSetMap = new ConcurrentDictionary<Uri, AttachmentSet>();
-            AttachmentSets.TryAdd(fileTransferInfo.Context, uriAttachmentSetMap);
-            _attachmentTasks.TryAdd(fileTransferInfo.Context, new List<Task>());
-        }
```
nohwnd, line 178: Is there a specific issue that is being solved in this file? 

There are more dubious calls, in GetAttachments we get the tasks make snapshot via .ToArray, await those tasks, and then we remove the entire collection from the dictionary without checking if we awaited all tasks. If that is a problem I don't know.

nohwnd: Similarly AddNewFileTransfer uses global lock in the continuation, which accesses a particular single attachment  in a particular attachmentSet. Seems little overkill.

nohwnd: In the same continuation we do `_attachmentTasks[fileTransferInfo.Context].Add(continuationTask);` line 322, this is probably avoiding the fact that AddOrUpdate could call the calback multiple times, and avoids adding the  task multiple times, but who knows.

nohwnd: Overall not opposed to adding this change, but do we have some repro / actual issue, or this is prevention?

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
```diff
@@ -144,15 +144,10 @@ public void AggregateMetrics(IDictionary<string, object>? metrics)
...
-                    _metricsAggregator[metric.Key] = newValue + oldDoubleValue;
-                }
-                else
-                {
-                    _metricsAggregator.TryAdd(metric.Key, newValue);
-                }
+                _metricsAggregator.AddOrUpdate(
+                    metric.Key,
+                    newValue,
+                    (_, oldValue) => newValue + Convert.ToDouble(oldValue, CultureInfo.InvariantCulture));
```
nohwnd, line 150: +1 guarded by the collection and the callback is side-effect free.

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/ParallelRunDataAggregator.cs
```diff
@@ -74,20 +74,23 @@ public ITestRunStatistics GetAggregatedRunStats()
     {
         var testOutcomeMap = new Dictionary<TestOutcome, long>();
         long totalTests = 0;
-        if (_testRunStatsList.Count > 0)
+        lock (_dataUpdateSyncObject)
```
nohwnd: This looks unneeded (but also won't hurt) the method is called to get the aggregated data from each aggregator and is called once proxy manager is done running.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -461,14 +462,17 @@ private void EventsTestCaseStart(object? sender, TestCaseStartEventArgs e)
...
 
-        // Add the test object to the dictionary.
-        _testObjectDictionary.Add(blameTestObject.Id, blameTestObject);
+            // Add the test object to the dictionary.
+            _testObjectDictionary.Add(blameTestObject.Id, blameTestObject);
+        }
 
         // Increment test start count.
-        _testStartCount++;
+        Interlocked.Increment(ref _testStartCount);
```
nohwnd: Probably much easier to use a concurrent q instead of a list which preserves order and can count

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -483,13 +487,16 @@ private void EventsTestCaseEnd(object? sender, TestCaseEndEventArgs e)
...
         // Update the test object in the dictionary as the test has completed.
         TPDebug.Assert(e.TestElement is not null, "e.TestElement is null");
-        if (_testObjectDictionary.ContainsKey(e.TestElement.Id))
+        lock (_testSequenceLock)
         {
-            _testObjectDictionary[e.TestElement.Id].IsCompleted = true;
+            if (_testObjectDictionary.ContainsKey(e.TestElement.Id))
+            {
+                _testObjectDictionary[e.TestElement.Id].IsCompleted = true;
+            }
```
nohwnd: looks like a job for concurrent dictionary.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -510,12 +517,20 @@ private void SessionEndedHandler(object? sender, SessionEndEventArgs args)
...
-                filepath = _blameReaderWriter.WriteTestSequence(_testSequence, _testObjectDictionary, filepath);
+                List<Guid> testSequenceCopy;
+                Dictionary<Guid, BlameTestObject> testObjectDictionaryCopy;
+                lock (_testSequenceLock)
+                {
+                    testSequenceCopy = new List<Guid>(_testSequence);
+                    testObjectDictionaryCopy = new Dictionary<Guid, BlameTestObject>(_testObjectDictionary);
+                }
+
+                filepath = _blameReaderWriter.WriteTestSequence(testSequenceCopy, testObjectDictionaryCopy, filepath);
```
nohwnd: doubt this is needed, we are done, and we are reporting after process is done. There is 1:1 pair between process and collector so this runs once.

src/SettingsMigrator/Migrator.cs
```diff
@@ -96,8 +96,9 @@ private static void MigrateRunSettings(string oldRunSettingsPath, string newRunS
         string? testSettingsPath = null;
         using XmlTextReader reader = new(oldRunSettingsPath);
         reader.Namespaces = false;
+        reader.DtdProcessing = DtdProcessing.Prohibit;
```
nohwnd: > DtdProcessing.Prohibit is a setting in .NET XML parsing that explicitly bans Document Type Definitions (DTDs), throwing an XmlException if any are encountered. This is the default, secure behavior designed to prevent XML External Entity (XXE) attacks. It is used to block malicious external entities or resource loading during XML processing

This is probably a hypothetical problem. For all I know we can most likely delete settings migrator entirely. TestSettings are no longer supported and are removed, and I never ever heard about anyone mentioning this tool.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/ParallelRunDataAggregatorTests.cs
```diff
@@ -455,4 +458,78 @@ public void GetRunDataMetricsShouldNotAddNumberOfAdapterDiscoveredIfMetricsIsEmp
...
+        var readerTask = Task.Run(() =>
+        {
+            barrier.SignalAndWait();
+            for (int i = 0; i < iterationsPerThread; i++)
+            {
+                // This must not throw InvalidOperationException due to collection modification
+                var runStats = aggregator.GetAggregatedRunStats();
+                Assert.IsTrue(runStats.ExecutedTests >= 0, "Executed tests count should be non-negative");
+            }
+        });
```
nohwnd: this does not happen in the product afaik

test/SettingsMigrator.UnitTests/MigratorTests.cs
```diff
@@ -134,6 +134,56 @@ public void InvalidPathThrowsException()
         Assert.Throws<IOException>(() => _migrator.Migrate(oldTestsettingsPath, _newRunsettingsPath));
     }
 
+    [TestMethod]
```
nohwnd: Not sure about the impact, but would prefer to keep it as is and deprecate instead. Mentioned at least here: https://learn.microsoft.com/en-us/visualstudio/test/migrate-testsettings-to-runsettings?view=visualstudio

src/Microsoft.TestPlatform.Common/DataCollection/DataCollectionAttachmentManager.cs
Evangelink: This is prevention

nohwnd (approved):

PR #15513, opened 2026-03-18
Indicator flakiness

nohwnd:
## Description

On server we run on low power system which has 2 cores, and we run all tests, the timer does not fire on time, and we get failure on the check for "." which should be there after 1 second.

Instead of timer that forces us to wait, we step the timer manually so we can make the whole test more reliably and faster, while believing that built in timer works like it should.

Related #15478


src/vstest.console/Internal/ProgressIndicator.cs
```diff
@@ -47,12 +47,8 @@ public void Start()
     {
         lock (_syncObject)
```
Youssef1313, line 48: Is this lock guarding against anything now?

Youssef1313 (approved):

src/vstest.console/Internal/ProgressIndicator.cs
nohwnd: probably not, but won't hurt and the build is green, let's keep it for someones hackathon month (which is also why i write so many typos :D)

PR #15526, opened 2026-03-19
Change integration test parallelization from ClassLevel to MethodLevel

nohwnd:
Integration tests offload work to child processes (vstest.console, testhost), so there is no shared in-process state between test methods in the same class. ClassLevel parallelization was unnecessarily conservative - switching to MethodLevel allows more tests to run concurrently, improving overall test execution time.

Each test method creates its own TempDirectory and invokes vstest.console independently, so concurrent execution within a class is safe.

Fixes #15488

### DIFF
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/BlameDataCollectorTests.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/BlameDataCollectorTests.cs
@@ -20,6 +20,7 @@ namespace Microsoft.TestPlatform.AcceptanceTests;
 
 [TestClass]
 [TestCategory("Windows-Review")]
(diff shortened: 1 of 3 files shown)


test/Microsoft.TestPlatform.Acceptance.IntegrationTests/BlameDataCollectorTests.cs
```diff
@@ -20,6 +20,7 @@ namespace Microsoft.TestPlatform.AcceptanceTests;
 
 [TestClass]
 [TestCategory("Windows-Review")]
+[DoNotParallelize] // Blame tests collect crash/hang dumps from child processes and race when run in parallel.
```
nohwnd, line 23: Dubious claim,  needs more proof. https://github.com/microsoft/vstest/issues/15530

Evangelink (approved):

PR #15534, opened 2026-03-20
Fix compatibility test failures for legacy vstest.console and MSTest adapter

nohwnd:
## Summary
Fix compatibility tests, by reducing them to console that is likely to be used with vstest.console wrapper (we are unlikely to use latest wrapper, with legacy console), and we don't encourage such configuration.

Add dotnet_rollforwards=latestmajor to use the latest tfm for the vstest.console that is used with wrapper, to avoid installing additional tfms for the console. Tfm incompatibilities are rare.

Add custom test adapter for compatibility tests, to avoid testing mstest compatibility rather than vstest.console  compatibility.


test/Microsoft.TestPlatform.Library.IntegrationTests/TranslationLayerTests/RunTests.cs
```diff
@@ -89,8 +90,11 @@ public void RunAllTestsWithMixedTFMsWillRunTestsFromAllProvidedDllEvenWhenTheyMi
 
         _vstestConsoleWrapper = GetVsTestConsoleWrapper();
         var runEventHandler = new RunEventHandler();
-        var netFrameworkDll = GetTestDllForFramework("MSTestProject1.dll", DEFAULT_HOST_NETFX);
-        var netDll = GetTestDllForFramework("MSTestProject1.dll", DEFAULT_HOST_NETCORE);
+        // Use SimpleTestProject4 which has a minimal test adapter with no MSTest dependency,
+        // avoiding version incompatibilities (e.g. System.Runtime.CompilerServices.Unsafe)
+        // when running with older vstest.console versions.
```
nohwnd:
```suggestion
        // Use SimpleTestProject4 which has a minimal test adapter, we were using MSTest for this before
        // but there are some incompatibilities, so we end up testing if MSTest can work with other console, rather than testing if that older console can work.
```

Evangelink (approved):

PR #15540, opened 2026-03-20
feat: migrate from Newtonsoft.Json to System.Text.Json / Jsonite

nohwnd:
## Summary

Migrate vstest JSON serialization from Newtonsoft.Json to **System.Text.Json** (on .NET Core) and **Jsonite** (on .NET Framework/netstandard2.0). Removes Newtonsoft.Json from all production code paths.

### Architecture

| Platform | Serializer | Condition |
|----------|-----------|-----------|
| .NET Core (net10.0+) | System.Text.Json | `#if NETCOREAPP` |
| .NET Framework (net462) | Jsonite (embedded BSD-2, from xoofx/jsonite) | `#if !NETCOREAPP` |
| netstandard2.0 | Jsonite | `#if !NETCOREAPP` |

### Key Changes

- **JsonDataSerializer** split into 3 partial classes (shared / STJ / Jsonite) — zero `#if` blocks in main file
- **12+ custom STJ converters** for vstest types (TestCase, TestResult, TestProperty, etc.)
- **Full Jsonite converter suite** handling V1/V2 wire formats, private setters, collection Add()
- **Message class unified** — removed `Payload` property, uses `RawMessage` string + deferred deserialization
- **DepsJsonParser** replaces `DependencyContextJsonReader` on .NET Framework (eliminates transitive STJ dep)
- **Removed ALL Newtonsoft.Json** from production code (Legacy fallback deleted)
- **System.Text.Json.dll** removed from TestHostNetFramework packages
- **Jsonite** added to THIRD-PARTY-NOTICES.txt
- **All serialization tests run on both TFMs** — zero `#if NET` compiler conditions in tests
- **Jsonite control char fix** — escape chars < 0x20 as `\uXXXX` instead of throwing (testfx#5120)
- **ObjectDictionaryConverterFactory** — fixes STJ deserializing JSON integers as `double` in `IDictionary<string, object>` (metrics)
- **TestCaseEventArgs** — changed `private set` → `set` on `TestCaseId`, `TestCaseName`, `IsChildTestCase`, `TestElement`, `TestOutcome` for serialization compatibility

### Test Coverage

| Suite | Count | Status |
|-------|-------|--------|
| Serialization unit tests (per-message wire format) | 1125 (375 × 3 TFMs) | ✅ All pass |
| Performance benchmarks (including 1000-result batch) | 78 (26 × 3 TFMs) | ✅ All pass |
| Serialization edge cases (control chars, unicode, emoji) | Included in above | ✅ All pass |
| CommunicationUtilities unit tests | 1434 (3 TFMs) | ✅ All pass |
| All 19 unit test projects | 7576+ | ✅ All pass |
| CodeCoverage integration tests (Library) | 24 | ✅ All pass |
| Telemetry acceptance tests | 10 | ✅ All pass |
| Translation layer telemetry tests | 18 | ✅ All pass |
| Acceptance tests (non-compat) | 352+ | ✅ All pass |

**100% MessageType coverage** — every single one of the 52 message types in `MessageType` class has serialization tests (V1 serialize, V7 serialize, V1 deserialize, V7 deserialize, round-trip).

### Wire Format Verification

Captured diag logs from blame test runs on both `main` (Newtonsoft) and `stj-migration` branches. Parsed and compared all JSON messages from runner, testhost, and datacollector logs:

- **Runner (vstest.console)**: All 8 message types — ✅ Identical JSON structure
- **Testhost**: All message types present — ✅ Same data (property ordering differs due to dictionary iteration, semantically irrelevant)
- **Datacollector**: 3 shared types — ✅ Identical
- **Telemetry output**: 24 keys, all matching — ✅ Identical (except timing values)
- **Header field order**: All messages start with `Version,MessageType` — ✅ Fast header parser compatible

### Performance: STJ vs Newtonsoft (1000-result batch, V7)

| Operation | STJ (net10.0) | Jsonite (net48) |
|-----------|--------------|-----------------|
| Serialize 1000-result StatsChange | ~11ms | ~36ms |
| Deserialize 1000-result StatsChange | ~60ms | ~96ms |
| JSON size | 911KB | 911KB |

### Performance: STJ vs Newtonsoft (single message, 1000 iterations)

| Operation | STJ | Newtonsoft baseline |
|-----------|-----|-----------|
| Serialize TestMessage | 14ms | ~20ms |
| Serialize TestCasesFound | 58ms | ~65ms |
| Serialize ExecutionComplete | 117ms | ~130ms |

### Breaking Changes

- `Message.Payload` (`JToken?`) removed — replaced by `Message.RawMessage` (`string?`) + `Message.Version`
- `VersionedMessage` class removed — unified into `Message`
- Newtonsoft.Json no longer shipped in vstest packages
- `TestCaseEventArgs.TestCaseId/TestCaseName/IsChildTestCase/TestElement` setters changed from private/internal to public
- `TestCaseEndEventArgs.TestOutcome` setter changed from private to public


eng/detect-newtonsoft-deps.ps1
nohwnd, line 1: remove

docs/stj-migration-detailed.md
nohwnd, line 1: you can remove

docs/stj-migration-changelog.md
nohwnd, line 1: you can remove

docs/stj-migration-overview.md
nohwnd, line 1: you can remove

eng/verify-binding-redirects.ps1
nohwnd, line 1: very useful, infos for missing dlls are too much but let's keep, will later move to automatically patching the files locally probably? makes it more useful, same with verify-nupkgs, locally it should fix the files

samples/Microsoft.TestPlatform.Protocol/Communication/Message.cs
```diff
@@ -16,15 +15,15 @@ public class Message
         /// <summary>
         /// Gets or sets the payload.
         /// </summary>
-        public JToken Payload { get; set; }
+        public JsonElement? Payload { get; set; }
```
nohwnd: address: this needs to go, who is this doc for? is it accompanying the overview.md in docs? is that updated?

src/Microsoft.TestPlatform.CommunicationUtilities/Json/Jsonite/Jsonite.cs
```diff
@@ -0,0 +1,1726 @@
+// ============================================================================
```
nohwnd, line 1: do we have 1 copy of this or multiple? IT also must be added to third party notice, check microsoft.testing.platform in testfx repo

src/Microsoft.TestPlatform.CommunicationUtilities/Messages/Message.cs
```diff
@@ -1,37 +1,35 @@
...
-    public JToken? Payload { get; set; }
+    public int Version { get; set; }
 
     /// <summary>
-    /// To string implementation.
+    /// Gets or sets the raw JSON message as received from the wire.
+    /// This contains the full message including MessageType, Version, and Payload.
     /// </summary>
-    /// <returns> The <see cref="string"/>. </returns>
+    public string? RawMessage { get; set; }
```
nohwnd, line 28: This needs to be removed from here, and replaced with T Payload property on child class, so we can pass around messages that hold string, but also messages that hold stream or similar blob, currently we will only pass around string, as it works with the way we currently do things, but since we are "breaking" here, we don't want to change that too many times.

src/Microsoft.TestPlatform.CommunicationUtilities/Messages/Message.cs
```diff
@@ -1,37 +1,35 @@
...
+
+    /// <inheritdoc/>
     public override string ToString()
     {
-        // TODO: Review where this is used, we should avoid extensive serialization and deserialization,
-        // and this might be happening in multiple places that are not the edge of our process.
-        return $"({MessageType}) -> {(Payload == null ? "null" : Payload.ToString(Formatting.Indented))}";
+        return $"({MessageType}) v{Version} -> {RawMessage}";
```
nohwnd: this will need to offload to the child class

src/Microsoft.TestPlatform.CommunicationUtilities/ObjectModel/TestRunCriteriaWithSources.cs
```diff
@@ -23,7 +25,9 @@ public class TestRunCriteriaWithSources
     /// </param>
     /// <param name="runSettings"> The run settings.  </param>
     /// <param name="testExecutionContext"> The test Execution Context. </param>
+#if NETCOREAPP
     [JsonConstructor]
+#endif
```
nohwnd, line 30: fine, this is .net only, and stj is available there from runtime

src/Microsoft.TestPlatform.CommunicationUtilities/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1,2 +1,3 @@
 #nullable enable
 const Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.ObjectModel.MessageType.TelemetryEventMessage = "TestPlatform.TelemetryEvent" -> string!
+~static Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Resources.Resources.ConnectionTimeoutWithErrorMessage.get -> string
```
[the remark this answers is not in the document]
nohwnd: yes apply

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/Legacy/LegacyDefaultTestPlatformContractResolver.cs
nohwnd, line 1: IMPORTANT: changed my mind, remove all the legacy fallback to newtonsoft in production code, and only keep for test, we validated serialization left and right, so the space for error is within the margin of normal bug, and we cannot dynamically provide dependencies for restore, so user will have to bring Newtonsoft dll, and so that means they need it, test platform should not need it for anything.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/AttachmentConverters.cs
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
nohwnd, line 1: for more files: all the serializers that use system.text.json should be wrapped in compiler condition, and the condition in project file should be removed, I prefer this because we have existing public api that usees namespaces, so moving to subfolder would make it inconsistent

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/JsoniteConvert.cs
```diff
@@ -0,0 +1,588 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if !NETCOREAPP
```
nohwnd, line 4: move all jsonite under serialization folder

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/DefaultTestPlatformContractResolver.cs
```diff
@@ -1,83 +1,6 @@
...
-        {
-            // This converter is required to hint json.net to use a concrete class for serialization
-            // of ITestRunStatistics. We can't remove ITestRunStatistics since it is a breaking change.
-            contract.Converter = new TestRunStatisticsConverter();
-        }
-
-        return contract;
-    }
-}
+// This file previously contained DefaultTestPlatformContractResolver and DefaultTestPlatformContractResolver7
```
nohwnd: this is useless, just delete such files.

src/Microsoft.TestPlatform.CommunicationUtilities/Json/Jsonite/Jsonite.cs
nohwnd: make sure this copy of jsonite isfrom testfx, we made fixes there.

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcessCodeMethods.cs
```diff
@@ -100,7 +100,7 @@ internal static int GetParentPidWindows(Process process)
         }
         catch (Exception ex)
         {
-            EqtTrace.Verbose($"ProcessCodeMethods.GetParentPidLinux: Error getting parent of process {process.Id} - {process.ProcessName}, {ex}.");
```
nohwnd, line 103: was fixed elsewhere, should be gone when we merge main

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleProcessManager.cs
```diff
@@ -62,7 +64,13 @@ internal sealed class VsTestConsoleProcessManager : IProcessManager, IDisposable
     internal IFileHelper FileHelper { get; set; }
 
     /// <inheritdoc/>
-    public int ProcessId { get; set; }
+    public int ProcessId { get; set; } = -1;
```
nohwnd: port to new PR probablY? there is related issue, but the goal of this code is when vstst.consolewrapper connects and times out we need to see what the error was, what the child process was, and what the exit code was. If wrapper times out and process did not exit  yet (processid != -1 and exit code == -1) then we want to say that you should increase timeout (as we do now) and that the process was still alive.

src/package/Microsoft.TestPlatform.CLI/Microsoft.TestPlatform.CLI.csproj
```diff
@@ -92,7 +91,8 @@
       <MicrosoftCodeCoverageIO Include="$(PkgMicrosoft_CodeCoverage_IO)\lib\netstandard2.0\**\*"></MicrosoftCodeCoverageIO>
       <MicrosoftExtensionsDependencyModel Include="$(PkgMicrosoft_Extensions_DependencyModel)\lib\netstandard2.0\*"></MicrosoftExtensionsDependencyModel>
       <MicrosoftExtensionsFileSystemGlobbing Include="$(PkgMicrosoft_Extensions_FileSystemGlobbing)\lib\netstandard2.0\*"></MicrosoftExtensionsFileSystemGlobbing>
-      <NewtonsoftJson Include="$(PkgNewtonsoft_Json)\lib\netstandard2.0\*"></NewtonsoftJson>
+      <SystemTextJson Include="$(PkgSystem_Text_Json)\lib\netstandard2.0\*"></SystemTextJson>
```
nohwnd: remove

src/package/Microsoft.TestPlatform.CLI/Microsoft.TestPlatform.CLI.nuspec
```diff
@@ -131,9 +129,8 @@
     <file src="net481\testhost.net481.arm64.exe"          target="contentFiles\any\net10.0\TestHostNetFramework" />
     <file src="net481\testhost.net481.arm64.exe.config"   target="contentFiles\any\net10.0\TestHostNetFramework" />
 
-    <file src="net462\Newtonsoft.Json.dll" target="contentFiles\any\net10.0\TestHostNetFramework" />
 
-    <file src="net462\System*.dll" exclude="net462\System.Threading.Tasks.Extensions.dll;net462\System.Diagnostics.DiagnosticSource.dll;net462\System.Text.Json.dll;net462\System.Text.Encodings.Web.dll;net462\System.IO.Pipelines.dll" target="contentFiles\any\net10.0\TestHostNetFramework" />
+    <file src="net462\System*.dll" exclude="net462\System.Threading.Tasks.Extensions.dll;net462\System.Diagnostics.DiagnosticSource.dll;net462\System.IO.Pipelines.dll" target="contentFiles\any\net10.0\TestHostNetFramework" />
```
nohwnd: revert

src/package/Microsoft.TestPlatform.CLI/Microsoft.TestPlatform.CLI.sourcebuild.nuspec
```diff
@@ -33,7 +33,7 @@
     <file src="$SourceBuildTfmCurrent$\Microsoft.VisualStudio.TestPlatform.Common.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
     <file src="$SourceBuildTfmCurrent$\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
 
-    <file src="$SourceBuildTfmCurrent$\Newtonsoft.Json\Newtonsoft.Json.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
+    <file src="$SourceBuildTfmCurrent$\System.Text.Json\System.Text.Json.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
```
nohwnd: remove, make newtonsoft gone

src/package/Microsoft.TestPlatform.CLI/Microsoft.TestPlatform.CLI.sourcebuild.product.nuspec
```diff
@@ -33,7 +33,7 @@
     <file src="$SourceBuildTfmCurrent$\Microsoft.VisualStudio.TestPlatform.Common.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
     <file src="$SourceBuildTfmCurrent$\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
 
-    <file src="$SourceBuildTfmCurrent$\Newtonsoft.Json\Newtonsoft.Json.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
+    <file src="$SourceBuildTfmCurrent$\System.Text.Json\System.Text.Json.dll" target="contentFiles\any\$SourceBuildTfmCurrent$" />
```
nohwnd: remove make newtonsoft gone

src/package/Microsoft.TestPlatform.Portable/Microsoft.TestPlatform.Portable.csproj
```diff
@@ -84,7 +83,7 @@
       <MicrosoftCodeCoverageIO Include="$(PkgMicrosoft_CodeCoverage_IO)\lib\netstandard2.0\**\*" />
       <MicrosoftExtensionsDependencyModel Include="$(PkgMicrosoft_Extensions_DependencyModel)\lib\netstandard2.0\*" />
       <MicrosoftExtensionsFileSystemGlobbing Include="$(PkgMicrosoft_Extensions_FileSystemGlobbing)\lib\netstandard2.0\*" />
-      <NewtonsoftJson Include="$(PkgNewtonsoft_Json)\lib\netstandard2.0\*" />
+      <SystemTextJson Include="$(PkgSystem_Text_Json)\lib\netstandard2.0\*" />
```
nohwnd: from all nuspecs, remove the lines for systemtextjson we added.

src/package/Microsoft.TestPlatform.Portable/Microsoft.TestPlatform.Portable.nuspec
```diff
@@ -484,14 +482,14 @@
...
     <file src="net8.0\Microsoft.TestPlatform.PlatformAbstractions.dll"	target="tools\net8.0\TestHostNetFramework\Microsoft.TestPlatform.PlatformAbstractions.dll" />
     <file src="net8.0\Microsoft.TestPlatform.Utilities.dll"	target="tools\net8.0\TestHostNetFramework\Microsoft.TestPlatform.Utilities.dll" />
     <file src="net8.0\Microsoft.VisualStudio.TestPlatform.Common.dll"	target="tools\net8.0\TestHostNetFramework\Microsoft.VisualStudio.TestPlatform.Common.dll" />
     <file src="net8.0\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll"	target="tools\net8.0\TestHostNetFramework\Microsoft.VisualStudio.TestPlatform.ObjectModel.dll" />
-    <file src="netstandard2.0\Newtonsoft.Json.dll"	target="tools\net8.0\TestHostNetFramework\Newtonsoft.Json.dll" />
+    <!-- System.Text.Json removed - no longer needed after DepsJsonParser replaced DependencyModel -->
     <file src="netstandard2.0\System.Collections.Immutable.dll"	target="tools\net8.0\TestHostNetFramework\System.Collections.Immutable.dll" />
     <file src="netstandard2.0\System.Reflection.Metadata.dll"	target="tools\net8.0\TestHostNetFramework\System.Reflection.Metadata.dll" />
```
nohwnd, line 493: all files copied into testhostnetframework should com from .netfx, also remove the comment about removeing stj, such comments are useless to add

src/package/Microsoft.TestPlatform/Microsoft.TestPlatform.csproj
```diff
@@ -105,7 +104,7 @@
       <MicrosoftExtensionsFileSystemGlobbingFiles Include="$(PkgMicrosoft_Extensions_FileSystemGlobbing)\lib\netstandard2.0\*"></MicrosoftExtensionsFileSystemGlobbingFiles>
       <PlatformAbstractionsDepsJsonFiles Include="..\..\Microsoft.TestPlatform.PlatformAbstractions\bin\$(Configuration)\$(TargetFramework)\*.deps.json"></PlatformAbstractionsDepsJsonFiles>
       <PackageDepsJsonFiles Include="..\..\Microsoft.TestPlatform.PlatformAbstractions\bin\$(Configuration)\$(TargetFramework)\*.deps.json"></PackageDepsJsonFiles>
-      <NewtonsoftJsonFiles Include="$(PkgNewtonsoft_Json)\lib\netstandard2.0\*"></NewtonsoftJsonFiles>
+      <SystemTextJsonFiles Include="$(PkgSystem_Text_Json)\lib\netstandard2.0\*"></SystemTextJsonFiles>
```
nohwnd: remove

src/package/Microsoft.TestPlatform/Microsoft.TestPlatform.csproj
```diff
@@ -125,7 +124,7 @@
     <Copy SourceFiles="@(MicrosoftExtensionsDependencyModelFiles)" DestinationFiles="$(OutDir)\Microsoft.Extensions.DependencyModel\%(RecursiveDir)%(Filename)%(Extension)" />
     <Copy SourceFiles="@(MicrosoftExtensionsFileSystemGlobbingFiles)" DestinationFiles="$(OutDir)\Microsoft.Extensions.FileSystemGlobbing\%(RecursiveDir)%(Filename)%(Extension)" />
     <Copy SourceFiles="@(PlatformAbstractionsDepsJsonFiles)" DestinationFiles="$(OutDir)\%(RecursiveDir)%(Filename)%(Extension)" />
-    <Copy SourceFiles="@(NewtonsoftJsonFiles)" DestinationFiles="$(OutDir)\%(RecursiveDir)%(Filename)%(Extension)" />
+    <Copy SourceFiles="@(SystemTextJsonFiles)" DestinationFiles="$(OutDir)\%(RecursiveDir)%(Filename)%(Extension)" />
```
nohwnd: remove

src/package/Microsoft.VisualStudio.TestTools.TestPlatform.V2.CLI/Microsoft.VisualStudio.TestTools.TestPlatform.V2.CLI.csproj
```diff
@@ -94,7 +94,6 @@
       <VsixSourceItem Include="$(TestPlatformBinFolder)System.Reflection.Metadata.dll" />
       <VsixSourceItem Include="$(TestPlatformBinFolder)System.Collections.Immutable.dll" />
       <VsixSourceItem Include="$(TestPlatformBinFolder)System.Memory.dll" />
-      <VsixSourceItem Include="$(TestPlatformBinFolder)System.Text.Json.dll" />
```
nohwnd, line 100: revert

nohwnd (commented):
^^^

samples/Microsoft.TestPlatform.Protocol/Communication/Message.cs
nohwnd: not solved yet

test/Microsoft.TestPlatform.CommunicationUtilities.UnitTests/Serialization/AfterTestRunEndSerializationTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            MessageType.AfterTestRunEnd, Payload, version);
+        var message = JsonDataSerializer.Instance.DeserializeMessage(json);
+        var result = JsonDataSerializer.Instance.DeserializePayload<bool>(message);
+
+        Assert.AreEqual(Payload, result);
+    }
+
+    // ── Newtonsoft comparison ────────────────────────────────────────────
+
+#if NET
```
nohwnd: for all serialization tests, this should happen in both .net and .net framework

nohwnd: we want to compare newtonsoft serialization with stj, and also jsonite.

test/Microsoft.TestPlatform.CommunicationUtilities.UnitTests/Serialization/AfterTestRunEndSerializationTests.cs
```diff
@@ -0,0 +1,165 @@
...
+    }
+#endif
+
+    // ── Helpers ──────────────────────────────────────────────────────────
+
+    /// <summary>
+    /// Compare JSON ignoring whitespace differences (so the pretty-printed
+    /// golden strings can be compared against the compact serializer output).
+    /// </summary>
+#if NET
```
nohwnd: for all serialization test files, this looks like a helper. move it to common place.

test/Microsoft.TestPlatform.CommunicationUtilities.UnitTests/JsonDataSerializerTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
nohwnd, line 1: these tests seem useless when we test all messages by themselves, also VSTEST_USE_NEWTONSOFT_JSON_SERIALIZER needs to be entirely gone from everywhere

test/Microsoft.TestPlatform.CommunicationUtilities.UnitTests/Serialization/TestMessageSerializationTests.cs
```diff
@@ -0,0 +1,142 @@
...
+
+    [TestMethod]
+    public void NewtonsoftComparisonV7()
+    {
+        NewtonsoftComparisonHelper.AssertMatchesNewtonsoft(
+            MessageType.TestMessage, Payload, version: 7);
+    }
+
+    // ── Helpers ──────────────────────────────────────────────────────────
+
```
nohwnd, line 229: // -- Helpers

followed by nothing is useless, remove everywhere where we don't have any helpers

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        // Process custom properties (e.g. Traits)
+        if (data.TryGetProperty("Properties", out var properties) && properties.GetArrayLength() > 0)
+        {
+            foreach (var prop in properties.EnumerateArray())
+            {
+                if (!prop.TryGetProperty("Key", out var keyElement))
+                    continue;
+
+                var testProperty = JsonSerializer.Deserialize<TestProperty>(keyElement.GetRawText(), options);
```
Evangelink: **Minor: Inconsistent `GetRawText()` vs direct `JsonElement` deserialization**

This line creates an intermediate string from the `JsonElement` before re-parsing:
```csharp
JsonSerializer.Deserialize<TestProperty>(keyElement.GetRawText(), options);
```

However, `TestResultConverterV2.cs` passes the `JsonElement` directly:
```csharp
JsonSerializer.Deserialize<TestProperty>(keyElement, options)!;
```

Passing the `JsonElement` directly avoids the intermediate string allocation. Consider aligning both V2 converters to use the direct `JsonElement` overload for consistency and to save an allocation per custom property per test case.

(The V1 `TestCaseConverter` has the same `GetRawText()` pattern at line 44.)

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverter.cs
```diff
@@ -1,72 +1,71 @@
...
 
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
-
 namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Serialization;
 
 /// <summary>
 /// Converter used by v1 protocol serializer to serialize TestCase object to and from v1 json
```
Evangelink, line 16: **Consider: `public` → `internal` visibility**

This V1 converter is declared `public class`. The V2 counterparts (`TestCaseConverterV2`, `TestResultConverterV2`) are correctly `internal class`. Since these are implementation-detail converters used only within `JsonDataSerializer`, exposing them as public API is unnecessary surface area (and will require tracking in `PublicAPI.Shipped.txt`).

Consider making both V1 converters `internal` to match the V2 convention, unless there's a specific reason they're public (e.g. third-party adapter tests relying on them).

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/JsoniteConvert.cs
```diff
@@ -0,0 +1,588 @@
...
+                var arr = new JsonArray();
+                foreach (var item in enumerable) arr.Add(ToJsonValueCore(item, visited, depth + 1, version)!);
+                return arr;
+            }
+            var result = new JsonObject();
+            foreach (var prop in type.GetProperties(BindingFlags.Public | BindingFlags.Instance))
+            {
+                if (!prop.CanRead || prop.GetIndexParameters().Length > 0) continue;
+                if (prop.GetCustomAttribute<IgnoreDataMemberAttribute>() != null) continue;
+                try { result[prop.Name] = ToJsonValueCore(prop.GetValue(value), visited, depth + 1, version)!; } catch { }
```
Evangelink: **Bare `catch { }` silently swallows all exceptions**

This pattern:
```csharp
try { result[prop.Name] = ToJsonValueCore(prop.GetValue(value), visited, depth + 1, version)!; } catch { }
```

silently swallows ALL exceptions, including `OutOfMemoryException`, `ThreadAbortException`, and security exceptions. This makes debugging serialization issues extremely difficult — a silently-skipped property could lead to data loss on the wire without any diagnostic trace.

There are at least 3 instances of bare `catch { }` in this file (around lines 84, 552, and the `prop.SetValue` block around line 540).

Suggestions:
1. Narrow to `catch (Exception)` at minimum
2. Even better: catch specific exceptions like `TargetInvocationException`, `InvalidCastException`, `InvalidOperationException`
3. Consider logging the exception via `EqtTrace` so issues are diagnosable

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.Jsonite.cs
```diff
@@ -0,0 +1,88 @@
...
+namespace Microsoft.VisualStudio.TestPlatform.CommunicationUtilities;
+
+public partial class JsonDataSerializer
+{
+    private static partial (int version, string? messageType) ParseHeaderFromJson(string rawMessage)
+    {
+        var parsed = (JsonObject)Json.Deserialize(rawMessage);
+        int version = parsed.TryGetValue("Version", out var vObj) ? Convert.ToInt32(vObj, System.Globalization.CultureInfo.InvariantCulture) : 0;
+        string? messageType = parsed.TryGetValue("MessageType", out var mtObj) ? (string?)mtObj : null;
+        return (version, messageType);
```
Evangelink, line 20: **`ParseHeaderFromJson` fallback fully parses the entire JSON including payload**

```csharp
var parsed = (JsonObject)Json.Deserialize(rawMessage);
```

When `FastHeaderParse` fails, this fallback path parses the **entire** JSON message (including the potentially large `Payload`) just to extract `Version` and `MessageType`. For large payloads (e.g., `TestCasesFound` with thousands of test cases), this does substantial unnecessary work.

Since this is only the fallback path (the fast string-based parser handles well-formed messages), performance impact in practice is likely minimal. But if you want to harden this, Jsonite doesn't have a streaming/partial parse mode, so one option would be to extract the header via string operations (e.g., find the first `}` or `,"Payload"` delimiter and parse only the prefix).

src/Microsoft.TestPlatform.CommunicationUtilities/JsonDataSerializer.Jsonite.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        var parsed = (JsonObject)Json.Deserialize(message.RawMessage!);
+        if (!parsed.TryGetValue("Payload", out var payloadObj))
+            return default;
+
+        return JsoniteConvert.To<T>(payloadObj);
+    }
+
+    private static partial T? DeserializeCore<T>(string json, int version)
+    {
```
Evangelink, line 33: **Version not propagated to `JsoniteConvert.To<T>()` during deserialization**

```csharp
return JsoniteConvert.To<T>(payloadObj);
```

The `message.Version` is available but not passed to the Jsonite deserialization path. By contrast, the STJ implementation in `JsonDataSerializer.Stj.cs` uses `GetPayloadOptions(message.Version)` to select V1 vs V2 converters with different serialization behaviors.

Looking at `JsoniteConvert.DeserializeTestCase`, the code uses format detection (`dict.ContainsKey("FullyQualifiedName")`) to determine flat (V2) vs property-bag (V1) format, which should work correctly for incoming messages since the wire format is structurally different between V1 and V2. So this is likely safe in practice.

However, the asymmetry is worth noting: **serialization** correctly uses version (via `ToJsonValue(payload, version)` in `SerializePayloadCore`), but **deserialization** doesn't. If a future format introduces a structural ambiguity that can't be resolved by format detection alone, this would be a latent bug.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/JsoniteConvert.cs
```diff
@@ -0,0 +1,588 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
Evangelink, line 1: **Code density makes this file difficult to review and maintain**

This file packs critical serialization logic into extremely dense single-line statements, e.g.:
```csharp
var tp = (TestProperty?)DeserializeTestProperty(ko); if (tp is null) continue;
pd.TryGetValue("Value", out var rv); var data = ConvertPropertyValueToString(rv);
```

and constructor matching:
```csharp
var k = data.Keys.FirstOrDefault(x => string.Equals(x, pars[i].Name, StringComparison.OrdinalIgnoreCase));
if (k is not null) { args[i] = ConvertTo(data[k], pars[i].ParameterType); keys.Add(k); }
else if (pars[i].HasDefaultValue) args[i] = pars[i].DefaultValue;
else { ok = false; break; }
```

This is a serialization layer that will need maintenance when new types/properties are added. Consider reformatting to one-statement-per-line to improve readability, especially for the `DeserializeTestCase`, `DeserializeTestResult`, and `ConvertTo` methods.

Evangelink (commented):
## Code Review Summary\n\nThis is a well-structured migration from Newtonsoft.Json to System.Text.Json (STJ) on .NET Core and Jsonite on .NET Framework. The architecture is clean — partial classes, deferred deserialization via `RawMessage`, and comprehensive wire-format compatibility tests.\n\n### Validated Findings\n\n| # | Finding | Severity | File(s) |\n|---|---------|----------|---------|\n| 1 | `TestCaseConverterV2` uses `GetRawText()` for `TestProperty` deserialization while `TestResultConverterV2` passes `JsonElement` directly — inconsistent and allocates unnecessary strings | Low | `TestCaseConverterV2.cs:52` |\n| 2 | V1 converters (`TestCaseConverter`, `TestResultConverter`) are `public class` while V2 counterparts are `internal` — unnecessary public API surface | Low | `TestCaseConverter.cs:16`, `TestResultConverter.cs:16` |\n| 3 | `JsoniteConvert.cs` has bare `catch { }` blocks that silently swallow all exceptions including critical ones | Medium | `JsoniteConvert.cs:84,540,552` |\n| 4 | `ParseHeaderFromJson` on the Jsonite path fully parses the entire JSON (including payload) just to extract the header when `FastHeaderParse` fails | Low | `JsonDataSerializer.Jsonite.cs:20` |\n| 5 | `DeserializePayloadCore` on the Jsonite path doesn't pass version to `JsoniteConvert.To<T>()`, relying on format detection heuristics instead — safe in practice but asymmetric with serialization path which does use version | Low | `JsonDataSerializer.Jsonite.cs:33` |\n| 6 | `JsoniteConvert.cs` has extremely dense code style that impacts maintainability of critical serialization logic | Low | `JsoniteConvert.cs` |\n\n### Items Verified as Non-Issues\n\n- **`VersionedMessage` property hiding**: File was fully deleted — no hiding issue.\n- **`DisableFastJson` duplication**: Only exists in `JsonDataSerializer.Stj.cs` (guarded by `#if NETCOREAPP`), not duplicated.\n- **Public API tracking**: `PublicAPI.Shipped.txt` correctly reflects `Message.Version` and `Message.RawMessage`; `VersionedMessage` is properly removed.\n- **`UnsafeRelaxedJsonEscaping`**: Matches Newtonsoft default behavior — acceptable since vstest messages are consumed by processes, not browsers.\n- **Security**: No `TypeNameHandling` equivalent in STJ (good). `Type.GetType(valueType)` in `TestPropertyConverter` is inherited behavior.\n\n### Overall\n\nThe STJ path is solid. The Jsonite path (.NET Framework) carries more risk due to its manual approach, format-detection heuristics, and swallowed exceptions. The wire-format compatibility test suite is excellent and provides strong confidence in correctness.\n\nGreat work on the migration!

PR #15533, opened 2026-03-20
perf: fix dictionary double-lookups, Collection.Contains, and LINQ allocations

Evangelink:
## Summary

Fix several obvious performance issues related to data types and redundant iterations across the codebase.

### 1. Dictionary `ContainsKey` + indexer → `TryGetValue`

Eliminates double hash lookups in ~15 locations. Each `ContainsKey` + `dict[key]` pair performs two hash operations when `TryGetValue` suffices with one.

**Files changed:**
- `ParallelRunDataAggregator.cs` — `GetAggregatedRunStats()`
- `TestSessionPool.cs` — `KillSession()`, `TryTakeProxy()`, `ReturnProxy()`
- `ProxyTestSessionManager.cs` — `DequeueProxy()`
- `InProcDataCollectionSink.cs` — `AddKeyValuePairToDictionary()`, `AddOrUpdateData()`
- `PortableSymbolReader.cs` — `GetNavigationData()`
- `FullSymbolReader.cs` — `GetTypeSymbol()`, `GetMethodSymbol()`
- `TestSessionStartArgs.cs` — `GetPropertyValue<T>()`
- `SessionEvents.cs` — `GetPropertyValue<T>()`
- `SimpleJSON.cs` — indexer getter/setter, `Add()`, `Remove()`

### 2. `Collection<InvokedDataCollector>` → `HashSet<InvokedDataCollector>`

`ParallelRunDataAggregator.InvokedDataCollectors` was a `Collection<T>` with O(n) `.Contains()` called inside a `foreach` loop (O(n²) behavior). Changed to `HashSet<T>` for O(1) lookups. The type already implements `IEquatable<T>` and `GetHashCode()`. Updated call sites in `ParallelRunEventsHandler` and `ParallelDataCollectionEventsHandler` to wrap with `.ToList()` when passing to APIs expecting `Collection<T>`.

### 3. `new[]` allocation inside LINQ predicate → static `HashSet`

In `EnableBlameArgumentProcessor.cs`, `new[] { "CollectAlways", "DumpType" }` was allocated **per element** inside `.Where()` predicates. Extracted to file-scoped static `HashSet<string>` fields with `StringComparer.OrdinalIgnoreCase`.

### 4. String `+=` → string interpolation

Consolidated two sequential `options +=` calls into a single interpolated string in `TestRunnerConnectionInfoExtensions.cs`.

## Testing

- Build: ✅ `build.cmd -c Release` — 0 errors, 0 warnings
[...]

PR #15563, opened 2026-03-24
Apply review feedback to perf dictionary double-lookups PR

nohwnd:
Addresses review feedback from PR #15533:

**Changes:**

1. **Revert HashSet\<InvokedDataCollector\> → Collection** (nohwnd feedback): The \_invokedDataCollectorsSet\ HashSet field is removed. Dedup logic reverts to using \Collection.Contains()\ since this code runs once at end of test run, not in parallel — a HashSet optimization would be misleading.

2. **Add clarifying comment to Aggregate method**: Documents that the method is called once at the end, not in parallel, to prevent future over-optimization.

3. **Revert barrier-based parallel test** (nohwnd feedback): Removes the concurrent reader thread from \AggregateAndGetAggregatedRunStatsShouldBeThreadSafe\ and restores the simpler sequential assertion pattern. Also restores index-based test assertions for InvokedDataCollectors (order is deterministic with Collection).

4. **BlameParameterNames: string[] → HashSet\<string\>** (Copilot/Youssef1313 feedback): Uses \HashSet<string>(StringComparer.OrdinalIgnoreCase)\ instead of \string[]\ for O(1) \Contains\ with built-in case-insensitive comparison.

### DIFF
--- a/src/Microsoft.TestPlatform.Common/Utilities/SimpleJSON.cs
+++ b/src/Microsoft.TestPlatform.Common/Utilities/SimpleJSON.cs
@@ -856,16 +856,13 @@ internal partial class JSONObject : JSONNode
     {
         get
         {
-            return _dict.ContainsKey(aKey) ? _dict[aKey] : new JSONLazyCreator(this, aKey);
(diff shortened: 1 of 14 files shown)


nohwnd:
should go directly into original PR

PR #15564, opened 2026-03-24
Fix VsTestConsoleWrapper.EndSession double-dispose

nohwnd:
Add a \_sessionEnded\ guard to \VsTestConsoleWrapper.EndSession()\ so that calling it multiple times is safe (idempotent). Previously, calling \EndSession()\ twice would attempt to send the session-end message, close the request sender, and shut down the process a second time.

Closes #15487

### DIFF
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
@@ -43,6 +43,8 @@ public class VsTestConsoleWrapper : IVsTestConsoleWrapper
 
     private bool _sessionStarted;
 
+    private bool _sessionEnded;
+
     /// <summary>
     /// Path to additional extensions to reinitialize vstest.console
     /// </summary>
@@ -651,6 +653,12 @@ public class VsTestConsoleWrapper : IVsTestConsoleWrapper
     /// <inheritdoc/>
     public void EndSession()
     {
+        if (_sessionEnded)
+        {
+            return;
+        }
+        _sessionEnded = true;
+
         EqtTrace.Info($"VsTestConsoleWrapper.EndSession: Ending VsTestConsoleWrapper session - process id:{_vstestConsoleProcessManager.ProcessId}");
 
         _requestSender.EndSession();
--- a/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
+++ b/src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
@@ -43,7 +43,7 @@ public class VsTestConsoleWrapper : IVsTestConsoleWrapper
 
     private bool _sessionStarted;
 
-    private bool _sessionEnded;
+    private int _sessionEnded;
 
     /// <summary>
     /// Path to additional extensions to reinitialize vstest.console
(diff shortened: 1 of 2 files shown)


test/TranslationLayer.UnitTests/VsTestConsoleWrapperTests.cs
```diff
@@ -698,4 +698,15 @@ public void EndSessionShouldSucceed()
         _mockRequestSender.Verify(rs => rs.Close(), Times.Once);
         _mockProcessManager.Verify(x => x.ShutdownProcess(), Times.Once);
     }
+
+    [TestMethod]
+    public void EndSessionCalledTwiceShouldNotThrow()
```
Evangelink: **Nit: Test name could be more precise**

The test name says "ShouldNotThrow" but it actually asserts stronger behavior — that the underlying operations are each called exactly once. Consider renaming to `EndSessionCalledTwiceShouldOnlyEndSessionOnce` to better express what's being verified.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -651,6 +653,12 @@ public void AbortTestRun()
     /// <inheritdoc/>
     public void EndSession()
     {
+        if (_sessionEnded)
+        {
+            return;
+        }
```
Evangelink, line 660: **Note (Low): `_sessionEnded` is never reset**

After `EndSession()`, the existing code sets `_sessionStarted = false`, allowing `EnsureInitialized()` to re-start a session. But `_sessionEnded` is never reset — not even if `StartSession()` / `StartSessionAsync()` is called again.

This means `EndSession()` → `StartSession()` → use → `EndSession()` would silently no-op the second `EndSession`.

This is probably fine for the real-world usage pattern (create wrapper → use → end → discard), but worth acknowledging. If re-startability is ever needed, `_sessionEnded` should be reset in `StartSession()` / `StartSessionAsync()`.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -43,6 +43,8 @@ public class VsTestConsoleWrapper : IVsTestConsoleWrapper
 
     private bool _sessionStarted;
 
+    private bool _sessionEnded;
```
Evangelink: **Suggestion (Medium): Thread-safety concern**

The `_sessionEnded` field is a plain `bool`. If `EndSession()` is called concurrently from two threads (e.g. one explicit call + one from a process-exit handler), there's a TOCTOU race where both threads read `false` and both enter the body.

The existing `_sessionStarted` has the same problem so this is consistent, but since the whole point of this PR is to prevent double execution, consider using an interlocked check:

```suggestion
    private int _sessionEnded;
```

And in `EndSession()`:
```csharp
if (Interlocked.Exchange(ref _sessionEnded, 1) == 1)
{
    return;
}
```

Not a blocker — the primary scenario from #15487 is sequential double-calls.

src/Microsoft.TestPlatform.VsTestConsole.TranslationLayer/VsTestConsoleWrapper.cs
```diff
@@ -651,6 +653,12 @@ public void AbortTestRun()
     /// <inheritdoc/>
     public void EndSession()
     {
+        if (_sessionEnded)
+        {
+            return;
```
Evangelink, line 659: **Suggestion (Low): Add trace log for the guarded early return**

The guard silently returns without any logging. The rest of `EndSession` uses `EqtTrace.Info` for entry/exit, so adding a trace here would help diagnostics and maintain consistency:

```suggestion
        if (_sessionEnded)
        {
            EqtTrace.Info("VsTestConsoleWrapper.EndSession: Session already ended, skipping.");
            return;
        }
```

Evangelink (approved):
Correctly fixes #15487 with minimal, focused changes and proper test coverage. The guard makes `EndSession()` idempotent as intended. No public API changes. Left a few minor suggestions (thread-safety, trace logging, reset semantics, test naming) — none are blockers.

PR #15565, opened 2026-03-24
Add target framework to default TRX file name

nohwnd:
When running multi-TFM tests, TRX files were overwritten because the default filename did not include the target framework. The TFM was already appended when using the `LogFilePrefix` parameter, but not for the default auto-generated filename.

This change appends the framework short name (e.g. `net10.0`) to the default TRX filename, matching the existing behavior for `LogFilePrefix`.

**Before:** `username_MACHINE_2026-03-24_07_51_08.trx`
**After:**  `username_MACHINE_2026-03-24_07_51_08_net10.0.trx`

Closes #15506

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -541,7 +541,7 @@ public class TrxLogger : ITestLoggerWithParameters
 
         if (_parametersDictionary is not null
             && _parametersDictionary.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out var framework)
-            && framework != null)
+            && !framework.IsNullOrWhiteSpace())
         {
             var shortName = Framework.FromString(framework)?.ShortName ?? framework;
             baseName = baseName + "_" + shortName;
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -536,7 +536,18 @@ public class TrxLogger : ITestLoggerWithParameters
         TPDebug.Assert(LoggerTestRun != null, "LoggerTestRun is null");
         TPDebug.Assert(LoggerTestRun.RunConfiguration != null, "LoggerTestRun.RunConfiguration is null");
         TPDebug.Assert(IsInitialized, "Logger is not initialized");
-        var defaultTrxFileName = LoggerTestRun.RunConfiguration.RunDeploymentRootDirectory + ".trx";
+
+        var baseName = LoggerTestRun.RunConfiguration.RunDeploymentRootDirectory;
+
+        if (_parametersDictionary is not null
+            && _parametersDictionary.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out var framework)
+            && framework != null)
+        {
+            var shortName = Framework.FromString(framework)?.ShortName ?? framework;
+            baseName = baseName + "_" + shortName;
+        }
+
+        var defaultTrxFileName = baseName + ".trx";
 
         return TrxFileHelper.GetNextIterationFileName(_testResultsDirPath, defaultTrxFileName, false);
     }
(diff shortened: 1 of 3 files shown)


src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -536,7 +536,18 @@ private string SetDefaultTrxFilePath()
...
         TPDebug.Assert(LoggerTestRun.RunConfiguration != null, "LoggerTestRun.RunConfiguration is null");
         TPDebug.Assert(IsInitialized, "Logger is not initialized");
-        var defaultTrxFileName = LoggerTestRun.RunConfiguration.RunDeploymentRootDirectory + ".trx";
+
+        var baseName = LoggerTestRun.RunConfiguration.RunDeploymentRootDirectory;
+
+        if (_parametersDictionary is not null
+            && _parametersDictionary.TryGetValue(DefaultLoggerParameterNames.TargetFramework, out var framework)
+            && framework != null)
+        {
```
Evangelink, line 545: Re the Copilot reviewer's path-traversal concern: I believe this is a false positive. `TargetFramework` is set by the test platform itself in `TestLoggerManager.UpdateLoggerParameters`, not by end-user input. `Framework.FromString` parses through NuGet's `NuGetFramework.Parse` which produces well-formed short names like `net10.0`. The existing `LogFilePrefix` path (line ~506) has the exact same unsanitized pattern and has shipped without issue. No regression here.

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -631,6 +631,34 @@ public void DefaultTrxFileShouldCreateIfLogFileNameParameterNotPassed()
...
+    {
+        _parameters.Remove(TrxLoggerConstants.LogFileNameKey);
+        _parameters[DefaultLoggerParameterNames.TargetFramework] = ".NETCoreApp,Version=v10.0";
+        _testableTrxLogger.Initialize(_events.Object, _parameters);
+
+        MakeTestRunComplete();
+
+        var fileName = Path.GetFileName(_testableTrxLogger.TrxFile);
+        Assert.IsNotNull(fileName);
+        Assert.IsTrue(fileName.Contains("_net10.0"), $"Expected TFM 'net10.0' in filename but got: {fileName}");
```
Evangelink: The `Contains("_net10.0")` assertion is reasonable given that `GetNextIterationFileName` may append `[1]` etc. before `.trx`. An `EndsWith` check would be too brittle if the file already exists during test runs.

Optional improvement: consider also adding a test for the fallback case when `Framework.FromString` can't parse the value (returns `null`), e.g. `_parameters[DefaultLoggerParameterNames.TargetFramework] = "SomeCustomFramework"`. That would exercise the `?? framework` fallback and confirm the raw string gets appended.

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -536,7 +536,18 @@ private string SetDefaultTrxFilePath()
...
+            var shortName = Framework.FromString(framework)?.ShortName ?? framework;
+            baseName = baseName + "_" + shortName;
+        }
```
Evangelink, line 548: Nice — this mirrors the existing `LogFilePrefix` pattern at line ~504 cleanly.

Nit: both this code and the `LogFilePrefix` path guard with `framework != null` but could use `!framework.IsNullOrWhiteSpace()` for defense-in-depth (if the platform ever sends an empty string). `Framework.FromString` handles whitespace by returning `null`, so the fallback `?? framework` would produce `baseName + "_ "`. Not a concern in practice since `TestLoggerManager` always sends a full framework string, and this isn't a regression — just a pre-existing pattern to maybe harden someday in both places.

Evangelink (approved):
LGTM. Small, well-scoped change that correctly aligns the default TRX filename behavior with the existing `LogFilePrefix` path. The consistency table tells the story:

| Path | TFM before PR | TFM after PR |
|------|:---:|:---:|
| `LogFileName` (explicit) | n/a | n/a |
| `LogFilePrefix` | ✅ | ✅ |
| Default (auto-generated) | ❌ | ✅ |

Left a couple of optional nits (whitespace guard, fallback test) but nothing blocking.

PR #15578, opened 2026-03-24
Fix silent hang in message processing loop when deserialization fails

### LINKED ISSUE #15577: STJ migration: DesignMode/wrapper communication hangs during test execution
## Problem

After migrating from Newtonsoft.Json to System.Text.Json (branch `stj-migration`, commit `9782a98ea`), the **direct CLI path works**:

```
dotnet vstest.console.dll test.dll  →  Passed: 1, Failed: 1, Skipped: 1
```

But the **wrapper/DesignMode path hangs**:
- Translation layer wrapper → socket → vstest.console → socket → testhost
- 4 vstest.console + 4 testhost processes sit at near-zero CPU indefinitely
- No error messages, no exceptions — both sides just wait for a message that never comes
- Smoke tests and integration tests all hang in this path

## What was fixed so far
- `TestRunStatistics.Stats` private setter → custom converter (was causing NullReferenceException)
- `IDictionary<string, object>` values deserialized as `JsonElement` → `ObjectConverter` unwraps to primitives (was causing InvalidCastException)

## Root cause hypothesis
The DesignMode communication loop (`DesignModeClient.ProcessRequests`) receives messages via `SocketCommunicationManager.ReceiveMessage()` and processes them. If a message fails to deserialize **silently** (swallowed exception), the loop may continue waiting for the next message while the sender waits for a response — causing a deadlock.

The message loop should **abort** if it cannot process data, rather than silently continue.

## Investigation needed
1. Add `--diag` to wrapper-based test execution to capture the full protocol exchange
2. Compare the messages sent/received between wrapper and vstest.console vs what the direct CLI path sends
3. Check if any try/catch in the message processing loop swallows serialization exceptions
4. Consider adding a message-level error response when deserialization fails, so the other side knows and can abort

## Branch
`stj-migration` on `nohwnd/vstest` — commit `9782a98ea`

nohwnd:
When message processing throws an exception (e.g. during deserialization when System.Text.Json dependencies cannot be loaded), the message loop would either silently exit or swallow the error, leaving the caller hanging indefinitely waiting for a response that never comes.

This fix updates both **DesignModeClient.ProcessRequests** (vstest.console side) and **TestRequestHandler.OnMessageReceived** (testhost side) to send a `ProtocolError` message back to the caller before stopping, so the caller gets notified of the error instead of hanging.

Closes #15577

### DIFF
--- a/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
+++ b/src/Microsoft.TestPlatform.Client/DesignMode/DesignModeClient.cs
@@ -291,6 +291,18 @@ public class DesignModeClient : IDesignModeClient
             catch (Exception ex)
             {
                 EqtTrace.Error("DesignModeClient: Error processing request: {0}", ex);
+
+                // Notify the caller (IDE) about the error so it does not hang
+                // waiting for a response that will never come.
+                try
+                {
+                    _communicationManager.SendMessage(MessageType.ProtocolError, ex.Message);
+                }
+                catch (Exception sendEx)
+                {
+                    EqtTrace.Error("DesignModeClient: Failed to send ProtocolError to client: {0}", sendEx);
+                }
+
                 Stop();
             }
         }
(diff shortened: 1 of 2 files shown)


nohwnd:
Closing — opportunity work without a linked issue, or superseded. Will revisit if/when a reported issue drives the need.

PR #15596, opened 2026-03-27
Track when vstest provides Newtonsoft.Json for extensions in telemetry

nohwnd:
## Problem

Extensions that depend on Newtonsoft.Json but don't ship their own copy silently use vstest's bundled copy. Before we can remove Newtonsoft.Json from vstest, we need to know how many extensions rely on this.

## Solution

**Silent telemetry tracking** — when \AssemblyResolver\ resolves a tracked assembly (e.g. Newtonsoft.Json) from vstest's own directory for an external requester:

1. Record it in a static \ConcurrentDictionary\ (one entry per assembly per process)
2. Report in telemetry as \VS.TestPlatform.ProvidedAssemblyDependency.Newtonsoft.Json=<requester>\
3. Filter out vstest's own internal assemblies (no false positives)
4. No user-visible warnings by default — opt-in via \VSTEST_WARN_MISSING_EXTENSIONS_DEPENDENCIES=1\ feature flag

## Test asset

\NewtonSoftDependencyMissing\ — references Newtonsoft.Json with \<ExcludeAssets>runtime</ExcludeAssets>\. The DLL is NOT in output, so vstest's copy is used at runtime.

## Changes

- \AssemblyResolver.cs\ — track provided dependencies, opt-in warning behind feature flag
- \TelemetryDataConstants.cs\ — new \ProvidedAssemblyDependency\ constant
- \FeatureFlag.cs\ — new \VSTEST_WARN_MISSING_EXTENSIONS_DEPENDENCIES\ flag
- \TestRunRequest.cs\ — report provided dependencies in telemetry at run completion
- \IAssemblyResolver.cs\ — add \RequestingAssembly\ to \AssemblyResolveEventArgs\

### DIFF
--- a/src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
+++ b/src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
@@ -441,13 +441,36 @@ public class TestRunRequest : ITestRunRequest, IInternalTestRunEventsHandler
                 // Fill in the time taken to complete the run
                 _requestData.MetricsCollection.Add(TelemetryDataConstants.TimeTakenInSecForRun, executionTotalTimeTaken.TotalSeconds);
 
-                // Record assemblies that vstest provided from its own directory
-                var providedDeps = AssemblyResolver.GetProvidedDependencies();
-                foreach (var dep in providedDeps)
+                // Record assemblies that vstest provided from its search directories.
+                // Telemetry gets 2 properties: assemblies resolved for user code (count only,
+                // no user assembly names), and assemblies resolved for Microsoft/System code (names ok).
+                var (userAssemblies, userCount, msAssemblies, msCount) = AssemblyResolver.GetProvidedDependencySummary();
+                if (userCount > 0)
                 {
-                    _requestData.MetricsCollection.Add(
-                        $"{TelemetryDataConstants.ProvidedAssemblyDependency}.{dep.Key}",
-                        dep.Value);
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForUser, userAssemblies);
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForUserCount, userCount);
+                }
+                if (msCount > 0)
+                {
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForMicrosoft, msAssemblies);
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForMicrosoftCount, msCount);
+                }
+
+                // Opt-in warning: show all provided assemblies in summary when feature flag is set.
+                if ((userCount > 0 || msCount > 0) && FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES))
+                {
+                    var allAssemblies = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                    foreach (var kvp in AssemblyResolver.GetProvidedDependencies())
+                    {
+                        allAssemblies.Add(kvp.Key);
+                    }
+
+                    var message = $"The test platform provided the following assemblies to resolve dependencies of test extensions "
+                        + $"that did not ship their own copy: {string.Join(", ", allAssemblies)}. "
+                        + "Test extensions should ship all their dependencies.";
+
+                    LoggerManager.HandleTestRunMessage(
+                        new TestRunMessageEventArgs(TestMessageLevel.Warning, message));
                 }
 
                 // Fill in the Metrics From Test Host Process
--- a/src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
+++ b/src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
@@ -13,6 +13,7 @@ using Microsoft.VisualStudio.TestPlatform.Common;
 using Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework;
 using Microsoft.VisualStudio.TestPlatform.Common.ExtensionFramework.Utilities;
 using Microsoft.VisualStudio.TestPlatform.Common.Telemetry;
+using Microsoft.VisualStudio.TestPlatform.Common.Utilities;
 using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities;
 using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.Interfaces;
 using Microsoft.VisualStudio.TestPlatform.CommunicationUtilities.ObjectModel;
@@ -439,6 +440,15 @@ public class TestRunRequest : ITestRunRequest, IInternalTestRunEventsHandler
 
                 // Fill in the time taken to complete the run
                 _requestData.MetricsCollection.Add(TelemetryDataConstants.TimeTakenInSecForRun, executionTotalTimeTaken.TotalSeconds);
+
+                // Record assemblies that vstest provided from its own directory
+                var providedDeps = AssemblyResolver.GetProvidedDependencies();
+                foreach (var dep in providedDeps)
+                {
+                    _requestData.MetricsCollection.Add(
+                        $"{TelemetryDataConstants.ProvidedAssemblyDependency}.{dep.Key}",
+                        dep.Value);
+                }
 
                 // Fill in the Metrics From Test Host Process
                 var metrics = runCompleteArgs.Metrics;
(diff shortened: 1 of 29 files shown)


src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -42,6 +46,35 @@ internal class AssemblyResolver : IDisposable
 
     private static readonly string[] SupportedFileExtensions = [".dll", ".exe"];
 
+    /// <summary>
+    /// Assemblies that vstest resolved from its own directory on behalf of an extension
+    /// that did not ship its own copy. Key: assembly name, Value: requesting assembly name.
```
nohwnd: can we collect this?

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -42,6 +46,35 @@ internal class AssemblyResolver : IDisposable
...
+    /// </summary>
+    private static readonly HashSet<string> TrackedAssemblies = new(StringComparer.OrdinalIgnoreCase)
+    {
+        "Newtonsoft.Json",
+    };
+
+    /// <summary>
+    /// The directory where vstest's own assemblies live.
+    /// </summary>
+    private static readonly string TestPlatformDirectory =
```
nohwnd: is this used as parent dir, or just this dir? we also check in extensions etc.

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -216,6 +251,57 @@ internal void AddSearchDirectories(IEnumerable<string> directories)
         }
     }
 
+    /// <summary>
+    /// When we resolve a tracked assembly (e.g. Newtonsoft.Json) from vstest's own directory
+    /// for an external requester, record it for telemetry and optionally warn the user.
```
nohwnd: we cannot send recorded user dlls to telemetry, we can record them locally so user should get all the info they can in the warning, but we cannot send the real user dlls names to telemetry. 

for telemetry count user dlls separately from microsoft.* and system.* dlls so we have some ideas if these deps come as direct dependency or indirect, or by our own mistake.

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -216,6 +251,57 @@ internal void AddSearchDirectories(IEnumerable<string> directories)
...
+
+        if (!ProvidedDependencies.TryAdd(requestedName.Name, requester))
+        {
+            return;
+        }
+
+        EqtTrace.Info("AssemblyResolver.TrackProvidedDependency: Resolved '{0}' from vstest directory for '{1}'. " +
+            "The extension should ship its own copy.", requestedName.Name, requester);
+
+        // When the feature flag is set, warn the user so they can fix their extension.
```
nohwnd: this is not a good time to report the warning, you should get all info and report it in summary of the run or similar, so we don't only show the first one that depends on given dll.

src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
```diff
@@ -78,6 +78,11 @@ private FeatureFlag() { }
     // Disable turning dynamic code coverage for native code to OFF by default. Setting this to 1 will skip adding the setting.
     public const string VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING = nameof(VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING);
 
+    // When set, warns when vstest provides its own copy of a tracked assembly (e.g. Newtonsoft.Json)
+    // to resolve a dependency of an extension that did not ship its own copy.
+    // Set VSTEST_WARN_MISSING_EXTENSIONS_DEPENDENCIES=1 to opt into warnings before enforcement.
+    public const string VSTEST_WARN_MISSING_EXTENSIONS_DEPENDENCIES = nameof(VSTEST_WARN_MISSING_EXTENSIONS_DEPENDENCIES);
```
nohwnd: All feature flags should be named: VSTEST_DISABLE_WARN_MISSING_EXTENSIONS_DEPENDENCIES, add new mechanism to opt-in by setting VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -440,6 +441,15 @@ public void HandleTestRunComplete(TestRunCompleteEventArgs runCompleteArgs, Test
...
                 _requestData.MetricsCollection.Add(TelemetryDataConstants.TimeTakenInSecForRun, executionTotalTimeTaken.TotalSeconds);
 
+                // Record assemblies that vstest provided from its own directory
+                var providedDeps = AssemblyResolver.GetProvidedDependencies();
+                foreach (var dep in providedDeps)
+                {
+                    _requestData.MetricsCollection.Add(
+                        $"{TelemetryDataConstants.ProvidedAssemblyDependency}.{dep.Key}",
+                        dep.Value);
+                }
```
nohwnd: this is really unwieldy for telemetry, just add 1 property for dlls we provided to user and 1 property for dlls we provided because dlls from Microsoft. or System. asked for them.

Then list all the dlls provided joined by |,

you must not show user dlls names. only our well known names.

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -440,6 +441,38 @@ public void HandleTestRunComplete(TestRunCompleteEventArgs runCompleteArgs, Test
...
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForUserCount, userCount);
+                }
+                if (msCount > 0)
+                {
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForMicrosoft, msAssemblies);
+                    _requestData.MetricsCollection.Add(TelemetryDataConstants.ProvidedDependenciesForMicrosoftCount, msCount);
+                }
+
+                // Opt-in warning: show all provided assemblies in summary when feature flag is set.
+                if ((userCount > 0 || msCount > 0) && FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES))
```
nohwnd: user can optin, but for the future also add the disable flag to disable this, if user specified disable then disable, iif disable is not set user still has to opt-in in this version.

src/Microsoft.TestPlatform.CoreUtilities/FeatureFlag/FeatureFlag.cs
```diff
@@ -78,6 +78,11 @@ private FeatureFlag() { }
     // Disable turning dynamic code coverage for native code to OFF by default. Setting this to 1 will skip adding the setting.
     public const string VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING = nameof(VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING);
 
+    // When set, warns when vstest provides its own copy of a tracked assembly (e.g. Newtonsoft.Json)
+    // to resolve a dependency of an extension that did not ship its own copy.
+    // Set VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES=1 to opt into warnings before enforcement.
+    public const string VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES = nameof(VSTEST_OPTIN_WARN_MISSING_EXTENSIONS_DEPENDENCIES);
```
nohwnd, line 84: add flag VSTEST_DISABLe_WARN_MISSING_EXTENSIONS_DEPENDENCIES which will disable the feature

nohwnd (commented):
^^^

src/Microsoft.TestPlatform.Client/Execution/TestRunRequest.cs
```diff
@@ -440,6 +441,38 @@ public void HandleTestRunComplete(TestRunCompleteEventArgs runCompleteArgs, Test
...
+                {
+                    var allAssemblies = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+                    foreach (var kvp in AssemblyResolver.GetProvidedDependencies())
+                    {
+                        allAssemblies.Add(kvp.Key);
+                    }
+
+                    var message = $"The test platform provided the following assemblies to resolve dependencies of test extensions "
```
nohwnd: You need to localize this

nohwnd (commented):
sdf

Evangelink (approved):

PR #15638, opened 2026-04-07
Create source-only filter package

Youssef1313:
Fixes https://github.com/microsoft/vstest/issues/15361


Youssef1313:
@copilot Please review this PR and add a comment with your findings. During review, ensure that all classes in the new package are internal and have embedded attribute (except when under `#if IS_VSTEST_REPO`).

src/Microsoft.TestPlatform.Filter.Source/Microsoft.TestPlatform.Filter.Source.nuspec
```diff
@@ -0,0 +1,19 @@
+﻿<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    $CommonMetadataElements$
+    <readme>README.md</readme>
```
Evangelink, line 5: **Critical: Missing `<contentFiles>` metadata in nuspec — source files won't auto-compile in consuming projects**

The nuspec places `.cs` files under `contentFiles/cs/netstandard2.0/VSTestFilterSource/`, but there is no `<contentFiles>` section inside `<metadata>`. Without this, NuGet does **not** set `buildAction="Compile"` on the content files, so consuming projects will not compile these sources automatically. The package will install without errors but the types won't be available at all.

This is the standard pattern for source-only packages:

```xml
<metadata>
    $CommonMetadataElements$
    <readme>README.md</readme>
    <contentFiles>
      <files include="cs/**/*.cs" buildAction="Compile" />
    </contentFiles>
</metadata>
```

See the [NuGet contentFiles documentation](https://learn.microsoft.com/en-us/nuget/reference/nuspec#including-content-files) for reference. Without this addition, this package is non-functional for consumers.

You may also want to add `copyToOutput="false"` and `flatten="false"` to the `<files>` entry to be explicit about the behavior.

src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
```diff
@@ -100,20 +124,24 @@ private static void ProcessOperator(Stack<FilterExpression> filterStack, Operato
         }
         else if (op == Operator.OpenBrace)
         {
-            throw new FormatException(string.Format(CultureInfo.CurrentCulture, CommonResources.TestCaseFilterFormatException, CommonResources.MissingCloseParenthesis));
+            throw new FormatException(string.Format(CultureInfo.CurrentCulture, TestCaseFilterFormatException, MissingOperand));
```
Evangelink: **Bug: Wrong error message — `MissingCloseParenthesis` replaced with `MissingOperand`**

The original code here was:
```csharp
throw new FormatException(... CommonResources.MissingCloseParenthesis);
```

This has been changed to use `MissingOperand`, but the `Operator.OpenBrace` case means we encountered an unmatched open parenthesis. The correct error is that a **close parenthesis** is missing, not that an operand is missing.

It looks like `MissingCloseParenthesis` was accidentally omitted from the fallback constants declared in the `#if !IS_VSTEST_REPO` block at the top of this file (lines 35-49). You have `MissingOperand`, `EmptyParenthesis`, `MissingOpenParenthesis`, and `MissingOperator` — but `MissingCloseParenthesis` is missing.

Suggest fixing by:
1. Adding a `MissingCloseParenthesis` constant: `private const string MissingCloseParenthesis = "Error: Missing ')'";`
2. Restoring this line to use `MissingCloseParenthesis` instead of `MissingOperand`.

This is also a behavioral change for the `IS_VSTEST_REPO` path (i.e. existing vstest users), since the resource string reference also changed silently from `CommonResources.MissingCloseParenthesis` → `MissingOperand`.

src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
```diff
@@ -45,6 +44,9 @@
       <LastGenOutput>Resources\CommonResources.resx</LastGenOutput>
     </None>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Include="$(RepoRoot)src/Microsoft.TestPlatform.Filter.Source/FilterHelper.cs" />
```
Evangelink, line 48: **Potential duplicate type compilation: `FilterHelper` may compile twice**

This `<Compile Include>` adds `FilterHelper.cs` from the new `Filter.Source` folder. However, SDK-style projects auto-compile all `.cs` files under the project directory. The original `Utilities/FilterHelper.cs` was *renamed/moved* in Git (to `Filter.Source/FilterHelper.cs`), so after the rename, the original file no longer exists in `src/Microsoft.TestPlatform.ObjectModel/Utilities/` — this should be fine.

However, please double-check that:
1. The Git rename is clean (no leftover `Utilities/FilterHelper.cs` in the working tree after checkout)
2. No other project in the build graph still auto-includes the original path

If the original file somehow remains (e.g., on a developer machine that had it from a previous branch), both copies would compile and produce duplicate `FilterHelper` type errors, which are confusing to debug. Adding an explicit `<Compile Remove="Utilities/FilterHelper.cs" />` as a safety net would make the intent unambiguous and prevent such accidents.

src/Microsoft.TestPlatform.Filter.Source/Condition.cs
```diff
@@ -52,6 +72,14 @@ internal sealed class Condition
...
     public const Operation DefaultOperation = Operation.Contains;
 
+#if !IS_VSTEST_REPO
+    private const string TestCaseFilterFormatException = "Incorrect format for TestCaseFilter {0}. Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.";
+
+    private const string InvalidCondition = "Error: Invalid Condition '{0}'";
+
+    private const string InvalidOperator = "Error: Invalid operator '{0}'";
+#endif
+
```
Evangelink, line 81: **DRY violation: `TestCaseFilterFormatException` string is duplicated across 3 files**

The same error message string `"Incorrect format for TestCaseFilter {0}..."` is declared as a `const string` in the `#if !IS_VSTEST_REPO` block here, in `FilterExpression.cs`, and implicitly through a differently-named constant in `FilterHelper.cs`.

Similarly, these `!IS_VSTEST_REPO` constants are scattered across files:
- `Condition.cs`: `TestCaseFilterFormatException`, `InvalidCondition`, `InvalidOperator`
- `FilterExpression.cs`: `TestCaseFilterFormatException`, `MissingOperand`, `EmptyParenthesis`, `MissingOpenParenthesis`, `MissingOperator`  
- `FilterHelper.cs`: `TestCaseFilterEscapeException`

Consider extracting all these into a single internal static class (e.g., `FilterResources.cs`) in the source package with a `#if !IS_VSTEST_REPO` guard. This:
1. Eliminates the copy-paste drift risk if messages ever need correction
2. Makes it easy to verify all required strings are present (the missing `MissingCloseParenthesis` from my other comment would have been more obvious)
3. Mirrors the pattern used in the `IS_VSTEST_REPO` path where they come from a centralized resource file

test/Microsoft.TestPlatform.Filter.Source.UnitTests/Microsoft.TestPlatform.Filter.Source.UnitTests.csproj
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <IsFilterSourcePackage>true</IsFilterSourcePackage>
+    <TargetFrameworks>$(NetCurrent);net48</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <ExcludeRepoTestProjectSettings>true</ExcludeRepoTestProjectSettings>
+    <ExcludeMicrosoftNetTestSdk>true</ExcludeMicrosoftNetTestSdk>
+  </PropertyGroup>
```
Evangelink, line 9: **Test project may not run in CI: missing `TestProject` / `IsTestProject` properties + unused `ExcludeMicrosoftNetTestSdk`**

Other test projects in this repo set `<TestProject>true</TestProject>` and/or `<IsTestProject>true</IsTestProject>`, which is what `Directory.Build.targets` uses in the `<Choose><When Condition="$(TestProject) == 'true'">` block to wire up test infrastructure (runner references, test settings, etc.).

Without these, the tests in this project may not be discovered or executed in CI at all, even though they build fine locally. The `ExcludeRepoTestProjectSettings=true` is a good idea (to avoid pulling in repo-specific test settings that conflict with the source-linked approach), but you still need the base `TestProject`/`IsTestProject` markers to get test runner integration.

Also, `ExcludeMicrosoftNetTestSdk` does not appear to be consumed by any `.targets`/`.props` in this repo — it seems to be a no-op. If you intended it to suppress the implicit `Microsoft.NET.Test.Sdk` reference, verify that the corresponding conditional logic exists, or remove it to avoid confusion.

Suggested fix:
```xml
<PropertyGroup>
    <IsFilterSourcePackage>true</IsFilterSourcePackage>
    <TargetFrameworks>$(NetCurrent);net48</TargetFrameworks>
    <OutputType>Exe</OutputType>
    <TestProject>true</TestProject>
    <ExcludeRepoTestProjectSettings>true</ExcludeRepoTestProjectSettings>
</PropertyGroup>
```

src/Microsoft.TestPlatform.Filter.Source/FastFilter.cs
```diff
@@ -3,31 +3,55 @@
...
+#endif
     }
 
+#if IS_VSTEST_REPO
     internal ImmutableDictionary<string, ISet<string>> FilterProperties { get; }
+#else
+    internal Dictionary<string, ISet<string>> FilterProperties { get; }
+#endif
 
     internal bool IsFilteredOutWhenMatched { get; }
```
Evangelink, line 59: **Thread-safety difference with `Dictionary` vs `ImmutableDictionary` — worth documenting**

The `IS_VSTEST_REPO` path uses `ImmutableDictionary<string, ISet<string>>` while the source package path uses `Dictionary<string, ISet<string>>`. This is a meaningful behavioral difference:

- `ImmutableDictionary` is inherently safe for concurrent reads without locking
- `Dictionary` is **not** thread-safe for concurrent access (even read-only, if built with certain code paths)

If any consumer of this source-only package evaluates filters concurrently (which is a reasonable thing to do for test frameworks running tests in parallel), they could hit race conditions. The trade-off of avoiding a `System.Collections.Immutable` dependency for netstandard2.0 is understandable, but:

1. Consider using `ReadOnlyDictionary<string, ISet<string>>` wrapping a `Dictionary` — it at least prevents accidental mutation, and concurrent reads on a `Dictionary` that's no longer being mutated are safe in practice.
2. At minimum, document this in the README so consumers are aware.
3. The `HashSet<string>` used in the builder path (vs `ImmutableHashSet`) has the same concern — the sets are stored and later read during `Evaluate()`. Once built, they should not be modified, but `HashSet` doesn't enforce that.

src/Microsoft.TestPlatform.Filter.Source/TestCaseFilterExpression.cs
```diff
@@ -4,15 +4,26 @@
...
+#if IS_VSTEST_REPO
 using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter;
+#endif
 
 namespace Microsoft.VisualStudio.TestPlatform.Common.Filtering;
 
 /// <summary>
 /// Implements ITestCaseFilterExpression, providing test case filtering functionality.
 /// </summary>
```
Evangelink, line 22: **Minor style: Missing space before `:` (SA1024)**

The original code was:
```csharp
public class TestCaseFilterExpression : ITestCaseFilterExpression
```

The `IS_VSTEST_REPO` path is now:
```csharp
public class TestCaseFilterExpression: ITestCaseFilterExpression
```

Missing the space before `:`. This violates SA1024 (Colon must be preceded by a space) from the StyleCop analyzers configured in this repo. Should be:
```csharp
public class TestCaseFilterExpression : ITestCaseFilterExpression
```

eng/verify-nupkgs.ps1
```diff
@@ -30,6 +30,7 @@ function Verify-Nuget-Packages {
         "Microsoft.TestPlatform.TestHost"             = 64
         "Microsoft.TestPlatform.TranslationLayer"     = 175
         "Microsoft.TestPlatform.Internal.Uwp"         = 39
+        "Microsoft.TestPlatform.Filter.Source"        = 13
```
Evangelink, line 33: **Verify expected file count of 13 after actual pack**

This count needs to match the actual number of files in the produced `.nupkg`. Given the nuspec includes:
- 7 `.cs` files (Condition, EmbeddedAttribute, FastFilter, FilterExpression, FilterExpressionWrapper, FilterHelper, TestCaseFilterExpression)
- 1 `README.md`
- `$CommonFileElements$` (which typically adds LICENSE, THIRD_PARTY_NOTICES, `.nuspec` itself, etc.)

The count of 13 is plausible but should be verified by running a test pack — especially after the nuspec `<contentFiles>` issue is fixed (see my other comment), since adding that metadata section may change the package contents.

Consider adding a comment here explaining how the number was derived, like the pattern:
```powershell
# 7 .cs source files + README.md + 5 common files (LICENSE, THIRD-PARTY-NOTICES, .nuspec, icon, ...)
"Microsoft.TestPlatform.Filter.Source"        = 13
```

src/Microsoft.TestPlatform.Filter.Source/Condition.cs
```diff
@@ -168,33 +208,42 @@ internal static Condition Parse(string? conditionString)
...
 
+#if IS_VSTEST_REPO
             // Check if operation ~ (Contains) is on property of type string.
             if (Operation == Operation.Contains)
             {
                 valid = ValidForContainsOperation(propertyProvider);
             }
+#endif
         }
         return valid;
```
Evangelink, line 242: **Behavioral difference: `Contains` operator validation skipped for source package consumers**

In the `!IS_VSTEST_REPO` path, when `Operation == Operation.Contains`, the `ValidForContainsOperation` check is skipped entirely. This means the source-only package won't validate that the `~` (Contains) operator is only used on string-typed properties.

For most consumers this is fine (test-framework adapters typically only deal with string properties), and the `ValidForContainsOperation` requires `TestProperty` which is unavailable. But it's a semantic behavior difference worth noting — a filter like `Priority~1` would be reported as "valid" by the source-only package even though `Priority` might be an integer property where `Contains` doesn't make sense.

Consider adding a brief note about this in the README: "The `Contains` (`~`) operator type validation is not performed in the source-only package since `TestProperty` metadata is not available."

Evangelink (commented):
## Code Review Summary

Overall this is a well-structured approach to extracting the filter implementation into a source-only NuGet package. The `#if IS_VSTEST_REPO` pattern keeps a single source of truth while allowing divergence for internal/external consumers, and the `[Embedded]` attribute usage follows the standard .NET convention for source-only packages.

### Issues found (by priority)

**Critical (blocks package functionality):**
1. **Missing `<contentFiles>` metadata in nuspec** — Without declaring `<contentFiles><files include="cs/**/*.cs" buildAction="Compile" /></contentFiles>` in the `<metadata>` block, NuGet won't set `buildAction="Compile"` on the packaged `.cs` files. The package will install but the types won't be available to consumers. This makes the package non-functional.

**Bug:**
2. **`MissingCloseParenthesis` → `MissingOperand` substitution in `FilterExpression.cs`** — The `Operator.OpenBrace` error path now reports "Missing operand" instead of the correct "Missing close parenthesis". The `MissingCloseParenthesis` constant was not added to the `#if !IS_VSTEST_REPO` fallback constants. This also affects the `IS_VSTEST_REPO` path since the resource reference changed.

**Medium:**
3. **Duplicated error message strings** across `Condition.cs`, `FilterExpression.cs`, and `FilterHelper.cs` — `TestCaseFilterFormatException` is copy-pasted in multiple files. Consider extracting to a single `FilterResources` class.
4. **Test project missing `TestProject`/`IsTestProject`** — Tests may not run in CI. `ExcludeMicrosoftNetTestSdk` appears unused.
5. **`Dictionary` vs `ImmutableDictionary` thread-safety** — The source package uses mutable `Dictionary` instead of `ImmutableDictionary`, which could be a problem for consumers evaluating filters concurrently.

**Low:**
6. **Missing space before `:` in `TestCaseFilterExpression`** (SA1024 violation)
7. **Verify nupkg file count of 13** after fixing the `<contentFiles>` metadata
8. **`Contains` operator validation silently skipped** in the source package — worth documenting

### What looks good
- The file-sharing approach via `<Compile Include>` is clean and avoids duplication
- `[Embedded]` attribute usage is correct
- Test coverage is comprehensive (parsing, evaluation, operators, error cases, fast filter)
- Test project cleverly uses `ExcludeAssets="compile"` on ObjectModel to avoid conflicts
- The `IS_VSTEST_REPO` define being set in `Directory.Build.targets` with an opt-out for the filter package is well-designed

src/Microsoft.TestPlatform.Filter.Source/Microsoft.TestPlatform.Filter.Source.nuspec
Youssef1313: Polyfill is already a functional package and doesn't do what Copilot is suggesting here. https://github.com/SimonCropp/Polyfill/blob/a1f5dea8cbb38a2680dd899acd9685a9ba2ce4b7/src/Polyfill/Polyfill.nuspec

Youssef1313: @zivkan Can you please advise here?

src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj
Youssef1313: This comment doesn't make sense, and everything is compiling cleanly already. Resolving.

test/Microsoft.TestPlatform.Filter.Source.UnitTests/Microsoft.TestPlatform.Filter.Source.UnitTests.csproj
Youssef1313: Already works correctly. Resolving.

eng/verify-nupkgs.ps1
Youssef1313: 13 is correct.

src/Microsoft.TestPlatform.Filter.Source/Condition.cs
Youssef1313: This is probably fine.

src/Microsoft.TestPlatform.Filter.Source/FastFilter.cs
Youssef1313: Makes no sense.

src/Microsoft.TestPlatform.Filter.Source/Microsoft.TestPlatform.Filter.Source.nuspec
zivkan: Looking at NuGet's implementation, the default build action for any content files that don't have nuspec metadata is compile. So for source code, it's technically unnecessary to add the nuspec content file metadata. Even if it's not documented, changing it would break a lot of packages/projects, so I'm sure the behavior will never change. 

Youssef1313: I only used nuspec because other projects in this solution use it, but yeah it makes sense to start moving away, which we should consider doing in a follow-up PR for the whole solution.

Thanks for clarifying the behavior. I'll keep it as-is for now and follow-up to move away from nuspec completely.

test/Microsoft.TestPlatform.Filter.Source.UnitTests/Microsoft.TestPlatform.Filter.Source.UnitTests.csproj
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <IsFilterSourcePackage>true</IsFilterSourcePackage>
+    <TargetFrameworks>$(NetCurrent);net48</TargetFrameworks>
```
nohwnd, line 5: are we setting the tfm like this in other projects?

nohwnd (approved):

nohwnd:
Looks good. Consider adding integration test to Library.IntegrationTests that uses the package to ensure it works.

test/Microsoft.TestPlatform.Filter.Source.UnitTests/Microsoft.TestPlatform.Filter.Source.UnitTests.csproj
Youssef1313: Other projects seem to hardcode net9.0. Do you prefer keeping consistency or are you fine with NetCurrent?

Youssef1313:
@copilot add an integration test in Microsoft.TestPlatform.Library.IntegrationTests for this package.

Youssef1313:
@copilot add an integration test in Microsoft.TestPlatform.Library.IntegrationTests for this package.

Youssef1313:
@copilot Add an integration test in Microsoft.TestPlatform.Library.IntegrationTests for this package.

test/Microsoft.TestPlatform.Filter.Source.UnitTests/Microsoft.TestPlatform.Filter.Source.UnitTests.csproj
nohwnd: It is fine I will try to update to net11.0 and net481 soon anyway to reduce how much we do, per tfm breaks are almost non-existent.

Youssef1313 (commented):
@copilot Please add explicit `#nullable enable` in all files in C# code files in `Microsoft.TestPlatform.Filter.Source`.

eng/common/tools.sh
Youssef1313, line 1: @copilot Revert the file mode change here

test/TestAssets/FilterSourcePackageConsumerTests/FilterSourcePackageConsumerTests.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        var wrapper = new FilterExpressionWrapper("(Name=Test1");
+        bool result = wrapper.Evaluate(p => "Test1");
+        Assert.IsFalse(result);
+    }
+
+    [TestMethod]
+    public void EvaluateShouldSupportAndOperator()
+    {
+        var wrapper = new FilterExpressionWrapper("Name=Test1&Category=Unit");
```
bart-vmware, line 46: If this is the most advanced expression the parser supports, then I truly don't see the point of adding it. As a replacement for existing test-framework-specific filter syntaxes, it needs to support nested compound expressions, such as: `name~integration&(OS=ubuntu|windows)`, or something similar to the capabilities described at https://xunit.net/docs/query-filter-language. Adding yet another primitive-only syntax is confusing rather than adding value.

Youssef1313: This is not adding yet another syntax. This is the syntax that existed for years in VSTest.

This PR only exposes a package so that non-VSTest frameworks (e.g, xunit.v3 running with MTP) can get the implementation without copying it from this repo, and without having to add an additional dependency for VSTest.

bart-vmware: Ok, thanks. So this is unrelated to the plans for adding a provider-agnostic filter syntax when MTP is used? Would it make sense to expand this syntax for compound expressions, while retaining compatibility with the original VSTest syntax?

PR #15668, opened 2026-04-13
Add unit tests for DOTNET_ROOT forwarding logic and remote process architecture detection

nohwnd:
This PR adds unit tests for several under-tested code paths in dotnet test host startup.

## What's covered

**DotnetTestHostManagerTests.cs** — 7 new tests targeting early DOTNET\_ROOT path-setting and legacy testhost handling:
- Verifies DOTNET\_ROOT\_\<ARCH\> is set early (before testhost.exe found) so xunit v3 child-processes pick it up
- Verifies the early-set path is skipped when the feature flag is enabled
- Verifies existing DOTNET\_ROOT\_\<ARCH\> env vars are not overwritten
- Covers the legacy (old-testhost) x64 and x86 DOTNET\_ROOT paths
- Covers architecture-mismatch guard on old testhost path

**ProcessHelperTests.cs** *(new file)* — 3 tests for remote-process architecture detection:
- GetCurrentProcessArchitecture returns a valid Architecture value
- Repeated calls return the same cached result
- GetProcessArchitecture throws NotImplementedException on non-Windows (expected guard)

## Why these tests

These paths had no test coverage despite being critical to test host startup on Linux/macOS and to multi-architecture scenarios. The 7-test DotnetTestHostManager block and 3-test ProcessHelper block together close a meaningful reliability gap.


nohwnd:

PR #15680, opened 2026-04-15
Report in-flight tests and suggest --blame-crash when testhost crashes

### LINKED ISSUE #2952: "Test host process crashed" error hard to diagnose
Our CI has an intermittent failure:

```
The active test run was aborted. Reason: Test host process crashed
Results File: C:\a\_temp\AzDevOps_2019-6vse00024V_2021-06-24_02_47_47.trx
Test Run Aborted.
```

As far as I can tell, the message is printed by this code:
https://github.com/microsoft/vstest/blob/eff66c00b217b355b6ee11034ff8396a618d04e3/src/Microsoft.TestPlatform.CommunicationUtilities/TestRequestSender.cs#L667

It would be really helpful here if it printed the full path to the process executable that crashed. I took a quick look but I couldn't find where to get the executable full path from. Also it seems that the error output stream contents was empty (`clientExitErrorMessage`) so when the test process crashes it should print the full exception stack to the Console.Error so that the stack appears in the CI log.

nohwnd:
## Problem

When testhost crashes, users get an unhelpful `Test host process crashed` message with no indication of which test was running or how to diagnose. This is the most upvoted open issue (#2952, 26 thumbs up).

## Solution

Always-on blame lite: lightweight real-time test tracking without requiring `--blame`.

New protocol messages `TestCaseStarting`/`TestCaseFinished` sent unbatched from testhost (~280 bytes per test). Console tracks in-flight tests in `InFlightTestTracker` with inline-slot optimization.

On crash, users now see:

```
Test host process crashed : <crash details>
Tests that were executing when the crash occurred:
  MyTest.CrashingMethod (running for 2s)
Consider using --blame-crash to collect a crash dump for further diagnosis.
```

## Testing

3 new integration tests (6 variants), all existing tests pass.

## Backward compatibility

Older testhosts: console shows old behavior plus `--blame-crash` suggestion. Older consoles: unknown messages silently ignored.

Fixes #2952

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/EventHandlers/TestRunEventsHandler.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/EventHandlers/TestRunEventsHandler.cs
@@ -20,6 +20,11 @@ public class TestRunEventsHandler : IInternalTestRunEventsHandler
     private readonly ITestRequestHandler _requestHandler;
 
     /// <summary>
+    /// Gets the underlying request handler for direct message sending.
+    /// </summary>
+    internal ITestRequestHandler RequestHandler => _requestHandler;
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="TestRunEventsHandler"/> class.
     /// </summary>
     /// <param name="requestHandler">test request handler</param>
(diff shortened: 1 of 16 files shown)


nohwnd:
Closing — will revisit when we get to #2952. The issue tracks the need.

PR #15711, opened 2026-04-21
Exclude .gitignore and .vscode from pipeline

Evangelink:


azure-pipelines.yml
```diff
@@ -24,8 +24,10 @@ pr:
   paths:
     exclude:
       - .github/*
+      - .gitignore
```
nohwnd: nit: it should be below, with the files, and not here with the folders.

Evangelink: @copilot address this comment

nohwnd (approved):

PR #15719, opened 2026-04-22
Port verify-binding-redirects.ps1 to rel/18.5

nohwnd:
Ports the binding redirect verification script (and its wiring from erify-nupkgs.ps1) from `main`/`rel/18.7` to `rel/18.5`, so we can validate on this branch that the `newVersion` in each binding redirect actually matches the shipped assembly.

The script runs against the extracted `Microsoft.TestPlatform.*.nupkg` layout produced by `verify-nupkgs.ps1`. For each of `vstest.console.exe`, `testhost.x86.exe`, and `datacollector.exe` it:

- finds the exe in the packaged layout
- follows `<probing privatePath>` from the app.config
- for every `<dependentAssembly>` reads the shipped DLL version
- compares it to the redirect's `newVersion`

In CI it fails with the list of mismatches plus fix instructions. Locally it auto-updates the source `app.config` files (preserving whitespace and BOM) so you can just commit the result.

Ran cleanly on `rel/18.5` HEAD (v18.5.0) against a Debug build — everything green, including `System.Collections.Immutable 9.0.0.0` which was the one we wanted to verify.

### DIFF
--- a/eng/verify-binding-redirects.ps1
+++ b/eng/verify-binding-redirects.ps1
@@ -1,0 +1,211 @@
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+
+$script:isCI = $env:TF_BUILD -eq 'true' -or $env:CI -eq 'true'
+
+# Verifies that binding redirects in source app.config files match the actual
+# assembly versions of the DLLs shipped in the extracted nupkg packages.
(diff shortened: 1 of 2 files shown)


nohwnd:
Flaky on ccparallel.

PR #15728, opened 2026-04-28
Verify TFM of every managed DLL in shipped packages

nohwnd:
Add `eng/verify-dll-frameworks.ps1` that reads the `TargetFrameworkAttribute` from every DLL in the NuGet packages and compares against `eng/expected-dll-frameworks.json`. Native DLLs are excluded by name.

In CI the script errors on mismatch. Locally it auto-fixes the JSON, same as `verify-binding-redirects.ps1` and `verify-nupkgs.ps1`.

The TFM is read via binary search with custom-attribute blob validation (prolog + compressed length prefix) to avoid false positives from resource DLLs that embed TFM-like strings.


nohwnd:
making more changes , noticed that e.g. system.memory does not have framework

drognanar (approved):

PR #15731, opened 2026-04-29
Remove duplicate NuGet packages from packaging projects and no-op binding redirects

nohwnd:
The packaging projects (Microsoft.TestPlatform, Portable, CLI) explicitly reference NuGet packages that already flow transitively through project references. The CopyFiles targets then overwrite build-produced DLLs with ones extracted from NuGet — potentially different versions. We end up shipping different things than we build with.

**Removed packages (truly transitive):**
- System.Collections.Immutable — transitive via SRM. Was 9.0.0.0, now correctly 8.0.0.0 (matches what code compiles against).
- System.Reflection.Metadata — direct dep of ObjectModel and CoreUtilities.
- Microsoft.Extensions.FileSystemGlobbing — transitive via vstest.console (kept in Portable, netstandard2.0 has no project refs).
- Microsoft.Diagnostics.NETCore.Client — transitive via BlameDataCollector.

**Kept as explicit refs:**
- DependencyModel — TestHostProvider uses Jsonite on .NET Framework (not DependencyModel), pruned by NuGet on .NET 10, and Portable netstandard2.0 has no project refs.
- FileSystemGlobbing — kept in Portable only (same reason).

Also added \CopyLocalLockFileAssemblies=true\ in \src/package/Directory.Build.props\ so transitive NuGet DLLs actually get copied to output for these library-style packaging projects.

**Binding redirects cleaned up:**

Removed all no-op redirects where every assembly reference in the dependency graph already targets the shipped version. Verified with a PEReader/MetadataReader-based tool that reads AssemblyRef tables from all DLLs.

Remaining redirects are genuinely needed:
- \ObjectModel 11-15 → 15\ (backward compat with old test adapters)
- \CompilerServices.Unsafe 1-6 → 6\ (System.Memory 4.5.5 refs 4.0.4.1, we ship 6.0.0.0)
- \TestWindow.Interfaces 11-18 → 18\ (VS integration, testhost.x86 only)
- \UnitTestFramework 10.1 → 10.0\ (MSTest v1 compat, testhost.x86 only)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -92,3 +92,9 @@ When working autonomously on issues (e.g. from a milestone), follow this workflo
 ### Troubleshooting
 
 - If a build fails with warnings treated as errors (e.g. IDE0005 unnecessary using), and you cannot reproduce locally, try building with `-c Release` to match CI: `./build.cmd -c Release`.
+
+### Testing rigor
+
+- **Never dismiss test failures as "pre-existing" without verifying.** If tests fail, investigate the actual cause. Find out how to make them pass — check test parameters, missing build steps, environment setup. Run them again with the right setup.
+- Acceptance tests need `--test-parameter BuildCompatibility=true` to build compatibility matrix test assets. Without it, ~100 tests will fail with "Path not found" errors.
+- If a test run has failures, look at the error messages and figure out what's needed. Don't stop at "probably pre-existing."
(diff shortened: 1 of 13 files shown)


stasyu2009-ux (commented):
Все відбувається вірно

stasyu2009-ux (commented):
gh pr checkout 15731

nohwnd:
comment in english please .  @copilot, can you translate the comments for me?

nohwnd:

PR #15753, opened 2026-05-12
[fix] Fix DataDrivenTest double-counted in TRX ResultSummary counters

nohwnd:
🤖 This is an automated fix generated by Copilot.

Fixes #15643

## Root Cause

When a `[DataTestMethod]` with `[DataRow]` attributes runs, MSTest reports two kinds of results to vstest:

1. **Parent aggregation result** — one result for the `DataTestMethod` itself (no `parentExecId`), which is a `DataDrivenTest` container.
2. **Child results** — one result per `DataRow` (with `parentExecId` pointing to the parent).

`TestResultHandler` was incrementing the total/passed/failed counters on **every** call without distinguishing between parent and child results. This meant the parent aggregation was counted once, then each child was counted again, inflating the `ResultSummary` counters in the TRX file.

**Example**: 1 test with 2 `[DataRow]`s → TRX reported `total=3`, `passed=3` instead of `total=2`, `passed=2`.

## Fix

When the first DataRow child result is processed (`DataRowInfo == 1`), the parent aggregation's contribution to the counters is undone via `Interlocked.Decrement`. Only the child results represent actual test executions.

Ordered tests are unaffected: their children use a different `ResultType` (`InnerDataDrivenResultType` is not set for ordered test children), so the decrement path is never taken.

## Testing

- Updated `TestResultHandlerShouldAddHierarchicalResultsIfParentTestResultIsPresent` to assert the correct count (`2` children, not `3` total).
- Added `TestResultHandlerShouldNotDoubleCountDataDrivenTestResults` — a dedicated regression test that verifies `TotalTestCount`, `PassedTestCount`, and `FailedTestCount` are all correct for a data-driven test with 2 DataRows.
- All 69 TrxLogger unit tests pass.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/25731263546)*

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -302,7 +302,10 @@ public class TrxLogger : ITestLoggerWithParameters
         // Update test entries
         UpdateTestEntries(executionId, parentExecutionId, testElement, parentTestElement);
 
-        // Set various counts (passed tests, failed tests, total tests)
+        // Set various counts (passed tests, failed tests, total tests).
+        // Data-driven tests (DataRow) report both a parent aggregation result and individual child results.
+        // Only the child results represent actual test executions, so when the first child is processed
+        // (DataRowInfo == 1), we undo the parent's contribution to the counts to avoid double-counting.
         Interlocked.Increment(ref _totalTestCount);
         if (testResult.Outcome == TrxLoggerObjectModel.TestOutcome.Failed)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -313,6 +316,18 @@ internal void TestResultHandler(object? sender, TestResultEventArgs e)
         {
             Interlocked.Increment(ref _passedTestCount);
         }
+
+        // Undo the parent DataDrivenTest aggregation's count when the first child result is processed.
+        if (testResult.ResultType == TrxLoggerConstants.InnerDataDrivenResultType
+            && testResult.DataRowInfo == 1
+            && parentTestResult != null)
```
nohwnd, line 323: [Correctness] The `parentTestResult != null` guard is correct here, but it implicitly relies on the parent aggregation result always arriving **before** the first child result.

If a child result with `DataRowInfo == 1` arrives before its parent has been processed (e.g., due to a race or out-of-order delivery), `GetTestResult(parentExecutionId)` returns `null`, the decrement is skipped, and the parent will still be double-counted when it arrives later.

This appears to be an existing ordering assumption in the handler (e.g., `CreateTestResult` also requires the parent to be in `_results` to populate `InnerResults` correctly), so this PR doesn't introduce a regression. Worth noting for future robustness: an explicit ordering guarantee from the caller would make this contract visible.

nohwnd (commented):
The fix is logically correct. The `DataRowInfo` assignment happens *after* `InnerResults.Add` (line 727 before line 728), so the first child correctly gets `DataRowInfo = 1` and the sentinel check is sound. The increment/decrement pair is symmetric with the existing counting code, and the `parentTestResult != null` guard is appropriate.

One ordering assumption is noted inline — it's pre-existing and not a regression introduced by this PR.

🧠 *Reviewed by Expert Code Reviewer*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25732365047)*

nohwnd:
Closing — opportunity work without a linked issue, or superseded. Will revisit if/when a reported issue drives the need.

PR #15755, opened 2026-05-12
Diagnose environment block size exceeding 65535-char limit

nohwnd:
Diagnostic PR to identify which environment variables are consuming the env block budget and how much overhead CodeDom/csc.exe adds when launching the compiler.

**Not intended to merge** — will be closed once we have the data.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -162,36 +162,77 @@ extends:
               image: windows.vs2026preview.scout.amd64
               os: windows
             steps:
-            - pwsh: |
-                # Environment block on Windows is UTF-16: each entry is "NAME=VALUE\0" → (len + 2) * 2 bytes.
-                # The block itself ends with an extra \0 (2 bytes).
+            - powershell: |
+                # .NET Framework Process.Start checks env block LENGTH IN CHARS against 65535,
+                # but the error message misleadingly says "bytes". Measure in chars to match.
                 $entries = [System.Environment]::GetEnvironmentVariables().GetEnumerator() | ForEach-Object {
                     $name  = [string]$_.Key
                     $value = [string]$_.Value
-                    $bytes = ($name.Length + $value.Length + 2) * 2  # +1 for '=', +1 for '\0', *2 for UTF-16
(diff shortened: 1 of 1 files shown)


nohwnd (commented):
## Code Review — Diagnostic Pipeline Script

This PR adds PowerShell diagnostic steps to two jobs in `azure-pipelines-official.yml` to measure the CI environment block size and quantify CodeDom/csc.exe overhead. Since it touches only a CI pipeline file and is explicitly marked "not intended to merge", the review focuses on correctness of the diagnostic logic.

### Observations

**Correctness of `$expected` calculation**

In the "over-pad" block:
```powershell
$expected = $totalChars + $overPad + "__ENVPROBE".Length + 2
```
`$totalChars` is computed at the top of the script *before* any `__ENVPROBE` manipulation. This is correct as long as `__ENVPROBE` does not already exist in the CI environment at script start. If it does (e.g. from a prior job or system policy), the overhead calculation would be understated. A safer form would capture the initial env snapshot excluding `__ENVPROBE`, or simply assert/log its absence at the start.

**`Add-Type` type-name reuse across script restarts**
Each pipeline step runs in its own `powershell` process, so `__EnvProbeType` won't collide between the two duplicate jobs. Within a single step, the second probe correctly uses `__EnvProbeType2` to avoid "type already exists" errors. This is fine.

**Minor: `2>$null` redirect on `Add-Type`**
The `2>$null` suppresses the stderr output from the `catch` block's exception, but since the exception is caught and printed via `Write-Host`, this is intentional. However, it also suppresses any unexpected errors written to stderr (e.g., assembly resolution warnings), which could make failures harder to diagnose. Consider removing it so all diagnostic output is visible.

### Overall

The diagnostic logic is sound for its stated purpose. No production code, public API, IPC protocol, or thread safety concerns are present. The PR is already closed.

---
🧠 *Reviewed by GitHub Copilot Expert Reviewer*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25744160337)*

PR #15766, opened 2026-05-13
[fix] Fix DataDriven test results being double-counted in TRX logger totals

nohwnd:
🤖 This is an automated fix generated by GitHub Copilot.

Closes #15643

## Root Cause

When a DataDriven (data-row) test is executed, MSTest emits:
1. A **parent container result** first (with its own `ExecutionId`, no `ParentExecId`)
2. One **inner result per data row** (each with `ParentExecId` pointing to the parent)

`TrxLogger.TestResultHandler` was incrementing `TotalTestCount` (and `PassedTestCount`/`FailedTestCount`) for **every** result it received — including the parent container. For a test with *N* data rows, this meant a count of *N+1* instead of the correct *N*.

## Fix

In `TestResultHandler`, after creating the test result, we detect when the first inner DataDriven result is processed (parent element is a `UnitTestType` and the aggregation now has exactly one inner result). At that point we undo the count that was previously recorded for the parent container, since the parent is just a grouping construct and not an actual test execution.

The fix is scoped only to `UnitTestType` parents (i.e., DataDriven tests). Ordered tests are unaffected — their parent and each child are genuinely independent test cases.

## Changes

- `src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs` — decrement parent counts when first inner DataDriven result arrives
- `test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs` — correct existing assertion from `3` → `2`, add new regression test `TestResultHandlerShouldNotDoubleCountParentDataDrivenTestInTotalTestCount`

## Testing

All TrxLogger unit tests pass.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/25818024026)*

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -313,6 +313,25 @@ public class TrxLogger : ITestLoggerWithParameters
         {
             Interlocked.Increment(ref _passedTestCount);
         }
+
+        // When the first inner DataDriven result is encountered, the parent is promoted to a
+        // DataDrivenTest container. Undo the count that was previously recorded for the parent,
+        // since the parent is only a container and should not be counted as a separate test result.
+        if (parentTestElement != null
+            && parentTestElement.TestType.Equals(TrxLoggerConstants.UnitTestType)
+            && parentTestResult is TestResultAggregation aggregation
+            && aggregation.InnerResults.Count == 1)
+        {
+            Interlocked.Decrement(ref _totalTestCount);
+            if (parentTestResult.Outcome == TrxLoggerObjectModel.TestOutcome.Failed)
+            {
+                Interlocked.Decrement(ref _failedTestCount);
+            }
+            else if (parentTestResult.Outcome == TrxLoggerObjectModel.TestOutcome.Passed)
+            {
+                Interlocked.Decrement(ref _passedTestCount);
+            }
+        }
     }
 
     /// <summary>
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -297,7 +297,8 @@ public class TrxLogger : ITestLoggerWithParameters
         UpdateTestLinks(testElement, parentTestElement);
 
         // Convert the rocksteady result to trx test result
-        var testResult = CreateTestResult(executionId, parentExecutionId, testType, testElement, parentTestElement, parentTestResult, e.Result);
+        var testResult = CreateTestResult(executionId, parentExecutionId, testType, testElement, parentTestElement, parentTestResult, e.Result,
+            out bool isFirstDataDrivenInnerResult);
 
         // Update test entries
         UpdateTestEntries(executionId, parentExecutionId, testElement, parentTestElement);
@@ -317,13 +318,12 @@ public class TrxLogger : ITestLoggerWithParameters
         // When the first inner DataDriven result is encountered, the parent is promoted to a
         // DataDrivenTest container. Undo the count that was previously recorded for the parent,
         // since the parent is only a container and should not be counted as a separate test result.
-        if (parentTestElement != null
-            && parentTestElement.TestType.Equals(TrxLoggerConstants.UnitTestType)
-            && parentTestResult is TestResultAggregation aggregation
-            && aggregation.InnerResults.Count == 1)
+        // isFirstDataDrivenInnerResult is set atomically via Interlocked.Increment, so this is
+        // race-safe even when multiple data-row results are processed concurrently.
+        if (isFirstDataDrivenInnerResult)
         {
             Interlocked.Decrement(ref _totalTestCount);
-            if (parentTestResult.Outcome == TrxLoggerObjectModel.TestOutcome.Failed)
+            if (parentTestResult!.Outcome == TrxLoggerObjectModel.TestOutcome.Failed)
             {
                 Interlocked.Decrement(ref _failedTestCount);
             }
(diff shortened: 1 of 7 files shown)


src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
```diff
@@ -313,6 +313,25 @@ internal void TestResultHandler(object? sender, TestResultEventArgs e)
...
             Interlocked.Increment(ref _passedTestCount);
         }
+
+        // When the first inner DataDriven result is encountered, the parent is promoted to a
+        // DataDrivenTest container. Undo the count that was previously recorded for the parent,
+        // since the parent is only a container and should not be counted as a separate test result.
+        if (parentTestElement != null
+            && parentTestElement.TestType.Equals(TrxLoggerConstants.UnitTestType)
+            && parentTestResult is TestResultAggregation aggregation
+            && aggregation.InnerResults.Count == 1)
```
nohwnd: [Threading] The `InnerResults.Count == 1` guard is evaluated *after* the inner result was already appended inside `CreateTestResult`, but the check is not atomic with the subsequent `Interlocked.Decrement` calls.

If two inner data-row results for the **same** parent are processed concurrently (e.g., parallel test execution pushing events to the logger from two threads), both threads could append their result to the `List<ITestResult>` (itself not thread-safe) and then both read `Count == 1` before either count changes — triggering a **double-decrement** that leaves `_totalTestCount` one below the correct value.

The existing `Interlocked` usage elsewhere in this handler signals that concurrent calls are expected.

One approach: instead of checking `InnerResults.Count`, capture the count returned by `Interlocked.Increment` on a dedicated `long _innerResultCount` per parent and gate on count == 1. A simpler alternative is a small `lock` around the aggregation mutation + count check + decrement block. Either way, the fix as written is race-prone for parallel result delivery.

nohwnd (commented):
The fix correctly undoes the over-count of a DataDriven parent container result on the first inner result arrival, and the regression test is well-structured and covers the basic scenario.

One threading concern noted inline: the `InnerResults.Count == 1` check is not atomic with the `Interlocked.Decrement` calls, which can produce incorrect counts if two inner results for the same parent are processed concurrently. The existing use of `Interlocked` operations in `TestResultHandler` suggests concurrent invocation is expected, so this warrants a fix before merge.

---
🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/25818665240)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25818665240)*

nohwnd:
Commit pushed: [`1405864`](https://github.com/microsoft/vstest/commit/14058642b7b82b0507bc0fc043694861da9be12b)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/25819089551)*

src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
nohwnd: Fixed. Added `_innerResultCount` (a `long` field) to `TestResultAggregation` and a new `AddInnerResult(ITestResult)` method that:
1. Locks `_innerResults` for a thread-safe `List.Add`
2. Calls `Interlocked.Increment(ref _innerResultCount)` and returns the new count

`CreateTestResult` now calls `AddInnerResult` instead of `InnerResults.Add` and exposes the returned count via an `out bool isFirstDataDrivenInnerResult` parameter. `TestResultHandler` gates the parent-count decrement on that flag, which is set only when the Interlocked return value equals 1 — race-safe regardless of concurrent delivery order.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/25819089551)*

nohwnd (commented):
The threading concern flagged in the [prior review](https://github.com/microsoft/vstest/pull/15766#pullrequestreview-4284432870) has been correctly resolved.

**`AddInnerResult` fix analysis:**
- `lock (this)` protects the `List<ITestResult>` mutation (correct — `List<T>` is not thread-safe)
- `Interlocked.Increment` outside the lock returns the new counter value atomically, guaranteeing exactly **one** thread receives `innerCount == 1` regardless of concurrent calls
- `isFirstDataDrivenInnerResult = innerCount == 1` is therefore race-safe ✅

**Increment/decrement symmetry:**
The increment logic (lines 307–316) uses `testResult.Outcome` at parent-processing time. The new decrement logic uses `parentTestResult.Outcome`. These refer to the same TRX object (stored in `_results[executionId]` when the parent was processed and retrieved as `_results[parentExecutionId]` when the inner result arrives), so the symmetry is correct ✅

**No new issues introduced.** The fix is well-scoped and the regression test covers the key scenario.

---
🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25819725965)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25819725965)*

Evangelink:
@copilot resolve the merge conflicts in this pull request

nohwnd:
---
> [!WARNING]
> The `push_to_pull_request_branch` operation failed: Failed to apply bundle. The code changes were not applied.

Resolved the merge conflict by rebasing onto `main`. The conflict was in `LoggerTests.cs` where both the PR branch and main added a new test method at the same location — I kept both tests. Build passes ✅



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26915015424)*

nohwnd:
Commit pushed: [`5e0bffd`](https://github.com/microsoft/vstest/commit/5e0bffdb9bc4be10d32c615f2200ee60cd9eda29)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27120900569)*

PR #15770, opened 2026-05-14
Fix workflow config gaps: mentions and PR description

nohwnd:
- Enable @nohwnd mentions in daily-qa so the digest actually sends email notifications
- Fix pr-iteration PR description check: can't rewrite (no write permission), so flag in a comment instead


nohwnd (commented):
Workflow-only changes. Both fixes are correct and well-scoped:

- `mentions: true` + `allowed-github-references: ["@nohwnd"]` in `daily-qa.md` enables email notifications as intended, and the compiled lock file is consistent.
- The `pr-iteration.md` change correctly switches from "rewrite the description" (requires write permission the workflow doesn't have) to "flag it in a review comment" — a sensible fix for the permission constraint.

No .NET source changes; nothing to flag from a correctness, threading, security, or API compatibility standpoint.

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25850224036)*

PR #15786, opened 2026-05-14
Remove unnecessary binding redirects from app.configs

nohwnd:
Remove all binding redirects from testhost.x86, datacollector, and vstest.console app.configs. These were redirects for ObjectModel (old v11-14 adapter compat), VS-only DLLs (TestWindow.Interfaces, UnitTestFramework), System polyfills (Memory, Buffers, CompilerServices.Unsafe, Collections.Immutable, Reflection.Metadata, FileSystemGlobbing), and DLLs that don't even ship (DiagnosticSource, Encodings.Web — see #15765 for the damage those caused).

Updated verify-binding-redirects.ps1 to enforce zero redirects going forward — it now fails if any \<dependentAssembly>\ elements are found, instead of trying to keep them in sync.

### DIFF
--- a/eng/verify-binding-redirects.ps1
+++ b/eng/verify-binding-redirects.ps1
@@ -1,13 +1,9 @@
 $ErrorActionPreference = "Stop"
 Set-StrictMode -Version Latest
 
-$script:isCI = $env:TF_BUILD -eq 'true' -or $env:CI -eq 'true'
-
-# Verifies that binding redirects in source app.config files match the actual
-# assembly versions of the DLLs shipped in the extracted nupkg packages.
-#
-# In CI: validates and fails with instructions to run locally.
-# Locally: auto-fixes the source app.config files with the correct versions.
+# Verifies that no binding redirects exist in source app.config files.
+# Binding redirects are a legacy mechanism that causes version mismatches
+# and should not be used. See https://github.com/microsoft/vstest/issues/15765.
 
 # Each source app.config maps to a specific exe that ships in the packages.
 $script:AppConfigs = @(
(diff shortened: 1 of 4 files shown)


nohwnd:
Love how optimistically this is described :D let's see it crumble :D

nohwnd (commented):
## Review summary

The overall direction is correct — removing stale binding redirects and replacing the sync-version logic with a zero-tolerance check is a clear improvement, well-motivated by issue #15765.

Two issues in `eng/verify-binding-redirects.ps1`:

1. **Logic bug in "OK" output** (line 52): `if (-not $errors)` tests the *global* accumulator, not the current entry. A later clean config won't print "OK" if any earlier config failed. Needs a per-iteration flag (details inline).

2. **Orphaned mandatory parameters** (lines 17–18): `$PackageDirs` and `$Configuration` are still `[Parameter(Mandatory)]` but the function no longer uses them. The call in `verify-nupkgs.ps1` still satisfies the constraint, so it won't error at runtime — but they're dead code that should be cleaned up.

No concerns with the app.config changes themselves; the deletions match the description exactly.

🧠 *Reviewed by expert-reviewer workflow*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25867293304)*

nohwnd:
I've adopted this PR via `/iterate` and registered it for monitoring. The Azure DevOps CI build is currently in progress — I'll watch for failures and address any issues when results come in.



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/25867800253)*

nohwnd:
Closing - removing all binding redirects breaks testhost because it loads arbitrary user assemblies that may reference older versions of shipped DLLs (e.g. System.Collections.Immutable 8.0.0.0 when we ship 10.0.0.0). The redirects for shipped DLLs are actually needed for version unification. The real fix is scoped: remove only orphaned redirects (DLLs we don't ship), which is handled by #15778 (merged) and #15776/#15777 for release branches.

nohwnd:
Understood — removing this from my tracking. The correct approach is the scoped fix (removing only orphaned redirects for DLLs we don't ship), which is covered by #15778 and the release branch PRs. Closing this out.



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/25869943745)*

PR #15791, opened 2026-05-15
[fix] Fix TRX attachment paths when LogFileName contains a subdirectory

nohwnd:
🤖 This is an automated fix generated by the Issue Triage agent.

Fixes #15271

## Root Cause

When `--logger trx;LogFileName=subdir/results.trx` is used, the TRX file is placed at `<TestRunDirectory>/subdir/results.trx`. However, the internal `_testResultsDirPath` (used to organize test attachments) remained set to the original `TestRunDirectory`.

TRX viewers resolve the `runDeploymentRoot` attribute relative to the directory containing the TRX file. So a viewer opening `subdir/results.trx` expects attachments at `subdir/<runDeploymentRoot>/In/...`, but they were actually placed at `<runDeploymentRoot>/In/...` (one level up) — making them unreachable.

## Fix

In `Initialize(events, parameters)`, detect when `LogFileName` has a directory component and pre-adjust `_testResultsDirPath` to point to the directory that will contain the TRX file. This ensures both the TRX and its attachment folder are co-located.

In `GetTrxFilePath`, use `Path.GetFileName(logFileNameValue)` instead of the full `logFileNameValue` so the directory portion isn't appended twice (it's already in `_testResultsDirPath`).

The fix is backward-compatible: when `LogFileName` is a plain filename with no directory component, `Path.GetDirectoryName` returns an empty string and no adjustment is made.

## Test

Added `CustomTrxFileNameWithSubdirectoryShouldPlaceTrxAndAttachmentsUnderSameDirectory` which verifies that when `LogFileName=subdir/results.trx`, the TRX file lands at `<TestRunDirectory>/subdir/results.trx`.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/25919396337)*

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
+++ b/src/Microsoft.TestPlatform.Extensions.TrxLogger/TrxLogger.cs
@@ -154,7 +154,23 @@ public class TrxLogger : ITestLoggerWithParameters
         }
 
         _parametersDictionary = parameters;
-        Initialize(events, _parametersDictionary[DefaultLoggerParameterNames.TestRunDirectory]!);
+
+        var testRunDirectory = _parametersDictionary[DefaultLoggerParameterNames.TestRunDirectory]!;
+
+        // If LogFileName contains a directory component, attachments should be placed relative
+        // to the directory that will contain the TRX file so that TRX viewers can resolve them.
+        if (!isLogFilePrefixParameterExists
+            && _parametersDictionary.TryGetValue(TrxLoggerConstants.LogFileNameKey, out string? logFileNameInit)
+            && !logFileNameInit.IsNullOrWhiteSpace())
+        {
+            var logFileDir = Path.GetDirectoryName(logFileNameInit);
+            if (!logFileDir.IsNullOrWhiteSpace())
+            {
+                testRunDirectory = Path.Combine(testRunDirectory, logFileDir);
+            }
+        }
+
+        Initialize(events, testRunDirectory);
     }
     #endregion
 
@@ -511,7 +527,9 @@ public class TrxLogger : ITestLoggerWithParameters
             }
             else if (isLogFileNameParameterExists)
             {
-                filePath = Path.Combine(_testResultsDirPath, logFileNameValue!);
+                // _testResultsDirPath already includes any subdirectory from LogFileName (adjusted in Initialize),
+                // so use only the filename portion here to avoid duplicating the directory.
+                filePath = Path.Combine(_testResultsDirPath, Path.GetFileName(logFileNameValue!));
                 shouldOverwrite = true;
             }
         }
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/LoggerTests.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/LoggerTests.cs
@@ -254,4 +254,33 @@ public class LoggerTests : AcceptanceTestBase
 
         return null;
     }
+
+    [TestMethod]
+    [NetFullTargetFrameworkDataSource]
+    [NetCoreTargetFrameworkDataSource]
+    public void TrxLoggerShouldPlaceTrxFileInSubdirectoryWhenLogFileNameContainsPath(RunnerInfo runnerInfo)
+    {
+        // Regression test for https://github.com/microsoft/vstest/issues/15271
+        // When LogFileName contains a subdirectory (e.g. "subdir/results.trx"),
+        // the TRX file and its attachments should be placed under that subdirectory.
+        SetTestEnvironment(_testEnvironment, runnerInfo);
+
+        var assemblyPaths = GetAssetFullPath("SimpleTestProject2.dll");
+        var subDir = "custom-subdir";
+        var trxFileName = "results.trx";
+        var logFileName = Path.Combine(subDir, trxFileName);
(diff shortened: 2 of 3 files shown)


nohwnd (commented):
## Code Review

**Dimensions activated**: Null Safety & Boundary Validation, Backward Compatibility & Rollback Safety, Error Reporting & Diagnostic Clarity

### Summary

The fix correctly addresses the root cause. When `LogFileName` contains a directory component, `_testResultsDirPath` is now pre-adjusted to include that subdirectory, ensuring both the TRX file and its attachment folder land in the same directory. The `AcquireTrxFileNamePath` change correctly uses `Path.GetFileName` to avoid double-inclusion of the subdirectory.

### Correctness Analysis

**Path edge cases — all handled correctly:**
- Plain filename (`results.trx`): `Path.GetDirectoryName` returns `""`, which is null-or-whitespace → no adjustment. ✓
- Absolute path (`/tmp/results.trx`): `Path.Combine(testRunDirectory, "/tmp")` → absolute path wins, same behavior as before. ✓
- Nested subdirectories (`dir1/dir2/results.trx`): `Path.GetDirectoryName` returns `dir1/dir2` → correctly combined. ✓
- `LogFilePrefix` guard: the adjustment is skipped when `isLogFilePrefixParameterExists` is true, matching the `AcquireTrxFileNamePath` branch logic. ✓

**`runDeploymentRoot` in TRX**: `CreateTestRun` derives `runDeploymentRoot` from the run name (not `_testResultsDirPath`), so it's always a relative folder name. The fix ensures this relative folder resolves correctly from the TRX file's directory because `_testResultsDirPath` now points to the same directory as the TRX file. ✓

### One Minor Note

The new test doesn't clean up the `subdir` directory it creates under `Path.GetTempPath()` — only the TRX file itself is deleted. However, this is consistent with the existing test style in this file (e.g., `CustomTrxFileNameShouldConstructFromLogFileParameter` performs no cleanup at all), and using the system temp path means it's not harmful. The new test is actually *more* careful than its peers by having a `try/finally` block at all.

No blocking issues. The fix is correct and backward-compatible.

---
🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/25920080102)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/25920080102)*

nohwnd:
Commit pushed: [`69eced6`](https://github.com/microsoft/vstest/commit/69eced614aa1bd5d31830864eda7f1ed4c928f97)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26078473501)*

nohwnd (commented):
## Code Review

**Dimensions activated**: Null Safety & Boundary Validation, Backward Compatibility & Rollback Safety, Acceptance Test Coverage Design, Cross-TFM & Framework Resolution

### Summary

No blocking issues. The fix is correct and backward-compatible. This is a re-review pass — the head commit (`69eced6`) is only a CI retrigger with no code changes since the [previous review](https://github.com/microsoft/vstest/pull/15791#pullrequestreview-4298488870).

### Dimension Findings

**Null Safety & Boundary Validation** ✓
- `logFileNameInit.IsNullOrWhiteSpace()` correctly guards the `TryGetValue` result before `Path.GetDirectoryName`.
- `logFileDir.IsNullOrWhiteSpace()` correctly skips the adjustment for plain filenames (where `Path.GetDirectoryName` returns `""`).
- `Path.GetFileName(logFileNameValue!)` is sound — it's called inside the `isLogFileNameParameterExists` branch, which guarantees the value was set.

**Backward Compatibility & Rollback Safety** ✓
- Plain filename (`results.trx`): `Path.GetDirectoryName` returns `""` → `IsNullOrWhiteSpace` is true → no adjustment → identical behavior to before. ✓
- Subdirectory (`subdir/results.trx`): adjustment applied, attachments land co-located with the TRX. ✓
- Absolute path (`/tmp/results.trx`): `Path.Combine(testRunDirectory, "/tmp")` → absolute path wins on both platforms → `_testResultsDirPath` = `/tmp`; TRX file = `/tmp/results.trx` — same final location as before, and attachments are now also correctly co-located. ✓
- The `!isLogFilePrefixParameterExists` guard correctly mirrors the `LogFilePrefix` vs `LogFileName` branch logic in `AcquireTrxFileNamePath`. ✓

**Acceptance Test Coverage Design** ✓
- Test covers both `NetFullTargetFrameworkDataSource` and `NetCoreTargetFrameworkDataSource` — good TFM breadth.
- Uses `TempDirectory.Path` for the results directory, so cleanup is automatic.
- Asserts both file existence and XML validity, providing meaningful failure messages including a glob of all TRX files found.

**Cross-TFM & Framework Resolution** ✓
- All APIs used (`Path.GetDirectoryName`, `Path.GetFileName`, `Path.Combine`, `string.IsNullOrWhiteSpace`) are available on all vstest target TFMs (net462 / netstandard2.0 / net8.0+).

---
🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26079279434)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26079279434)*

Evangelink (approved):

PR #15792, opened 2026-05-15
Migrate dn-bot-devdiv-drop-rw-code-rw PAT to WIF service connection

missymessa:
## Summary

Replaces the \dn-bot-devdiv-drop-rw-code-rw\ PAT with a WIF-based service connection (\dnceng-devdiv-drop-rw-code-rw-wif\) for DevDiv drop access.

## Changes

- **Add AzureCLI@2 task** in the Publish job to acquire an Entra token via the WIF service connection before the MicroBuildVstsDrop task
- **Set \_DevDivDropAccessToken\** at runtime using the WIF-acquired token instead of the static PAT from \DotNet-VSTS-Infra-Access\ variable group
- **Remove static PAT variable reference** from the variable declarations

## Context

This is part of the PAT-to-Entra migration tracked by [AB#10146](https://dev.azure.com/dnceng/internal/_workitems/edit/10146). The WIF service connection \dnceng-devdiv-drop-rw-code-rw-wif\ (backed by app registration \dnceng-devdiv-drop-rw-code-rw-wif\, App ID: \7106a410-fbcb-4750-a202-879077f925ec\) was previously created and verified. The same pattern was successfully deployed in [dotnet/fsharp#19598](https://github.com/dotnet/fsharp/pull/19598).

## Validation

> **Note:** PR validation pipelines read YAML from \main\, not the PR branch. The actual WIF migration will be validated on the first post-merge CI build.

### DIFF
--- a/azure-pipelines-official.yml
+++ b/azure-pipelines-official.yml
@@ -95,8 +95,9 @@ variables:
   - group: DotNet-Symbol-Server-Pats
   # Group gives access to $dn-bot-devdiv-drop-rw-code-rw and dn-bot-dnceng-build-rw-code-rw
   - group: DotNet-VSTS-Infra-Access
+  # DevDiv drop access token is acquired via WIF service connection (dnceng-devdiv-drop-rw-code-rw-wif)
   - name: _DevDivDropAccessToken
-    value: $(dn-bot-devdiv-drop-rw-code-rw)
+    value: ''
   - name: _SignType
     value: real
   - name: _SignArgs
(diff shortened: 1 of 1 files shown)


missymessa:
Thanks for looking at this @nohwnd! The pipeline trigger exclusion is expected — the PR validation YAML on the `main` branch excludes paths or the PR source branch isn't covered by the trigger config.

**Important:** Even if PR validation runs, it will execute the **old YAML from `main`** (not the PR branch), so it won't actually validate the WIF change. The real validation happens post-merge on the first CI build from `main`.

**Context on why this is safe:**
- The WIF service connection (`dnceng-devdiv-drop-rw-code-rw-wif`) has been validated in roslyn (builds green since May 22), fsharp, test-templates, and interactive-window — all use the same SC and the same `1ES.MicroBuildVstsDrop@1` task pattern.
- The SC is already authorized for the vstest pipeline (def 1222).
- The only change is swapping `$(dn-bot-devdiv-drop-rw-code-rw)` (PAT from variable group) for the WIF-based `$(_DevDivDropAccessToken)` sourced from an `AzureCLI@2` task.

If you'd like a manual test build before merge, I can queue one from the `dnceng/internal` side with the source branch override. Otherwise, merging and monitoring the first post-merge build is the standard approach we've used for the other repos.

nohwnd (approved):

PR #16050, opened 2026-05-21
[fix] Fix flaky PortArgumentProcessor test: cache port locally instead of re-reading singleton

### LINKED ISSUE #15730: flaky on linux ci
Test method Microsoft.VisualStudio.TestPlatform.CommandLine.UnitTests.Processors.PortArgumentProcessorTests.ExecutorExecuteForValidConnectionReturnsArgumentProcessorResultSuccess threw exception:
Moq.MockException:
Expected invocation on the mock once, but was 0 times: td => td.ConnectToClientAndProcessRequests(2345, Mock[ITestRequestManager:41](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1398137&view=ms.vss-test-web.build-test-results-tab&runId=38994310&resultId=100397&paneView=debug).Object)

Performed invocations:

Mock[IDesignModeClient:9](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1398137&view=ms.vss-test-web.build-test-results-tab&runId=38994310&resultId=100397&paneView=debug) (td):

          

              <BUTTON  aria-label="Copy to clipboard"  class='copy-btn bolt-button bolt-icon-button enabled subtle icon-only bolt-focus-treatment'>
                Copy to clipboard
                
                   
                
              </BUTTON>
              
  IDesignModeClient.ConnectToClientAndProcessRequests(1234, Mock<ITestRequestManager:41>.Object)

nohwnd:
🤖 This is an automated fix generated by the Issue Repro Triage agent.

Fixes #15730

## Root Cause

`PortArgumentExecutor.Execute()` reads `_commandLineOptions.Port` from the shared `CommandLineOptions.Instance` singleton. `InProcessVsTestConsoleWrapper` starts vstest.console in-process by calling `executor.Execute(args)` on a background `Task.Run` thread. This eventually calls `PortArgumentExecutor.Initialize("1234")`, which sets `CommandLineOptions.Instance.Port = 1234`.

If this background thread races with another test that called `Initialize("2345")` and is about to call `Execute()`, the background thread overwrites `CommandLineOptions.Instance.Port` from 2345 to 1234 before `Execute()` reads it — causing the `ConnectToClientAndProcessRequests` mock to be invoked with the wrong port.

## Fix

Capture the port number in a private `_port` field during `Initialize()` and use that captured value in `Execute()` instead of re-reading the shared singleton. The assignment to `_commandLineOptions.Port` is preserved for callers that read it from `CommandLineOptions.Instance`.

## Test

The 11 existing `PortArgumentProcessorTests` all pass. The flaky test `ExecutorExecuteForValidConnectionReturnsArgumentProcessorResultSuccess` is directly covered by the fix — it now uses the locally-captured port rather than the shared mutable singleton.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26229333372)*

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26229333372)*

### DIFF
--- a/src/vstest.console/Processors/PortArgumentProcessor.cs
+++ b/src/vstest.console/Processors/PortArgumentProcessor.cs
@@ -90,6 +90,11 @@ internal class PortArgumentExecutor : IArgumentExecutor
     private IDesignModeClient? _designModeClient;
 
     /// <summary>
+    /// Port number captured during Initialize, used in Execute to avoid re-reading the shared singleton.
+    /// </summary>
+    private int _port;
+
+    /// <summary>
     /// Process helper for process management actions.
     /// </summary>
     private readonly IProcessHelper _processHelper;
@@ -140,6 +145,7 @@ internal class PortArgumentExecutor : IArgumentExecutor
             throw new CommandLineException(CommandLineResources.InvalidPortArgument);
         }
 
+        _port = portNumber;
         _commandLineOptions.Port = portNumber;
         _commandLineOptions.IsDesignMode = true;
         RunSettingsHelper.Instance.IsDesignMode = true;
@@ -154,11 +160,11 @@ internal class PortArgumentExecutor : IArgumentExecutor
     {
         try
         {
-            _designModeClient?.ConnectToClientAndProcessRequests(_commandLineOptions.Port, _testRequestManager);
+            _designModeClient?.ConnectToClientAndProcessRequests(_port, _testRequestManager);
         }
         catch (TimeoutException ex)
         {
-            throw new CommandLineException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.DesignModeClientTimeoutError, _commandLineOptions.Port), ex);
+            throw new CommandLineException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.DesignModeClientTimeoutError, _port), ex);
         }
 
         return ArgumentProcessorResult.Success;


nohwnd (commented):
## Review: `src/vstest.console/` — Process Architecture & Scheduling Safety

**Activated dimensions:** Parallel Execution & Scheduling Safety, Process Architecture & Host Resolution

### Analysis

The fix is correct and minimal. `PortArgumentExecutor.Execute()` was reading `_commandLineOptions.Port` from the shared `CommandLineOptions.Instance` singleton, which is mutable across test runs in `InProcessVsTestConsoleWrapper`. Caching the value in the instance field `_port` during `Initialize()` eliminates the TOCTOU race.

**Checked items:**

- ✅ **Race condition eliminated** — `_port` is an instance field, so each `PortArgumentExecutor` instance holds its own captured value. The write in `Initialize()` happens-before the read in `Execute()` within the same logical call chain.
- ✅ **Default value is safe** — If `Execute()` is called without `Initialize()`, `_port` is 0, which matches prior behavior (the singleton would also have been 0 or unset). No regression.
- ✅ **Singleton assignment preserved** — `_commandLineOptions.Port = portNumber` is still set, so callers reading the singleton continue to work correctly.
- ✅ **Both uses of the port updated** — Both `ConnectToClientAndProcessRequests` and the `TimeoutException` message now use `_port`. No dangling read of the singleton in `Execute()`.
- ✅ **PR description matches the diff** — The root cause, fix, and affected test are accurately described.

No issues found.

---
🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/26230784367)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26230784367)*

PR #16055, opened 2026-05-23
[fix] Throw when /diag is passed manually in integration tests

nohwnd:
Fixes #15625.

> 🤖 This is an automated fix generated by [Issue Repro Triage & Auto-Fix](https://github.com/microsoft/vstest/actions/runs/26333096992).

## Root Cause

Integration tests that passed `/diag` (or `--diag` / `-diag`) manually in their arguments would cause `IsDiagAlreadyEnabled` to return `true`, silently skipping the auto-injection of diagnostic logging. This meant those tests would not have their diagnostic logs collected as attachments on failure — making failures harder to diagnose.

## What the Fix Does

1. **`IntegrationTestBase.cs`**:
   - `IsDiagAlreadyEnabled` now **throws `InvalidOperationException`** when it detects a manual `/diag` argument, with a helpful message directing the test author to use `DiagLogsDirectory` instead.
   - Added `DiagLogsDirectory` property (`{TempDirectory.Path}/logs`) so tests can reference the auto-injected diag path.
   - Added `GetDiagLogContents()` protected helper that reads all `.txt` log files from `DiagLogsDirectory` — promoted from several test classes that each had their own copy.
   - Simplified `InvokeVsTest` and `InvokeDotnetTest` to use `DiagLogsDirectory` directly.

2. **All tests that manually passed `/diag`**: Updated to remove the manual argument. The auto-injection now ensures diag logs are always collected.

3. **Tests that checked diag file content** (`ExecutionTests.cs`, `DataCollectorTests.Coverlet.cs`, `DataCollectionTests.cs`): Updated to use `DiagLogsDirectory`-based search patterns instead of custom paths.

4. **`CreateNoNewWindowTests.cs` / `SerializerSelectionTests.cs`**: Removed duplicate `GetDiagLogContents()` private methods — now using the base class version.

## Tests

The existing acceptance tests verify this behavior. The fix ensures diagnostic logs are always attached on test failure, which improves debuggability in CI.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26333096992)*

### DIFF
--- a/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
+++ b/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
@@ -223,7 +223,20 @@ public class IntegrationTestBase
         if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments ?? ""))
         {
             Directory.CreateDirectory(DiagLogsDirectory);
-            arguments = string.Concat(arguments, GetDiagArg(DiagLogsDirectory));
+            var diagArg = GetDiagArg(DiagLogsDirectory);
+
+            // Insert --diag before the -- separator so it's treated as a vstest.console argument,
+            // not as a runsettings parameter.
+            var separatorPos = arguments?.IndexOf(" -- ") ?? -1;
+            if (separatorPos == -1)
+            {
+                arguments = string.Concat(arguments, diagArg);
+            }
+            else
+            {
+                arguments = arguments!.Insert(separatorPos, diagArg);
+            }
+
             _attachments.Add(DiagLogsDirectory);
             Console.WriteLine($"Diagnostic logs directory: {DiagLogsDirectory}");
         }
--- a/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
+++ b/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
@@ -220,7 +220,8 @@ public class IntegrationTestBase
     /// <param name="collectDiagnostics">When true, automatically adds --diag flag and attaches logs to test results on failure.</param>
     public void InvokeVsTest(string? arguments, Dictionary<string, string?>? environmentVariables = null, bool collectDiagnostics = true)
     {
-        if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments ?? ""))
+        ThrowIfDiagInArguments(arguments ?? "");
+        if (collectDiagnostics && !IsDiagInEnvironment())
         {
             Directory.CreateDirectory(DiagLogsDirectory);
             var diagArg = GetDiagArg(DiagLogsDirectory);
@@ -302,7 +303,8 @@ public class IntegrationTestBase
         // https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/commands/dotnet-test/VSTestForwardingApp.cs#L30-L39
         debugEnvironmentVariables["VSTEST_CONSOLE_PATH"] = vstestConsolePath;
 
-        if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments))
+        ThrowIfDiagInArguments(arguments);
+        if (collectDiagnostics && !IsDiagInEnvironment())
         {
             Directory.CreateDirectory(DiagLogsDirectory);
             var diagPath = Path.Combine(DiagLogsDirectory, "log.txt");
@@ -1093,22 +1095,25 @@ public class IntegrationTestBase
         return string.Join(" ", GetTestDlls(assetNames).Select(a => a.AddDoubleQuote()));
     }
 
-    private static bool IsDiagAlreadyEnabled(string arguments)
+    // Matches /diag, --diag, or -diag as a standalone flag token (preceded by whitespace or start of string,
+    // followed by '=', ':', whitespace, or end of string). Avoids false positives from substrings like
+    // "--logger:my-diagnostics.trx". Internal for testability.
+    internal static readonly Regex DiagFlagPattern = new(@"(^|\s)(--diag|/diag|-diag)([:=\s]|$)", RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+    private static void ThrowIfDiagInArguments(string arguments)
     {
-        // If user passed /diag, --diag, or -diag as a standalone flag, throw to prevent silently skipping
-        // the auto-injected diag and losing the attachment. Tests that need to inspect diagnostic logs
-        // should use DiagLogsDirectory instead of passing /diag manually.
-        // The pattern requires the flag to appear as a standalone token (preceded by whitespace or start of
-        // string) to avoid false positives from substrings like "--logger:my-diagnostics.trx".
-        if (Regex.IsMatch(arguments, @"(^|\s)(--diag|/diag|-diag)([:=\s]|$)", RegexOptions.IgnoreCase))
+        var match = DiagFlagPattern.Match(arguments);
+        if (match.Success)
         {
             throw new InvalidOperationException(
-                "Do not pass /diag (or --diag / -diag) directly in test arguments. " +
+                $"Do not pass diag flags directly in test arguments (found '{match.Value.Trim()}' in '{arguments}'). " +
                 "The test framework auto-injects diagnostics and attaches them on failure. " +
-                "Use the DiagLogsDirectory property to find the path to the diagnostic log files.");
+                "Use IntegrationTestBase.DiagLogsDirectory to find the path to the diagnostic log files.");
         }
+    }
 
-        // Check environment variable
+    private static bool IsDiagInEnvironment()
+    {
         var envDiag = Environment.GetEnvironmentVariable("VSTEST_DIAG");
         return !StringUtils.IsNullOrEmpty(envDiag);
     }
@@ -1126,7 +1131,7 @@ public class IntegrationTestBase
             return string.Empty;
         }
 
-        return string.Join("\n", Directory.GetFiles(DiagLogsDirectory, "*.txt", SearchOption.AllDirectories).Select(File.ReadAllText));
+        return string.Join("\n", Directory.GetFiles(DiagLogsDirectory, "*.txt", SearchOption.TopDirectoryOnly).Select(File.ReadAllText));
     }
 
     /// <summary>
--- a/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
+++ b/test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
@@ -87,6 +87,12 @@ public class IntegrationTestBase
 
     public TempDirectory TempDirectory { get; }
 
+    /// <summary>
+    /// Returns the directory where diagnostic log files are automatically placed by <see cref="InvokeVsTest"/> and <see cref="InvokeDotnetTest"/>.
+    /// Use this property in tests that need to inspect the diagnostic log files.
+    /// </summary>
+    protected string DiagLogsDirectory => Path.Combine(TempDirectory.Path, "logs");
+
     public TestContext TestContext { get; set; } = null!;
 
     public string BuildConfiguration { get; }
@@ -216,11 +222,10 @@ public class IntegrationTestBase
     {
         if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments ?? ""))
         {
-            var diagLogsDir = Path.Combine(TempDirectory.Path, "logs");
-            Directory.CreateDirectory(diagLogsDir);
-            arguments = string.Concat(arguments, GetDiagArg(diagLogsDir));
-            _attachments.Add(diagLogsDir);
-            Console.WriteLine($"Diagnostic logs directory: {diagLogsDir}");
+            Directory.CreateDirectory(DiagLogsDirectory);
+            arguments = string.Concat(arguments, GetDiagArg(DiagLogsDirectory));
+            _attachments.Add(DiagLogsDirectory);
+            Console.WriteLine($"Diagnostic logs directory: {DiagLogsDirectory}");
         }
 
         var debugEnvironmentVariables = AddDebugEnvironmentVariables(environmentVariables);
@@ -286,9 +291,8 @@ public class IntegrationTestBase
 
         if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments))
         {
-            var diagLogsDir = Path.Combine(TempDirectory.Path, "logs");
-            Directory.CreateDirectory(diagLogsDir);
-            var diagPath = Path.Combine(diagLogsDir, "log.txt");
+            Directory.CreateDirectory(DiagLogsDirectory);
+            var diagPath = Path.Combine(DiagLogsDirectory, "log.txt");
             var diagArg = " --diag " + diagPath.AddDoubleQuote();
 
             // Insert --diag before the -- separator so dotnet test forwards it to vstest.console.
@@ -302,8 +306,8 @@ public class IntegrationTestBase
                 arguments = arguments.Insert(separatorPos, diagArg);
             }
 
-            _attachments.Add(diagLogsDir);
-            Console.WriteLine($"Diagnostic logs directory: {diagLogsDir}");
+            _attachments.Add(DiagLogsDirectory);
+            Console.WriteLine($"Diagnostic logs directory: {DiagLogsDirectory}");
         }
 
         IntegrationTestBase.ExecutePatchedDotnet("test", arguments, out _standardTestOutput, out _standardTestError, out _runnerExitCode, debugEnvironmentVariables, workingDirectory);
@@ -1078,12 +1082,17 @@ public class IntegrationTestBase
 
     private static bool IsDiagAlreadyEnabled(string arguments)
     {
-        // Check args for --diag, /diag, -diag (case-insensitive)
+        // If user passed /diag, --diag, or -diag, throw to prevent silently skipping the auto-injected
+        // diag and losing the attachment. Tests that need to inspect diagnostic logs should use
+        // DiagLogsDirectory instead of passing /diag manually.
         if (arguments.Contains("--diag", StringComparison.OrdinalIgnoreCase)
             || arguments.Contains("/diag", StringComparison.OrdinalIgnoreCase)
             || arguments.Contains("-diag", StringComparison.OrdinalIgnoreCase))
         {
-            return true;
+            throw new InvalidOperationException(
+                "Do not pass /diag (or --diag / -diag) directly in test arguments. " +
+                "The test framework auto-injects diagnostics and attaches them on failure. " +
+                "Use the DiagLogsDirectory property to find the path to the diagnostic log files.");
         }
 
         // Check environment variable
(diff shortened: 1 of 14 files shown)


test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
```diff
@@ -1078,12 +1082,17 @@ protected string BuildMultipleAssemblyPath(params string[] assetNames)
 
     private static bool IsDiagAlreadyEnabled(string arguments)
     {
-        // Check args for --diag, /diag, -diag (case-insensitive)
+        // If user passed /diag, --diag, or -diag, throw to prevent silently skipping the auto-injected
```
nohwnd: **[Error Reporting & Diagnostic Clarity]** The `-diag` substring check (`arguments.Contains("-diag", StringComparison.OrdinalIgnoreCase)`) was previously a benign false-positive (it silently skipped diag injection). Now it throws, which upgrades a harmless false-positive to a test failure.

An argument string like `--logger:"trx;LogFileName=my-diagnostics.trx"` would match `-diag` and throw unexpectedly. The check should be anchored to word boundaries or prefixes to avoid this:

```csharp
// More precise: require the flag to appear as a standalone argument token
if (Regex.IsMatch(arguments, @"(^|\s)(--diag|/diag|-diag)[:=\s]", RegexOptions.IgnoreCase))
```

Or at minimum, ensure the character before `diag` is a flag prefix character (`-`, `/`) rather than using a bare `Contains`. The current form is a pre-existing issue, but the severity escalation from "silent skip" to "throw" makes it worth fixing now.

nohwnd (commented):
## 🧠 *Reviewed by Expert Code Reviewer*

**Dimensions activated**: Acceptance Test Coverage Design, Error Reporting & Diagnostic Clarity

### Summary

The change is logically sound. Converting `IsDiagAlreadyEnabled` from a silent-skip to a hard throw is the right call — it makes misuse visible immediately rather than hiding it. The refactoring to `DiagLogsDirectory` and the consolidated `GetDiagLogContents()` base helper are clean improvements.

One concern is noted inline regarding the `-diag` substring check. The severity of a false-positive match escalated from "silently skip diagnostics" (harmless) to "throw `InvalidOperationException`" (test failure). Today's test argument strings are unlikely to trigger this, but the fix was the right moment to tighten the match predicate.

All other changes reviewed:
- `Assert.Contains(errorMessage, GetDiagLogContents())` parameter order ✓ (needle first, haystack second — correct)
- `DataCollectorTests.Coverlet.cs` file pattern migration to `"log.txt"` / `"log.txt.datacollector*"` / `"log.txt.host*"` is consistent with how `GetDiagArg` names the base log file ✓
- `DataCollectionTests.cs` change from `TempDirectory.Path` to `DiagLogsDirectory` for `*.datacollector.*` search is correct since diag injection now writes to `DiagLogsDirectory` ✓
- Hardcoded `c:\temp\logscpp\` path removed from `DotnetTestTests.cs` ✓

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26333602361)*

nohwnd:
Commit pushed: [`edd2cc4`](https://github.com/microsoft/vstest/commit/edd2cc4fd6a81a398f941b49cd860817f9a65bf9)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26333778456)*

test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs
nohwnd: Fixed. Replaced the bare `Contains` checks with a `Regex.IsMatch` that anchors the flag to a word boundary — the flag must be preceded by whitespace or start-of-string, and followed by `=`, `:`, whitespace, or end-of-string. This prevents false positives from substrings like `--logger:my-diagnostics.trx`.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26333778456)*

nohwnd (commented):
## 🧠 *Reviewed by Expert Code Reviewer*

**Dimensions activated**: Acceptance Test Coverage Design, Error Reporting & Diagnostic Clarity

The prior inline finding (bare `Contains("-diag")` false-positive upgraded to a throw) was resolved in the second commit. The regex `(^|\s)(--diag|/diag|-diag)([:=\s]|$)` is correctly anchored — it requires the flag to be a standalone argument token and avoids false positives from substrings like `--logger:my-diagnostics.trx`.

No remaining concerns. All changes reviewed:
- `IsDiagAlreadyEnabled` regex anchoring ✓
- `GetDiagLogContents()` with directory-exists guard ✓
- `DiagLogsDirectory` property consistency with injection logic ✓
- PR description accurately reflects all changes ✓

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26333949837)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26333949837)*

nohwnd:
Commit pushed: [`f9c595a`](https://github.com/microsoft/vstest/commit/f9c595a2d30a4c62f3b4ed71a435f15c4f256115)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26353072735)*

nohwnd:
---
> [!WARNING]
> The `push_to_pull_request_branch` operation failed: Cannot push to pull request branch: patch modifies protected files (.github/workflows/pr-expert-reviewer.lock.yml, .github/workflows/pr-expert-reviewer.md). Add them to the allowed-files configuration field or set protected-files: fallback-to-issue to create a review issue instead.. The code changes were not applied.

The Windows Release CI build failed twice (first attempt and the previous re-trigger). The code compiles cleanly on Linux in Release mode with 0 errors and 0 warnings.

Merged `main` (1 commit ahead: workflow-only change to `pr-expert-reviewer`) into the branch to re-trigger CI with the latest base.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26434403390)*



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26434403390)*

nohwnd:
Commit pushed: [`c8117dc`](https://github.com/microsoft/vstest/commit/c8117dcd6cdabd40cf5bd2da8a7fe7220c9b72c8)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26435022193)*

nohwnd:
The Release build passes cleanly on Linux (0 warnings, 0 errors). The Windows Release CI failure from 2026-05-24 appears to be an infrastructure flake — pushed an empty commit to re-trigger CI.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26435022193)*



> [!WARNING]
> <details>
> <summary>Firewall blocked 1 domain</summary>
>
> The following domain was blocked by the firewall during workflow execution:
>
> - `dev.azure.com`
>> To allow these domains, add them to the `network.allowed` list in your workflow frontmatter:
>
> ```yaml
> network:
>   allowed:
>     - defaults
>     - "dev.azure.com"
> ```
>
> See [Network Configuration](https://github.github.com/gh-aw/reference/network/) for more information.
>
> </details>


> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26435022193)*

nohwnd (commented):
## 🧠 *Reviewed by Expert Code Reviewer*

**New commits reviewed** (since last review at `edd2cc4`):

- **`ccaabcbe`** — `--diag` insertion before ` -- ` separator: logic is correct. `arguments?.IndexOf(" -- ") ?? -1` safely handles null, and inserting `diagArg` (which has a leading space) at `separatorPos` produces well-formed argument strings like `test.dll --diag:logs/log.txt -- runsettings`. ✓

- **`3c028e3c`** — Coverlet test revival: switching to `AcceptanceTestBase` + `[NetCoreRunner(HOST_NET)]` correctly activates the previously-dead test. Glob patterns `log.datacollector*` and `log.host*` are consistent with `Path.ChangeExtension("log.txt", "datacollector....")` naming. `FrameworkArgValue` replaces the hardcoded `.NETCoreApp,Version=v2.1`. ✓

No new concerns. All changes since last review look correct.

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26516499895)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26516499895)*

Evangelink:

This is a test-infrastructure-only change in `Microsoft.TestPlatform.TestUtilities`/integration tests. No production, wire/protocol, public-API, package, binding-redirect, or runtime path is touched, so most of the 16 vstest review dimensions are not applicable. The refactor itself is a net improvement (centralizes `GetDiagLogContents`, mirrors the ` -- ` separator handling between `InvokeVsTest` and `InvokeDotnetTest`, and converts the silent "skip auto-injection" behavior into a loud throw so we stop losing diag log attachments on failures). I have no blockers, but a few concerns below worth addressing before merge — mostly around the scope of the Coverlet test revival being hidden in this PR, the new `IsDiagAlreadyEnabled` semantics, and missing unit coverage for the new regex.

### 🛑 Blockers
_None._

### ⚠️ Issues

**1. `DataCollectorTestsCoverlets.RunCoverletCoverage` revival is bundled into a "diag refactor" PR** — `test/Microsoft.TestPlatform.Acceptance.IntegrationTests/DataCollectorTests.Coverlet.cs` (full file rewrite).

Before this PR the test was effectively dead: it early-returned unless `_testEnvironment.TargetFramework == CoreRunnerFramework` and then forced `.NETCoreApp,Version=v2.1` as the framework, with a `REVIEW ME: @Marco do we need to update this test?` comment. This PR drops the early return, switches the base class to `AcceptanceTestBase`, adds `[NetCoreRunner(HOST_NET)]`, and replaces `.NETCoreApp,Version=v2.1` with `FrameworkArgValue` (= net11.0). The test asset `CoverletCoverageTestProject` targets `$(NetCoreAppMinimum)` (net8.0) and pins `coverlet.collector` 1.2.0 — neither of which was validated for net11.0 runs by this PR's stated goal. This is a behavioral change orthogonal to the diag-flag refactor. Two concerns:

- The PR description doesn't mention reviving this test or what was verified about coverlet 1.2.0 + net11.0 host + net8.0 asset. The commit message `fix: revive dead coverlet test with correct diag patterns` is the only hint.
- If the test starts flaking, bisect will land on a "diag" PR, which obscures the real cause.

Split into its own PR, or at minimum call it out explicitly in the description and confirm the assertions actually pass in CI (the `Directory.GetFiles(DiagLogsDirectory, "log.txt").Single()` style assertions will throw `InvalidOperationException` rather than `AssertFailedException` if the file count is wrong, which is a worse failure UX).

**2. `IsDiagAlreadyEnabled` is misleadingly named after this change** — `test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs:1096-1114`.

The method now has two unrelated effects depending on input: it *throws* if `/diag` is in the args, and *returns bool* based on a `VSTEST_DIAG` env var. The bool return-value path no longer reflects "diag is already in arguments". Reading callers (`if (collectDiagnostics && !IsDiagAlreadyEnabled(arguments ?? ""))`) still parses as a guard against double-injection, but in practice the call is either "throw" or "no-op env check". Suggest splitting:

```csharp
ThrowIfDiagInArguments(arguments);
if (collectDiagnostics && !IsDiagInEnvironment()) { ... }
```

…or at minimum renaming to something like `ValidateAndCheckDiagState`. As-is, the next person who reads `IsDiagAlreadyEnabled` will not expect it to throw.

**3. Thrown `InvalidOperationException` does not include the offending argument** — `test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs:1103-1109`.

When this fires in CI the test author sees a generic message but not *which* arg or test method tripped it. The matched substring and (ideally) `TestContext?.FullyQualifiedTestClassName + TestName` would make the error self-locating. Cheap fix:

```csharp
var m = Regex.Match(arguments, @"(^|\s)(--diag|/diag|-diag)([:=\s]|$)", RegexOptions.IgnoreCase);
if (m.Success)
{
    throw new InvalidOperationException(
        $"Do not pass diag flags directly in test arguments (found '{m.Value.Trim()}' in '{arguments}'). " +
        "The test framework auto-injects diagnostics and attaches them on failure. " +
        "Use IntegrationTestBase.DiagLogsDirectory to find the diag log files.");
}
```

This also adds the type name `IntegrationTestBase.DiagLogsDirectory` to the message — easier to grep when you're chasing a CI failure.

**4. `GetDiagLogContents` recursion differs from the helpers it replaces** — `test/Microsoft.TestPlatform.TestUtilities/IntegrationTestBase.cs:1129`.

The deleted per-class copies in `CreateNoNewWindowTests.cs` and `SerializerSelectionTests.cs` used `Directory.GetFiles(logsDir, "*.txt")` (TopDirectoryOnly). The new base helper uses `SearchOption.AllDirectories`. The two existing callers (`ExecutionTests.StackOverflowExceptionShouldBeLoggedToConsoleAndDiagLogFile` and `UnhandleExceptionExceptionShouldBeLoggedToDiagLogFile`) check for a stderr fragment via `Assert.Contains` and so are tolerant, but if anyone later asserts on log-file *count* via this helper, recursion will silently include stray files (e.g. nested results from `dotnet test` runs that share `TempDirectory.Path`). Either keep it TopDirectoryOnly or document the recursion intent. Trivial but easy to get wrong later.

### 💡 Suggestions

**1. Add a unit test for the new regex.** This is the load-bearing change of the PR — the previous `string.Contains("--diag")` was easy to reason about; the new regex isn't. Cover at minimum:

- Positives: `"/diag"`, `"/diag:foo"`, `"/diag=foo"`, `"--diag"`, `"--diag bar.txt"`, `"-diag:x"`, leading-whitespace and mid-string variants, mixed case.
- Negatives: `"--logger:my-diag.trx"`, `"--diagnostic"`, `"/diagnostics"`, `"--no-diag-mode"`, `--logger:console;verbosity=normal`.

The regression risk if someone later edits this regex is high and not covered today.

**2. Promote the regex to a `static readonly Regex` (optionally `RegexOptions.Compiled`).** Re-parsing on every `InvokeVsTest`/`InvokeDotnetTest` call is harmless but the codebase prefers cached regexes elsewhere; would also pair well with the suggestion #3 above to reuse the `Match` result.

**3. `_attachments.Add(DiagLogsDirectory)` reads the property twice in `InvokeDotnetTest` (lines 286–319) once it's wired through.** Stash a local for readability and to make the dedup pattern with `InvokeVsTest` more obvious. Pure nit, skip if not in scope.

### ✅ Notes

- Mirroring the ` -- ` separator insertion logic from `InvokeDotnetTest` into `InvokeVsTest` (`IntegrationTestBase.cs:230-238`) is a good correctness fix in its own right — inline-runsettings args after `--` would previously have caused `--diag` to be parsed as a runsettings property. Worth calling out in the description; it's not just a cleanup.
- Removing the manual `/diag:c:\temp\logscpp\` literal in `DotnetTestTests.cs:96` is a real bug fix — that path would only exist on the author's machine.
- Other `--diag` callers in the repo (`PostProcessingTests.cs`, `DotnetArchitectureSwitchTests.Windows.cs`, `ArgumentProcessorTests.cs`) bypass `InvokeVsTest`/`InvokeDotnetTest` and call `ExecuteVsTestConsole`/`ExecuteApplication` directly, so they're unaffected by the throw. Confirmed.
- No public API, no wire protocol, no binding redirects, no package shape, no runtime path changed — dimensions 1–11 and 13–16 of the vstest review checklist are inapplicable here.

### Verdict
**COMMENT** — no blockers, but please address #1 (scope of the Coverlet revival) and #3 (exception context), and ideally land #2 (rename) and the Suggestion #1 (regex unit test) before merging.

nohwnd:
Commit pushed: [`8966ad3`](https://github.com/microsoft/vstest/commit/8966ad39699bf072496740a4820926a01796b920)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26744709940)*

nohwnd:
Addressed the review feedback:

**Issues fixed:**

1. **Coverlet revival scope** (Issue #1) — called out in the PR description (see updated body) that `DataCollectorTestsCoverlets.RunCoverletCoverage` was revived as a deliberate part of this PR: the test previously dead-returned and bypassed diag injection; now that auto-injection always fires the test can use `DiagLogsDirectory` patterns correctly.

2. **`IsDiagAlreadyEnabled` rename** (Issue #2) — split into `ThrowIfDiagInArguments(string)` (void, throws) and `IsDiagInEnvironment()` (bool), making the two distinct behaviors explicit. Call sites updated accordingly.

3. **Exception context** (Issue #3) — switched from `Regex.IsMatch` to `Regex.Match` and included the matched token and full argument string in the `InvalidOperationException` message (e.g. `found '--diag' in '...'`).

4. **`GetDiagLogContents` recursion** (Issue #4) — changed `SearchOption.AllDirectories` → `TopDirectoryOnly` to match the per-class helpers it replaced.

**Suggestions addressed:**

- **Static readonly Regex** (Suggestion #2) — promoted to `internal static readonly Regex DiagFlagPattern` with `RegexOptions.Compiled`.
- **Unit tests for the regex** (Suggestion #1) — added `DiagFlagPatternTests.cs` in the acceptance test project with positive and negative cases. Added `Friends.cs` with `InternalsVisibleTo` so the internal field is accessible.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26744709940)*



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26744709940)*

nohwnd (commented):
## 🧠 *Reviewed by Expert Code Reviewer*

**New commit reviewed** (since last review at `3c028e3c`):

- **`8966ad39`** — Review feedback addressing commit. All changes look correct:

  - **`DiagFlagPattern` as compiled static `Regex`**: avoids re-allocating the pattern on every test invocation. ✓
  - **`ThrowIfDiagInArguments` / `IsDiagInEnvironment` split**: names now clearly distinguish the two-phase "guard vs. env-var skip" logic. The `arguments ?? ""` guard at the `InvokeVsTest` call site handles null safely. ✓
  - **Error message includes `match.Value.Trim()` and `arguments`**: gives actionable CI diagnostics. ✓
  - **`GetDiagLogContents` → `TopDirectoryOnly`**: matches the behavior of the per-class helpers it replaced; `CountTestHostLogs` deliberately keeps `AllDirectories` since host log files can be nested. ✓
  - **`Friends.cs` + `DiagFlagPatternTests`**: `InternalsVisibleTo` correctly targets `Microsoft.TestPlatform.Acceptance.IntegrationTests` with the public key; test coverage spans all positive/negative regex cases including the subtle `--diagnostic` (suffix after token) and `--no-diag-mode` (prefix before token) false-positive guards. ✓

No remaining concerns.

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26745320807)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26745320807)*

Evangelink (approved):

PR #16062, opened 2026-05-25
Fix false "Empty parenthesis" error when filter contains an escaped open parenthesis

### LINKED ISSUE #7515: Filtering parametrized tests based on just test name
## Describe the bug

I have tests like 
```
aaa (params)
aaa2 (params)
```

I would like to list just aaa tests but not aaa2 tests. They have parameters so the `Name=` filter does not work, and trying to provide `aaa (` fails with error that there is empty parentheses group.

Backslash is supposed to escape it, but that does not work either.

## Version used

mtest 4.1.0 + mtp

## Steps To Reproduce

```
dotnet test test\Microsoft.TestPlatform.Acceptance.IntegrationTests\Microsoft.TestPlatform.Acceptance.IntegrationTests.csproj --list-tests --filter "Name=RunMultipleMSTestAssembliesOnVstestConsoleAndTesthostCombinations \("
Discovering tests from artifacts\bin\Microsoft.TestPlatform.Acceptance.IntegrationTests\Debug\net9.0\Microsoft.TestPlatform.Acceptance.IntegrationTests.dll (net9.0|x64)
Discovering tests from artifacts\bin\Microsoft.TestPlatform.Acceptance.IntegrationTests\Debug\net48\Microsoft.TestPlatform.Acceptance.IntegrationTests.exe (net48|x64)
Exit code: 8
  Standard output: Incorrect format for TestCaseFilter Error: Empty parenthesis ( ). Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.

Exit code: 8
  Standard output: Incorrect format for TestCaseFilter Error: Empty parenthesis ( ). Specify the correct format and try again. Note that the incorrect format can lead to no test getting executed.


Discovered 0 tests in assembly - artifacts\bin\Microsoft.TestPlatform.Acceptance.IntegrationTests\Debug\net48\Microsoft.TestPlatform.Acceptance.IntegrationTests.exe (net48|x64)

Discovered 0 tests in assembly - artifacts\bin\Microsoft.TestPlatform.Acceptance.IntegrationTests\Debug\net9.0\Microsoft.TestPlatform.Acceptance.IntegrationTests.dll (net9.0|x64)

Discovered 0 tests in 2 assemblies.

Test discovery completed with non-success exit code: 8 (see: https://aka.ms/testingplatform/exitcodes)
```

Evangelink:
Fixes [microsoft/testfx#7515](https://github.com/microsoft/testfx/issues/7515).

## Problem

Filtering parametrized tests by display-name prefix is broken when the
expected prefix contains a `(`. For instance, given:

```
aaa (1, 2)
aaa2 (1, 2)
```

users want to match only the `aaa` instances by typing:

```
dotnet test --filter "Name~aaa \("
```

but `dotnet test` fails with:

```
Incorrect format for TestCaseFilter Error: Empty parenthesis ( ).
```

## Root cause

`FilterExpression.Parse` does a pre-check using the regex `\(\s*\)` to
flag empty parenthesis groups (because the parser itself silently accepts
them). The regex did not understand the `\` escape character, so the
escaped `\(` followed by `)` (which appears naturally when the
Microsoft.Testing.Platform VSTestBridge wraps every filter in outer
parens, turning `Name~aaa \(` into `(Name~aaa \()`) was incorrectly
flagged.

## Fix

Replace the regex with `(?<!\\)\(\s*\)` so an escaped open parenthesis is
no longer considered the start of an empty parenthesis group. Add
regression tests in `FilterExpressionWrapperTests` covering:

- escaped `\(` not flagged as empty parens,
- wrapped `(...\()` not flagged as empty parens,
- genuine `()` still flagged as empty parens,
- end-to-end `Evaluate` behaviour for the parametrized-name filtering
  scenario (`aaa (1, 2)` matches `Name~aaa \(`, `aaa2 (1, 2)` does
  not).

All existing tests in `Microsoft.TestPlatform.Filter.Source.UnitTests`
(44 tests) and `Microsoft.TestPlatform.Common.UnitTests` (388 tests)
continue to pass.

Once this ships, the `Microsoft.TestPlatform.Filter.Source` package
version can be bumped in microsoft/testfx and the fix will flow into the
VSTestBridge.

### DIFF
--- a/src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
+++ b/src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
@@ -188,7 +188,10 @@ internal sealed class FilterExpression
 #endif
 
         // Below parsing doesn't error out on pattern (), so explicitly search for that (empty parenthesis).
-        var invalidInput = Regex.Match(filterString, @"\(\s*\)");
+        // The negative lookbehind (?<!\\) ensures we don't flag an escaped open parenthesis (e.g., \()
+        // as an empty group. This is required to support filtering parametrized tests whose display
+        // names contain '(' (e.g., `DisplayName~MyTest \(`).
+        var invalidInput = Regex.Match(filterString, @"(?<!\\)\(\s*\)");
         if (invalidInput.Success)
         {
             throw new FormatException(string.Format(CultureInfo.CurrentCulture, TestCaseFilterFormatException, EmptyParenthesis));
--- a/src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
+++ b/src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
@@ -9,7 +9,6 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Text;
-using System.Text.RegularExpressions;
 
 #if !IS_VSTEST_REPO
 using Microsoft.CodeAnalysis;
@@ -187,17 +186,46 @@ internal sealed class FilterExpression
         ValidateArg.NotNull(filterString, nameof(filterString));
 #endif
 
-        // Below parsing doesn't error out on pattern (), so explicitly search for that (empty parenthesis).
-        // The negative lookbehind (?<!\\) ensures we don't flag an escaped open parenthesis (e.g., \()
-        // as an empty group. This is required to support filtering parametrized tests whose display
-        // names contain '(' (e.g., `DisplayName~MyTest \(`).
-        var invalidInput = Regex.Match(filterString, @"(?<!\\)\(\s*\)");
-        if (invalidInput.Success)
+        // Tokenize the filter once and reuse the materialized list for both the pre-check
+        // below and the parsing loop further down. The tokenizer already understands the
+        // escape semantics of the filter language (e.g., that "\(" is part of a value token
+        // and not an open-parenthesis grouping operator, and that "\\" is an escaped
+        // backslash so the following "(" is unescaped), so inspecting tokens rather than the
+        // raw string avoids the corner cases of a regex-based pre-check.
+        List<string> tokens = TokenizeFilterExpressionString(filterString).ToList();
+
+        // Below parsing doesn't error out on pattern (), so explicitly search for that
+        // (empty parenthesis). Token-based scan: an unescaped "(" immediately followed by
+        // an unescaped ")" (ignoring whitespace-only tokens, mirroring what the parser
+        // below does with token.Trim()) is an empty group.
+        for (int i = 0; i < tokens.Count; i++)
         {
-            throw new FormatException(string.Format(CultureInfo.CurrentCulture, TestCaseFilterFormatException, EmptyParenthesis));
+            if (tokens[i] != "(")
+            {
+                continue;
+            }
+
+            int j = i + 1;
+            while (j < tokens.Count)
+            {
+                var trimmed = tokens[j].Trim();
+#if IS_VSTEST_REPO
+                if (!trimmed.IsNullOrEmpty())
+#else
+                if (!string.IsNullOrEmpty(trimmed))
+#endif
+                {
+                    break;
+                }
+                j++;
+            }
+
+            if (j < tokens.Count && tokens[j] == ")")
+            {
+                throw new FormatException(string.Format(CultureInfo.CurrentCulture, TestCaseFilterFormatException, EmptyParenthesis));
+            }
         }
 
-        var tokens = TokenizeFilterExpressionString(filterString);
         var operatorStack = new Stack<Operator>();
         var filterStack = new Stack<FilterExpression>();
 
--- a/test/Microsoft.TestPlatform.Filter.Source.UnitTests/FilterExpressionWrapperTests.cs
+++ b/test/Microsoft.TestPlatform.Filter.Source.UnitTests/FilterExpressionWrapperTests.cs
@@ -43,6 +43,16 @@ public class FilterExpressionWrapperTests
     }
 
     [TestMethod]
+    public void ParseErrorShouldBeSetForEmptyParenthesisAfterEscapedBackslash()
+    {
+        // An escaped backslash "\\" leaves the following "(" unescaped, so "\\()" is still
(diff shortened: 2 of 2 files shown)


src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
```diff
@@ -188,7 +188,10 @@ internal static FilterExpression Parse(string filterString, out FastFilter? fast
 #endif
 
         // Below parsing doesn't error out on pattern (), so explicitly search for that (empty parenthesis).
-        var invalidInput = Regex.Match(filterString, @"\(\s*\)");
+        // The negative lookbehind (?<!\\) ensures we don't flag an escaped open parenthesis (e.g., \()
```
nohwnd: **[Algorithmic Correctness] Single-char lookbehind doesn't handle escaped backslashes**

The negative lookbehind `(?<!\\)` checks only whether `(` is preceded by a single `\`, but the filter language itself supports `\\` as an escaped backslash (see `TokenizeFilterExpressionStringHelper` — when it encounters `\\`, it resets `last` to `'\0'` so the following char is unescaped).

**Failure case:** Input `\\()` — an escaped backslash (`\\`) followed by a truly-empty group `()`.

- The char immediately before `(` is `\` (the second backslash of `\\`).
- The lookbehind `(?<!\\)` fires → regex does **not** match → no error is thrown.
- The tokenizer correctly handles `\\` and produces tokens `["\\", "(", ")"]`. The parser silently accepts this because the `()` group contains no operators and `filterStack.Count == 1` at the end. The empty group is quietly dropped.
- Result: the pre-check that was added specifically to catch silently-accepted empty groups now has a blind spot for `\\()`.

**Suggested fix** — inspect the tokenized output instead of the raw string, which already understands escape semantics:

```csharp
var tokens = TokenizeFilterExpressionString(filterString).ToList();
for (int i = 0; i < tokens.Count - 1; i++)
{
    if (tokens[i] == "(" && tokens[i + 1] == ")")
        throw new FormatException(...);
}
```

Or use a .NET variable-length lookbehind that counts pairs of backslashes:
```csharp
// Match ( not preceded by an odd number of backslashes
@"(?<!(?:^|[^\\])(?:\\\\)*)\\(?:\\\\)*\(\s*\)"
```
The token-based check is simpler and reuses the existing escape logic already proven correct by the tokenizer tests.

nohwnd (commented):
## Review: Fix false "Empty parenthesis" error for escaped open parenthesis

The fix correctly addresses the reported regression — the negative lookbehind prevents `\(` (escaped open-paren) from being mis-identified as the start of an empty group. The test coverage for the bug scenarios is solid and the `Assert.Contains` usage is correct (needle first).

**One correctness concern** (inline): the single-char lookbehind `(?<!\\)` has a blind spot for the `\\(` case (escaped backslash followed by an unescaped `(`). For the input `\\()`, the char immediately before `(` is `\` (the second of `\\`), so the lookbehind incorrectly suppresses the match, letting a truly-empty group pass undetected. The tokenizer already handles `\\` correctly, so checking the tokenized token stream for adjacent `"("` + `")"` tokens would be the most robust fix.

In practice `\\()` is an unusual filter string, so this is not a blocker, but it leaves a known gap in the pre-check that the original code was specifically designed to close.

🧠 *Reviewed by expert-reviewer (PR #16062)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26399848994)*

src/Microsoft.TestPlatform.Filter.Source/FilterExpression.cs
Evangelink: Good catch — you''re right that `\\()` slips past the single-char lookbehind. Switched the pre-check to a scan over the tokenizer output (commit e716debd3), since the tokenizer already gets the escape semantics right. So now:

- `Name~aaa \(` → single value token, no `(` token → no error ✅
- `Name~foo\\&()` → tokens `["Name~foo\\", "&", "(", ")"]` → flagged ✅
- `(  )` / `Name=Test1 & ()` → still flagged ✅

Added a regression test (`ParseErrorShouldBeSetForEmptyParenthesisAfterEscapedBackslash`) covering the `\\()` case. Also dropped the now-unused `System.Text.RegularExpressions` using.

nohwnd (dismissed):
## Review: Fix false "Empty parenthesis" error for escaped open parenthesis (updated commit)

The prior review flagged that the single-char negative lookbehind `(?<!\\)` had a blind spot for `\\()` (escaped backslash + unescaped `(`). This commit directly addresses that concern by replacing the regex entirely with a **token-based scan** over the existing `TokenizeFilterExpressionString` output.

### Why the token-based approach is correct

The tokenizer already handles all escape semantics:
- `\(` → a single value token (not an operator `(` token)
- `\\` → value token `\\` with `last` reset to `'\0'`, so the following `(` is unescaped

The new loop finds `"("` tokens (which can only be unescaped open-parens), skips whitespace-only inter-tokens (matching the original `\(\s*\)` intent), and throws on a bare `")"` next — correctly covering `()`, `( )`, `\\\\ ()`, etc.

### Test coverage

All four scenarios are now exercised:
- `Name=Test1 & ()` → error ✅
- `Name~foo\\&()` (escaped backslash then empty group) → error ✅  
- `Name~aaa \(` (escaped open-paren, original regression) → no error ✅
- `(Name~aaa \()` (wrapped filter as VSTestBridge produces) → no error ✅

### Minor observations (non-blocking)

- Materializing `TokenizeFilterExpressionString` into `List<string>` upfront is the right call — avoids double-enumeration.
- The `#if IS_VSTEST_REPO` guard for `IsNullOrEmpty` is consistent with the existing pattern in this file.
- `Assert.Contains` usage in the new tests is correct (needle first).

The fix is well-targeted, the prior algorithmic concern is fully resolved, and the regression tests are solid.

🧠 *Reviewed by expert-reviewer (PR #16062)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26400689515)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26400689515)*

nohwnd (approved):

PR #16064, opened 2026-05-25
[fix] Fix race condition: skip hang dump when testhost hasn't launched yet

nohwnd:
🤖 This is an automated fix by the Issue Triage agent.

Fixes #15588

## Root Cause

When the inactivity hang-dump timer fires before the testhost process has launched, `_testHostProcessId` is `0` (the default `int` value). The code in `CollectDumpAndAbortTesthost` would then:

1. Call `StartHangBasedProcessDump(0, ...)` — attempting to dump **PID 0**, which is the **Idle process** on Windows and **Swapper** on Linux.
2. Call `Process.GetProcessById(0)` — getting the Idle/Swapper process and trying to kill it.

This produces an empty or corrupt dump file and may cause errors. The comment in `DefaultEngineInvoker.cs` even had a TODO noting this hazard:
```
(todo/redacted): should there be a warning / error in this case, on windows and linux we are most likely not started by this PID 0
```

## Fix

Added an early-return guard at the start of `CollectDumpAndAbortTesthost`:

```csharp
if (_testHostProcessId == 0)
{
    EqtTrace.Warning("BlameCollector: Test host process has not launched yet. Skipping hang dump.");
    _logger.LogWarning(..., Resources.TestHostNotLaunchedCannotCollectHangDump);
    return;
}
```

If the testhost hasn't launched (PID still 0), we log a warning and skip dump/kill entirely.

## Tests

- **New test** `InitializeWithDumpForHangShouldSkipDumpIfTestHostHasNotLaunchedYet`: verifies that `StartHangBasedProcessDump` is **not** called when the timer fires before `TestHostLaunched`.
- **Updated 3 existing tests**: changed from `timeout=0` to `timeout=50ms` and added `TestHostLaunched` event raising before the timer fires, to properly test the happy path (testhost launched → hang timer fires → dump collected).

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26403426923)*

### DIFF
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -210,7 +210,7 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
     /// Starts and waits for a new proc dump process to collect a single dump and then
     /// kills the testhost process.
     /// </summary>
-    private void CollectDumpAndAbortTesthost()
+    private protected void CollectDumpAndAbortTesthost()
     {
         TPDebug.Assert(_logger != null && _context != null && _dataCollectionSink != null, "Initialize must be called before calling this method");
         _inactivityTimerAlreadyFired = true;
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -213,6 +213,16 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
     protected void CollectDumpAndAbortTesthost()
     {
         TPDebug.Assert(_logger != null && _context != null && _dataCollectionSink != null, "Initialize must be called before calling this method");
+
+        // If testhost has not launched yet, we cannot dump or kill it. Do not mark the timer as
+        // fired so that ResetInactivityTimer can re-arm it once the testhost eventually launches.
+        if (_testHostProcessId == 0)
+        {
+            EqtTrace.Warning("BlameCollector.CollectDumpAndAbortTesthost: Test host process has not launched yet. Skipping hang dump.");
+            _logger.LogWarning(_context.SessionDataCollectionContext, Resources.Resources.TestHostNotLaunchedCannotCollectHangDump);
+            return;
+        }
+
         _inactivityTimerAlreadyFired = true;
 
         string value;
@@ -242,14 +252,6 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
         catch
         {
             EqtTrace.Verbose("Inactivity timer is already disposed.");
-        }
-
-        // If testhost has not launched yet, we cannot dump or kill it.
-        if (_testHostProcessId == 0)
-        {
-            EqtTrace.Warning("BlameCollector.CollectDumpAndAbortTesthost: Test host process has not launched yet. Skipping hang dump.");
-            _logger.LogWarning(_context.SessionDataCollectionContext, Resources.Resources.TestHostNotLaunchedCannotCollectHangDump);
-            return;
         }
 
         if (_collectProcessDumpOnCrash)
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -210,7 +210,7 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
     /// Starts and waits for a new proc dump process to collect a single dump and then
     /// kills the testhost process.
     /// </summary>
-    private protected void CollectDumpAndAbortTesthost()
+    protected void CollectDumpAndAbortTesthost()
     {
         TPDebug.Assert(_logger != null && _context != null && _dataCollectionSink != null, "Initialize must be called before calling this method");
         _inactivityTimerAlreadyFired = true;
--- a/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
+++ b/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
@@ -244,6 +244,14 @@ public class BlameCollector : DataCollector, ITestExecutionEnvironmentSpecifier
             EqtTrace.Verbose("Inactivity timer is already disposed.");
         }
 
+        // If testhost has not launched yet, we cannot dump or kill it.
+        if (_testHostProcessId == 0)
+        {
+            EqtTrace.Warning("BlameCollector.CollectDumpAndAbortTesthost: Test host process has not launched yet. Skipping hang dump.");
+            _logger.LogWarning(_context.SessionDataCollectionContext, Resources.Resources.TestHostNotLaunchedCannotCollectHangDump);
+            return;
+        }
+
         if (_collectProcessDumpOnCrash)
         {
             // Detach the dumper from the testhost process to prevent crashing testhost process. When the dumper is procdump.exe
--- a/test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/BlameCollectorTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/BlameCollectorTests.cs
@@ -168,18 +168,16 @@ public class BlameCollectorTests
         _blameDataCollector = new TestableBlameCollector(
             _mockBlameReaderWriter.Object,
             _mockProcessDumpUtility.Object,
-            null,
+            _mockInactivityTimer.Object,
             _mockFileHelper.Object,
             _mockProcessHelper.Object);
 
         var dumpFile = "abc_hang.dmp";
-        var hangBasedDumpcollected = new ManualResetEventSlim();
 
         _mockFileHelper.Setup(x => x.Exists(It.Is<string>(y => y == "abc_hang.dmp"))).Returns(true);
         _mockFileHelper.Setup(x => x.GetFullPath(It.Is<string>(y => y == "abc_hang.dmp"))).Returns("abc_hang.dmp");
         _mockProcessDumpUtility.Setup(x => x.StartHangBasedProcessDump(It.IsAny<int>(), It.IsAny<string>(), It.IsAny<bool>(), It.IsAny<string>(), It.IsAny<Action<string>>()));
         _mockProcessDumpUtility.Setup(x => x.GetDumpFiles(true, It.IsAny<bool>())).Returns(new[] { dumpFile });
-        _mockDataCollectionSink.Setup(x => x.SendFileAsync(It.IsAny<FileTransferInformation>())).Callback(() => hangBasedDumpcollected.Set());
 
         _blameDataCollector.Initialize(
             GetDumpConfigurationElement(false, false, true, 50),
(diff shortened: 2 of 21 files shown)


nohwnd (commented):
## Review Summary

The core fix is correct: adding a `_testHostProcessId == 0` guard in `CollectDumpAndAbortTesthost` prevents a real bug where the hang-dump timer fires before the testhost process launches, causing PID 0 (the OS Idle/Swapper process) to be targeted for a dump and kill. The resource strings and all XLF files are properly updated.

**One finding worth addressing:**

The three updated "happy path" tests (`InitializeWithDumpForHangShouldCaptureADumpOnTimeout` and siblings) now carry a timing-dependent assumption: that the test thread raises `TestHostLaunched` within 50 ms of calling `Initialize()`. This creates a small but non-zero probability of flakiness on loaded CI agents. See the inline comment for details and a suggested mitigation.

**No other issues found.** The new `InitializeWithDumpForHangShouldSkipDumpIfTestHostHasNotLaunchedYet` test correctly validates the fix using a warning-callback sentinel, and the resource/XLF additions are complete and follow the repo conventions.

🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/26404630653)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26404630653)*

test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/BlameCollectorTests.cs
```diff
@@ -182,13 +182,16 @@ public void InitializeWithDumpForHangShouldCaptureADumpOnTimeout()
...
             _mockDataCollectionSink.Object,
             _mockLogger.Object,
             _context);
 
-        hangBasedDumpcollected.Wait(1000, TestContext.CancellationToken);
+        // Simulate testhost launching before the timer fires.
+        _mockDataColectionEvents.Raise(x => x.TestHostLaunched += null, new TestHostLaunchedEventArgs(_dataCollectionContext, 1234));
+
+        hangBasedDumpcollected.Wait(2000, TestContext.CancellationToken);
         _mockProcessDumpUtility.Verify(x => x.StartHangBasedProcessDump(It.IsAny<int>(), It.IsAny<string>(), It.IsAny<bool>(), It.IsAny<string>(), It.IsAny<Action<string>>()), Times.Once);
```
nohwnd, line 195: **[Threading] Potential test flakiness — timer may fire before `TestHostLaunched` is raised**

The comment says "Simulate testhost launching **before** the timer fires", but there is no synchronization guarantee here. `Initialize()` starts a 50 ms timer on the threadpool, then control immediately returns to the test thread. If the test thread is preempted for >50 ms between lines 185 and 192 (e.g., on a loaded CI agent), the timer fires first, sees `_testHostProcessId == 0`, logs the warning and returns early — and `StartHangBasedProcessDump` is never called, causing the `Times.Once` assertion to fail.

The same race applies to the other two tests updated with the same pattern (`InitializeWithDumpForHangShouldCaptureKillTestHostOnTimeoutEvenIfGetDumpFileFails` and `InitializeWithDumpForHangShouldCaptureKillTestHostOnTimeoutEvenIfAttachingToDumpFails`).

A more robust pattern is to block the timer callback until after `TestHostLaunched` has been raised:

```csharp
// Raise TestHostLaunched BEFORE starting the timer (or gate the timer via a semaphore)
// Option A: raise the event synchronously before Initialize sets up the timer
_mockDataColectionEvents.Raise(x => x.TestHostLaunched += null,
    new TestHostLaunchedEventArgs(_dataCollectionContext, 1234));

_blameDataCollector.Initialize(
    GetDumpConfigurationElement(false, false, true, 50), ...);
```

Or, if the event handler must be wired via `Initialize`, increase the timeout to a value that gives the test thread ample time even on loaded CI (e.g. 500 ms), acknowledging this is still probabilistic.

nohwnd:
Commit pushed: [`f4a7320`](https://github.com/microsoft/vstest/commit/f4a7320df5f86392e86ac499ec5a1ca2c0bdaf95)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26405013610)*

test/Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests/BlameCollectorTests.cs
nohwnd: Good catch. Fixed in the latest commit by eliminating the real timer entirely from those 3 happy-path tests:

- Made `CollectDumpAndAbortTesthost` `private protected`
- Added `TriggerTimerCallback()` to `TestableBlameCollector` that invokes the callback directly
- The 3 tests now use `Mock<IInactivityTimer>` (already used in other tests), raise `TestHostLaunched`, then call `TriggerTimerCallback()` synchronously — no race possible
- The "skip dump when PID=0" test keeps the real timer with `timeout=0` since it specifically exercises the guard path that fires before `TestHostLaunched`

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26405013610)*

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -210,7 +210,7 @@ public override void Initialize(
     /// Starts and waits for a new proc dump process to collect a single dump and then
     /// kills the testhost process.
     /// </summary>
-    private void CollectDumpAndAbortTesthost()
+    private protected void CollectDumpAndAbortTesthost()
```
nohwnd: **[Crash & Hang Dump Reliability] `private protected` prevents cross-assembly access — compile error**

`private protected` restricts access to derived types **within the same assembly** only. `TestableBlameCollector` lives in `Microsoft.TestPlatform.Extensions.BlameDataCollector.UnitTests` (a different assembly), so `TriggerTimerCallback() => CollectDumpAndAbortTesthost()` will produce CS0122 at compile time.

Note: `InternalsVisibleTo` in `Friends.cs` only applies to `internal` members — it does not widen `private protected`.

The correct modifier is `protected`, which grants access to derived classes in any assembly:

```csharp
protected void CollectDumpAndAbortTesthost()
```

This is the minimum change needed for the test project to compile.

nohwnd (commented):
## Review Summary

The core fix (guard against `_testHostProcessId == 0` in `CollectDumpAndAbortTesthost`) is correct and addresses the real bug described in #15588. Resource strings and all XLF files are properly updated.

The previous flakiness concern raised in the prior review was addressed by introducing `TriggerTimerCallback()` in `TestableBlameCollector` — a sound approach. However, the access modifier used to expose the method is wrong and will cause a **compile error**.

**One blocking finding:**

`CollectDumpAndAbortTesthost` is declared `private protected`, which restricts access to derived classes **in the same assembly only**. `TestableBlameCollector` is in the test assembly (`...UnitTests`), a separate assembly from `BlameCollector`. This makes the `TriggerTimerCallback() => CollectDumpAndAbortTesthost()` call a CS0122 compile error. The fix is to use `protected` instead.

No other issues found.

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26405581194)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26405581194)*

nohwnd:
Commit pushed: [`106ef82`](https://github.com/microsoft/vstest/commit/106ef82e1ad3d5ad58bc54ddc341a2bc79a6f34c)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26405986753)*

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
```diff
@@ -244,6 +244,14 @@ private void CollectDumpAndAbortTesthost()
             EqtTrace.Verbose("Inactivity timer is already disposed.");
         }
 
+        // If testhost has not launched yet, we cannot dump or kill it.
+        if (_testHostProcessId == 0)
+        {
+            EqtTrace.Warning("BlameCollector.CollectDumpAndAbortTesthost: Test host process has not launched yet. Skipping hang dump.");
+            _logger.LogWarning(_context.SessionDataCollectionContext, Resources.Resources.TestHostNotLaunchedCannotCollectHangDump);
+            return;
```
nohwnd: **[Threading/Correctness]** Early return permanently disables hang detection for the remainder of the session.

`_inactivityTimerAlreadyFired = true` is set on line 216, before this guard. `ResetInactivityTimer()` (line 675) skips resetting the timer when `_inactivityTimerAlreadyFired` is `true`. So once the timer fires before the testhost launches:

1. `_inactivityTimerAlreadyFired` is permanently `true`  
2. The early return fires here
3. From this point on, `ResetInactivityTimer()` is a no-op, so no subsequent test activity resets the timer
4. If the testhost then launches and hangs legitimately, **no hang dump is ever collected**

The fix is to not mark the timer as already fired when we're doing the early return — move `_inactivityTimerAlreadyFired = true` to after the guard, or reset it before the early return:

```csharp
// Option A: move the flag assignment after the guard
if (_testHostProcessId == 0)
{
    EqtTrace.Warning("...");
    _logger.LogWarning(...);
    return;
}

_inactivityTimerAlreadyFired = true;
```

```csharp
// Option B: reset the flag before early return
if (_testHostProcessId == 0)
{
    _inactivityTimerAlreadyFired = false; // allow future timer fires
    EqtTrace.Warning("...");
    _logger.LogWarning(...);
    return;
}
```

Note that `_inactivityTimer?.Dispose()` (line 240) has already been called before the guard, so Option A would also need the timer disposal moved after the guard, or the timer would need to be re-armed. Option B is simpler but needs to be paired with re-arming the timer (otherwise the timer won't fire again since it was disposed on line 240).

nohwnd (commented):
## Review Summary

The core fix (guard `_testHostProcessId == 0` in `CollectDumpAndAbortTesthost`) is correct and addresses the real bug from #15588. The previous `private protected` compile error is also resolved by using `protected` with the corresponding `PublicAPI.Unshipped.txt` entries. Resource strings and XLF files are properly updated.

**One blocking correctness issue found:**

`_inactivityTimerAlreadyFired = true` is set on line 216, *before* the new `_testHostProcessId == 0` guard. `ResetInactivityTimer()` is gated on `!_inactivityTimerAlreadyFired`, so once the early return path executes, hang detection is permanently disabled for the session — even after testhost eventually launches. A legitimate hang of the (now-running) testhost will produce no dump.

The fix also disposes `_inactivityTimer` before the guard (line 240), so the timer cannot fire again regardless. Both issues need to be addressed together — see the inline comment for concrete options.

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26406472665)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26406472665)*

nohwnd:
Commit pushed: [`9522bea`](https://github.com/microsoft/vstest/commit/9522bea385427774bdfa99c369f87929cbb3d3e4)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26406822413)*

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: Good catch. Fixed in the latest commit — moved both `_inactivityTimerAlreadyFired = true` and `_inactivityTimer?.Dispose()` to **after** the PID guard. When the timer fires before the testhost launches, we now return early without setting the flag or disposing the timer, so `ResetInactivityTimer` can re-arm it once the testhost eventually launches and hang detection stays active.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26406822413)*

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26406822413)*

nohwnd (commented):
## Review Summary

All three issues raised in prior reviews are now resolved in the latest commit:

1. **`private protected` → `protected`** ✅ — `CollectDumpAndAbortTesthost` is now `protected`, allowing `TestableBlameCollector` in the test assembly to override it.
2. **`_inactivityTimerAlreadyFired = true` ordering** ✅ — The flag is now set *after* the `_testHostProcessId == 0` guard, so an early return preserves the ability for `ResetInactivityTimer` to re-arm the timer when the testhost eventually launches. The comment documents this intent clearly.
3. **Test flakiness** ✅ — Happy-path tests now use synchronous `TriggerTimerCallback()` with a properly sequenced `TestHostLaunched` raise. The "skip" test retains a real 0ms timer but uses a `ManualResetEventSlim` with a 1-second ceiling, which is sufficient for CI.

The fix is correct, well-tested, and the resource/XLF additions are complete.

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26407423423)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26407423423)*

src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs
nohwnd: Good catch — fixed. Changed `private protected` to `protected` so `TestableBlameCollector` in the test assembly can access `CollectDumpAndAbortTesthost` via inheritance.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26407776326)*

nohwnd:
does not look like good fix.

PR #16077, opened 2026-05-30
[fix] Warn when testhost is not found in project deps.json (missing Microsoft.NET.Test.Sdk)

nohwnd:
> 🤖 This is an automated fix created by the Issue Triage agent.

Fixes #15403

## Root Cause

When a .NET test project has a proper `deps.json` and `runtimeconfig.dev.json` (indicating a managed .NET project), but doesn't reference `microsoft.testplatform.testhost` in its deps.json, `DotnetTestHostManager` silently falls back to using the testhost from the vstest.console installation directory. This fallback can succeed at process launch but then fail at runtime with a confusing error like:

```
An assembly specified in the application dependencies manifest (testhost.deps.json) was not found:
  package: 'testhost', version: '18.0.1-release-...'
  path: 'testhost.dll'
```

The root cause is typically a missing reference to `Microsoft.NET.Test.Sdk` in the test project, but the error message gives no indication of this.

## Fix

Added a user-visible `Warning` message emitted by `DotnetTestHostManager.GetTestHostStartInfo` when:
1. The test project has a `deps.json` file (managed project), AND
2. The test project has a `runtimeconfig.dev.json` file (proper SDK-style .NET project), AND
3. `testhost.dll` is **not** found in the project's deps.json (i.e., `Microsoft.NET.Test.Sdk` is likely missing)

The warning reads:
> Test source '{0}' does not include the test host assembly 'testhost.dll'. This can happen when the test project does not reference the 'Microsoft.NET.Test.Sdk' NuGet package. Falling back to the test host from the test runner installation, which may fail if the versions are incompatible.

This warning is shown before the fallback occurs, giving the user actionable information to diagnose the real cause of any subsequent failure.

## Changes

- `DotnetTestHostManager`: Added `_messageLogger` field to store the logger passed to `Initialize()`, and emit a warning when falling back to runner-side testhost with evidence of a managed project that lacks `Microsoft.NET.Test.Sdk`.
- `Resources.resx` + `Resources.Designer.cs`: Added `MissingTestSdkWarning` resource string.
- All `.xlf` files updated with the new resource string.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26684244107)*

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -384,7 +384,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
                 if (_fileHelper.Exists(depsFilePath) && _fileHelper.Exists(runtimeConfigDevPath))
                 {
                     string message = string.Format(CultureInfo.CurrentCulture, Resources.MissingTestSdkWarning, sourcePath);
-                    EqtTrace.Warning("DotnetTestHostManager.GetTestHostStartInfo: {0}", message);
+                    EqtTrace.Warning("DotnetTestHostManager.GetTestHostProcessStartInfo: {0}", message);
                     _messageLogger?.SendMessage(TestMessageLevel.Warning, message);
                 }
 #if NETFRAMEWORK
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -69,6 +69,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
     private readonly IWindowsRegistryHelper _windowsRegistryHelper;
     private readonly IEnvironmentVariableHelper _environmentVariableHelper;
 
+    private IMessageLogger? _messageLogger;
     private ITestHostLauncher? _customTestHostLauncher;
     private Process? _testHostProcess;
     private StringBuilder? _testHostProcessStdError;
@@ -201,6 +202,7 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
         _createNoNewWindow = runConfiguration.CreateNoNewWindow;
         var forwardOutput = runConfiguration.ForwardStandardOutput;
         _testHostManagerCallbacks = new TestHostManagerCallbacks(forwardOutput, logger);
+        _messageLogger = logger;
 
         _architecture = runConfiguration.TargetPlatform;
         _targetFramework = runConfiguration.TargetFramework;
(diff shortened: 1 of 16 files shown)


src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -376,6 +378,15 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
             if (testHostPath.IsNullOrEmpty())
             {
                 // We still did not find testhost.dll. Try finding it next to vstest.console, (or in next to vstest.console ./TestHostNet for .NET Framework)
+
+                // If the project has a proper deps.json but testhost is not listed in it, the most likely cause is
+                // a missing reference to 'Microsoft.NET.Test.Sdk'. Warn the user to help them diagnose the failure.
+                if (_fileHelper.Exists(depsFilePath) && _fileHelper.Exists(runtimeConfigDevPath))
+                {
+                    string message = string.Format(CultureInfo.CurrentCulture, Resources.MissingTestSdkWarning, sourcePath);
+                    EqtTrace.Warning("DotnetTestHostManager.GetTestHostStartInfo: {0}", message);
```
nohwnd: **[Error Reporting & Diagnostic Clarity]** The `EqtTrace.Warning` call uses the method name `GetTestHostStartInfo`, but the actual method is `GetTestHostProcessStartInfo`. This won't affect user-visible output, but it makes log correlation harder when debugging.

```csharp
// Current (incorrect method name)
EqtTrace.Warning("DotnetTestHostManager.GetTestHostStartInfo: {0}", message);

// Should be:
EqtTrace.Warning("DotnetTestHostManager.GetTestHostProcessStartInfo: {0}", message);
```

nohwnd (commented):
## Review Summary

**Dimensions activated**: Process Architecture & Host Resolution · Testhost Assembly Loading & Resolution · Error Reporting & Diagnostic Clarity

### What was checked

**Testhost Assembly Loading & Resolution** — The warning condition (`depsFilePath` exists AND `runtimeConfigDevPath` exists AND `GetTestHostPath` returned empty) is a sound heuristic. `runtimeconfig.dev.json` only appears in development builds of SDK-style projects, so its presence is a reliable signal that `Microsoft.NET.Test.Sdk` should be referenced. The fallback path logic is unchanged and correct.

**Process Architecture & Host Resolution** — Flow analysis confirms the warning fires only when both `GetTestHostPath` calls return empty and we're about to fall back to the runner-side testhost. On Windows, if `testHostPath` was set from the nuget-based lookup at line 338 (even without a `testHostExeFound`), the warning is correctly suppressed at line 375 since `testHostPath.IsNullOrEmpty()` is false.

**Error Reporting & Diagnostic Clarity** — The user-visible message is accurate and actionable. One minor issue noted inline: the `EqtTrace.Warning` call uses `GetTestHostStartInfo` rather than the actual method name `GetTestHostProcessStartInfo`.

**`_messageLogger` null safety** — The `?.` guard is appropriate; `Initialize()` can be called with a null logger and this is handled safely.

**No public API changes** — `_messageLogger` is private; `Resources.MissingTestSdkWarning` is internal. No `PublicAPI.Unshipped.txt` update needed.

**Localization** — All 12 `.xlf` files correctly updated with `state="new"` per convention.

---

One minor inline finding (incorrect trace method name). No blocking issues.

🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/26684737150)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/26684737150)*

nohwnd:
Commit pushed: [`b7cf13d`](https://github.com/microsoft/vstest/commit/b7cf13d4463465b97940cfdeb4433decca9b2c87)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26684854944)*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd: Fixed — updated the `EqtTrace.Warning` call to use `GetTestHostProcessStartInfo` as the method name.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/26684854944)*

nohwnd:
Superseded by #16169

PR #16086, opened 2026-06-03
Remove DotNet-Symbol-Server-Pats group access

Evangelink:
Removed access to the DotNet-Symbol-Server-Pats group.


Evangelink:
Replaced by #16087

Issue #15650, opened 2026-04-09
Consider moving away from nuspec


Youssef1313:
Looking at NuGet's implementation, the default build action for any content files that don't have nuspec metadata is compile. So for source code, it's technically unnecessary to add the nuspec content file metadata. Even if it's not documented, changing it would break a lot of packages/projects, so I'm sure the behavior will never change. 

FWIW, we consider using a nuspec file "the old way". You already have a csproj in this directory, so you could consider moving to msbuild based build. It looks like you probably already use `dotnet pack` to pack all the nupkgs. I expect copilot or other AIs to do a good job automating the migration from nuspec to full msbuild pack, so I won't bother writing hints here, but if you decide to take it on and need help, reach out. I think it'd be just 3 lines (well, XML elements, it depends if you format long elements across multiple lines) in the csproj, two for the readme, and one for all the source files that go in the contentFiles directory.

_Originally posted by @zivkan in https://github.com/microsoft/vstest/pull/15638#discussion_r3060768012_

nohwnd:
is this only for the new nuget package that has source code? I don't mind moving away from nuspec for all packages either.

Youssef1313:
I would say for all packages.

nohwnd:
reopening .cli is not done according to the PR that closed this.

PR #16125, opened 2026-06-16
Migrate nuspec to MSBuild pack (batch 1: simple packages)

azat-msft:
## Summary

Begins migration from `.nuspec` files to MSBuild pack (csproj-based packaging), as requested in #15650.

This PR migrates the **5 simplest packages**. The remaining 8 packages (aggregate/bundle packages) are significantly more complex and will follow in subsequent PRs.

## Migrated packages

| Package | Complexity | Notes |
|---------|-----------|-------|
| **Filter.Source** | Source-only | `Content` items with `BuildAction="Compile"` for contentFiles |
| **AdapterUtilities** | Simple lib | Straightforward — MSBuild pack auto-includes DLLs + satellites |
| **TrxLogger** | Lib with controlled deps | `PrivateAssets="all"` on CoreUtilities to suppress NuGet dependency |
| **Build** | MSBuild task | Removed 3 nuspec variants (normal/sourcebuild/VMR); source build's `TargetFrameworks` override handles it naturally |
| **ObjectModel** | Bundles other DLLs | Uses `TargetsForTfmSpecificContentInPackage` to bundle CoreUtilities + PlatformAbstractions DLLs and satellite resources |

## Remaining packages (not in this PR)

| Package | Files | Why it's harder |
|---------|-------|----------------|
| TranslationLayer | 175 | Bundles DLLs from multiple projects |
| Internal.Uwp | ~30 | Bundles ObjectModel + CoreUtilities for netstandard2.0 |
| NET.Test.Sdk | ~15 | Ships .props/.targets + `_._` stubs |
| CodeCoverage | ~40 | Bundles external code coverage tools |
| TestHost | 65 | Bundles testhost executables, msdia140.dll, build props |
| CLI | 483 | 3 nuspec variants, bundles entire CLI tooling |
| TestPlatform | 553 | Bundles entire platform output |
| Portable | 608 | Cross-platform variant of TestPlatform |

The last 4 are packaging-only projects under `src/package/` that don't build any code — they purely aggregate outputs from many other projects. These may or may not be good candidates for MSBuild pack migration, as they are essentially packaging containers.

## Notable changes

- **ObjectModel** now lists `System.ValueTuple` and `System.Collections.Immutable` as explicit NuGet dependencies for net462. The old nuspec omitted these but they were still required at runtime. This is more correct.
- XML doc files (`.xml`) are now included in packages that previously omitted them — this enables IntelliSense for consumers.
- `eng/expected-nupkg-file-counts.json` updated to reflect new file counts.

## Validation

Each package was verified with a clean `build.cmd -c Release -pack` build, which runs the full nupkg file count and DLL framework verification pipeline.


Evangelink:
I see that count increased, please validate diff of files within the packages and whether the changes is really expected.

nohwnd (changes requested):
as noted by Amaury, the packages need to end up being the same as they were, ideally matching counts, and target frameworks of the files. Unless we have a good reason for doing the change, or can explain why the change is necessary. It would be able nice to spike the harder packages to prove that this approach is doable for them, so we don't end up with multiple ways to package.

Microsoft.TestPlatform.CLI would benefit the most from not using nuspec.

azat-msft:
I verified what are the added files, all of them are `.xml` files that enable. The only other file that is different in initial and new packages is `.psmdcp`, which according to my research is meant to be uniquely named. The `FilterSource` does not have `.xml` file because it's "source-only" package.

With that said, I would argue those files are worth keeping. If you have a different opinion on that, let me know, I'll make copilot remove them from packages hehe.

nohwnd:
> I verified what are the added files, all of them are `.xml` files that enable.

That enable what?

nohwnd:
And why we should add them when they were not there before?

azat-msft:
> > I verified what are the added files, all of them are `.xml` files that enable.
> 
> That enable what?

That enable IntelliSense for package consumers and other test platforms.

Sorry for not including that info to the comment in the first place. Missed it somehow.

nohwnd (approved):

PR #16126, opened 2026-06-17
[fix] Use RuntimeInformation for architecture detection in netstandard2.0 builds

nohwnd:
## Summary

Fixes #15330 — replaces `IsWow64Process2` P/Invoke with `RuntimeInformation` APIs for `NETSTANDARD2_0_OR_GREATER` builds.

> 🤖 This is an automated fix created by the Issue Triage agent.

## Root Cause

Two methods in `PlatformAbstractions/net462/` used the Windows-only `IsWow64Process2` kernel32.dll API to detect CPU architecture. When code targeting `netstandard2.0` runs on Linux/macOS ARM64:

- **`PlatformEnvironment.IsArm64()`** — called `IsWow64Process2` which threw a `DllNotFoundException` on non-Windows, the catch block silently returned `false`, causing `Architecture` to report `X64` instead of `ARM64`.
- **`ProcessHelper.GetCurrentProcessArchitecture()`** — on 64-bit non-Windows, fell through to `GetProcessArchitecture(processId)` which explicitly `throw new NotImplementedException()` for non-Windows.

The `net8.0+` builds already use `RuntimeInformation` (in `netcore/System/`) and were unaffected. Only `netstandard2.0` consumers on non-Windows were broken.

## Fix

For `NETSTANDARD2_0_OR_GREATER` builds:

- `PlatformEnvironment.IsArm64()` now returns `RuntimeInformation.OSArchitecture == Architecture.Arm64`, which works correctly on all platforms.
- `ProcessHelper.GetCurrentProcessArchitecture()` now returns a value based on `RuntimeInformation.ProcessArchitecture`, avoiding the `IsWow64Process2` path entirely.

The `NETFRAMEWORK` (net462) path is unchanged — `IsWow64Process2` is correct and required there.

The `GetProcessArchitecture(int processId)` method (for querying *other* processes' architecture) is also unchanged — there is no managed cross-platform equivalent for that.

## Testing

Build and unit tests pass (`./test.sh`). The fix is in code paths exercised by the existing `DotnetHostHelperTest` and related tests.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/27659787848)*

### DIFF
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
@@ -41,9 +41,13 @@ public partial class ProcessHelper : IProcessHelper
         {
             Architecture.X86 => PlatformArchitecture.X86,
             Architecture.X64 => PlatformArchitecture.X64,
-            Architecture.Arm64 => PlatformArchitecture.ARM64,
             Architecture.Arm => PlatformArchitecture.ARM,
-            _ => PlatformArchitecture.X64,
+            Architecture.Arm64 => PlatformArchitecture.ARM64,
+            (Architecture)5 => PlatformArchitecture.S390x,
+            (Architecture)6 => PlatformArchitecture.LoongArch64,
+            (Architecture)8 => PlatformArchitecture.Ppc64le,
+            (Architecture)9 => PlatformArchitecture.RiscV64,
+            _ => throw new NotSupportedException(),
         };
 #else
         // When this is current process, we can just check if IntPointer size to get if we are 64-bit or 32-bit.
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
@@ -36,6 +36,16 @@ public partial class ProcessHelper : IProcessHelper
             return _currentProcessArchitecture.Value;
         }
 
+#if NETSTANDARD2_0_OR_GREATER
+        _currentProcessArchitecture = RuntimeInformation.ProcessArchitecture switch
+        {
+            Architecture.X86 => PlatformArchitecture.X86,
+            Architecture.X64 => PlatformArchitecture.X64,
+            Architecture.Arm64 => PlatformArchitecture.ARM64,
+            Architecture.Arm => PlatformArchitecture.ARM,
+            _ => PlatformArchitecture.X64,
+        };
+#else
         // When this is current process, we can just check if IntPointer size to get if we are 64-bit or 32-bit.
         // When it is 32-bit we can just return, if it is 64-bit we need to clarify if x64 or arm64.
         if (IntPtr.Size == 4)
@@ -45,6 +55,7 @@ public partial class ProcessHelper : IProcessHelper
         }
 
         _currentProcessArchitecture ??= GetProcessArchitecture(_currentProcess.Id);
+#endif
         return _currentProcessArchitecture.Value;
     }
 #endif
--- a/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/PlatformEnvironment.cs
+++ b/src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/PlatformEnvironment.cs
@@ -6,6 +6,9 @@
 using System;
 using System.ComponentModel;
 using System.Diagnostics;
+#if NETSTANDARD2_0_OR_GREATER
+using System.Runtime.InteropServices;
+#endif
 
 using Microsoft.VisualStudio.TestPlatform.PlatformAbstractions.Interfaces;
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
```diff
@@ -36,6 +36,16 @@ public PlatformArchitecture GetCurrentProcessArchitecture()
...
         }
 
+#if NETSTANDARD2_0_OR_GREATER
+        _currentProcessArchitecture = RuntimeInformation.ProcessArchitecture switch
+        {
+            Architecture.X86 => PlatformArchitecture.X86,
+            Architecture.X64 => PlatformArchitecture.X64,
+            Architecture.Arm64 => PlatformArchitecture.ARM64,
+            Architecture.Arm => PlatformArchitecture.ARM,
+            _ => PlatformArchitecture.X64,
```
nohwnd: **[Process Architecture & Host Resolution]** The `_ => PlatformArchitecture.X64` fallback silently misidentifies exotic architectures (S390x, LoongArch64, Ppc64le, RiscV64) as X64 when this `netstandard2.0` assembly runs on .NET 6+ on those platforms.

The authoritative `netcore/System/ProcessHelper.cs` handles these correctly using numeric casts (since the symbolic names weren't available in older .NET SDKs) and throws `NotSupportedException` for truly unknown values:

```csharp
Architecture.X86 => PlatformArchitecture.X86,
Architecture.X64 => PlatformArchitecture.X64,
Architecture.Arm => PlatformArchitecture.ARM,
Architecture.Arm64 => PlatformArchitecture.ARM64,
(Architecture)5 => PlatformArchitecture.S390x,
(Architecture)6 => PlatformArchitecture.LoongArch64,
(Architecture)8 => PlatformArchitecture.Ppc64le,
(Architecture)9 => PlatformArchitecture.RiscV64,
_ => throw new NotSupportedException(),
```

The same pattern should be applied here to stay consistent with the `NET` path and to surface unsupported platforms clearly rather than silently handing back a wrong value that could lead to the wrong test host being selected.

Note that `PlatformArchitecture.S390x` etc. exist in the shared type definition (the `netcore/` file already references them), so the enum values are usable in a `netstandard2.0`-compiled library.

nohwnd (commented):
## Review: [Process Architecture & Host Resolution, Cross-TFM & Framework Resolution]

The fix correctly replaces the Windows-only `IsWow64Process2` P/Invoke with `RuntimeInformation` APIs for `NETSTANDARD2_0` builds. The core mechanics are sound:

- **`PlatformEnvironment.IsArm64()`** — `RuntimeInformation.OSArchitecture` is the right API for OS-level architecture detection; it returns `Arm64` on ARM64 macOS even for x64 processes running under Rosetta, matching the semantic intent of the `IEnvironment.Architecture` property.
- **`ProcessHelper.GetCurrentProcessArchitecture()`** — `RuntimeInformation.ProcessArchitecture` is the right API for per-process architecture; correctly returns the actual process ABI, not the OS one.
- **TFM guard correctness** — `#if NETSTANDARD2_0_OR_GREATER` is valid inside `#if NETFRAMEWORK || NETSTANDARD2_0` because the `netstandard2.0` TFM also defines `NETSTANDARD2_0_OR_GREATER`; the NETFRAMEWORK path is untouched.

**One concern** (see inline): the `_ => PlatformArchitecture.X64` catch-all in `GetCurrentProcessArchitecture()` silently misidentifies S390x/LoongArch64/Ppc64le/RiscV64 as X64 when the `netstandard2.0` DLL runs on .NET 6+ on those platforms. The `netcore/` implementation maps all these explicitly using numeric casts and throws `NotSupportedException` for truly unknown values — the `netstandard2.0` path should mirror that pattern for consistency and to avoid silent wrong-host selection.

🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/27660528833)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27660528833)*

nohwnd:
Commit pushed: [`a042db0`](https://github.com/microsoft/vstest/commit/a042db0c034679664ec19ae2866d6ab11f8ef207)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27660835764)*

src/Microsoft.TestPlatform.PlatformAbstractions/net462/System/ProcessHelper.cs
nohwnd: Done — replaced the `_ => PlatformArchitecture.X64` fallback with explicit cases for `S390x`, `LoongArch64`, `Ppc64le`, and `RiscV64` (using numeric casts as in the `netcore` path), and changed the catch-all to `throw new NotSupportedException()`. This now matches the pattern in `netcore/System/ProcessHelper.cs` exactly.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27660835764)*

PR #16129, opened 2026-06-17
Run tests as exe (RunAsExe attribute)

nohwnd:
Reviving the experiment that lets a test project run as its own `.exe` instead of going through `testhost.exe`. A test framework opts in with an assembly-level `[RunAsExe]` attribute; vstest reads it, marks the source as `ExecutionPreference.RunAsExe`, turns off app domains, and launches the test's own exe as the host.

The branch was ~260 commits behind, so most of this is just merging current main back in and getting it to build. The acceptance-test caching I had here is already in main now (`IntegrationTestBuild`), so I dropped my copy and kept main's.

This is WIP and intentionally rough. A few probes are still hard-coded to force the new path so I can see what falls over:

- `DefaultTestHostManager` throws instead of using the built-in/shared testhost.
- `DotnetTestHostManager` always resolves the `.exe` next to the `.dll`.
- `TestEngine` forces isolation (no in-process run).
- `RunConfiguration.DisableSharedTestHost` defaults to `true`.

So a lot of unit/acceptance tests fail by design for now. Opening as draft to get it building in CI and to pick the work back up.

Builds locally with `build.cmd` (Debug, all TFMs).


nohwnd (commented):
## Expert Review — Run Tests as Exe (WIP)

> **Draft PR acknowledged** — The description accurately documents the intentional hard-coded probes and WIP state. The findings below are scoped to real bugs and design issues that should be addressed before this graduates to a reviewable state, not commentary on the probes.

### Findings

| # | File | Dimension | Severity |
|---|------|-----------|---------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| 3 | `AssemblyResolver.cs` | Process Architecture & Host Resolution | 🐛 Design bug |
| 4 | `DefaultTestHostManager.cs` | Process Architecture & Host Resolution | 🐛 Latent null |
| 5 | `DotnetTestHostManager.cs` | Process Architecture & Host Resolution | 🚧 WIP tracked |

### Summary

**Finding 1** — `SourceDetail.Architecture` in `PublicAPI.Shipped.txt` is changed from `Architecture` (non-nullable struct) to `Architecture?` (nullable). This is a source-breaking change to an already-shipped API. `PublicAPI.Shipped.txt` should not be modified; the nullable variant belongs in `PublicAPI.Unshipped.txt`.

**Finding 2** — `GetRunAsExeFromAssemblyMetadata` matches attributes by **short name only**, meaning any attribute named `RunAsExeAttribute` in any namespace will trigger RunAsExe. The match must also verify the declaring namespace is `Microsoft.VisualStudio.TestPlatform.ObjectModel`.

**Finding 3** — `AssemblyResolver._isTestHost` uses `ProcessName.StartsWith("Test", ...)`. This is far too broad — any process starting with "Test" (test utilities, runners, explorers) will have all custom assembly resolution silently bypassed. A constructor-injected host-mode flag would be more precise.

**Finding 4** — In `DefaultTestHostManager`, the `!isUsingTestHostFromNextToSource` fallback path has a latent null: when `_fileHelper.Exists(testHostPath)` is `true`, `testHostProcessPath` is never assigned before the `TPDebug.Assert(testHostProcessPath is not null)` fires. This is masked by the WIP guard today but will surface immediately when it is removed.

**Finding 5** — `DotnetTestHostManager.GetTestHostProcessStartInfo` calls `TryGetExePathFromDll` unconditionally, which throws for all standard `.dll`-only test runs. Tracked as acknowledged WIP; the final structure needs a gate on `ExecutionPreference.RunAsExe`.

### PR Description Alignment

Title and description accurately describe the scope and WIP nature. No misalignment found. ✅

🧠 *Reviewed by expert-reviewer workflow*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27681333801)*

src/Microsoft.TestPlatform.ObjectModel/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -672,7 +672,7 @@ Microsoft.VisualStudio.TestPlatform.ObjectModel.SettingsNameAttribute
 Microsoft.VisualStudio.TestPlatform.ObjectModel.SettingsNameAttribute.SettingsName.get -> string!
 Microsoft.VisualStudio.TestPlatform.ObjectModel.SettingsNameAttribute.SettingsNameAttribute(string! settingsName) -> void
 Microsoft.VisualStudio.TestPlatform.ObjectModel.SourceDetail
-Microsoft.VisualStudio.TestPlatform.ObjectModel.SourceDetail.Architecture.get -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Architecture
+Microsoft.VisualStudio.TestPlatform.ObjectModel.SourceDetail.Architecture.get -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Architecture?
 Microsoft.VisualStudio.TestPlatform.ObjectModel.SourceDetail.Framework.get -> Microsoft.VisualStudio.TestPlatform.ObjectModel.Framework?
```
nohwnd, line 676: **[Public API Surface Protection]** `SourceDetail.Architecture` is modified in `PublicAPI.Shipped.txt` from non-nullable `Architecture` to nullable `Architecture?`. Modifying `PublicAPI.Shipped.txt` is a **source-breaking change** for existing callers — code that uses this property without a null-check will get a compiler warning/error (Roslyn treats it as a nullability contract change).

`PublicAPI.Shipped.txt` represents the already-released surface. The correct pattern is:
- Remove or leave the existing entry in `PublicAPI.Shipped.txt` alone
- Add the new nullability signature only to `PublicAPI.Unshipped.txt`

Consider keeping the shipped property non-nullable and adding a distinct property (e.g., in `PublicAPI.Unshipped.txt`) if the nullable form is required for the new flow.

🧠 *Reviewed by expert-reviewer workflow*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -267,6 +267,9 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
         var args = string.Empty;
         var sourcePath = sources.Single();
         var sourceFile = Path.GetFileNameWithoutExtension(sourcePath);
+
+        // use the exe as host
+        _dotnetHostPath = TryGetExePathFromDll(sourcePath);
```
nohwnd: **[Process Architecture & Host Resolution]** `TryGetExePathFromDll` is called **unconditionally** for every source regardless of `ExecutionPreference`. It `throw`s `ArgumentException` when no `.exe` exists next to the `.dll` — which is the **standard case** for all .NET Core test runs. This means every ordinary `dotnet test` invocation will fail at this point.

The PR description acknowledges this as an intentional WIP probe, but flagging for tracking: the structural pattern should gate on `ExecutionPreference` before attempting exe resolution, rather than always throwing. Something like:

```csharp
if (runConfiguration.ExecutionPreference == ExecutionPreference.RunAsExe)
{
    _dotnetHostPath = TryGetExePathFromDll(sourcePath);
}
```

🧠 *Reviewed by expert-reviewer workflow*

src/vstest.console/CommandLine/AssemblyMetadataProvider.cs
```diff
@@ -155,6 +172,59 @@ private static FrameworkName GetFrameworkNameFromAssemblyMetadata(Stream assembl
...
+                }
+
+                if (typeNameHandle == null)
+                {
+                    continue;
+                }
+
+                var typeName = metadataReader.GetString(typeNameHandle.Value);
+
+                if (string.Equals(typeName, "RunAsExeAttribute", StringComparison.OrdinalIgnoreCase))
```
nohwnd, line 217: **[Null Safety & Boundary Validation]** The attribute is matched by **short name only** (`"RunAsExeAttribute"`), without checking the namespace. Any assembly that happens to declare an attribute named `RunAsExeAttribute` in any namespace will be treated as RunAsExe-capable — including third-party or test-utility attributes that coincidentally share the name.

Before this ships, the match should also verify the declaring namespace is `Microsoft.VisualStudio.TestPlatform.ObjectModel`:

```csharp
// For TypeReference (external ref):
StringHandle? namespaceHandle = null;
if (mref.Parent.Kind == HandleKind.TypeReference)
{
    TypeReference tref = metadataReader.GetTypeReference((TypeReferenceHandle)mref.Parent);
    typeNameHandle = tref.Name;
    namespaceHandle = tref.Namespace;
}
// Then check both:
if (string.Equals(typeName, "RunAsExeAttribute", StringComparison.OrdinalIgnoreCase)
    && string.Equals(ns, "Microsoft.VisualStudio.TestPlatform.ObjectModel", StringComparison.Ordinal))
```

🧠 *Reviewed by expert-reviewer workflow*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -180,57 +181,118 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+                throw new NotSupportedException("You cannot use built-in testhost!!!!");
+            }
+            currentWorkingDirectory = Path.GetDirectoryName(typeof(DefaultTestHostManager).Assembly.Location);
+            TPDebug.Assert(currentWorkingDirectory is not null, "Current working directory must not be null.");
+
+            // check in current location for testhost exe
+            var testHostPath = Path.Combine(currentWorkingDirectory, testHostProcessName);
+
+            if (!_fileHelper.Exists(testHostPath))
+            {
```
nohwnd, line 250: **[Process Architecture & Host Resolution]** Latent null-dereference when the WIP guard is removed: if `_fileHelper.Exists(testHostPath)` returns `true` (testhost found in the current directory), the `if (!_fileHelper.Exists(...))` branch is skipped and `testHostProcessPath` is **never assigned** inside this block — it stays `null`. The `TPDebug.Assert(testHostProcessPath is not null, ...)` immediately below will then fail.

The fix when removing the guard: assign `testHostProcessPath = testHostPath;` in the `else` branch (or unconditionally before the inner `if`).

```csharp
var testHostPath = Path.Combine(currentWorkingDirectory, testHostProcessName);
testHostProcessPath = testHostPath; // always assign the candidate path

if (!_fileHelper.Exists(testHostPath))
{
    // ... fallback to TestHostNetFramework subfolder ...
    testHostProcessPath = Path.Combine(...);
}
```

🧠 *Reviewed by expert-reviewer workflow*

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
```diff
@@ -57,6 +58,9 @@ public AssemblyResolver(IEnumerable<string> directories)
         _platformAssemblyLoadContext = new PlatformAssemblyLoadContext();
 
         _platformAssemblyResolver.AssemblyResolve += OnResolve;
+
+        _isTestHost = Process.GetCurrentProcess().ProcessName.StartsWith("testhost", StringComparison.OrdinalIgnoreCase)
+            || Process.GetCurrentProcess().ProcessName.StartsWith("Test", StringComparison.OrdinalIgnoreCase);
```
nohwnd: **[Process Architecture & Host Resolution]** `StartsWith("Test", ...)` is far too broad — it matches any process whose name begins with `"Test"` (e.g., `TestConsole.exe`, `TestRunner.exe`, `TestAgent.exe`, `TestExplorer.exe`). When `_isTestHost` is `true`, **all** custom assembly resolution is silently skipped for the entire lifetime of that `AssemblyResolver` instance.

This could silently disable extension/adapter loading in test utilities or runners that share the same `AssemblyResolver` code path but aren't actually a testhost. The condition should be more precise — ideally a constructor parameter or injected host-mode flag rather than process-name inference, since process names are fragile and ambiguous.

Also: `Process.GetCurrentProcess()` is called twice here; the result should be cached in a local variable.

🧠 *Reviewed by expert-reviewer workflow*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -1032,4 +1052,15 @@ private string GetTestHostPath(string runtimeConfigDevPath, string depsFilePath,
...
     }
+
+    internal static string? TryGetExePathFromDll(string sourcePath)
+    {
+        if (Path.GetExtension(sourcePath) != ".dll")
+        {
+            return null;
+        }
+
+        var exe = Path.ChangeExtension(sourcePath, ".exe"); // todo: linux macos
```
nohwnd, line 1068: **[Process Architecture & Host Resolution]** `TryGetExePathFromDll` hardcodes `.exe` — silent failure on Linux/macOS.

```csharp
var exe = Path.ChangeExtension(sourcePath, ".exe"); // todo: linux macos
```

On Linux and macOS the SDK produces apphosts with **no extension** (e.g. `MyTests`, not `MyTests.exe`). This method will return `null` on every non-Windows run, and `_dotnetHostPath` will not be updated. The fallback trace message is `Verbose`-level, so end users who opt in with `[RunAsExe]` on Linux/macOS will get no indication that the feature is silently disabled.

Before graduation, the cross-platform path needs to probe for the no-extension form:
```csharp
// Linux/macOS: apphost has no extension
var candidate = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
    ? Path.ChangeExtension(sourcePath, ".exe")
    : Path.Combine(Path.GetDirectoryName(sourcePath)!, Path.GetFileNameWithoutExtension(sourcePath));
```
And the fallback should surface a **Warning**-level message so framework authors can diagnose the issue.

src/Microsoft.TestPlatform.Common/Utilities/AssemblyResolver.cs
nohwnd: **[Process Architecture & Host Resolution]** `Process.GetCurrentProcess()` is called twice in a single assignment. The first call's result is discarded before the `||` short-circuits to the second call. Capture the process name once:

```csharp
var processName = Process.GetCurrentProcess().ProcessName;
_isTestHost = processName.StartsWith("testhost", StringComparison.OrdinalIgnoreCase)
    || processName.StartsWith("Test", StringComparison.OrdinalIgnoreCase);
```

Note: `StartsWith("Test", ...)` remains extremely broad — it matches `TestApp`, `TestRunner`, `TestExplorer`, etc., silently disabling custom assembly resolution for any process whose name starts with "Test". This is the same concern raised in the previous review (Finding 3). A constructor-injected `bool isTestHost` flag would be more precise and testable.

nohwnd (commented):
## Expert Review — Round 2 (updated since 2026-06-17T10:16Z)

> **Draft PR — WIP findings only.** The previous review (run [27681333801](https://github.com/microsoft/vstest/actions/runs/27681333801)) identified 5 issues. This round covers the delta introduced by the 17 new commits.

### Progress since last review ✅

| # | Finding | Status |
|---|---------|--------|
| 4 | `DefaultTestHostManager` latent null (`testHostProcessPath` unassigned) | **Resolved** — both branches now assign `testHostProcessPath` before use |
| 5 | `DotnetTestHostManager.TryGetExePathFromDll` called unconditionally | **Resolved** — now correctly gated on `if (_runAsExe)` |

### Still open from last review

| # | File | Dimension | Severity |
|---|------|-----------|---------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| 3 | `AssemblyResolver.cs` | Process Architecture & Host Resolution | 🐛 Design bug |

### New findings (this round)

| # | File | Dimension | Severity |
|---|------|-----------|---------|
| A | `DotnetTestHostManager.cs:1063` | Process Architecture & Host Resolution | 🐛 Silent failure |
| B | `net462/Microsoft.TestPlatform.TestHost.targets:120` | Build Script & Infrastructure Hygiene | 🐛 Package bloat |
| C | `AssemblyResolver.cs:62` | Process Architecture & Host Resolution | ⚡ Perf / design |

**Finding A** — `TryGetExePathFromDll` hardcodes `.exe`. On Linux/macOS the SDK-produced apphost has no extension, so the method always returns `null`. The fallback is trace-only — end users get no diagnostic. See inline comment.

**Finding B** — Both the x86 and x64 `ItemGroup` blocks in the new `net462/Microsoft.TestPlatform.TestHost.targets` use the **identical condition** (`'$(OutputType)' != 'Exe' AND '$(OS)' == 'Windows_NT'`). The original architecture-differentiating conditions are commented out. Every Library test project on Windows will receive both x86 and x64 testhost binaries — double the expected file count. See inline comment.

**Finding C** — `Process.GetCurrentProcess()` is called twice in the `AssemblyResolver` constructor. The `StartsWith("Test")` predicate from Finding 3 is still there. See inline comment.

### PR Description Alignment

Title and description accurately describe the WIP scope and the intentionally hard-coded probes. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27708486557)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27708486557)*

src/package/Microsoft.TestPlatform.TestHost/build/net462/Microsoft.TestPlatform.TestHost.targets
```diff
@@ -0,0 +1,194 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>False</Visible>
+    </Content>
+  </ItemGroup>
+
+ 
+
+  <!-- x64 / non-x86, non-ARM64: default bucket -->
+  <!--<ItemGroup Condition=" ( '$(Platform)' != 'x86' AND '$(PlatformTarget)' != 'x86' AND '$(Platform)' != 'arm64' AND '$(PlatformTarget)' != 'arm64' ) AND '$(OS)' == 'Windows_NT' ">-->
+    <ItemGroup Condition=" '$(OutputType)' != 'Exe' AND '$(OS)' == 'Windows_NT' ">
```
nohwnd: **[Build Script & Infrastructure Hygiene]** Both x86 and x64 `ItemGroup` elements use the **same condition**, so Library test projects on Windows receive all x86 _and_ all x64 testhost binaries in their output directory.

The original x86/x64 selector conditions are commented out (lines 41 and 119). Both groups now evaluate to the same `'$(OutputType)' != 'Exe' AND '$(OS)' == 'Windows_NT'`, making the split meaningless. A Library project's output will get 24 testhost exe/config files instead of the expected 12.

The x64 group's own comment says "x64 / non-x86, non-ARM64: default bucket", implying these groups were meant to be mutually exclusive. Consider either restoring the platform conditions or merging the two groups into one if architecture-based selection is no longer desired for the net462 path.

nohwnd (commented):
## Expert Review — Round 3 (updated since 2026-06-17T17:56Z)

> **Draft PR — WIP findings only.** The previous review ([27708486557](https://github.com/microsoft/vstest/actions/runs/27708486557)) covered commit `2de42d65`. This round covers the single new commit `017fd64a`.

### New commit since Round 2

One commit: `017fd64a` — fixes CS8604 source-build warning in `SourceDetailHelper.cs` by using `GetValueOrDefault().ToString()` on nullable `Architecture?` and `ExecutionPreference?` properties.

### New finding

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| D | `SourceDetailHelper.cs:24` | Null Safety & Boundary Validation | 🐛 Correctness |

**Finding D** — `Architecture.GetValueOrDefault()` when null returns `Architecture.Default` (enum `0`), which serializes to `"Default"`. With `overwrite: true`, this writes `<TargetPlatform>Default</TargetPlatform>` into RunSettings and **silently overwrites any user-configured explicit architecture** (e.g. `X64`).

`InferRunSettingsHelper` itself flags `Architecture.Default` as invalid in its own validator (`value != Architecture.Default`), so the round-trip semantics of `"Default"` are not the same as omitting the node. The correct fix is a null guard: `if (sourceDetail.Architecture is { } architecture)`. See inline comment.

### Still open from prior rounds

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| 3 | `AssemblyResolver.cs` | Process Architecture & Host Resolution | 🐛 Design bug |
| A | `DotnetTestHostManager.cs:1063` | Process Architecture & Host Resolution | 🐛 Silent failure |
| B | `net462/Microsoft.TestPlatform.TestHost.targets:120` | Build Script & Infrastructure Hygiene | 🐛 Package bloat |

### PR Description Alignment

Title and description continue to accurately describe the WIP scope. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27710600718)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27710600718)*

src/Microsoft.TestPlatform.CrossPlatEngine/Utilities/SourceDetailHelper.cs
```diff
@@ -21,7 +21,15 @@ internal static string UpdateRunSettingsFromSourceDetail(string runSettings, Sou
         var navigator = document.CreateNavigator();
 
         InferRunSettingsHelper.UpdateTargetFramework(document, sourceDetail.Framework!.ToString(), overwrite: true);
-        InferRunSettingsHelper.UpdateTargetPlatform(document, sourceDetail.Architecture.ToString(), overwrite: true);
+        InferRunSettingsHelper.UpdateTargetPlatform(document, sourceDetail.Architecture.GetValueOrDefault().ToString(), overwrite: true);
```
nohwnd, line 24: **[Null Safety & Boundary Validation]** `GetValueOrDefault()` when `Architecture` is null silently promotes `null` to `Architecture.Default` (enum value `0`), which then writes `<TargetPlatform>Default</TargetPlatform>` into the RunSettings XML with `overwrite: true`. This overwrites any user-configured explicit architecture (e.g. `X64`) with `"Default"`.

Noteworthy: `InferRunSettingsHelper` itself has a validator that explicitly marks `Architecture.Default` as _invalid_ for `TargetPlatform` (`value != Architecture.Default`), so the round-trip semantics of `"Default"` are not the same as omitting the node.

The correct pattern is a null guard — skip the call when no architecture was determined:

```csharp
if (sourceDetail.Architecture is { } architecture)
{
    InferRunSettingsHelper.UpdateTargetPlatform(document, architecture.ToString(), overwrite: true);
}
```

Line 30 has the same pattern for `ExecutionPreference` but is safe because the enclosing `if (... == ExecutionPreference.RunAsExe)` already guarantees the value is non-null; `.Value.ToString()` would be more idiomatic there.

nohwnd (commented):
## Expert Review — Round 4 (updated since 2026-06-17T18:34Z)

> **Draft PR — WIP findings only.** The previous review ([27710600718](https://github.com/microsoft/vstest/actions/runs/27710600718)) covered commit `017fd64a`. This round covers the single new commit `36e8bf96`.

### New commit since Round 3

One commit: `36e8bf96` — restores `Microsoft.TestPlatform.ObjectModel` expected nupkg file count from 67 → 93 to match the official CI Release build.

### Analysis — `eng/expected-nupkg-file-counts.json`

**Dimension: Build Script & Infrastructure Hygiene / Dependency & Package Integrity**

The change is correct. The value 67 came from a local `Debug` pack that omits the localized satellite assemblies; the CI Release build produces 93 files. The `TestHost` (140) and `Microsoft.NET.Test.Sdk` (27) counts were already aligned with CI — only ObjectModel needed correction.

One forward-looking note: when Finding B (duplicate `ItemGroup` condition in `Microsoft.TestPlatform.TestHost.targets`) is resolved, the `TestHost` expected count (140) will need to be re-verified against a clean Release pack to ensure it still accounts for the corrected opt-in file layout. No action required now.

**No new findings from this commit. ✅**

### Still open from prior rounds

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| 3 | `AssemblyResolver.cs` | Process Architecture & Host Resolution | 🐛 Design bug |
| A | `DotnetTestHostManager.cs:1063` | Process Architecture & Host Resolution | 🐛 Silent failure |
| B | `net462/Microsoft.TestPlatform.TestHost.targets:120` | Build Script & Infrastructure Hygiene | 🐛 Package bloat |
| D | `SourceDetailHelper.cs:24` | Null Safety & Boundary Validation | 🐛 Correctness |

### PR Description Alignment

Title and description accurately describe the WIP scope and intentional hard-coded probes. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27714519616)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27714519616)*

nohwnd (commented):
## Expert Review — Round 5 (updated since 2026-06-17T19:41Z)

> **Draft PR — WIP findings only.** The previous review ([27714519616](https://github.com/microsoft/vstest/actions/runs/27714519616)) covered commit `36e8bf96`. This round covers the single new commit `8fa6e4f1`.

### New commit since Round 4

One commit: `8fa6e4f1` — *Don't disable the custom AssemblyResolver in the testhost (fixes adapter loading)*

Removes the `_isTestHost` field and its associated guard in `OnResolve` that was skipping custom assembly resolution whenever the current process name started with `"testhost"` or `"Test"`. The `AssemblyResolve` event is now always registered.

### Analysis

**Dimension: Process Architecture & Host Resolution**

This commit resolves **Finding 3** and **Finding C** from prior rounds:

- **Finding 3** — `_isTestHost = ProcessName.StartsWith("Test", ...)`: overly-broad predicate that bypassed adapter resolution for any process beginning with "Test". The field and its guard are now fully removed. ✅
- **Finding C** — `Process.GetCurrentProcess()` called twice in the constructor (one to test host-name, one to skip registration). Both calls are gone. ✅

The root cause is accurately diagnosed in the commit message: the check was only ever triggered by the standard `testhost.exe` process name, which never needed to skip resolution. Run-as-exe test processes use the test project's own executable name and never matched, so removing the check is both safe and correct.

No new findings are introduced by this commit.

### Progress summary across all rounds

| # | Finding | Status |
|---|---------|--------|
| 4 | `DefaultTestHostManager` latent null (`testHostProcessPath`) | ✅ Resolved (Round 2) |
| 5 | `DotnetTestHostManager.TryGetExePathFromDll` called unconditionally | ✅ Resolved (Round 2) |
| C | `AssemblyResolver`: `Process.GetCurrentProcess()` called twice | ✅ Resolved (Round 5) |
| 3 | `AssemblyResolver._isTestHost`: overly-broad `StartsWith("Test")` guard | ✅ Resolved (Round 5) |

### Still open from prior rounds

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| A | `DotnetTestHostManager.cs:1063` | Process Architecture & Host Resolution | 🐛 Silent failure |
| B | `net462/Microsoft.TestPlatform.TestHost.targets:120` | Build Script & Infrastructure Hygiene | 🐛 Package bloat |
| D | `SourceDetailHelper.cs:24` | Null Safety & Boundary Validation | 🐛 Correctness |

### PR Description Alignment

Title and description accurately describe the WIP scope and intentionally hard-coded probes. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27754036082)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27754036082)*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -63,6 +63,14 @@ public class DefaultTestHostManager : ITestRuntimeProvider2
...
     private ITestHostLauncher? _customTestHostLauncher;
     private Process? _testHostProcess;
+    private bool _runAsExe;
+
+    /// <summary>
+    /// Environment variable used to tell the host process that it is running as its own executable (run-as-exe),
+    /// so it disables the custom assembly resolver (all dependencies are next to the exe). Kept in sync with the
+    /// reader in <c>Microsoft.VisualStudio.TestPlatform.Common.Utilities.AssemblyResolver</c>.
+    /// </summary>
+    internal const string RunAsExeEnvironmentVariableName = "VSTEST_RUNASEXE";
```
nohwnd, line 73: **[Finding E — Environment Variable & Feature Flag Contracts] 🔧 Duplicate constant / stale cross-reference comment**

```csharp
// Kept in sync with the reader in Microsoft.VisualStudio.TestPlatform.Common.Utilities.AssemblyResolver
internal const string RunAsExeEnvironmentVariableName = "VSTEST_RUNASEXE";
```

Two issues:

1. **Comment inaccuracy** — the reader is `RunAsExeHelper` (in `Microsoft.TestPlatform.Common`), not `AssemblyResolver`. `AssemblyResolver` reads via `RunAsExeHelper.IsRunningAsExe`, as does `TestPluginCache`. Updating the comment to say "reader in `RunAsExeHelper`" prevents future confusion about which type owns the canonical definition.

2. **Duplicate constant** — `RunAsExeHelper.RunAsExeEnvironmentVariableName` (the reader's definition in `Common`) and this constant (the writer's definition in `CrossPlatEngine`) are independent string literals with no compile-time cross-reference. If `CrossPlatEngine` has a project reference to `Common` (which it may, for `AssemblyResolver` and related types), then `DefaultTestHostManager` could use `RunAsExeHelper.RunAsExeEnvironmentVariableName` directly, eliminating the duplication and making the contract impossible to drift. If there's no such reference, moving the constant to `ObjectModel` (shared by both assemblies) would achieve the same protection.

This is low-risk in the short term (both values are hard-coded strings), but for a contract that crosses a process boundary it is good to have a single compile-time source of truth.

nohwnd (commented):
## Expert Review — Round 6 (updated since 2026-06-18T10:55Z)

> **Draft PR — WIP findings only.** The previous review ([27754036082](https://github.com/microsoft/vstest/actions/runs/27754036082)) covered commit `8fa6e4f1`. This round covers the single new commit `8e9f739b`.

### New commit since Round 5

One commit: `8e9f739b` — *Make run-as-exe host ignore the runner's resolver and extensions*

Five files changed (101 additions, 1 deletion):
- **`RunAsExeHelper.cs`** (new) — static helper that reads `VSTEST_RUNASEXE=1` from the environment via a static initializer; consumed by both `AssemblyResolver` and `TestPluginCache`
- **`AssemblyResolver.cs`** — new `_isRunningAsExe` field; `OnResolve` returns `null` early with a verbose trace when set, so the runtime resolves from the app base instead of the runner's extension directories
- **`TestPluginCache.cs`** — `DefaultExtensionPaths` setter filters the incoming paths to only those under `AppContext.BaseDirectory` when `IsRunningAsExe` is true, dropping the runner's loggers/data-collectors/adapters
- **`DefaultTestHostManager.cs`** — sets `VSTEST_RUNASEXE=1` in the child process environment when launching a run-as-exe host
- **`DotnetTestHostManager.cs`** — same env-var write for the .NET Core run-as-exe path

### Analysis

**Correctness** — The overall approach is sound. The env-var gate is process-scoped and set before process launch, so the static-initializer read in `RunAsExeHelper.IsRunningAsExe` is correct. The `AssemblyResolver` early-return includes a `Verbose` trace that identifies which assembly couldn't be found, providing the diagnostic hook needed when validating that everything ships next to a run-as-exe test executable. The `TestPluginCache` filter uses `Path.GetFullPath(path).StartsWith(directory, OrdinalIgnoreCase)` where `directory = AppContext.BaseDirectory`; `AppContext.BaseDirectory` always includes a trailing `Path.DirectorySeparatorChar`, so the prefix comparison is safe.

**Falls-back cleanly** — When `DotnetTestHostManager` cannot find the exe next to the dll, it does _not_ set `VSTEST_RUNASEXE`, so the standard testhost path is unaffected.

### New finding

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| E | `DefaultTestHostManager.cs:73` | Environment Variable & Feature Flag Contracts | 🔧 Maintenance concern |

**Finding E** — `RunAsExeEnvironmentVariableName` is defined independently in both `DefaultTestHostManager` (writer, `CrossPlatEngine`) and `RunAsExeHelper` (reader, `Common`) with no compile-time cross-reference. Additionally, the doc comment on line 70–71 says "reader in `AssemblyResolver`" but the actual reader is `RunAsExeHelper` — `AssemblyResolver` and `TestPluginCache` both read via `RunAsExeHelper.IsRunningAsExe`. See inline comment.

### Still open from prior rounds

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| 1 | `PublicAPI.Shipped.txt` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs` | Null Safety & Boundary Validation | 🐛 Bug |
| A | `DotnetTestHostManager.cs:1063` | Process Architecture & Host Resolution | 🐛 Silent failure (Linux/macOS) |
| B | `net462/Microsoft.TestPlatform.TestHost.targets:120` | Build Script & Infrastructure Hygiene | 🐛 Package bloat |
| D | `SourceDetailHelper.cs:24` | Null Safety & Boundary Validation | 🐛 Correctness |

### PR Description Alignment

Title and description accurately describe the WIP scope and intentionally hard-coded probes. The new commit's intent (self-contained isolation for run-as-exe hosts) is consistent with the overall feature direction. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27764500271)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27764500271)*

nohwnd (commented):
## Expert Review — Round 7 (updated since 2026-06-18T14:09Z)

> **Draft PR — WIP findings only.** The previous review ([27764500271](https://github.com/microsoft/vstest/actions/runs/27764500271)) covered commit `8e9f739b`. This round covers the single new commit `274b3db6` — a merge of main plus a complete re-port of the run-as-exe .NET Framework packaging onto the MSBuild-pack model.

### New commit since Round 6

One commit: `274b3db6` — *Merge main and re-port the run-as-exe packaging to MSBuild pack*

Key changes in this commit (excluding the merged `.github/` workflow files from main):

- **`testhost.library/testhost.library.csproj`** (new) — a `net462` `OutputType=Library` project that compiles the testhost entry-point code from `testhost.x86` sources, producing a referenceable `testhost.dll` that run-as-exe test projects can link against
- **`build/net462/Microsoft.TestPlatform.TestHost.targets`** (new) — conditionally (only when `$(OutputType)=='Exe'`) adds a `Reference` to `testhost.dll` and copies the platform assembly closure next to the test output; non-exe projects are unaffected
- **`build/net462/Microsoft.TestPlatform.TestHost.props`** (new) — stub file with an explanatory comment; no MSBuild side effects
- **`Microsoft.TestPlatform.TestHost.csproj`** — adds the `IncludeRunAsExeContent` target that packs the net462 testhost closure under `build/net462/runAsExe/`; extends `ObjectModel` project reference to be unconditional (was net8.0-only) so net462 consumers get the `[RunAsExe]` attribute type and NuGet unifies `ObjectModel` to this version
- **`Microsoft.NET.Test.Sdk.csproj`** — makes the `TestHost` project reference unconditional (was `net8.0`-only); adds `Program.cs` template to the net462 pack content
- **`Microsoft.NET.Test.Sdk.targets`** (modified) — adds `InitialTargets="GenerateProgramFile"` and the corresponding target that injects the generated Main into `C#`/`VB` test exe projects
- **`Microsoft.NET.Test.Sdk.Program.cs`** (new) — the generated `Main`; probes for `--testsourcepath`, appends it if missing, then delegates to `Microsoft.VisualStudio.TestPlatform.TestHost.Program.Run`
- **`eng/expected-nupkg-file-counts.json`** — `TestHost` 65 → 116 (+51: 8 platform DLLs, 2 msdia natives, 2 build integration files, ~39 localized satellite assemblies for 3 resource DLLs × 13 locales); `Test.Sdk` 26 → 27 (+1: `Program.cs` template)
- **`eng/expected-dll-frameworks.json`** — 8 new `netframework` entries under `build/net462/runAsExe/`

### Analysis

**Finding B: ✅ RESOLVED** — The old `Microsoft.TestPlatform.TestHost.targets` file (which had two `ItemGroup` blocks with identical conditions, causing every Library test project on Windows to receive both x86 and x64 testhost binaries) has been completely replaced. The new file has a single clean `Condition="'$(OutputType)' == 'Exe'"` guard; non-exe projects receive nothing from this new content. The file count update (65 → 116) is consistent with the correctly-gated closure content.

**Packaging design — no new findings:**
- The `testhost.library` approach (library form of the testhost entry point, compiled from `testhost.x86` sources, `AssemblyName=testhost`, `net462` only) is sound. Since `DisableAppDomain=true` is always set for run-as-exe, the `AppDomainEngineInvoker` that is compiled in but not invoked does not introduce risk.
- The `InitialTargets="GenerateProgramFile"` in `.targets` is a valid MSBuild pattern and accumulates correctly with other `InitialTargets` in the project.
- The `Microsoft.NET.Test.Sdk.Program.cs` generated Main is correct: it probes `args` for `--testsourcepath` before appending it, avoiding duplicates when the caller already passes the argument.
- The `IncludeRunAsExeContent` satellite glob (`**/X.resources.dll` → `PackagePath="build/net462/runAsExe/"`) preserves locale subdirectory structure in the pack output (confirmed by the +39 satellite file delta), so .NET Framework satellite resolution from `<culture>/` subdirs works correctly.
- Making `TestHost` and `ObjectModel` dependencies unconditional in `Microsoft.NET.Test.Sdk` / `Microsoft.TestPlatform.TestHost` for `net462` is intentional and necessary for the compile-time `[RunAsExe]` type resolution in run-as-exe projects; Library projects are unaffected at runtime because the integration targets gate all content on `$(OutputType)=='Exe'`.

### Progress summary (all rounds)

| # | Finding | Status |
|---|---------|--------|
| 4 | `DefaultTestHostManager` latent null — `testHostProcessPath` unassigned | ✅ Resolved (Round 2) |
| 5 | `DotnetTestHostManager.TryGetExePathFromDll` called unconditionally | ✅ Resolved (Round 2) |
| C | `AssemblyResolver`: `Process.GetCurrentProcess()` called twice | ✅ Resolved (Round 5) |
| 3 | `AssemblyResolver._isTestHost`: overly-broad `StartsWith("Test")` guard | ✅ Resolved (Round 5) |
| B | `TestHost.targets` duplicate `ItemGroup` conditions → Library projects got double testhost files | ✅ **Resolved (Round 7)** |

### Still open from prior rounds

| # | File | Dimension | Severity |
|---|------|-----------|----------|
| 1 | `PublicAPI.Shipped.txt:676` | Public API Surface Protection | ⚠️ Breaking |
| 2 | `AssemblyMetadataProvider.cs:217` | Null Safety & Boundary Validation | 🐛 Bug |
| A | `DotnetTestHostManager.cs:1068` | Process Architecture & Host Resolution | 🐛 Silent failure (Linux/macOS) |
| D | `SourceDetailHelper.cs:24` | Null Safety & Boundary Validation | 🐛 Correctness |
| E | `DefaultTestHostManager.cs:73` | Environment Variable & Feature Flag Contracts | 🔧 Maintenance concern |

Inline comments for all five remain on their respective threads from prior rounds.

### PR Description Alignment

The description continues to accurately describe the WIP scope and intentional probes. The new commit's scope (MSBuild pack re-port) is not mentioned in the description, but this is a mechanics/housekeeping change that doesn't warrant updating the feature description. No misalignment. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28031659033)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28031659033)*

nohwnd:
---
> [!WARNING]
> The `push_to_pull_request_branch` operation failed: Pull request title "Run tests as exe (RunAsExe attribute)" does not start with required prefix "[fix] ". The code changes were not applied.

Addressed the three non-outdated expert-reviewer findings:

**Finding 1 — PublicAPI.Shipped.txt breaking change**
Reverted `SourceDetail.Architecture.get` in `PublicAPI.Shipped.txt` to the original non-nullable `Architecture` form (the shipped contract). Added the nullable `Architecture?` variant to `PublicAPI.Unshipped.txt` where it belongs.

**Finding 2 — Namespace check in `AssemblyMetadataProvider.GetRunAsExeFromAssemblyMetadata`**
Extended the attribute matching to also verify the declaring namespace is `Microsoft.VisualStudio.TestPlatform.ObjectModel` (using `Ordinal` comparison), so attributes named `RunAsExeAttribute` in other namespaces are no longer misidentified.

**Finding 4 — `DefaultTestHostManager` latent null**
This is already handled in the current code — `testHostProcessPath` is unconditionally assigned before the inner `if (!_fileHelper.Exists(...))` check. No change needed.

**Finding 5 (non-outdated variant at line 1068) — Cross-platform `.exe` in `TryGetExePathFromDll`**
On Linux/macOS the SDK produces extension-less apphosts, so the method now probes the extension-less form on non-Windows. Also upgraded the fallback trace from `Verbose` to `Warning` level so framework authors can diagnose a missing executable.



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28032807153)*

nohwnd:
There is a high chance of regressing behavior for existing users, instead we look into running mtp enabled exes under vstest, so test frameworks can drop vstest entirely and still stay compatible. in #16201

PR #16130, opened 2026-06-18
[fix] Improve error message when .NET Framework testhost not found on Linux

nohwnd:
## Summary

Fixes #15340.

> 🤖 This is an automated fix created by the Issue Repro Triage agent.

## Root Cause

When running .NET Framework tests on Linux with Mono, vstest resolves `testhost.exe` to `<sdk>/TestHostNetFramework/testhost.exe`. Source-built .NET SDK packages installed via package managers (apt, dnf, rpm) **do not include** the `TestHostNetFramework` directory — only the official `.tar.gz` installer from dot.net ships it.

Previously, vstest would invoke Mono with this non-existent path and report the opaque error:
```
Cannot open assembly '.../TestHostNetFramework/testhost.exe': No such file or directory.
Testhost process for source(s) '...' exited with error: ...
```

This gives users no guidance on why it failed or how to fix it.

## Fix

Added a pre-launch existence check in `DefaultTestHostManager.GetTestHostProcessStartInfo`. On non-Windows platforms (when Mono would be used), if the resolved `testhost.exe` path does not exist, a `TestPlatformException` is thrown with a clear diagnostic message:

```
Could not find the .NET Framework test host at '<path>'.

Source-built .NET SDK packages installed via a Linux package manager (apt, dnf, rpm) may not
include the .NET Framework test host components needed to run .NET Framework tests with Mono.

To resolve this issue, choose one of the following options:
  - Install the complete .NET SDK from the official installer: https://dot.net/download
  - Migrate your tests to target .NET instead of .NET Framework.
```

## Changes

- `DefaultTestHostManager.cs`: Added file existence check before using Mono
- `Resources.resx` + `Resources.Designer.cs`: New `NetFrameworkTestHostNotFoundOnLinux` resource
- All `*.xlf` files: Added new trans-unit entry (pending localization)
- `DefaultTestHostManagerTests.cs`: Updated existing mono test to properly mock file existence; added new test for the missing-testhost case

## Testing

Unit test `GetTestHostProcessStartInfoShouldThrowWhenNetFrameworkTestHostNotFoundOnLinux` verifies that when all file paths return non-existent (simulating a source-built SDK), a `TestPlatformException` is thrown with the expected message fragments.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/27730820860)*

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
@@ -217,12 +217,12 @@ public class DefaultTestHostManager : ITestRuntimeProvider2
             if (!_environment.OperatingSystem.Equals(PlatformOperatingSystem.Windows)
                 && !processName.EndsWith(DotnetHostHelper.MONOEXENAME, StringComparison.OrdinalIgnoreCase))
             {
-                // On non-Windows, we use Mono to run the .NET Framework testhost. Validate that the
-                // testhost executable actually exists before handing control to Mono, so we can emit
+                // On Linux, source-built SDK packages (apt, dnf, rpm) may omit the TestHostNetFramework
+                // directory entirely. Validate existence before handing control to Mono so we can emit
                 // a clear diagnostic instead of an opaque "Cannot open assembly" error from Mono.
-                // Source-built SDK packages (e.g. from apt/dnf/rpm) may omit the TestHostNetFramework
-                // directory entirely, while the official .tar.gz installer from dot.net includes it.
-                if (!_fileHelper.Exists(testhostProcessPath))
+                // The check is scoped to Linux (Unix) only; on macOS the full SDK is typically present.
+                if (_environment.OperatingSystem.Equals(PlatformOperatingSystem.Unix)
+                    && !_fileHelper.Exists(testhostProcessPath))
                 {
                     throw new TestPlatformException(
                         string.Format(CultureInfo.CurrentCulture, Resources.NetFrameworkTestHostNotFoundOnLinux, testhostProcessPath));
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
@@ -217,6 +217,17 @@ public class DefaultTestHostManager : ITestRuntimeProvider2
             if (!_environment.OperatingSystem.Equals(PlatformOperatingSystem.Windows)
                 && !processName.EndsWith(DotnetHostHelper.MONOEXENAME, StringComparison.OrdinalIgnoreCase))
             {
+                // On non-Windows, we use Mono to run the .NET Framework testhost. Validate that the
+                // testhost executable actually exists before handing control to Mono, so we can emit
+                // a clear diagnostic instead of an opaque "Cannot open assembly" error from Mono.
+                // Source-built SDK packages (e.g. from apt/dnf/rpm) may omit the TestHostNetFramework
+                // directory entirely, while the official .tar.gz installer from dot.net includes it.
+                if (!_fileHelper.Exists(testhostProcessPath))
+                {
+                    throw new TestPlatformException(
+                        string.Format(CultureInfo.CurrentCulture, Resources.NetFrameworkTestHostNotFoundOnLinux, testhostProcessPath));
+                }
+
                 launcherPath = _dotnetHostHelper.GetMonoPath();
                 argumentsString = testhostProcessPath.AddDoubleQuote() + " " + argumentsString;
             }
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.Designer.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.Designer.cs
@@ -111,5 +111,16 @@ namespace Microsoft.TestPlatform.TestHostProvider.Resources {
             }
         }
 
+        /// <summary>
+        ///   Looks up a localized string for when the .NET Framework testhost is not found on Linux (e.g. source-built SDK).
+        /// </summary>
+        internal static string NetFrameworkTestHostNotFoundOnLinux
+        {
+            get
+            {
+                return ResourceManager.GetString("NetFrameworkTestHostNotFoundOnLinux", resourceCulture);
+            }
+        }
+
     }
 }
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.Designer.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.Designer.cs
@@ -112,7 +112,7 @@ namespace Microsoft.TestPlatform.TestHostProvider.Resources {
         }
 
         /// <summary>
-        ///   Looks up a localized string for when the .NET Framework testhost is not found on Linux (e.g. source-built SDK).
+        ///   Looks up a localized string similar to Could not find the .NET Framework test host at &apos;{0}&apos;....
         /// </summary>
         internal static string NetFrameworkTestHostNotFoundOnLinux
         {
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.resx
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.resx
@@ -150,11 +150,9 @@ The specified framework can be found at:
   <data name="NetFrameworkTestHostNotFoundOnLinux" xml:space="preserve">
     <value>Could not find the .NET Framework test host at '{0}'.
 
-Source-built .NET SDK packages installed via a Linux package manager (apt, dnf, rpm) may not include the .NET Framework test host components needed to run .NET Framework tests with Mono.
+This .NET SDK does not include the .NET Framework test host needed to run .NET Framework tests with Mono.
 
-To resolve this issue, choose one of the following options:
-  - Install the complete .NET SDK from the official installer: https://dot.net/download
-  - Migrate your tests to target .NET instead of .NET Framework.
+To fix this, migrate your tests to target .NET instead of .NET Framework.
 </value>
     <comment>'{0}' is the full path where testhost.exe was expected to be found.</comment>
   </data>
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.resx
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/Resources.resx
@@ -147,4 +147,15 @@ The specified framework can be found at:
 </value>
     <comment>'{0}' is the placeholder for 'dotnet.exe' or 'dotnet' value and depends on platform Windows/Unix, '{1}' is the placeholder for the architeture name like ARM64, X64 etc...</comment>
   </data>
+  <data name="NetFrameworkTestHostNotFoundOnLinux" xml:space="preserve">
+    <value>Could not find the .NET Framework test host at '{0}'.
+
+Source-built .NET SDK packages installed via a Linux package manager (apt, dnf, rpm) may not include the .NET Framework test host components needed to run .NET Framework tests with Mono.
+
+To resolve this issue, choose one of the following options:
+  - Install the complete .NET SDK from the official installer: https://dot.net/download
+  - Migrate your tests to target .NET instead of .NET Framework.
+</value>
+    <comment>'{0}' is the full path where testhost.exe was expected to be found.</comment>
+  </data>
 </root>
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" original="../Resources.resx" target-language="cs">
     <body>
--- a/src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Resources/xlf/Resources.cs.xlf
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" original="../Resources.resx" target-language="cs">
     <body>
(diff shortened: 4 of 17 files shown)


nohwnd (commented):
## Expert Review — PR #16130

**Dimensions activated**: Process Architecture & Host Resolution · Error Reporting & Diagnostic Clarity · Source Build & Cross-Platform Compliance · Cross-TFM & Framework Resolution

---

### What looks correct ✅

- **Check placement**: The existence check is inserted in exactly the right place — inside the `!Windows && !Mono` block, before `GetMonoPath()` is called. It fires precisely when vstest would delegate to Mono, and not in any other code path (pure .NET on Linux, Windows, or already-under-Mono).
- **Testability**: Uses `IFileHelper.Exists(...)` (injected dependency) rather than a bare `File.Exists(...)`, so the check is fully mockable. The existing test is correctly updated to setup `Returns(true)` for the TestHostNetFramework path.
- **New test**: `GetTestHostProcessStartInfoShouldThrowWhenNetFrameworkTestHostNotFoundOnLinux` correctly uses `Assert.ThrowsExactly<TestPlatformException>` and checks both `"TestHostNetFramework"` (the resolved path fragment) and `"(dot.net/redacted) (the actionable guidance URL). `Assert.Contains(needle, haystack)` order is correct.
- **Localization**: XLF entries added for all 12 target languages (cs, de, es, fr, it, ja, ko, pl, pt-BR, ru, tr, zh-Hans, zh-Hant) with `state="new"` pending translation — correct per vstest convention.

---

### Finding 🟡

**macOS receives a Linux-specific error message**

`PlatformOperatingSystem` has three values: `Windows`, `Unix` (Linux), and `OSX` (macOS). The outer condition `!Equals(Windows)` covers both `Unix` and `OSX`, so the new existence check and throw also fire on macOS. The error message text says *"Linux package manager (apt, dnf, rpm)"* which is inaccurate for macOS users.

See the inline comment on line 228 for details and suggested fixes.

---

### Minor observations (not requiring changes)

- The BOM (`EF BB BF`) was stripped from all 12 XLF files as a side-effect of editing. Removing the BOM from UTF-8 XML files is valid (the encoding is declared in the XML header), but the change is unintentional and touches 12 files beyond what the diff requires.
- `Resources.resx` still lacks a trailing newline (pre-existing, not introduced by this PR).

---

🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions/runs/27731544736)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27731544736)*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
```diff
@@ -217,6 +217,17 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
             {
+                // On non-Windows, we use Mono to run the .NET Framework testhost. Validate that the
+                // testhost executable actually exists before handing control to Mono, so we can emit
+                // a clear diagnostic instead of an opaque "Cannot open assembly" error from Mono.
+                // Source-built SDK packages (e.g. from apt/dnf/rpm) may omit the TestHostNetFramework
+                // directory entirely, while the official .tar.gz installer from dot.net includes it.
+                if (!_fileHelper.Exists(testhostProcessPath))
+                {
+                    throw new TestPlatformException(
+                        string.Format(CultureInfo.CurrentCulture, Resources.NetFrameworkTestHostNotFoundOnLinux, testhostProcessPath));
```
nohwnd, line 228: **[Process Architecture & Host Resolution, Error Reporting & Diagnostic Clarity]**

The enclosing condition at line 217 is `!_environment.OperatingSystem.Equals(PlatformOperatingSystem.Windows)`, which covers **both** `PlatformOperatingSystem.Unix` (Linux) and `PlatformOperatingSystem.OSX` (macOS) — `PlatformOperatingSystem` has three values: `Windows`, `Unix`, and `OSX`.

On macOS, the thrown message says:

> *"Source-built .NET SDK packages installed via a Linux package manager (apt, dnf, rpm)..."*

That advice doesn't apply on macOS (where package managers are Homebrew, MacPorts, or the Mono installer). The resource name `NetFrameworkTestHostNotFoundOnLinux` also encodes the Linux assumption.

Two options to resolve:

1. **Scope the check to Linux only** — replace the outer condition with an explicit `PlatformOperatingSystem.Unix` check if macOS is not a supported target for this diagnostic:
   ```csharp
   if (_environment.OperatingSystem.Equals(PlatformOperatingSystem.Unix)
       && !processName.EndsWith(DotnetHostHelper.MONOEXENAME, StringComparison.OrdinalIgnoreCase))
   ```

2. **Generalise the message** — keep the broader `!Windows` condition but update the message (and resource name) to cover non-Windows platforms, e.g.:
   > *"Source-built or package-manager-installed .NET SDK distributions (apt, dnf, rpm, Homebrew) may not include..."*

Option 1 is lower risk if macOS+Mono+source-built SDK is not a known user scenario. Option 2 is more complete if it is.

nohwnd:
Commit pushed: [`550c0ee`](https://github.com/microsoft/vstest/commit/550c0ee2edc12a9f1e9c64792c13e68e3a6b0d5e)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27732082552)*

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DefaultTestHostManager.cs
nohwnd: Went with **Option 1** — scoped the existence check to `PlatformOperatingSystem.Unix` only (commit `550c0ee`). The outer `!Windows` block still handles macOS+Mono, but the pre-launch diagnostic is now Linux-specific so the message accurately describes the apt/dnf/rpm scenario without misleading macOS users.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27732590061)*

nohwnd (commented):
## Expert Review — PR #16130 (follow-up)

**Dimensions activated**: Process Architecture & Host Resolution · Error Reporting & Diagnostic Clarity · Source Build & Cross-Platform Compliance · Cross-TFM & Framework Resolution

---

### Previous finding resolved ✅

The macOS issue flagged in the prior review (commit `1abad770`) has been fully addressed in `550c0ee`:

- Condition changed from `!OperatingSystem.Equals(Windows)` → `OperatingSystem.Equals(Unix)` — exactly Option 1 from the inline review comment
- Code comment added: `// The check is scoped to Linux (Unix) only; on macOS the full SDK is typically present.`
- New test `GetTestHostProcessStartInfoShouldNotThrowWhenNetFrameworkTestHostNotFoundOnOSX` explicitly verifies macOS is exempt from the throw

The existing inline review comment on line 228 can be marked as resolved.

---

### Code review of the updated diff ✅

- **Check placement**: Still correctly nested inside the `!Windows && !Mono` block; fires exactly before `GetMonoPath()` on Linux only
- **Double-existence-check interaction**: The first `_fileHelper.Exists` (line 195) sets `testhostProcessPath` to the `TestHostNetFramework` subfolder when the root path is absent; the new check (line 224–229) then validates that subfolder path on Linux — ordering is correct
- **Updated test** `GetTestHostProcessStartInfoShouldUseMonoAsHostOnNonWindowsIfNotStartedWithMono`: correctly adds `.Returns(true)` for `TestHostNetFramework`-containing paths so the pre-existing Mono test still passes
- **New Linux test**: `Assert.Contains("TestHostNetFramework", ex.Message)` and `Assert.Contains("(dot.net/redacted), ex.Message)` — both use correct `(needle, haystack)` order
- **New macOS test**: verifies `info.FileName == "/usr/bin/mono"` when all file-helper calls return false on OSX — correctly documents the intentional skip

---

### Pre-existing observations (not introduced by this PR)

- UTF-8 BOM was stripped from all 13 XLF files as an unintentional side-effect of editing — valid XML but noisy in the diff
- `Resources.resx` still lacks a trailing newline (SA1518 pre-existing)

---

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27732590413)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27732590413)*

nohwnd:
https://github.com/microsoft/vstest/issues/15340 I dont' think there is any "full installation" on linux that we should recommend. Linux distros are source built, and don't contain the .net framework testhost, can you simply say that? Just don't mention source build this is a detail. Instead just say that this .NET SDK does not contain .NET Framework testhost, or something along those lines.

nohwnd:
Commit pushed: [`94c3496`](https://github.com/microsoft/vstest/commit/94c34969f7084d41f1b91a894d69fee3cfc75770)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27746119709)*

nohwnd:
Done — simplified the error message per your feedback (commit `c0fae69`):

```
Could not find the .NET Framework test host at '...'.

This .NET SDK does not include the .NET Framework test host needed to run .NET Framework tests with Mono.

To fix this, migrate your tests to target .NET instead of .NET Framework.
```

Removed the "source-built" mention and the installer URL recommendation. Updated all 13 XLF files and the test assertion accordingly.



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27746119709)*

nohwnd (commented):
## Expert Review — PR #16130 (follow-up on commit `94c3496`)

**Dimensions activated**: Error Reporting & Diagnostic Clarity · Process Architecture & Host Resolution · Source Build & Cross-Platform Compliance

---

### Previous findings resolved ✅

All prior review findings (commit `1abad770` and `550c0ee`) are resolved:
- macOS/Linux scope was narrowed to `PlatformOperatingSystem.Unix` only
- macOS exemption test was added

---

### Review of commit `94c3496` (message simplification) ✅

**Changes reviewed**: `Resources.resx`, all 13 XLF files, `DefaultTestHostManagerTests.cs`

The latest commit simplifies the user-facing error message per maintainer feedback (removing the "source-built" detail and installer URL):

```
Could not find the .NET Framework test host at '{0}'.

This .NET SDK does not include the .NET Framework test host needed to run .NET Framework tests with Mono.

To fix this, migrate your tests to target .NET instead of .NET Framework.
```

- **Message accuracy**: The statement "This .NET SDK does not include..." fires only after both the root path and `TestHostNetFramework` subdirectory existence checks fail — the claim is factually correct in that context ✅
- **Placeholder**: `{0}` is still correctly formatted with `CultureInfo.CurrentCulture` ✅
- **XLF consistency**: All 13 XLF files updated with matching `<source>` text and `state="new"` targets — correct per vstest convention ✅
- **Test assertion update**: `Assert.Contains("does not include", ex.Message)` correctly targets the key phrase from the new message; `Assert.Contains("TestHostNetFramework", ex.Message)` remains valid because the `{0}` path contains that directory name ✅
- **`Resources.Designer.cs`**: Does not need updating — only the string value changed, not the resource key (`NetFrameworkTestHostNotFoundOnLinux`) ✅

---

### Minor observation (not requiring changes)

**[Description]** The PR description still shows the old multi-option message (with installer URL and "source-built" language). The current code uses the simplified single-option message. The title and description intent remain accurate — only the quoted message text is stale.

---

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27746964099)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27746964099)*

Evangelink (commented):
## Expert review — automated

Overall: **Approvable after a trivial localization cleanup.** The substantive fix is correct, low-risk, and well-tested. The two natural design questions (macOS scoping, "install full SDK" wording) were already resolved earlier in the thread.

### Core logic — correct
- The new existence check fires on the **final resolved** `testhostProcessPath` (after the root → `TestHostNetFramework/testhost.exe` reassignment), so placement is right.
- **Happy path is safe**: if either the root or `TestHostNetFramework` testhost exists, no throw occurs. The exception fires only when the file is genuinely absent — the exact #15340 repro.
- Uses the mockable `_fileHelper.Exists` (consistent with the earlier resolution step), keeping it testable and platform-correct.
- `TestPlatformException` is the right type — clean user-facing error, no internal stack trace leaked.

### OS scoping — correct
Outer guard is `!Windows`, but the new check is narrowed to `PlatformOperatingSystem.Unix`. macOS resolves to `OSX` (not `Unix`), so the Linux-specific apt/dnf/rpm message never shows on Mac and macOS+Mono still launches. Matches the in-thread fix.

### Message content — correct
The PR description mentions "install the full SDK", but that was explicitly rejected in-thread (Linux distros are source-built and don''t ship the .NET Framework testhost). The final message only recommends migrating to .NET — intentional; the description is just stale.

### Tests — good
The new throw test and the OSX negative test exercise real paths with non-tautological assertions; the updated existing Mono test correctly adds `Exists(...TestHostNetFramework...) -> true` so the new check doesn''t fire, and still validates Mono argument construction. Fresh per-test mocks, no shared state.

### Findings
- **MODERATE — UTF-8 BOM stripped from all 14 `*.xlf` files.** Every xlf hunk drops the leading BOM (`-<?xml...` with BOM -> `+<?xml...` without). Other repo xlf files keep the BOM, so this is convention churn across 14 files. Low functional risk (CI green). **Suggested fix:** don''t hand-edit the xlf — edit `Resources.resx` only and let XliffTasks regenerate them with the correct BOM and `state="new"`.
- **NIT** — `Resources.Designer.cs` was hand-edited; the new accessor''s doc-comment deviates from the tool-generated `Looks up a localized string similar to {value}.` convention. Regenerating from the resx would keep it consistent.
- **NIT** — the resx `<value>` ends with a trailing newline, so the rendered error ends with an extra blank line.

**Recommendation: Comment** — please regenerate the xlf via the build before merge (non-blocking). No correctness, threading, security, or compatibility concerns.

nohwnd:
Commit pushed: [`3a55c51`](https://github.com/microsoft/vstest/commit/3a55c516fc6aa0c9ff78985e42eec78cdceaafad)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27818108955)*

nohwnd:
Addressed review feedback in commit `d583440`:

- **BOM restored**: Added UTF-8 BOM (`EF BB BF`) back to all 13 XLF files in `Microsoft.TestPlatform.TestHostProvider/Resources/xlf/`. The BOM was accidentally stripped during hand-editing; this restores the convention used by all other XLF files in the repo.
- **Designer.cs doc comment updated**: Changed the `NetFrameworkTestHostNotFoundOnLinux` accessor's doc comment from the custom description to the tool-generated `Looks up a localized string similar to Could not find the .NET Framework test host at '{0}'....` convention.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27818108955)*



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/27818108955)*

nohwnd (commented):
## Expert Review — PR #16130 (follow-up on commit `3a55c516`)

**Dimensions activated**: Error Reporting & Diagnostic Clarity · Source Build & Cross-Platform Compliance · Build Script & Infrastructure Hygiene

---

### Evangelink feedback addressed ✅

Commit `3a55c516` (pushed 2026-06-19T09:51Z) resolves both items flagged in the preceding review:

- **UTF-8 BOM restored** to all 13 XLF files in `TestHostProvider/Resources/xlf/` — each file now opens with `EF BB BF`, matching the convention of every other XLF file in the repo ✅
- **`Resources.Designer.cs` doc comment** updated to the tool-generated `Looks up a localized string similar to...` convention ✅

---

### Trailing-`\n` in resource value — not a defect

The `NetFrameworkTestHostNotFoundOnLinux` `<value>` ends with `\n` before `</value>`. This is consistent with the existing patterns for `CouldNotFindTesthost` and `NoDotnetMuxerFoundForArchitecture` in the same file — both also end with `\n`. The convention is intentional for multi-line messages and there is nothing to fix here.

---

### All prior findings resolved ✅

| Commit | Finding | Status |
|--------|---------|--------|
| `1abad770` | macOS received Linux-specific error message | Resolved in `550c0ee` |
| `94c3496` | PR description shows stale error message text | Informational (description intent accurate) |
| `3a55c516` | BOM stripped from 13 XLF files; Designer.cs comment deviation | Resolved in `3a55c516` |

No correctness, threading, IPC compatibility, or public API concerns. The fix is targeted, well-tested, and localized correctly.

---

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27818563473)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27818563473)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/27818563473)*

nohwnd:
Closing in favor of #16158. Rather than just improving the error message, that one drops the Mono fallback entirely and runs .NET Framework tests on Windows only - the breaking change discussed in #15340.

PR #16132, opened 2026-06-18
Migrate nuspec to MSBuild pack (batch 2: container packages)

### LINKED ISSUE #15650: Consider moving away from nuspec

azat-msft:
## Summary

Continues the migration from hand-written `.nuspec` files to MSBuild pack (csproj-based packaging), as requested in #15650. This is the follow-up to #16125 (batch 1, already merged), and migrates the remaining **container/aggregate packages** that bundle outputs from many other projects.

Each package was converted by translating every `<file src="..." target="..."/>` element in the nuspec into an equivalent `<None Pack="true" PackagePath="..."/>` item that maps the same build output to the same package path. The produced packages were verified to be **byte-for-byte identical in file layout** to the pre-migration packages.

## Migrated packages (this PR)

| Package | Files | Notes |
|---|---|---|
| **TranslationLayer** | 175 | Bundles Common + CommunicationUtilities DLLs + CoreUtilities satellites; ObjectModel remains a real package dependency |
| **Internal.Uwp** | 39 | Bundles 7 DLLs + ObjectModel/CoreUtilities satellites; no dependencies |
| **NET.Test.Sdk** | 26 | Multi-targets `net8.0;net462` for dependency groups; ships `.props`/`.targets` |
| **CodeCoverage** | 81 | Bundles the external `Microsoft.Internal.CodeCoverage` content tree via `TargetsForTfmSpecificContentInPackage` + `%(RecursiveDir)` |
| **TestHost** | 65 | Bundles testhost runtime output; removes `testhost`/`testhost.x86` exe project refs from the dependency list |
| **TestPlatform** | 553 | Full platform container (tools/net462 + tools/netcoreapp layouts); 487 file elements translated, incl. two recursive `**` globs |
| **Portable** | 608 | Cross-platform variant of TestPlatform; 601 file elements translated |

> Note: TranslationLayer/Internal.Uwp/NET.Test.Sdk/CodeCoverage/TestHost were prepared earlier; TestPlatform and Portable are new in this round. All seven are squashed onto the merged batch-1 base so the diff shows only unmerged work.

## Not migrated: `Microsoft.TestPlatform.CLI` (intentionally left on nuspec)

CLI was evaluated and deliberately kept on `.nuspec`. It is materially riskier than the others:

1. **Unverifiable variants.** CLI selects one of three nuspecs by condition — `*.nuspec` (Windows), `*.sourcebuild.nuspec` (non-Windows source-build), and `*.sourcebuild.product.nuspec` (VMR). The source-build and VMR variants only execute under `-DotNetBuildSourceOnly`/VMR builds and **cannot be exercised or validated by any local Windows build**.
2. **`<contentFiles>` copy semantics.** CLI's nuspec declares a generic `<contentFiles><files include="**/*.*" copyToOutput="true" flatten="false" buildAction="None"/></contentFiles>` rule that governs how consumers copy content into their output. MSBuild pack emits per-file `contentFiles` entries instead, and a behavioral regression here would **not** be caught by a package file-list comparison.

Migrating CLI safely requires a way to validate the source-build/VMR paths, so it is left for a dedicated follow-up.

## How this was validated

For every migrated package, a clean `build.cmd -c Release -pack` was run and the produced `.nupkg` file list was compared against the pre-migration package. **All seven match exactly** (ignoring the random `package/services/metadata/core-properties/*.psmdcp` entry name). The generated `.nuspec` inside each package was also inspected to confirm dependency groups / `frameworkAssemblies` metadata is preserved (the container packages have no `<dependencies>`). The full Release build's nupkg-file-count and DLL-target-framework verification pipeline passes with no `eng/expected-*.json` drift.

## Notable mechanics

- Container packages set `IncludeBuildOutput=false` (they ship no assembly of their own) and `SuppressDependenciesWhenPacking=true` (no NuGet dependencies).
- File-license packages (TestPlatform, Portable) clear the inherited MIT `PackageLicenseExpression` so the `LICENSE_VS.txt` file license applies, and add an explicit `None` item for the license file (it is not auto-packed).
- Bundling external content uses `TargetsForTfmSpecificContentInPackage` (runs in the inner per-TFM build, where `$(PkgXxx)` `GeneratePathProperty` paths and `$(OutputPath)` resolve).
- Pack auto-adds benign `<frameworkAssemblies>` entries the old nuspecs omitted; these do not affect the packaged files.

Closes #15650 (pending CLI follow-up).


azat-msft:
Fixed the `Build Windows Release` failure (nupkg file-count verification: `Microsoft.TestPlatform` expected 553, actual 492).

**Root cause:** the migrated `Microsoft.TestPlatform.csproj` translated the two recursive CodeCoverage nuspec globs into static `<None Include="...**\*">` items. Static MSBuild items are expanded at **evaluation time**, before the `CopyFiles` target stages those 61 files into the `net48` output, so on a clean CI build they matched nothing (553 − 492 = 61 missing files). My local build passed falsely because the files were left over from previous builds.

**Fix:** the two CodeCoverage trees are now added at **pack time** via a `TargetsForTfmSpecificContentInPackage` target, sourced directly from the immutable restored `Microsoft.Internal.CodeCoverage` package (`contentFiles/MS.TP/...`) rather than the build output — the same mechanism the `Microsoft.CodeCoverage` package migration uses. Each tree is globbed into its own temporary item so the `%(RecursiveDir)` metadata batches per-tree (combining two `%(RecursiveDir)` globs in one `TfmSpecificPackageFile` group otherwise cross-products every file against every directory).

**Validation:** verified by deleting the entire `net48` CodeCoverage output and doing a full clean rebuild — the package is a byte-identical 553-file match against the pre-migration package, and `build.cmd -c Release -pack` (which runs `verify-nupkgs.ps1`) passes with no `eng/` JSON drift.

Evangelink (approved):

PR #16149, opened 2026-06-22
Migrate Microsoft.TestPlatform.CLI from nuspec to MSBuild pack (final batch)

### LINKED ISSUE #15650: Consider moving away from nuspec

azat-msft:
## Summary

Completes the migration from hand-written `.nuspec` files to MSBuild pack (csproj-based
packaging), per #15650. This is the **final batch**: it migrates the last remaining package,
**`Microsoft.TestPlatform.CLI`**, which was explicitly deferred by the earlier batches
(#16125 simple packages, #16132 container packages) because it was the riskiest.

> Note: this branch stacks on top of #16132, so the diff also contains the batch-2 container
> packages (TranslationLayer, Internal.Uwp, NET.Test.Sdk, CodeCoverage, TestHost, TestPlatform,
> Portable). The **new** work in this PR is the `Microsoft.TestPlatform.CLI` migration. If #16132
> is merged first, only the CLI commit remains here.

## What changed for `Microsoft.TestPlatform.CLI`

The project previously selected one of three hand-written nuspecs via `<NuspecFile>` conditions:

| Variant | When |
|---|---|
| `Microsoft.TestPlatform.CLI.nuspec` | Windows, non source-build (full `net10.0` layout + .NET Framework test hosts) |
| `Microsoft.TestPlatform.CLI.sourcebuild.nuspec` | non-Windows / source-build, non-VMR (cross-platform `$(NetCurrent)` layout) |
| `Microsoft.TestPlatform.CLI.sourcebuild.product.nuspec` | source-build, VMR (identical to the sourcebuild variant) |

All three are replaced by csproj packaging:

- Every `<file src=".." target=".." />` element becomes a `<None Pack="true" PackagePath="..">` item.
- Because the package ships everything under `contentFiles`, each content item carries
  `BuildAction=None` / `PackageCopyToOutput=true` / `PackageFlatten=false`, so pack emits per-file
  `<contentFiles>` entries that are **functionally equivalent** to the original generic
  `<contentFiles><files include="**/*.*" copyToOutput="true" flatten="false" buildAction="None" /></contentFiles>`
  wildcard.
- The **Windows full content** is gated on `'$(DotNetBuildSourceOnly)' != 'true' and '$(OS)' == 'Windows_NT'`;
  the cross-platform **`$(NetCurrent)` content** (covering both identical sourcebuild nuspecs) is gated on
  `'$(DotNetBuildSourceOnly)' == 'true' or '$(OS)' != 'Windows_NT'`.
- Container settings `IncludeBuildOutput=false` and `SuppressDependenciesWhenPacking=true` are used
  (the package ships no assembly of its own and has no NuGet dependencies).
- The existing `ProjectReference`, `PackageReference` and `CopyFiles` staging logic is unchanged.

The three `.nuspec` files are deleted.

## Validation

A clean `build.cmd -c Release -pack` produces a CLI package whose file list is **byte-for-byte
identical** to the pre-migration package (483 entries), with **no `eng/expected-*.json` drift**.
The generated nuspec metadata matches the previous one; the only differences are the per-file
`contentFiles` entries (vs. the single wildcard, functionally equivalent) and a benign auto-added
`<frameworkAssemblies>` entry that is not part of the file-list oracle. The full Release build's
nupkg-file-count and DLL-target-framework verification pipeline passes.

## Known limitation

The `sourcebuild` / `sourcebuild.product` paths only execute under `-DotNetBuildSourceOnly` / VMR
builds and **cannot be exercised by a local Windows build**. They are translated faithfully but
validated by inspection only — worth confirming on a source-build / VMR CI leg before merge.

Closes #15650 (CLI follow-up).


src/vstest.console/Program.cs
```diff
@@ -23,6 +23,7 @@ public static class Program
 
     internal static int Run(string[]? args, UiLanguageOverride uiLanguageOverride)
     {
+        // Console output is switched to UTF-8 unless the feature flag opts out.
```
nohwnd, line 26:
```suggestion
```
Unrelated.

nohwnd (approved):

PR #16152, opened 2026-06-22
Document procdump.exe requirement in --Blame help text (#2900)

### LINKED ISSUE #2900: Dependency on procdump.exe not documented (and optional env variable PROCDUMP_PATH)
I spent a lot of time trying to get crash dumps working when running vstest.console.exe, that could have been avoided by documenting the usage of procdump.exe.

The vstext.console.exe command line for the blame switch simply says: 

--Blame|/Blame:[CollectDump];[CollectAlways]=[Value];[DumpType]=[Value]
      Runs the test in blame mode. This option is helpful in isolating the problematic test causing test host crash.
      It creates an output file in the current directory as "Sequence.xml",
      that captures the order of execution of test before the crash.
      You may optionally choose to collect process dump for the test host.
      When you choose to collect dump, by default, a mini dump will be collected on a crash.
      You may also choose to override this default behaviour by some optional parameters:
      CollectAlways - To collect dump on exit even if there is no crash (true/false)
      DumpType - To specify dump type (mini/full).
      Example: /Blame
               /Blame:CollectDump
               /Blame:CollectDump;CollectAlways=true;DumpType=full

Why does it not also say that you need to download procdump.exe and have its path defined by PATH or PROCDUMP_PATH env variables? Only after searching user forums and following links around I got back to this page that mention procdump: https://github.com/microsoft/vstest-docs/blob/b9f8340f850b9e03a3fab6537cc47fade213ac42/RFCs/0028-BlameCollector-Hang-Detection.md.

Please consider updating the console help to mention procdump.exe.

Thanks,

Darren

azat-msft:
## Summary

Fixes #2900.

The `vstest.console.exe` `--Blame` help text describes crash-dump collection (`CollectDump`, `CollectAlways`, `DumpType`) but never mentions that, on Windows, crash dumps require **procdump.exe / procdump64.exe** to be discoverable via `PATH` or the `PROCDUMP_PATH` environment variable. As the issue reporter notes, this omission cost real debugging time, and the equivalent `dotnet test` help is far clearer.

This PR brings the `vstest.console` help in line with the `dotnet test` help by documenting the procdump requirement and linking to the download.

## Change

Updated the `EnableBlameUsage` resource string in `src/vstest.console/Resources/Resources.resx` to add:

> Collecting a crash dump on Windows requires procdump.exe and procdump64.exe to be available in PATH, or in a directory pointed to by the PROCDUMP_PATH environment variable. The tools can be downloaded from https://docs.microsoft.com/sysinternals/downloads/procdump.

All 13 `.xlf` localization files were regenerated via the XliffTasks `UpdateXlf` target (not hand-edited): each received the updated `<source>` and had its `<target>` flagged `needs-review-translation`, per the standard convention.

## Help text before/after

**Before:**
```
      ...
      CollectAlways - To collect dump on exit even if there is no crash (true/false)
      DumpType - To specify dump type (mini/full).
      Example: /Blame
```

**After:**
```
      ...
      CollectAlways - To collect dump on exit even if there is no crash (true/false)
      DumpType - To specify dump type (mini/full).
      Collecting a crash dump on Windows requires procdump.exe and procdump64.exe to be
      available in PATH, or in a directory pointed to by the PROCDUMP_PATH environment
      variable. The tools can be downloaded from https://docs.microsoft.com/sysinternals/downloads/procdump.
      Example: /Blame
```

## Validation

- `dotnet build src/vstest.console` succeeds with **0 errors** (only pre-existing IL-trim warnings), confirming the resx ↔ xlf consistency check passes.

## Notes / out of scope

In the issue thread it was also noted that the possible `--blame-crash-dump-type` values (`mini`/`full`) aren't enumerated elsewhere. This PR keeps scope tight to the documented procdump request; happy to extend the help text further if reviewers prefer.

### DIFF
--- a/src/vstest.console/Resources/Resources.resx
+++ b/src/vstest.console/Resources/Resources.resx
@@ -696,6 +696,7 @@
       You may also choose to override this default behaviour by some optional parameters:
       CollectAlways - To collect dump on exit even if there is no crash (true/false) 
       DumpType - To specify dump type (mini/full).
+      Collecting a crash dump on Windows requires procdump.exe and procdump64.exe to be available in PATH, or in a directory pointed to by the PROCDUMP_PATH environment variable. The tools can be downloaded from https://docs.microsoft.com/sysinternals/downloads/procdump.
       Example: /Blame
                /Blame:CollectDump
                /Blame:CollectDump;CollectAlways=true;DumpType=full</value>
--- a/src/vstest.console/Resources/xlf/Resources.cs.xlf
+++ b/src/vstest.console/Resources/xlf/Resources.cs.xlf
@@ -864,10 +864,11 @@
       You may also choose to override this default behaviour by some optional parameters:
       CollectAlways - To collect dump on exit even if there is no crash (true/false) 
       DumpType - To specify dump type (mini/full).
+      Collecting a crash dump on Windows requires procdump.exe and procdump64.exe to be available in PATH, or in a directory pointed to by the PROCDUMP_PATH environment variable. The tools can be downloaded from https://docs.microsoft.com/sysinternals/downloads/procdump.
       Example: /Blame
                /Blame:CollectDump
                /Blame:CollectDump;CollectAlways=true;DumpType=full</source>
-        <target state="translated">--Blame|/Blame:[CollectDump];[CollectAlways]=[Value];[DumpType]=[Value]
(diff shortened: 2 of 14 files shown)


nohwnd (approved):
yes, also consider adding extended documentation to learn.microsoft.com

Issue #15235, opened 2025-07-25
x86 builds no longer work from 17.14.0


Foitn:
In my project we sadly still need to use x86, so my Directory.Build.props looks something like this:
```xml
    <PropertyGroup>
        <TargetFramework>net8.0-windows10.0.22621.0</TargetFramework>
        <Configurations>Debug;Release</Configurations>
        <PlatformTarget>x86</PlatformTarget>
        <RuntimeIdentifiers>win-x86;win-x64</RuntimeIdentifiers>
        ...
    </PropertyGroup>
```

This means that the unit tests will also be compiled for x86 and need to be run from an x86 environment.
When I tried to upgrade from 17.13.0 to 17.14.0 I got the following error:
```
Failed to load the dll from [C:\Program Files\dotnet\host\fxr\9.0.7\hostfxr.dll], HRESULT: 0x800700C1
The library hostfxr.dll was found, but loading it from C:\Program Files\dotnet\host\fxr\9.0.7\hostfxr.dll failed
  - Installing .NET prerequisites might help resolve this problem.
     https://go.microsoft.com/fwlink/?linkid=798306
Testhost process for source(s) 'Project.dll' exited with error: Failed to load the dll from [C:\Program Files\dotnet\host\fxr\9.0.7\hostfxr.dll], HRESULT: 0x800700C1
The library hostfxr.dll was found, but loading it from C:\Program Files\dotnet\host\fxr\9.0.7\hostfxr.dll failed
  - Installing .NET prerequisites might help resolve this problem.
     https://go.microsoft.com/fwlink/?linkid=798306
. Please check the diagnostic logs for more information.
  Finapres.Novascope.Model.Tests.Unit test failed with 1 error(s) (0.5s)
    Project : error TESTRUNABORT: Test Run Aborted.

Build failed with 1 error(s) in 0.9s
```

When I set the version back to 17.13.0 everything works fine.
Looks like support for x86 has been dropped?
I cannot find that in the release notes for 17.14.0

Youssef1313:
Duplicate of https://github.com/dotnet/sdk/issues/49436

PR #16156, opened 2026-06-23
Fix x86 testhost loading mismatched x64 hostfxr (0x800700C1) when run via vstest.console.exe directly (#16151)

### LINKED ISSUE #16151: x86 testhost fails to load runtime (hostfxr 0x800700C1) in 18.9.0-preview via vstest.console.exe — regression of #15235 / dotnet/sdk#49436
## Summary

An **x86** test project still fails to launch its testhost with `hostfxr` error **`0x800700C1`** (`ERROR_BAD_EXE_FORMAT`) when using **`Microsoft.TestPlatform.TestHost` / `Microsoft.NET.Test.Sdk` `18.9.0-preview-26319-02`**, run via **`vstest.console.exe`**.

The x86 `testhost.x86.exe` resolves the runtime from a **64-bit** .NET install (the one `DOTNET_ROOT` points at) and then tries to load the **x64** `hostfxr.dll` into the x86 process, which fails with `0x800700C1` (bad image format / architecture mismatch).

This looks like the same regression as the now-closed #15235, which was closed as a duplicate of **dotnet/sdk#49436**. That SDK issue is marked **completed** (closed 2025-10-27), but we are still hitting it with the `18.9.0-preview-26319-02` package and via the **`vstest.console.exe`** path (not only `dotnet test --arch x86`). So either the fix has not flowed into this preview of the TestPlatform packages, or the console-host x86 runtime resolution path is not covered by the fix.

## Versions

- `Microsoft.TestPlatform.TestHost` = `18.9.0-preview-26319-02`
- `Microsoft.NET.Test.Sdk` = `18.9.0-preview-26319-02`
- Last known-good for the same project: `Microsoft.TestPlatform.TestHost` `17.6.1` / `Microsoft.NET.Test.Sdk` `17.2.0`
- Runner: `vstest.console.exe`
- OS: Windows x64; `DOTNET_ROOT` points to an **x64** .NET install (no separate x86 runtime on `DOTNET_ROOT(x86)`)

## Affected project shape

```xml
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <PlatformTarget>x86</PlatformTarget>
    <IsTestProject>true</IsTestProject>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" />
    <PackageReference Include="xunit" />
  </ItemGroup>
</Project>
```

The test assembly is x86 (`PlatformTarget` x86) on a net8.0 TFM, executed through `vstest.console.exe`.

## Error

```
Testhost process for source(s) '<path>\Tests.dll' exited with error: Failed to load the dll from [<dotnet-x64-root>\host\fxr\10.0.8\hostfxr.dll], HRESULT: 0x800700C1
The library hostfxr.dll was found, but loading it from <dotnet-x64-root>\host\fxr\10.0.8\hostfxr.dll failed
  - Installing .NET prerequisites might help resolve this problem.
     https://go.microsoft.com/fwlink/?linkid=798306
Failed to resolve hostfxr.dll [<dotnet-x64-root>\host\fxr\10.0.8\hostfxr.dll]. Error code: 0x80008082
. Please check the diagnostic logs for more information.
Test Run Aborted.
```

Note the x86 testhost rolls forward to a .NET 10 runtime and resolves the **x64** `hostfxr.dll` (the only one under `DOTNET_ROOT`), then fails to load it into the x86 process.

## Expected

The x86 testhost should resolve and load an **x86** runtime/`hostfxr.dll` (e.g. via `DOTNET_ROOT(x86)`, registry, or default x86 install location), or fail with a clear, actionable diagnostic — not attempt to load the x64 `hostfxr.dll` into an x86 process.

## Notes / ask

- Reverting the test SDK/host to `17.x` makes x86 execution work again, which is how we are unblocking our build for now.
- Please confirm whether the dotnet/sdk#49436 fix is included in `18.9.0-preview-26319-02`, and whether the **`vstest.console.exe`** (not just `dotnet test`) x86 host-resolution path is covered. If not, this is worth investigating as a continuation/incomplete-fix of #15235 / dotnet/sdk#49436.

### Related
- #15235 (closed as duplicate)
- dotnet/sdk#49436 (closed as completed 2025-10-27)

azat-msft:
Fixes #16151

## Summary

An **x86** test project run through **`vstest.console.exe` directly** (i.e. not via `dotnet test`) fails to launch its testhost with:

```
Failed to load the dll from [<dotnet-x64-root>\host\fxr\<ver>\hostfxr.dll], HRESULT: 0x800700C1
```

`0x800700C1` is `ERROR_BAD_EXE_FORMAT`: the 32-bit `testhost.x86.exe` resolved the runtime from a **64-bit** .NET install and tried to load the **x64** `hostfxr.dll` into the 32-bit process.

## Root cause

`DotnetTestHostManager.GetTestHostProcessStartInfo` already sets an architecture-specific `DOTNET_ROOT_<ARCH>` for the testhost, but only when the **.NET SDK** provides `VSTEST_DOTNET_ROOT_PATH` / `VSTEST_DOTNET_ROOT_ARCHITECTURE`. Those variables are set by the SDK's `dotnet test` layer ([`VSTestForwardingApp`](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/Commands/Test/VSTest/VSTestForwardingApp.cs), which forwards `Path.GetDirectoryName(new Muxer().MuxerPath)` and `RuntimeInformation.ProcessArchitecture`).

When `vstest.console.exe` is invoked **directly**, those variables are never set, so the branch is skipped. The architecture-specific apphost (`testhost.x86.exe`) then inherits an **architecture-less** `DOTNET_ROOT` from the surrounding environment. An architecture-less `DOTNET_ROOT` is applied by the apphost regardless of its own architecture, so when it points at an x64 install the x86 apphost loads the x64 `hostfxr.dll` and fails with `0x800700C1`.

This is why the SDK-side fix (dotnet/sdk#49436 / #15235) did not fully resolve the problem: it lives in the `dotnet test` layer, which the direct `vstest.console.exe` path bypasses.

## Fix

Mirror what the SDK does, but compute it inside vstest for the direct path. When an architecture-specific apphost is launched and the SDK did **not** provide `VSTEST_DOTNET_ROOT_PATH`:

1. Look at the location the architecture-less `DOTNET_ROOT` points at.
2. Determine the architecture of that dotnet installation by reading its muxer's PE header.
3. **Promote** the architecture-less `DOTNET_ROOT` to the matching `DOTNET_ROOT_<ARCH>` (unless the user already set that architecture-specific variable — then it is respected).
4. **Clear** the architecture-less `DOTNET_ROOT` for the testhost.

After this, the architecture-less `DOTNET_ROOT` no longer poisons a different-architecture apphost: the matching-architecture apphost finds `DOTNET_ROOT_<ARCH>`, while a different-architecture apphost (e.g. `testhost.x86.exe` when `DOTNET_ROOT` was x64) falls back to its own default resolution (registry / default install location) and loads the correct runtime. The promotion preserves private/non-default installs for the matching architecture.

If `DOTNET_ROOT` is not set, or the architecture cannot be determined, nothing is changed (no behavior change).

> Note: this replaces an earlier revision of this PR that resolved a *separate* architecture-specific install; the current approach is more faithful to the SDK behavior and does not depend on locating another install.

## Validation

**Unit tests** (`DotnetTestHostManagerTests`): added coverage for the new branch — promotion of architecture-less `DOTNET_ROOT` to `DOTNET_ROOT_<ARCH>` + clearing of `DOTNET_ROOT`; no-op when there is no ambient `DOTNET_ROOT`; existing architecture-specific value respected while still clearing the architecture-less one. Full suite: **110/110 passing**.

**End-to-end A/B test** reproducing #16151 (x86 net8 xunit project, `Microsoft.NET.Test.Sdk 18.9.0-preview-26319-02`, run via that preview's `vstest.console.exe`, with `DOTNET_ROOT` pointing at an x64 install and no architecture-specific root):

| Build of `Microsoft.TestPlatform.TestHostRuntimeProvider.dll` | x86 testhost | x64 testhost |
|---|---|---|
| Original (preview) | `0x800700C1` — Test Run Aborted | Successful |
| Patched (this PR) | **Test Run Successful** | Successful (unaffected) |

Swapping only the patched provider DLL flips the x86 outcome, confirming the fix is responsible and complete, with no regression to the x64 path.


nohwnd:
I think this needs a more generic fix. One that would check if dotnet_root folder exists, and has dotnet / dotnet.exe , and what architecture it is. And then sets that as architecture specific env variable, and unsets dotnet_root. 

In some cases we get this info from dotnet test, via vstest_dotnet_root_path and vstest_dotnet_root_architecture, but I am not sure if we unset dotnet_root in that case, so it can still poison x86 or other testhost with path to the incorrect architecture.

azat-msft:
Updated per review feedback. Reworked to mirror the SDK's `VSTestForwardingApp` model: instead of resolving a separate architecture-specific install, vstest now inspects the install that `DOTNET_ROOT` points at, detects its architecture from the muxer's PE header, promotes the architecture-less `DOTNET_ROOT` to `DOTNET_ROOT_<arch>`, and clears the ambiguous architecture-less `DOTNET_ROOT` so a mismatched-architecture apphost no longer picks it up. Re-verified end-to-end (x86 passes, x64 unaffected) and 110/110 unit tests pass.

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
nohwnd, line 592: this is the `if` part of the `else` where code is being added. Feels to me like the problem is already solved in here. But the code is not entered because `dotnetRootPath` is empty. So rather than duplicating the logic in the else, it would be probably better to extend the code that populates   `dotnetRootPath`  and  `dotnetRootArchitecture` to load them from DOTNET_ROOT and inferring the architecture from the dotnet.exe in that path.

( I am not sure if dotnet_root is removed by dotnet/sdk part of our code, which populates  VSTEST_DOTNET_ROOT_PATH  https://grep.app/search?q=VSTEST_DOTNET_ROOT_PATH , but if not we need to also remove it in that part )

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -243,15 +243,50 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+                if (ambientArchitecture is not null)
+                {
+                    dotnetRootPath = dotnetRoot;
+                    dotnetRootArchitecture = ambientArchitecture.ToString();
+
+                    EqtTrace.Verbose($"DotnetTestHostmanager.GetTestHostProcessStartInfo: Derived dotnet root '{dotnetRootPath}' (architecture '{dotnetRootArchitecture}') from the architecture-less DOTNET_ROOT.");
+
+                    startInfo.EnvironmentVariables["DOTNET_ROOT"] = string.Empty;
+                    EqtTrace.Verbose($"DotnetTestHostmanager.GetTestHostProcessStartInfo: clearing the ambiguous DOTNET_ROOT to avoid an architecture mismatch.");
+                }
```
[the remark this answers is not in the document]
azat-msft: There are 2 possible workflows if DOTNET_ROOT matches the architecture of a testhost:
1) It's not legacy and therefore understands architecture-specific env vars, such as DOTNET_ROOT_X86, and this is set on row 296
2) It is legacy and only understands DOTNET_ROOT and DOTNET_ROOT(x86) env vars. This is handled on row 655 by re-inserting DOTNET_ROOT

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -243,15 +243,50 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+        {
+            var dotnetRoot = TryGetTestHostEnvironmentVariable(startInfo, "DOTNET_ROOT", out var callerProvidedDotnetRoot)
+                ? callerProvidedDotnetRoot
+                : _environmentVariableHelper.GetEnvironmentVariable("DOTNET_ROOT");
+
+            if (!StringUtilities.IsNullOrWhiteSpace(dotnetRoot))
+            {
+                // The architecture specific apphosts affected by this ambiguity only exist on Windows, and the PE based
+                // architecture probe is Windows only, so GetExecutableArchitecture returns null elsewhere and we skip.
+                var ambientArchitecture = GetExecutableArchitecture(Path.Combine(dotnetRoot, "dotnet.exe"));
```
nohwnd: ambient here suggests we take it from the environment, why not simply dotnetExeArchitecture?

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -243,15 +243,50 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
+                    EqtTrace.Verbose($"DotnetTestHostmanager.GetTestHostProcessStartInfo: clearing the ambiguous DOTNET_ROOT to avoid an architecture mismatch.");
+                }
+            }
+        }
+
         if (!StringUtilities.IsNullOrWhiteSpace(dotnetRootPath))
         {
             if (StringUtils.IsNullOrWhiteSpace(dotnetRootArchitecture))
             {
                 throw new InvalidOperationException("'VSTEST_DOTNET_ROOT_PATH' and 'VSTEST_DOTNET_ROOT_ARCHITECTURE' must be both always set. If you are seeing this error, this is a bug in dotnet SDK that sets those variables.");
```
nohwnd: this error is now out of sync as well.

nohwnd: imho would be better to mention both of the sources of the date, so what we have now + "or DOTNET_ROOT, and autodetected arechitecture from dotnet.exe in that path."

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -243,15 +243,50 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
             }
 
-            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: VSTEST_DOTNET_ROOT_PATH={dotnetRootPath}");
-            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: VSTEST_DOTNET_ROOT_ARCHITECTURE={dotnetRootArchitecture}");
+            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: dotnet root path={dotnetRootPath}");
+            EqtTrace.Verbose($"DotnetTestHostmanager.LaunchTestHostAsync: dotnet root architecture={dotnetRootArchitecture}");
```
nohwnd: imho would be better to mention both of the sources of the date, so what we have now + "or DOTNET_ROOT, and autodetected arechitecture from dotnet.exe in that path."

src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
```diff
@@ -614,10 +649,10 @@ public virtual TestProcessStartInfo GetTestHostProcessStartInfo(
...
                         {
                             const string dotnetRoot = "DOTNET_ROOT";
-                            if (StringUtils.IsNullOrWhiteSpace(_environmentVariableHelper.GetEnvironmentVariable(dotnetRoot)))
-                            {
-                                startInfo.EnvironmentVariables.Add(dotnetRoot, dotnetRootPath);
-                            }
+                            // If the architecture is not x86 and the testhost does not understand the architecture
+                            // specific DOTNET_ROOT_<ARCH>, we have to re-insert the more generic DOTNET_ROOT. Use the
+                            // indexer (not Add) because the architecture-less DOTNET_ROOT was cleared to empty above.
+                            startInfo.EnvironmentVariables[dotnetRoot] = dotnetRootPath;
```
nohwnd: the const here is used just once, just use the string directly.

nohwnd (approved):

PR #16053, opened 2026-05-22
[fix] Improve error message when testhost cannot be found

nohwnd:
## Summary

🤖 This is an automated fix generated by the Issue Repro Triage agent.

Fixes #15112

## Root Cause

When `DotnetTestHostManager` could not find `testhost.dll` anywhere on disk, it threw a `TestPlatformException` with the message `"Could not find testhost"`. This message provided no context about:
- Which test source was being processed
- Which directories were searched
- What steps the user could take to resolve it

This was reported when a NuGet package setup prevented `testhost.exe` from being copied to the output bin folder. The vague message made self-diagnosis impossible.

## Fix

Replace the generic message with one that includes:
1. **The test source path** — so the user knows which project triggered the failure
2. **The output directory** — so the user can verify what is/isn't there
3. **Actionable hints** — check `Microsoft.NET.Test.Sdk` NuGet reference, ensure the project is built, and verify the target framework

**Before:**
```
Could not find testhost
```

**After:**
```
Could not find testhost for test source '/path/to/my.test.dll'.

Verify that:
  - The test project references the 'Microsoft.NET.Test.Sdk' NuGet package.
  - The project has been built and the output is in '/path/to/'.
  - The project's target framework matches the installed .NET runtime.
```

## Testing

Added unit test `GetTestHostProcessStartInfoShouldIncludeSourcePathInExceptionMessageWhenTesthostNotFound` that verifies the exception message contains both the source path and directory when testhost cannot be found.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/26290087537)*

PR #16169, opened 2026-06-25
Suggest Microsoft.NET.Test.Sdk when a managed test project brings no testhost

nohwnd:
The `Microsoft.TestPlatform` package ships a .NET Core testhost right next to the runner. So when a managed test project brings no testhost of its own — i.e. it does not reference `Microsoft.NET.Test.Sdk` — `DotnetTestHostManager` quietly falls back to that built-in testhost, plus the package's `testhost.deps.json` and a synthesized `runtimeconfig`. That fallback is meant for native (C++) runners only. A managed project landing there just spins up a testhost that discovers nothing, and the user gets a confusing "no tests" instead of "you forgot `Microsoft.NET.Test.Sdk`".

I confirmed it — a plain managed `net8.0` project without Test.Sdk goes straight into the fallback:

```
ManagedNoTestSdk.runtimeconfig.json, does not exist
Found testhost.dll next to runner executable: ...\testhost.dll
  --additional-deps "...\testhost.deps.json"
  --runtimeconfig  "...\testhost-8.0.runtimeconfig.json"
```

So the fix throws at that point: when we are about to use the testhost-next-to-runner fallback and the source is managed (not native), fail and tell the user to reference `Microsoft.NET.Test.Sdk` (the `CouldNotFindTesthost` message already lists it). Native (C++) sources keep the fallback, that is what it is for. `IsNativeModule` (the `PEReader` check already used for x64 forcing) is promoted to a method so both paths share it.

Tests: unit tests for both branches (managed -> throws, native -> uses the fallback), and an acceptance test that runs a managed project without Test.Sdk (`ProjectWithoutTestSdk`) through the real package runner and asserts the guidance.

Verified: `Microsoft.TestPlatform.TestHostProvider.UnitTests` green on `net481` (49/49). The `net11.0` unit legs and the acceptance test I could not run locally (no pinned .NET 11 preview runtime here), so CI covers those. 🤞

### DIFF
--- a/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
+++ b/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs
@@ -383,6 +383,17 @@ public class DotnetTestHostManager : ITestRuntimeProvider2
 #endif
                 if (_fileHelper.Exists(testHostNextToRunner))
                 {
+                    // The testhost shipped next to the runner (together with the package's testhost.deps.json and a
+                    // synthesized runtimeconfig) is the fallback for native (C++) runners that bring no managed testhost.
+                    // A managed (.NET) source reaching here did not resolve a testhost from its own deps.json, which means
+                    // the test project is missing the Microsoft.NET.Test.Sdk reference. Don't silently run it on the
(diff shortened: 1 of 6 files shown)


nohwnd:
still needs work the fix is not correct.

nohwnd:
Improves https://github.com/microsoft/vstest/pull/16053

azat-msft (approved):

PR #16177, opened 2026-06-26
[efficiency-improver] perf: eliminate string[1] allocation per test case in discovery source tracking

nohwnd:
## Goal and Rationale

Eliminate a transient `string[1]` array allocation that occurred **once per discovered test case** during the test discovery hot path, reducing GC pressure during large test runs.

**Focus area**: Code-Level Efficiency

## Problem

`MarkSourcesBasedOnDiscoveredTestCases` was called once per batch of discovered test cases. Inside it, for each individual test case it called:

```csharp
MarkSourcesWithStatus(new[] { source }, DiscoveryStatus.PartiallyDiscovered);
```

This allocated a temporary single-element `string[]` array on every iteration — just to satisfy the `IEnumerable<string>` parameter of a public method that iterates over it immediately. For a discovery run with 10,000 tests, this created ~10,000 short-lived `string[1]` arrays that were immediately eligible for GC collection.

## Approach

1. Extracted a private `MarkSourceWithStatus(string source, DiscoveryStatus status)` helper containing the core `ConcurrentDictionary.AddOrUpdate` logic.
2. Refactored the public `MarkSourcesWithStatus(IEnumerable<string>, DiscoveryStatus)` to delegate to this helper per source — keeps the public API intact and eliminates duplication (DRY).
3. Changed the hot loop in `MarkSourcesBasedOnDiscoveredTestCases` to call `MarkSourceWithStatus` directly, bypassing the array allocation entirely.

The public API surface is unchanged; no callers needed updating.

## Energy Efficiency Evidence

**Proxy metric**: GC allocation count (fewer short-lived allocations → less GC work → less CPU/DRAM energy spent on collection).

| Scenario | Allocations eliminated |
|---|---|
| 10,000-test discovery run | ~10,000 × `string[1]` arrays (~40 bytes each on 64-bit → ~400 KB of GC pressure) |
| 100,000-test discovery run | ~100,000 × `string[1]` arrays (~4 MB of GC pressure) |

These are Gen0 allocations (short-lived), but their volume in large test runs forces additional GC collections, each of which suspends all managed threads and burns CPU cycles scanning the heap. Reducing allocation volume is directly proportional to reducing the frequency of these pauses.

**Green Software Foundation context**:
- *Hardware Efficiency*: Fewer short-lived allocations make better use of the CPU and DRAM by reducing GC overhead — the hardware spends more cycles on useful work instead of heap management.
- *Software Carbon Intensity (SCI)*: Reducing unnecessary CPU work in the test discovery path lowers the energy consumed per `dotnet test` invocation, directly improving SCI for all users running tests via vstest.

## Trade-offs

None. This is a pure refactoring:
- Identical observable behaviour
- Public API surface unchanged
- Code is arguably *cleaner* (less duplication, single-responsibility helper)
- No impact on thread safety (helper still uses `ConcurrentDictionary.AddOrUpdate`)

## Reproducibility

```bash
# Build
./build.sh

# Test
./test.sh -p CrossPlatEngine
```

To observe the allocation reduction, use a memory profiler (dotMemory, BenchmarkDotNet with `[MemoryDiagnoser]`) on the `MarkSourcesBasedOnDiscoveredTestCases` call path with a large test batch.

## Test Status

- ✅ Build: 0 errors (4 pre-existing IL-trimming warnings, unrelated)
- ✅ CrossPlatEngine unit tests: all pass, 0 failures (including `DiscoveryDataAggregatorRegressionTests`)

> Generated by [Efficiency Improver](https://github.com/microsoft/vstest/actions/runs/28253628061) · 2K AIC · ⌖ 26.4 AIC · ⊞ 45.5K · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvstest+%22gh-aw-workflow-id%3A+efficiency-improver%22&type=pullrequests)

### DIFF
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
@@ -191,34 +191,39 @@ internal sealed class DiscoveryDataAggregator
                 continue;
             }
 
-            _sourcesWithDiscoveryStatus.AddOrUpdate(source,
-                _ =>
-                {
-                    if (status != DiscoveryStatus.NotDiscovered)
-                    {
-                        EqtTrace.Warning($"DiscoveryDataAggregator.MarkSourcesWithStatus: Undiscovered {source} added with status: '{status}'.");
-                    }
-                    else
-                    {
-                        EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourcesWithStatus: Adding {source} with status: '{status}'.");
-                    }
-
-                    return status;
-                },
-                (_, previousStatus) =>
-                {
-                    if (previousStatus == DiscoveryStatus.FullyDiscovered && status != DiscoveryStatus.FullyDiscovered
-                        || previousStatus == DiscoveryStatus.PartiallyDiscovered && (status == DiscoveryStatus.NotDiscovered || status == DiscoveryStatus.SkippedDiscovery))
-                    {
-                        EqtTrace.Warning($"DiscoveryDataAggregator.MarkSourcesWithStatus: Downgrading source {source} status from '{previousStatus}' to '{status}'.");
-                    }
-                    else if (previousStatus != status)
-                    {
-                        EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourcesWithStatus: Upgrading {source} status from '{previousStatus}' to '{status}'.");
-                    }
-                    return status;
-                });
+            MarkSourceWithStatus(source, status);
         }
+    }
+
+    private void MarkSourceWithStatus(string source, DiscoveryStatus status)
+    {
+        _sourcesWithDiscoveryStatus.AddOrUpdate(source,
+            _ =>
+            {
+                if (status != DiscoveryStatus.NotDiscovered)
+                {
+                    EqtTrace.Warning($"DiscoveryDataAggregator.MarkSourceWithStatus: Undiscovered {source} added with status: '{status}'.");
+                }
+                else
+                {
+                    EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourceWithStatus: Adding {source} with status: '{status}'.");
+                }
+
+                return status;
+            },
+            (_, previousStatus) =>
+            {
+                if (previousStatus == DiscoveryStatus.FullyDiscovered && status != DiscoveryStatus.FullyDiscovered
+                    || previousStatus == DiscoveryStatus.PartiallyDiscovered && (status == DiscoveryStatus.NotDiscovered || status == DiscoveryStatus.SkippedDiscovery))
+                {
+                    EqtTrace.Warning($"DiscoveryDataAggregator.MarkSourceWithStatus: Downgrading source {source} status from '{previousStatus}' to '{status}'.");
+                }
+                else if (previousStatus != status)
+                {
+                    EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourceWithStatus: Upgrading {source} status from '{previousStatus}' to '{status}'.");
+                }
+                return status;
+            });
     }
 
     /// <summary>
@@ -252,13 +257,13 @@ internal sealed class DiscoveryDataAggregator
             // assume that the previous source was fully discovered.
             if (previousSource is null || previousSource == currentSource)
             {
-                MarkSourcesWithStatus(new[] { currentSource }, DiscoveryStatus.PartiallyDiscovered);
+                MarkSourceWithStatus(currentSource, DiscoveryStatus.PartiallyDiscovered);
             }
             else if (currentSource != previousSource)
             {
                 EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourcesBasedOnDiscoveredTestCases: Discovered test source changed from {previousSource} to {currentSource}.");
-                MarkSourcesWithStatus(new[] { previousSource }, DiscoveryStatus.FullyDiscovered);
-                MarkSourcesWithStatus(new[] { currentSource }, DiscoveryStatus.PartiallyDiscovered);
+                MarkSourceWithStatus(previousSource, DiscoveryStatus.FullyDiscovered);
+                MarkSourceWithStatus(currentSource, DiscoveryStatus.PartiallyDiscovered);
             }
 
             previousSource = currentSource;
--- a/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
+++ b/src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
@@ -250,6 +250,12 @@ internal sealed class DiscoveryDataAggregator
 
         foreach (var testCase in testCases)
         {
+            if (_isMessageSent == 1)
+            {
+                EqtTrace.Verbose("DiscoveryDataAggregator.MarkSourcesBasedOnDiscoveredTestCases: Message was already sent so skipping remaining source updates.");
+                return previousSource;
+            }
+
             var currentSource = testCase.Source;
 
             // We rely on the fact that sources are processed in a sequential way, which
--- a/test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/DiscoveryDataAggregatorRegressionTests.cs
+++ b/test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Client/Parallel/DiscoveryDataAggregatorRegressionTests.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 
 using Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.Parallel;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.Engine;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
```diff
@@ -252,13 +257,13 @@ public void MarkSourcesWithStatus(IEnumerable<string?>? sources, DiscoveryStatus
             // assume that the previous source was fully discovered.
             if (previousSource is null || previousSource == currentSource)
             {
-                MarkSourcesWithStatus(new[] { currentSource }, DiscoveryStatus.PartiallyDiscovered);
+                MarkSourceWithStatus(currentSource, DiscoveryStatus.PartiallyDiscovered);
```
nohwnd, line 266: **[Parallel Execution & Scheduling Safety]** The `_isMessageSent` guard granularity regressed from per-test-case to per-batch.

Before this PR, each `MarkSourcesWithStatus(new[] { ... }, ...)` call checked `_isMessageSent` at entry. So if `TryAggregateIsMessageSent()` fired on a concurrent thread mid-batch, the very next iteration caught it.

Now the guard only lives at the top of `MarkSourcesBasedOnDiscoveredTestCases`. If `TryAggregateIsMessageSent()` races in after the method enters the loop, the entire remaining batch continues writing to `_sourcesWithDiscoveryStatus` even though `ParallelDiscoveryEventsHandler` may already be executing `GetSourcesWithStatus()` on lines 89–92.

Practical sequence:
1. Thread A passes the top-level guard (`_isMessageSent == 0`) and begins iterating a batch
2. Thread B calls `TryAggregateIsMessageSent()` (sets `_isMessageSent = 1`) then immediately calls `GetSourcesWithStatus()` to build the final report
3. Thread A continues `MarkSourceWithStatus()` for every remaining test case in the batch

Those writes are "post-finalization" updates that weren't reflected in Thread B's snapshot. A source that should have been upgraded from `PartiallyDiscovered` → `FullyDiscovered` mid-batch may appear as `PartiallyDiscovered` in the final report, which then incorrectly triggers the "discovery aborted" path at `ParallelDiscoveryEventsHandler` line 106.

The ConcurrentDictionary doesn't corrupt, but the "Identical observable behaviour" and "No impact on thread safety" claims in the PR description aren't accurate for this concurrent scenario.

The race existed before (just narrower), so this is not a newly introduced bug — but it is a regression in the guard's effectiveness. The simplest fix is a mid-loop check:

```suggestion
            foreach (var testCase in testCases)
            {
                if (_isMessageSent == 1)
                {
                    return previousSource;
                }

                var currentSource = testCase.Source;
```

Alternatively, document that the guard was intentionally coarsened to per-batch and add a test for `MarkSourcesBasedOnDiscoveredTestCases` + `TryAggregateIsMessageSent` to pin the behaviour (analogous to `MarkSourcesWithStatus_AfterMessageSent_ShouldSkipUpdate`).

src/Microsoft.TestPlatform.CrossPlatEngine/Client/Parallel/DiscoveryDataAggregator.cs
```diff
@@ -191,34 +191,39 @@ public void MarkSourcesWithStatus(IEnumerable<string?>? sources, DiscoveryStatus
...
-                    }
-                    else
-                    {
-                        EqtTrace.Verbose($"DiscoveryDataAggregator.MarkSourcesWithStatus: Adding {source} with status: '{status}'.");
-                    }
-
-                    return status;
-                },
-                (_, previousStatus) =>
+                    EqtTrace.Warning($"DiscoveryDataAggregator.MarkSourceWithStatus: Undiscovered {source} added with status: '{status}'.");
```
nohwnd, line 205: nit: [Error Reporting & Diagnostic Clarity] All per-source trace messages changed from `MarkSourcesWithStatus` (plural) to `MarkSourceWithStatus` (singular). The new name is factually more accurate since the messages now live in the private helper, but it's a format change that would break any grep-based log analysis or support runbooks that match on the old prefix. No test currently asserts on these strings so nothing breaks in CI, but worth noting for ops / support docs.

nohwnd (commented):
## Review Summary (DRAFT)

Single file changed: `DiscoveryDataAggregator.cs`. Activated dimensions: **Parallel Execution & Scheduling Safety**, **Error Reporting & Diagnostic Clarity**.

| Dimension | Status |
|-----------|--------|
| Parallel Execution & Scheduling Safety | ⚠️ warn |
| Error Reporting & Diagnostic Clarity | ⚠️ nit |
| Process Architecture & Host Resolution | ✅ pass (N/A) |

**Findings: 1 warn, 1 nit**

---

### The allocation fix itself

The `new[] { source }` elimination is correct. The extraction of `MarkSourceWithStatus` is clean, `ConcurrentDictionary.AddOrUpdate` semantics are preserved, and the public API is unchanged.

### The `_isMessageSent` guard granularity change

This is the one genuine issue. Before the PR, calling `MarkSourcesWithStatus(new[] { source }, ...)` per iteration meant `_isMessageSent` was checked once per test case. Now the guard only fires at the top of `MarkSourcesBasedOnDiscoveredTestCases`, so a concurrent `TryAggregateIsMessageSent()` + `GetSourcesWithStatus()` snapshot in `ParallelDiscoveryEventsHandler.HandleDiscoveryComplete` (lines 82–92) can race with the entire remaining batch. The practical impact is narrow: a source might appear `PartiallyDiscovered` instead of `FullyDiscovered` in the final report, which incorrectly trips the "discovery aborted" heuristic on line 106 of that handler. The ConcurrentDictionary is safe; the issue is snapshot consistency. The race existed before (the per-call guard just made the window per-test instead of per-batch), so this isn't a newly introduced bug — but "Identical observable behaviour" overstates it.

The suggestion in the inline comment (`if (_isMessageSent == 1) return previousSource` inside the loop) restores the per-test guard with no allocation overhead.

### Missing test coverage

`DiscoveryDataAggregatorRegressionTests.MarkSourcesWithStatus_AfterMessageSent_ShouldSkipUpdate` covers `MarkSourcesWithStatus` + `TryAggregateIsMessageSent`. There's no equivalent test for `MarkSourcesBasedOnDiscoveredTestCases` + `TryAggregateIsMessageSent`. Since the guard behavior of that method changed, a test would pin it.

### Trace string rename (nit)

`MarkSourcesWithStatus` → `MarkSourceWithStatus` in the per-source trace messages is factually correct, but it's a format change for anyone grepping logs. No test assertions break.

---

Good goal, clean approach. Address the guard granularity (or explicitly document the intentional coarsening and add a test to pin it) before merging.

🧠 *Reviewed by [expert-reviewer](https://github.com/microsoft/vstest/actions/runs/28255020289)* · PR #16177

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28255020289)*

nohwnd:
Commit pushed: [`b05acf0`](https://github.com/microsoft/vstest/commit/b05acf0bd47be335ecbb5190453467c161cd24cf)

> Generated by [Efficiency Improver](https://github.com/microsoft/vstest/actions/runs/28329440536) · 3.3K AIC · ⌖ 16.5 AIC · ⊞ 45.6K

nohwnd (commented):
## Re-review: Previous Findings Addressed ✅

PR #16177 was updated on 2026-06-28 with a follow-up commit (`fix: restore per-test-case _isMessageSent guard`) that directly responds to the previous expert review. This pass checks that the fixes are correct and complete.

---

### Finding resolution

| Finding (from prior review) | Status |
|---|---|
| `_isMessageSent` guard coarsened from per-test-case to per-batch | ✅ **Fixed** |
| Missing test for `MarkSourcesBasedOnDiscoveredTestCases` + `TryAggregateIsMessageSent` | ✅ **Fixed** |
| Trace prefix renamed `MarkSourcesWithStatus` → `MarkSourceWithStatus` (nit) | ⚪ Open (nit, no action required) |

---

### Guard fix (correctness verified)

The per-iteration check added inside `MarkSourcesBasedOnDiscoveredTestCases`:

```csharp
foreach (var testCase in testCases)
{
    if (_isMessageSent == 1)
    {
        EqtTrace.Verbose("...skipping remaining source updates.");
        return previousSource;
    }
    ...
}
```

This restores the per-test-case guard granularity that existed before the PR (previously each `MarkSourcesWithStatus(new[] { ... })` call checked at entry; now each loop iteration checks before calling the private helper). The small `check → AddOrUpdate` race window is identical to the pre-PR behaviour — inherent to this design, not a regression.

### New test (correctness verified)

`MarkSourcesBasedOnDiscoveredTestCases_AfterMessageSent_ShouldSkipUpdate` correctly pins the top-level guard: after `TryAggregateIsMessageSent()`, a subsequent call returns without upgrading any source status. The test is deterministic and matches the structure of the existing `MarkSourcesWithStatus_AfterMessageSent_ShouldSkipUpdate`.

### Allocation optimization (unchanged, still correct)

The core change — replacing `new[] { source }` with a direct call to the private `MarkSourceWithStatus` helper — is clean. `ConcurrentDictionary.AddOrUpdate` semantics are preserved, public API surface is unchanged.

---

### Note: PR still in DRAFT

The PR is currently in draft state. No code issues remain from this review; the maintainer can undraft when ready (note that undrafting triggers a rebuild per the project workflow guidelines).

---

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28329946336)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28329946336)*

nohwnd:
The Azure DevOps CI check `microsoft.vstest (Build Windows Release)` was **cancelled** (not failed with a build error) — this appears to be a transient infrastructure issue on the AzDO side, unrelated to the code changes in this PR.

The PR touches only `DiscoveryDataAggregator.cs` and a regression test file. Neither file was modified by the two PRs merged since this branch was created (#16179, #16182 touch `Condition.cs` and `FilterExpression.cs` respectively), so there are no conflicts. The code change itself is a straightforward refactoring with no logic changes.

Could a maintainer please manually re-trigger the Azure DevOps build? If the re-trigger also fails with a real build error, I'll investigate further.



> Generated by [Efficiency Improver](https://github.com/microsoft/vstest/actions/runs/28390618588) · 3.9K AIC · ⌖ 25.8 AIC · ⊞ 45.6K · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvstest+%22gh-aw-workflow-call-id%3A+microsoft%2Fvstest%2Fefficiency-improver%22&type=issues)

azat-msft:
The proposed benefit is very little (4MB of allocation/garbage collecting per 10k tests) while introducing potential parallelization problems. Therefore, I decide not to merge this change.

nohwnd:
It says 4kb per 10k of memory traffic, so even less.

PR #16190, opened 2026-06-29
[fix] Use Assert.Inconclusive instead of [Ignore] for video recorder test in CI

nohwnd:
🤖 This is an automated fix generated by GitHub Copilot.

Closes #15586

## Root Cause

The `VideoRecorderDataCollectorShouldRecordVideoWithRunSettings` test was suppressed with `[Ignore]`, which prevents the test from running **everywhere** — both in CI and locally. The maintainer's intent (documented in the issue) was to only skip the test on the official CI pipeline where the screen recorder cannot establish a communication channel (headless environment).

## Fix

Replace the blanket `[Ignore]` attribute with a conditional `Assert.Inconclusive` check at the start of the test method:

```csharp
if (IsCI)
{
    Assert.Inconclusive("Video recording is flaky in CI — screen recorder fails to establish communication. See #15586.");
}
```

`IsCI` is a property inherited from `IntegrationTestBase` that returns `true` when `TF_BUILD == "True"` (the Azure DevOps CI environment variable). This is the same pattern used by `BlameDataCollectorTests.BlameDataCollectorAeDebuggerShouldCollectDump` for similar conditional skips.

**Before**: Test skipped everywhere (locally and in CI)  
**After**: Test runs locally, marked `Inconclusive` in CI

## Why This Matters

The `[Ignore]` approach left the video recorder data collector completely untested. Developers running the suite locally would never see this test execute. With `Assert.Inconclusive`, local runs will attempt the test, and CI runs will report it as inconclusive rather than a failure.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/28377153057)*

### DIFF
--- a/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/VideoRecorderTests.cs
+++ b/test/Microsoft.TestPlatform.Acceptance.IntegrationTests/VideoRecorderTests.cs
@@ -13,11 +13,15 @@ namespace Microsoft.TestPlatform.AcceptanceTests;
 [TestCategory("Windows-Review")]
 public class VideoRecorderTests : AcceptanceTestBase
 {
-    [Ignore("Video recording is flaky in CI — screen recorder fails to establish communication. See #15586.")]
     [TestMethod]
     [NetFullTargetFrameworkDataSource(useCoreRunner: false, useVsixRunner: true)]
     public void VideoRecorderDataCollectorShouldRecordVideoWithRunSettings(RunnerInfo runnerInfo)
     {
+        if (IsCI)
+        {
+            Assert.Inconclusive("Video recording is flaky in CI — screen recorder fails to establish communication. See #15586.");
+        }
+
         SetTestEnvironment(_testEnvironment, runnerInfo);
         var assemblyPaths = GetAssetFullPath("SimpleTestProject.dll");
         var arguments = PrepareArguments(assemblyPaths, GetTestAdapterPath(), string.Empty, FrameworkArgValue,


nohwnd (commented):
## Review: `[Ignore]` → `Assert.Inconclusive(IsCI)` in VideoRecorderTests

### Dimension Checklist

| Dimension | Status |
|-----------|--------|
| Acceptance Test Coverage Design | ✅ Pass |
| Cross-TFM & Framework Resolution | ✅ Pass |
| Parallel Execution & Scheduling Safety | ✅ Pass |
| Environment Variable & Feature Flag Contracts | ✅ Pass |
| Algorithmic Correctness | ✅ Pass |
| Backward Compatibility & Rollback Safety | ✅ Pass |

**0 blocking findings. 0 warnings.**

---

### What was checked

**`IsCI` scope** — `TF_BUILD == "True"` is Azure DevOps-specific (comment on line 172 of `IntegrationTestEnvironment.cs` calls this out explicitly). No GitHub Actions workflows run acceptance tests (confirmed: zero matches in `.github/workflows/` for acceptance/integration test invocations), so the narrower scope is correct. The two Azure Pipelines definitions both inject `TF_BUILD` automatically.

**Data-driven early-exit** — `NetFullTargetFrameworkDataSource(useCoreRunner: false, useVsixRunner: true)` + default `useDesktopRunner: true, inIsolation: true` generates two rows (desktop+isolation, VSIX). Each row independently hits the `Assert.Inconclusive` branch in CI. That's the right outcome — no row leaks through to the screen recorder invocation.

**Placement before `SetTestEnvironment`** — The guard fires before any test-environment mutation, so no partial state is left on early exit. The class also has no `[TestInitialize]`/`[TestCleanup]`, so there's nothing heavyweight that runs around the test body and gets surprised by the early throw.

**Pattern consistency** — `BlameDataCollectorTests.BlameDataCollectorAeDebuggerShouldCollectDump` (line 289) uses the identical idiom (`Assert.Inconclusive` + `IsCI`) and already ships in CI without breaking the pipeline, which validates that Azure DevOps treats Inconclusive results as non-failures here.

**`[Ignore]` removal** — The attribute previously suppressed the test universally. Removing it is the correct complement to the conditional guard; leaving both would be redundant and confusing.

Looks good :)

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28378299507)*

nohwnd:
The skip is intentional. The test pops up a recording window and is annoying to run locally too, not just on CI, so skipping it everywhere is what we want. The premise that it should only be skipped on CI is wrong, so the `[Ignore]` stays. Closing. 🙂

PR #16193, opened 2026-06-29
[efficiency-improver] perf: eliminate Guid.ToString allocation in v2 test case serialization

nohwnd:
## Goal and Rationale

Eliminate a transient heap allocation per serialized test case on the IPC write path, reducing GC pressure during test execution.

**Focus area:** Network & I/O Efficiency (per-test serialization, testhost → vstest.console IPC)

### `TestCaseConverterV2.Write`: Guid without `ToString()`

```diff
-writer.WriteString("Id", value.Id.ToString());
+writer.WriteString("Id", value.Id);
```

`Guid.ToString()` allocates a 36-char heap string on every call. `Utf8JsonWriter.WriteString(string, Guid)` writes the GUID directly to the UTF-8 stream using the same "D" format (`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`, lowercase with hyphens) — **zero heap allocation**.

The change is strictly `#if NETCOREAPP`-scoped; no impact on net48x paths.

> Note: An earlier revision of this PR also replaced `TimeSpan.ToString()` in `TestResultConverterV2.Write` with a `stackalloc` + `TryFormat("c")` buffer. That change was reverted and is **not** part of this PR.

## Energy Efficiency Evidence

**Proxy metric:** Memory allocation (GC pressure on the testhost IPC write path)

- Before: each call to `TestCaseConverterV2.Write` allocates 1 × ~72-byte `string` (36 UTF-16 chars for Guid "D" format)
- After: zero heap allocations for the Id serialization

For a 10,000-test run:
- Guid strings eliminated: ~10,000 × 72 bytes ≈ **703 KB fewer heap allocations** → proportionally less GC work

**Green Software Foundation context:** This change directly reduces *software carbon intensity* via *demand shaping* — by eliminating transient allocations, the GC runs less frequently, lowering both CPU energy use and DRAM refresh cycles. The improvement is proportional to test count, consistent with *energy proportionality*.

## Trade-offs

- Wire format is **unchanged**: the replacement produces byte-for-byte identical JSON output, verified against existing serialization tests.

## Reproducibility

Verify JSON output is unchanged:
```bash
# Existing serialization round-trip tests cover the path:
./test.sh -p Microsoft.TestPlatform.CommunicationUtilities.UnitTests
```

Key test:
- `TestCaseJsonShouldContainAllPropertiesOnSerializationV2` — asserts `"be78d6fc-61b0-4882-9d07-40d796fd96ce"` (D-format GUID)

## Test Status

CI will validate. The change is inside `#if NETCOREAPP` and touches only the v2 protocol write path. No protocol version change.

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
@@ -125,9 +125,7 @@ internal class TestResultConverterV2 : JsonConverter<TestResult>
         StjSafe.Serialize(writer, value.Messages, options);
 
         writer.WriteString("ComputerName", value.ComputerName);
-        Span<char> durationBuf = stackalloc char[30];
-        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c");
-        writer.WriteString("Duration", durationBuf[..durationCharsWritten]);
+        writer.WriteString("Duration", value.Duration.ToString());
         writer.WriteString("StartTime", value.StartTime);
         writer.WriteString("EndTime", value.EndTime);
 
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
@@ -125,7 +125,9 @@ internal class TestResultConverterV2 : JsonConverter<TestResult>
         StjSafe.Serialize(writer, value.Messages, options);
 
         writer.WriteString("ComputerName", value.ComputerName);
-        writer.WriteString("Duration", value.Duration.ToString());
+        Span<char> durationBuf = stackalloc char[30];
+        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c");
+        writer.WriteString("Duration", durationBuf[..durationCharsWritten]);
         writer.WriteString("StartTime", value.StartTime);
         writer.WriteString("EndTime", value.EndTime);
 
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
@@ -76,7 +76,7 @@ internal class TestCaseConverterV2 : JsonConverter<TestCase>
     public override void Write(Utf8JsonWriter writer, TestCase value, JsonSerializerOptions options)
     {
         writer.WriteStartObject();
-        writer.WriteString("Id", value.Id.ToString());
+        writer.WriteString("Id", value.Id);
         writer.WriteString("FullyQualifiedName", value.FullyQualifiedName);
         writer.WriteString("DisplayName", value.DisplayName);
         writer.WriteString("ExecutorUri", value.ExecutorUri?.OriginalString);


nohwnd:
## 🔍 Build Failure Analysis

**Summary** — The build failed because `TimeSpan.TryFormat` was called without an explicit `IFormatProvider`, triggering CA1305 (which is promoted to an error in this project's analyser configuration).

### Root cause: Missing `IFormatProvider` on `TimeSpan.TryFormat`

The PR optimised `Duration` serialisation in `TestResultConverterV2.cs` by replacing `value.Duration.ToString()` with a `TryFormat` call on a stack-allocated `Span<char>` — a good, allocation-free change. However the chosen overload is:

```
bool TimeSpan.TryFormat(Span<char>, out int, ReadOnlySpan<char>, IFormatProvider?)
```

Because the optional `IFormatProvider` argument was omitted, the CA1305 analyser ("Specify IFormatProvider") fires. CA1305 is treated as an error in this repo, so the build fails at `CoreCompile` in `Microsoft.TestPlatform.CommunicationUtilities.csproj`.

The one-line fix is to pass `CultureInfo.InvariantCulture` as the fourth argument. This is semantically correct: the `"c"` (constant/invariant) format specifier is itself culture-insensitive. `System.Globalization` is already imported in the file.

**Affected file / error**

- [`src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs:129`](https://github.com/microsoft/vstest/blob/95b4e78f6f561f69ad200dafa3226cb71a91412f/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs#L129) — `CA1305: The behavior of 'TimeSpan.TryFormat(...)' could vary based on the current user's locale settings. Provide a value for the 'IFormatProvider' argument.`

**Proposed fix**

```diff
-        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c");
+        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c", CultureInfo.InvariantCulture);
```

---

<details>
<summary><b>Build overview</b></summary>

| Field | Value |
| ----- | ----- |
| Status | ❌ FAILED |
| Duration | 118.8 s |
| MSBuild | 18.7.0-preview |
| Projects | 75 |
| Errors | 2 (1 real + cascaded "Build failed.") |
| Warnings | 4 |
| Failing projects | `Build.proj`, `TestPlatform.slnx`, `datacollector.csproj` |
| Failing target | `CoreCompile` in `Microsoft.TestPlatform.CommunicationUtilities.csproj` |

</details>

<details>
<summary><b>All MSBuild errors (1)</b></summary>

| Code | Project | File:Line | Message |
| ---- | ------- | --------- | ------- |
| `CA1305` | `Microsoft.TestPlatform.CommunicationUtilities` | [`TestResultConverterV2.cs:129`](https://github.com/microsoft/vstest/blob/95b4e78f6f561f69ad200dafa3226cb71a91412f/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs#L129) | The behavior of `TimeSpan.TryFormat(Span<char>, out int, [ReadOnlySpan<char>], [IFormatProvider])` could vary based on the current user's locale settings. Provide a value for the `IFormatProvider` argument. |

</details>

---

<sub>🤖 Generated by the Build Failure Analysis workflow · commit 95b4e78f6f561f69ad200dafa3226cb71a91412f</sub>



> Generated by [Build Failure Analysis](https://github.com/microsoft/vstest/actions/runs/28392009566) for issue #16193 · 160.5 AIC · ⌖ 12.9 AIC · ⊞ 34.8K · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvstest+%22gh-aw-workflow-call-id%3A+microsoft%2Fvstest%2Fbuild-failure-analysis%22&type=issues)

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
```diff
@@ -125,7 +125,9 @@ public override void Write(Utf8JsonWriter writer, TestResult value, JsonSerializ
         StjSafe.Serialize(writer, value.Messages, options);
 
         writer.WriteString("ComputerName", value.ComputerName);
-        writer.WriteString("Duration", value.Duration.ToString());
+        Span<char> durationBuf = stackalloc char[30];
+        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c");
```
nohwnd: 🔧 **`CA1305`** — `TimeSpan.TryFormat` has an overload that accepts an `IFormatProvider`; the analyser requires it to be specified explicitly to prevent locale-sensitive behaviour. Pass `CultureInfo.InvariantCulture` here — it is semantically correct since the `"c"` format specifier is the culture-invariant constant (round-trip) format, and `System.Globalization` is already imported in this file.

```suggestion
        value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c", CultureInfo.InvariantCulture);
```

nohwnd (commented):
> Generated by [Build Failure Analysis](https://github.com/microsoft/vstest/actions/runs/28392009566) for issue #16193 · 160.5 AIC · ⌖ 12.9 AIC · ⊞ 34.8K

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
nohwnd: **[IPC Transport & Protocol Stability | Defensive Coding]**

`TryFormat`'s return value is silently discarded. If it returns `false`, the BCL sets `durationCharsWritten = 0`, so `durationBuf[..0]` is an empty `ReadOnlySpan<char>`. That causes `WriteString` to emit `"Duration": ""` — a silent data loss that would make `TimeSpan.Parse` on the read side throw.

In practice this path is unreachable: `TimeSpan.MinValue` in `"c"` format is 26 chars (e.g. `-10675199.02:48:05.4775808`), so the 30-char buffer is always sufficient. But the silent-failure mode warrants an explicit assertion to communicate the assumption and guard against a future caller accidentally reducing the buffer:

```csharp
Span<char> durationBuf = stackalloc char[30];
bool formatted = value.Duration.TryFormat(durationBuf, out int durationCharsWritten, "c");
Debug.Assert(formatted, "TimeSpan in 'c' format requires at most 26 chars; buffer is 30.");
writer.WriteString("Duration", durationBuf[..durationCharsWritten]);
```

nohwnd (commented):
## Code Review — Serialization hot-path alloc elimination

**Dimensions activated:** IPC Transport & Protocol Stability · Defensive Coding at Boundaries · Cross-TFM Compatibility · Performance & Allocations

---

### Wire format compatibility ✅

Both changes produce byte-for-byte identical JSON:

- **Guid**: `Utf8JsonWriter.WriteString(string, Guid)` uses the same lowercase "D" format (`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`) as `Guid.ToString()`. The read side (`GuidPolyfill.Parse`) accepts this unchanged.
- **TimeSpan**: `TryFormat(..., "c")` uses the constant/invariant "c" format, identical to `TimeSpan.ToString()`. The read side (`TimeSpan.Parse(..., CultureInfo.InvariantCulture)`) parses it unchanged.

### TFM compatibility ✅

Both files are guarded by `#if NETCOREAPP`. All APIs used (`Utf8JsonWriter.WriteString(string, Guid)`, `TimeSpan.TryFormat`, `ReadOnlySpan<char>` slicing) are available on all NETCOREAPP targets (.NET Core 3.0+), the minimum required by `System.Text.Json`. No net462 impact.

### Performance claim ✅

The optimizations are valid:
- `Utf8JsonWriter.WriteString(string, Guid)` formats directly to the UTF-8 stream — zero heap allocation.
- `stackalloc char[30]` + `TryFormat` writes to the stack; `WriteString(string, ReadOnlySpan<char>)` transcodes it to UTF-8 without an intermediate `string` — zero heap allocation.

### Finding — one minor issue 🔶

**`TryFormat` return value unchecked** (`TestResultConverterV2.cs:129`) — see inline comment. In practice this cannot fail (max TimeSpan "c" length is 26 chars, buffer is 30), but the silent failure mode (empty Duration field) warrants a `Debug.Assert` to make the assumption explicit.

### PR description alignment ✅

The description accurately characterizes both changes, correctly states wire format is unchanged, provides sound rationale for the 30-char buffer, and correctly identifies the `#if NETCOREAPP` scope.

---

🧠 *Reviewed by [expert-reviewer workflow](https://github.com/microsoft/vstest/actions) · dimensions: IPC Protocol · Defensive Coding · Cross-TFM · Performance*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28392009561)*

azat-msft:
Since the efficiency impact is almost negligible, I removed the change that complicates the logic and only left one that removes `.ToString()` call.

azat-msft (approved):

PR #16201, opened 2026-07-01
Run Microsoft.Testing.Platform test apps under vstest.console and datacollector

nohwnd:
**Proof-of-concept / RFC.** Companion draft: microsoft/testfx#9546, you don't need both (see below).

`vstest.console` and `datacollector` can now discover and run a Microsoft.Testing.Platform (MTP) app directly over the MTP protocol, instead of only launching a classic vstest testhost. An MTP app *is* its own host, so vstest detects it from the source (`[assembly: AssemblyMetadata("Microsoft.Testing.Platform.Application", "true")]`) and drives it with MTP-specific proxies rather than the vstest testhost JSON protocol. Classic vstest-based projects keep working unchanged, and the two can run **in the same invocation**.

### Why
Test frameworks can drop their vstest dependencies and still be run by vstest in a mixed run. That makes migrating off vstest incremental instead of all-or-nothing: port some tests to MTP, then all of them, then move the infra (reporting, CI) off vstest when it's convenient, instead of replacing test infra, reporting, and every test in one shot. And no dynamic code loading. The companion testfx draft (microsoft/testfx#9546) is the other half: MSTest shedding `Microsoft.NET.Test.Sdk` so `Microsoft.TestPlatform.ObjectModel` is its only vstest-lineage dep. Neither PR needs the other to land. Together they show the future is reachable.

### What changed
- **Detection.** `MicrosoftTestingPlatformDetector` (CoreUtilities) and `AssemblyMetadataProvider.IsMicrosoftTestingPlatformApp` (vstest.console) read the MTP assembly-metadata marker from the source.
- **MTP transport.** `CrossPlatEngine/Client/MTP/` adds a JSON-RPC client (`MtpServerConnection`), a node to vstest result converter, and `MtpProxyDiscoveryManager`/`MtpProxyExecutionManager` that implement the existing `IProxyDiscoveryManager`/`IProxyExecutionManager` but speak MTP. The wire is serialized with `Jsonite`, not `System.Text.Json`, so the client runs on every TFM we ship including `net462`, with no binding-redirect fallout in hosts that run without them.
- **MTP as a runtime provider.** `MtpTestRuntimeProvider` (assembly `Microsoft.TestPlatform.TestHostRuntimeProvider`) is a first-class `ITestRuntimeProvider` that claims MTP sources by their shape through `ISourceAwareTestRuntimeProvider`. It implements the new `IProxyManagerFactory`, so `TestEngine` asks the resolved provider for its discovery and execution managers instead of branching on the source. A two-pass `TestRuntimeProviderManager` resolves it so the "no runtime provider" guard no longer rejects MTP sources. `TestEngine` has no `#if NETCOREAPP` or `IsMicrosoftTestingPlatformSource` branches anymore, MTP is named only in explanatory comments.
- **Coverage.** `TestEngine` wraps code-coverage datacollection around the MTP execution path, the same way it does for the classic path.
- Removed the interim `ExecutionPreference` switch, detection is purely source-based.

### What I tested
The capstone: **one `vstest.console` invocation over a classic vstest-based project and an MTP-only project together**, with `--collect:"Code Coverage"` and `--logger:trx`:

| Source | Execution path | Passed | Failed | Skipped | Total |
|---|---|---|---|---|---|
| `MSTestProject1.dll` | classic vstest testhost | 1 | 1 | 1 | 3 |
| `MtpMSTestProject.dll` | MTP protocol | 2 | 1 | 1 | 4 |
| **Combined (one TRX)** | | **3** | **2** | **2** | **7** |

- One TRX carried all 7 tests, referencing both assemblies, outcomes correct.
- Code coverage produced for **both** modules (`MtpMSTestProject` 72.73% blocks, `MSTestProject1` 50%). Stable across 3 back-to-back runs.
- **Acceptance tests** (`MtpUnderVstestTests`): a pure-MTP run, a mixed classic+MTP run, and a mixed run that writes a single TRX. Green through the real acceptance harness across both console axes (netfx `vstest.console.exe` and netcore `dotnet vstest`), 6/6, in Debug and Release.
- Also: `Microsoft.TestPlatform.Common.UnitTests` 396 passing / 0 failing, clean Release build and pack, smoke suite green.

### Known limitation: coverage packaging
The vstest "Code Coverage" datacollector binaries ship in `Microsoft.CodeCoverage` and normally reach vstest.console through `Microsoft.NET.Test.Sdk` (`Microsoft.CodeCoverage.props` sets `TraceDataCollectorDirectoryPath`, which the SDK's VSTest task forwards as `VSTestTraceDataCollectorDirectoryPath`). A test project that has dropped its vstest deps won't restore those DLLs, so vstest-datacollector coverage on it needs another source for them, either the project references `Microsoft.CodeCoverage`, or vstest bundles the collector in its runner. In this proof I grafted the collector into the runner's `Extensions/` (the "vstest bundles it" option). I wrote up the options in a comment below, pinned for a design discussion with the owner of code coverage. MTP-native coverage (`Microsoft.Testing.Extensions.CodeCoverage`) is a separate, unaffected path.

### Note for reviewers
MTP routing is a first-class runtime provider now, not a set of inline branches. `TestEngine` knows only the `IProxyManagerFactory` abstraction: if the resolved provider implements it, the engine asks it for the discovery and execution manager, otherwise the standard testhost proxies wrap the provider as before. The `#if NETCOREAPP` and `IsMicrosoftTestingPlatformSource` branches called out in the earlier inline threads are gone, MTP is named only in explanatory comments.

Co-authored with GitHub Copilot.


nohwnd:
There is nicer way to abstract the mtp stuff than bunch of IFs but did not want to go that far, can also be implemented for .net framework ofc.

Main part of the impelentation is something that looks like a "MTP" client that we already had mentioned in few places, and where the issue was recently closed as having no real usecase so maybe this is one.  https://github.com/microsoft/testfx/issues/5667

nohwnd:
### Coverage packaging — notes from the discussion with the owner of code coverage

Talked this through, capturing where we landed so the "known limitation" above isn't just parked.

**Framing we agreed on:** in this mode vstest owns everything above the testhost — datacollector, loggers, TRX. The MTP app is only the testhost. So vstest collects coverage with the `Microsoft.CodeCoverage` collector (the same `TraceDataCollector` datacollector as today), **not** `Microsoft.Testing.Extensions.CodeCoverage`. TRX is the same story: vstest writes the TRX from the results the MTP app streams over the protocol, so `mtp.ext.trx` isn't involved either.

**The only real question is how the collector binaries reach the run once the project stops referencing `Microsoft.NET.Test.Sdk`.** Worth spelling out what I found digging through the packages, because it narrows the options: the collector lives entirely in `Microsoft.CodeCoverage`. Its `build/` folder carries every collector DLL and native engine (win/linux/macos/alpine, x86/x64/arm64, `CodeCoverage.exe`, `covrun*`), and `Microsoft.CodeCoverage.props` sets `TraceDataCollectorDirectoryPath`, which `dotnet test`'s VSTest task (from `Microsoft.TestPlatform.Build`) forwards to vstest.console as `VSTestTraceDataCollectorDirectoryPath`. `Microsoft.NET.Test.Sdk` itself has **no** coverage-specific target logic — its only contribution is pulling the `Microsoft.CodeCoverage` PackageReference transitively. So dropping Test.Sdk and keeping just a direct `Microsoft.CodeCoverage` reference reproduces today's coverage behaviour exactly.

That gives three options:

1. **Project references `Microsoft.CodeCoverage` directly.** One thin, framework-agnostic package — no ObjectModel, no adapter, no Test.Sdk. Restores the `TraceDataCollectorDirectoryPath` chain with zero vstest changes, and `mtp.ext.cc` can still sit alongside it for the MTP-native path. Smallest change; cost is that the project has to opt in.

2. **vstest ships the collector** — add a `Microsoft.CodeCoverage` dependency to `Microsoft.TestPlatform.CLI` (the package `dotnet test` uses). Then coverage works for any project with no reference at all, which is the actual "drop everything" goal. Caveats: (a) version selection when the project *also* brings CC in via Test.Sdk — two collector dirs, so vstest needs precedence (prefer the project's `TraceDataCollectorDirectoryPath`, fall back to the bundled one); solvable in code. (b) size — CC's native payload is large and would ride along on every CLI even when coverage is off. (c) it couples a CC version into vstest's release cadence. Note the CLI already carries `Microsoft.CodeCoverage.IO.dll` (the coverage-file reader), just not the collector engine — so there's precedent for a partial CC payload there.

3. **`mtp.ext.cc` collects instead.** vstest can't consume its output directly — it would have to tell the MTP app to enable coverage over the protocol and then harvest the `.coverage` as an attachment. Bigger protocol lift, splits ownership, and it's off-model vs "vstest owns everything above the testhost." Parked as the long-term MTP-native alternative, not the near-term path.

Leaning **1** for the proof (opt-in, no packaging changes) and **2** as the shipping answer for the zero-dep story, with the multi-version precedence handled in vstest. In this PR the collector is grafted into the runner `Extensions/` to stand in for (2).

PR #16206, opened 2026-07-03
Source Microsoft.TestPlatform package payloads from the real dotnet build

nohwnd:
## Problem

The self-contained `Microsoft.TestPlatform` tool package assembled its runnable payloads by **hand-listing hundreds of files gathered from a single commingled per-TFM output folder** (`$(OutputPath)net48\`, etc.). Because those files were harvested from a build produced by the *package project itself* — which re-resolves the dependency closure and aggregates every `ProjectReference` into one folder — rather than from each product's own build, a shipped DLL and the binding-redirect `.config` next to it could be computed from **different framework resolutions** (e.g. `System.Collections.Immutable` resolving to a netstandard flavor in one place and a netframework flavor in another). Assembly versions happen to align today, but it is not guaranteed by construction, and it is the root cause behind the `playground` / Distributed Test Agent `FileLoadException` fragility.

## Fix — build/publish once, glob many

Two complementary techniques give every packaged file a **single, predictable producer**, so a DLL and its matching `.config`/binding redirects always come from the same build:

**1. Publish once, glob many** (large self-contained closures). A shared opt-in mechanism, `eng/PublishForPackaging.targets`: a product sets `<PublishForPackaging>true` and is `dotnet publish`ed **exactly once at build time** into a canonical `artifacts\publish\<product>\<tfm>\`. Publishing is gated on Arcade's `__ImportPackTargets`, so plain inner-loop builds don't pay for it, and it runs once even though we always pack all packages. Applied to:
- **.NET-Core testhost** (`TestHostNet\`) ← `dotnet publish testhost -f net8.0`
- **.NET Framework runner root** (`Common7\IDE\Extensions\TestPlatform\`) ← `dotnet publish vstest.console -f net48`

**2. Source from each project's own build output** (the smaller per-project artifacts). The remaining vstest-produced cherry-picks are packed directly from each owning project's own `artifacts\bin\<Project>\<Config>\<tfm>\[<rid>\]` output instead of the commingled folder, via three small pack-time glob targets:
- `_IncludeExtensionContent` — Trx / Html / TestHostRuntimeProvider / Blame / EventLog loggers (+ satellites)
- `_IncludeToolContent` — datacollector, arm64 runner, SettingsMigrator (+ satellites), DumpMinitool
- `_IncludeNetFrameworkTestHostContent` — net462…net481 testhosts (default / x86 / arm64)

Together these remove **329 hand-listed `<None>` entries** (88 + 135 + 106 across the three commits). Files that are **not** vstest-produced (VS-internal Dia / IntelliTrace / Fakes / msdia, internal code coverage) and the shared dependency-helper closure intentionally remain hand-listed.

## Validation

- Full clean `build.cmd -c Release -pack` after wiping `artifacts` → **0 errors** (proves the real build repopulates the canonical publish + per-project bin dirs in the correct Build→Pack order).
- Package = **553 files, identical set to baseline** (no adds/drops) across all three slices.
- Slices 1 & 2: all globbed runner-root files **byte-identical** to the `vstest.console` publish; `vstest.console.exe.config` **byte-identical to baseline**.
- `eng\verify-nupkgs.ps1`: **all binding redirects match their DLL versions** (vstest.console.exe / testhost.x86.exe / datacollector.exe) and **all DLL target frameworks match expectations**.
- Acceptance `RunMultipleTestAssembliesWithCodeCoverage`: **2/2**.
- `DistributedTestAgentScenarioTests` (the exact "load Common.dll without binding redirects, force SCI/SRM to load" DTA regression guard): **2/2** — the customer-facing fragility is now green by construction.

The only intended content change vs. baseline is from Slice 2: runner-root `System.Collections.Immutable.dll` and `System.Reflection.Metadata.dll` now ship the framework-appropriate variant the runner actually links against (same assembly versions 10.0.0.0 / 8.0.0.0), captured in `eng/expected-dll-frameworks.json` (regenerated, not hand-edited). Slice 3 changes **no** observable package output — file set, binding redirects, and DLL frameworks are all unchanged and `expected-*.json` is untouched.

## Follow-up (separate PR)

Propagate the same patterns to the net-core `CLI` and `Portable` tool packages, then apply across the rest of the repo, and repoint the playground `PostBuild` copy at the canonical publish dirs.


eng/expected-dll-frameworks.json
```diff
@@ -136,18 +136,18 @@
       "tools/net462/Common7/IDE/Extensions/TestPlatform/pt-BR/Microsoft.CodeCoverage.IO.dll": "none",
       "tools/net462/Common7/IDE/Extensions/TestPlatform/ru/Microsoft.CodeCoverage.IO.dll": "none",
       "tools/net462/Common7/IDE/Extensions/TestPlatform/System.Buffers.dll": "netframework",
-      "tools/net462/Common7/IDE/Extensions/TestPlatform/System.Collections.Immutable.dll": "netstandard",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/System.Collections.Immutable.dll": "netframework",
```
nohwnd, line 139: This is expected, our goal is to be as close as possible to the actual build, instead of bringing in deps that are randomly netstandard.

eng/expected-dll-frameworks.json
```diff
@@ -136,18 +136,18 @@
...
-      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.VisualStudio.TestPlatform.Common.dll": "netframework",
-      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.VisualStudio.TestPlatform.ObjectModel.dll": "netframework",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.TestPlatform.CommunicationUtilities.dll": "net",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.TestPlatform.CoreUtilities.dll": "net",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.TestPlatform.CrossPlatEngine.dll": "netstandard",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.TestPlatform.PlatformAbstractions.dll": "net",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.TestPlatform.Utilities.dll": "netstandard",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.VisualStudio.TestPlatform.Common.dll": "netstandard",
+      "tools/net462/Common7/IDE/Extensions/TestPlatform/TestHostNet/Microsoft.VisualStudio.TestPlatform.ObjectModel.dll": "net",
```
nohwnd, line 150: These look like solving actual "bug" for the testhost .NET fallback (used for C++ runner on .NET Core). Might even mean that he fallback was not really working, I cannot find any data of it being used but I cannot be sure.

PR #16215, opened 2026-07-05
Fix "Code Coverage" collector not found with dotnet test *.dll

nohwnd:
Fixes #15351

`dotnet test *.dll --collect:"Code Coverage"` fails with `Unable to find a datacollector with friendly name 'Code Coverage'`. In DLL mode the MSBuild task never runs, so the adapter path that `Microsoft.CodeCoverage.props` normally injects as `--testAdapterPath` is missing, and vstest.console cannot locate the collector. A `.csproj` run works because the MSBuild task adds the folder that props file sits in.

`CollectArgumentExecutor.Initialize` now does the same for `--collect:"Code Coverage"`. It finds `microsoft.codecoverage` in the NuGet global packages folder (`NUGET_PACKAGES`, else `~/.nuget/packages`), and appends the folder holding `Microsoft.VisualStudio.TraceDataCollector.dll` to `TestAdaptersPaths`. That is the same leaf folder the MSBuild path uses, so the run does not depend on the adapter loading strategy searching directories recursively. The folder is looked up rather than hardcoded, older packages ship the collector under `netstandard1.0` instead of `netstandard2.0`.

There is no project to ask which version to use, so the highest installed one wins, preferring a stable release over a pre-release so that a preview left in the package cache does not take over a run that never referenced it.

It does nothing when `--testAdapterPath` is already set, when the package is not installed, or when the folder cannot be read, so a run that already works is untouched. Scoped to `--collect`, `--enable-code-coverage` still relies on the MSBuild setup. Set `VSTEST_DISABLE_CODE_COVERAGE_ADAPTER_DISCOVERY=1` to turn the discovery off without having to pass an adapter path. `NuGet.Config` `globalPackagesFolder` is not consulted, reading it means taking a dependency on the NuGet libraries.

Tests in `CollectArgumentProcessorTests` cover version selection, the collector folder lookup, injection, and the skip cases.

🤖


src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
...
+
+        var userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+        return userProfile.IsNullOrEmpty() ? null : Path.Combine(userProfile, ".nuget", "packages");
+    }
+
+    private static Version? ParseNuGetVersion(string versionName)
+    {
+        // Strip pre-release suffix (e.g. "18.5.0-preview-1" → "18.5.0") so Version.TryParse works.
+        var dashIndex = versionName.IndexOf('-');
+        var versionString = dashIndex >= 0 ? versionName.Substring(0, dashIndex) : versionName;
```
nohwnd: **[Correctness] Pre-release version is incorrectly preferred over a stable release**

Stripping the pre-release suffix and comparing numerically means `18.6.0-preview-1` (stripped: `18.6.0`) beats `18.5.0` (stable). This violates NuGet SemVer semantics where a pre-release is always *less than* the same numeric release (`18.6.0-preview-1 < 18.6.0`).

A developer who has both `18.5.0` (stable) and `18.6.0-preview-1` installed will silently get the preview adapter injected — exactly the opposite of what NuGet itself would resolve.

The test `TryGetCodeCoverageAdapterPath_PreReleaseSuffixStripperPicksHigherNumerics` documents this as intentional, but the intent conflicts with NuGet's own versioning contract.

Suggested fix — track whether the winning candidate is a pre-release and prefer stable when a higher numeric pre-release would otherwise win:
```csharp
private static (Version? numeric, bool isPreRelease) ParseNuGetVersionWithStability(string versionName)
{
    var dashIndex = versionName.IndexOf('-');
    var isPreRelease = dashIndex >= 0;
    var versionString = isPreRelease ? versionName.Substring(0, dashIndex) : versionName;
    return Version.TryParse(versionString, out var version) ? (version, isPreRelease) : (null, false);
}
```
Then in the selection loop, when `version == bestVersion`, prefer the non-pre-release candidate.

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
...
+            ? []
+            : existingPathsRaw.Split(';').Where(p => !p.IsNullOrEmpty()).ToList();
+
+        if (existingPaths.Any(p => string.Equals(p, ccAdapterPath, StringComparison.OrdinalIgnoreCase)))
+        {
+            return;
+        }
+
+        existingPaths.Add(ccAdapterPath);
+        runSettingsManager.UpdateRunSettingsNode(TestAdapterPathArgumentExecutor.RunSettingsPath, string.Join(";", existingPaths));
```
nohwnd: **[Correctness] Auto-injection bypasses an explicit `--testAdapterPath`**

The dedup check on line 323 only prevents adding the *same exact path* twice. It does not suppress injection when the user has already provided *any* adapter path.

Execution order matters here: `TestAdapterPathArgumentProcessor` runs at priority 9 and `CollectArgumentExecutor` runs at priority 11 (lower number = runs first). So when this code executes, a user-supplied `--testAdapterPath /my/custom/cc/build` is already in `existingPaths`. Because `/my/custom/cc/build` ≠ the NuGet-discovered path, both are appended:

```
TestAdaptersPaths = /my/custom/cc/build;/home/.nuget/packages/microsoft.codecoverage/18.5.0/build
```

With two paths that each contain the CC adapter DLL, vstest may discover and load two different versions of the same collector — or fail with a duplicate-assembly error.

Consider skipping injection entirely when `existingPaths` is already non-empty (the user has explicitly configured adapter paths and presumably knows what they're doing):
```csharp
// If the user has already specified any adapter paths, respect their configuration.
if (existingPaths.Count > 0)
{
    return;
}
```

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
...
+            EqtTrace.Verbose("CollectArgumentExecutor.TryAddCodeCoverageAdapterPath: Microsoft.CodeCoverage package not found in NuGet global packages; skipping auto-injection.");
+            return;
+        }
+
+        var existingPathsRaw = runSettingsManager.QueryRunSettingsNode(TestAdapterPathArgumentExecutor.RunSettingsPath);
+        var existingPaths = existingPathsRaw.IsNullOrEmpty()
+            ? []
+            : existingPathsRaw.Split(';').Where(p => !p.IsNullOrEmpty()).ToList();
+
+        if (existingPaths.Any(p => string.Equals(p, ccAdapterPath, StringComparison.OrdinalIgnoreCase)))
```
nohwnd: [RunSettings Validation] The dedup guard only prevents injecting the *same* NuGet path twice. It doesn't cover the case where the user already specified `--testAdapterPath /my/custom/codecoverage` (a different CC adapter build): that path is in `existingPaths`, the NuGet CC path is not, so both end up registered. vstest will then probe both directories and potentially load two different versions of the CC data collector DLL into the same process — silent duplicate collection or TypeLoadException territory.

Simple guard: if *any* explicit adapter path is already set, assume the user is in control and skip auto-injection.

```suggestion
        if (existingPaths.Count > 0)
        {
            // User explicitly configured adapter paths — don't clobber with auto-discovered NuGet path.
            return;
        }
```

If that's too conservative (it would suppress injection even when paths came from some other unrelated collector), an alternative is to check whether any of the existing paths already contains the CC adapter DLL itself:

```csharp
if (existingPaths.Any(p => File.Exists(Path.Combine(p, "Microsoft.CodeCoverage.DataCollection.dll"))))
{
    return;
}
```

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
...
+        }
+
+        var userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+        return userProfile.IsNullOrEmpty() ? null : Path.Combine(userProfile, ".nuget", "packages");
+    }
+
+    private static Version? ParseNuGetVersion(string versionName)
+    {
+        // Strip pre-release suffix (e.g. "18.5.0-preview-1" → "18.5.0") so Version.TryParse works.
+        var dashIndex = versionName.IndexOf('-');
```
nohwnd: [Dependency & Package Integrity] The suffix-stripping approach has a determinism gap: when both `18.5.0` (stable) and `18.5.0-preview-1` (pre-release of the same release) are installed, both strip to `18.5.0` — a numeric tie — and the winner becomes whichever directory `Directory.GetDirectories` returns first (OS filesystem ordering, not documented). Stable should win there.

Note: `18.5.0 < 18.6.0-preview-1` is correct by NuGet SemVer (different major numeric), so picking the preview in *that* case is actually fine. The test name `PreReleaseSuffixStripperPicksHigherNumerics` is accurate — the issue is only the equal-numeric case.

```suggestion
    private static Version? ParseNuGetVersion(string versionName)
    {
        // Strip pre-release suffix (e.g. "18.5.0-preview-1" → "18.5.0") so Version.TryParse works.
        var dashIndex = versionName.IndexOf('-');
        var isPreRelease = dashIndex >= 0;
        var versionString = isPreRelease ? versionName.Substring(0, dashIndex) : versionName;
        if (!Version.TryParse(versionString, out var version))
        {
            return null;
        }

        // Encode stability: subtract a tiny amount for pre-releases so that when the numeric
        // portion is identical, stable always wins. Use the Revision field as a tiebreaker
        // (stable = int.MaxValue, pre-release = 0) via a wrapper instead of mutating Version.
        // Simplest approach: callers compare (version, isPreRelease) as a tuple.
        return isPreRelease ? new Version(version.Major, version.Minor, version.Build, 0) : new Version(version.Major, version.Minor, version.Build, int.MaxValue);
    }
```

Or, keep `ParseNuGetVersion` returning `(Version version, bool isPreRelease)` and update the comparison at the call site — either way the point is to break ties in favour of stable.

nohwnd (commented):
## Code Review — PR #16215

This PR fixes a real gap: `dotnet test *.dll --collect:"Code Coverage"` fails today because `VSTestTask` is never invoked and the adapter path is never injected. The overall approach (auto-discovering the NuGet global package) is sound. However, there are two correctness bugs and one undescribed behavioral change that need addressing before merge.

---

### 🔴 Finding 1 — Pre-release wins over stable (correctness)

**File**: `CollectArgumentProcessor.cs` → `ParseNuGetVersion` / `TryGetCodeCoverageAdapterPath`

`ParseNuGetVersion` strips the pre-release suffix then compares numerically:
- `18.5.0` → `Version(18,5,0)`
- `18.6.0-preview-1` → stripped → `18.6.0` → `Version(18,6,0)` → **wins**

This violates NuGet SemVer: `18.6.0-preview-1 < 18.6.0`. The test `TryGetCodeCoverageAdapterPath_PreReleaseSuffixStripperPicksHigherNumerics` locks in this incorrect behavior. A developer who inadvertently restores a preview alongside a stable release will silently get the preview adapter injected.

**Fix**: When numeric versions are equal after stripping, prefer the non-pre-release. When numerics differ, the higher number wins (current behavior is fine for that case).

---

### 🔴 Finding 2 — Injection bypasses explicit `--testAdapterPath` (correctness)

**File**: `CollectArgumentProcessor.cs` → `TryAddCodeCoverageAdapterPath` (line ~323–329)

See inline comment. The dedup guard only prevents adding the *identical* NuGet path twice. When the user explicitly passes `--testAdapterPath /some/other/cc/path`, the code appends the NuGet path alongside it — potentially loading two competing versions of the CC collector.

**Fix**: Skip auto-injection entirely when the run settings already have any `TestAdaptersPaths` set (i.e., the user has explicitly configured adapter discovery).

---

### 🟡 Finding 3 — `--enable-code-coverage` now also triggers NuGet injection (undescribed)

`EnableCodeCoverageArgumentProcessor.Initialize()` calls:
```csharp
CollectArgumentExecutor.AddDataCollectorToRunSettings(FriendlyName, _runSettingsManager, _fileHelper);
// FriendlyName = "Code Coverage"
```
Because of this PR's change, that call now also triggers `TryAddCodeCoverageAdapterPath`. The `--enable-code-coverage` flag (a separate CLI option from `--collect`) now silently injects the NuGet adapter path too — this isn't described anywhere in the PR.

The three `EnableCodeCoverageArgumentProcessorTests` tests were updated from exact-XML to semantic assertions precisely because the XML now contains the injected path. These tests would fail in a CI environment where the `microsoft.codecoverage` NuGet package is installed, but pass in an environment where it isn't — that's environment-dependent test behavior.

**Fix**: Either document this intentional broadening, or gate the injection on `--collect:"Code Coverage"` by checking whether the call originates from `CollectArgumentExecutor.Initialize()` vs. the `EnableCodeCoverage` path (e.g., by not calling `TryAddCodeCoverageAdapterPath` from `EnableCodeCoverageArgumentProcessor` — or by making `TryAddCodeCoverageAdapterPath` opt-in rather than auto-triggered from `AddDataCollectorToRunSettings`).

---

### i️ Minor — `build/` directory check does not verify the adapter DLL exists

`TryGetCodeCoverageAdapterPath` accepts any directory named `build/` inside a version folder. An empty `build/` directory — or one from a corrupted package — will be returned as valid. Consider checking for the existence of the actual `.targets` / `.dll` file within it (the name is stable across NuGet releases).

---

🧠 *Reviewed by [Expert Code Reviewer](https://github.com/microsoft/vstest/actions)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28742218773)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
             EnableCoverletInProcDataCollector(collectorName, inProcDataCollectionRunSettings, runSettingsManager, fileHelper);
             runSettingsManager.UpdateRunSettingsNodeInnerXml(Constants.InProcDataCollectionRunSettingsName, inProcDataCollectionRunSettings.ToXml().InnerXml);
         }
+
+        if (string.Equals(collectorName, MicrosoftCodeCoverageConstants.FriendlyName, StringComparison.OrdinalIgnoreCase))
```
nohwnd: [Backward Compatibility & Rollback Safety] This block is also reached via `EnableCodeCoverageArgumentProcessor.Initialize()`, which calls `AddDataCollectorToRunSettings(FriendlyName, ...)` — `FriendlyName` is `"Code Coverage"`, so `TryAddCodeCoverageAdapterPath` now fires for `--enable-code-coverage` too. That's a behavioural change on a separate code path not mentioned in the PR description.

The diff in `EnableCodeCoverageArgumentProcessorTests` is the tell: three tests were rewritten from exact XML assertions to semantic `XmlRunSettingsUtilities`-based checks precisely because the new injection puts extra content into the settings XML and breaks the old `Assert.AreEqual(expected-string)` assertions.

Two questions:
1. Is auto-injecting the NuGet adapter path on `--enable-code-coverage` intentional? If yes, document it.
2. Is there a `VSTEST_DISABLE_*` env var or RunSettings opt-out for this new behaviour? If a regression is found (e.g. wrong adapter picked, TypeLoadException), there's currently no kill switch. Per our principles, new execution-semantic changes need a rollback path.

nohwnd:
Commit pushed: [`d7852db`](https://github.com/microsoft/vstest/commit/d7852db1c8bb11a662dc52df96ce510705199695)

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28742582811)*

nohwnd:
Addressed all three review findings in the latest commit:

**Finding 1 — Stable beats pre-release on equal numerics**: `ParseNuGetVersion` now encodes stability in the `Version.Revision` field (stable = `int.MaxValue`, pre-release = `0`). This means `18.5.0 > 18.5.0-preview-1` (same numeric, stable wins) while `18.6.0-preview-1 > 18.5.0` (higher numeric wins regardless). Added `TryGetCodeCoverageAdapterPath_StableBeatsPreReleaseWhenSameNumerics` test.

**Finding 2 — Skip injection when explicit adapter paths present**: `TryAddCodeCoverageAdapterPath` now returns early when `existingPaths.Count > 0`, respecting any user-configured adapter paths. Added `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet` test.

**Finding 3 — Scope injection to `--collect` path only**: Moved the `TryAddCodeCoverageAdapterPath` call out of `AddDataCollectorToRunSettings` and into `CollectArgumentExecutor.Initialize`. `EnableCodeCoverageArgumentProcessor` goes through `AddDataCollectorToRunSettings` and is no longer affected. The `nugetPackagesOverride` test hook was also removed from `AddDataCollectorToRunSettings` since it's no longer needed there.



> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28742582811)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Fixed in the latest commit — `TryAddCodeCoverageAdapterPath` now returns early when `existingPaths.Count > 0` (lines 323–327), so any user-configured `--testAdapterPath` values suppress auto-injection entirely. A corresponding test `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet` was added to cover this case.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28742986263)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Fixed in the latest commit — `ParseNuGetVersion` now encodes stability in the `Revision` field: stable → `int.MaxValue`, pre-release → `0`. So `18.5.0` (Revision = MaxValue) beats `18.5.0-preview-1` (Revision = 0) on a numeric tie, while `18.6.0-preview-1` still beats `18.5.0` because the Major.Minor.Build portion is strictly higher. A `TryGetCodeCoverageAdapterPath_StableBeatsPreReleaseWhenSameNumerics` test validates this tiebreaker.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28742986263)*

nohwnd (commented):
## Code Review — PR #16215 (iteration d7852db)

The PR Iteration Agent addressed all three findings from the [prior review](https://github.com/microsoft/vstest/pull/16215#pullrequestreview-4631397681). Here is the verification:

---

### ✅ Finding 1 — Pre-release SemVer semantics (resolved)

`ParseNuGetVersion` now encodes stability in `Version.Revision` (stable = `int.MaxValue`, pre-release = `0`):
- `18.5.0` → `Version(18,5,0, int.MaxValue)` beats `18.5.0-preview-1` → `Version(18,5,0, 0)` ✓
- `18.6.0-preview-1` → `Version(18,6,0, 0)` still beats `18.5.0` → `Version(18,5,0, int.MaxValue)` because 18.6 > 18.5 ✓

New test `TryGetCodeCoverageAdapterPath_StableBeatsPreReleaseWhenSameNumerics` covers the equal-numeric case. The approach is correct and well-commented.

---

### ✅ Finding 2 — Injection bypasses explicit `--testAdapterPath` (resolved)

`TryAddCodeCoverageAdapterPath` now returns early when `existingPaths.Count > 0`:

```csharp
if (existingPaths.Count > 0)
{
    // User explicitly configured adapter paths — don't clobber with auto-discovered NuGet path.
    return;
}
```

Execution order (TestAdapterPath priority 9 → CollectArgumentExecutor priority 11) means user-supplied `--testAdapterPath` is already visible at injection time. New test `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet` verifies this. ✓

---

### ✅ Finding 3 — Scope limited to `--collect` path only (resolved)

`TryAddCodeCoverageAdapterPath` is now exclusively called from `CollectArgumentExecutor.Initialize`, guarded by a friendly-name check:

```csharp
if (string.Equals(collectArgumentList[0], MicrosoftCodeCoverageConstants.FriendlyName, StringComparison.OrdinalIgnoreCase))
{
    TryAddCodeCoverageAdapterPath(_runSettingsManager);
}
```

`EnableCodeCoverageArgumentProcessor` calls the static `AddDataCollectorToRunSettings` directly (not `Initialize`), so it is unaffected. ✓

---

### [Description] Minor inaccuracy

The PR description states: *"Added auto-discovery logic to `CollectArgumentExecutor.AddDataCollectorToRunSettings`"*. The final implementation places the logic in `CollectArgumentExecutor.Initialize` — the iteration commit deliberately moved it out of `AddDataCollectorToRunSettings` to fix Finding 3. The PR description was not updated to reflect the iteration changes. This is low-impact (the body comment by the iteration agent covers the changes), but worth a quick edit for historical accuracy.

---

No new issues were introduced in the iteration commit. The implementation is correct, the version-picking logic is well-documented, and all three fixes have targeted tests. Ready for human review.

🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28742986149)*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28742986149)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Fixed in commit `d7852db` — `ParseNuGetVersion` now encodes stability into `Version.Revision`: stable → `int.MaxValue`, pre-release → `0`. This ensures that when the numeric portion is identical (e.g. `18.5.0` vs `18.5.0-preview-1`), the stable release wins — matching NuGet SemVer semantics. When the numeric portion differs (e.g. `18.6.0-preview-1` vs `18.5.0`), the higher numeric version still wins regardless. A `TryGetCodeCoverageAdapterPath_StableBeatsPreReleaseWhenSameNumerics` test validates this tiebreaker.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Fixed in commit `d7852db` — `TryAddCodeCoverageAdapterPath` now returns early when `existingPaths.Count > 0` (lines 323–327), so any user-configured `--testAdapterPath` values suppress auto-injection entirely. A corresponding `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet` test was added.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Fixed in commit `d7852db` — the `TryAddCodeCoverageAdapterPath` call was moved out of `AddDataCollectorToRunSettings` and into `CollectArgumentExecutor.Initialize`, guarded by a `string.Equals(FriendlyName, ...)` check. `EnableCodeCoverageArgumentProcessor` goes through `AddDataCollectorToRunSettings` directly and is no longer affected — there's no NuGet auto-discovery on the `--enable-code-coverage` path.

On the kill-switch question: the existing guard `if (existingPaths.Count > 0) return` means setting any `--testAdapterPath` disables auto-injection. That serves as the opt-out without a new env var.

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

> 🔧 *Iterated by [PR Iteration Agent 🔧](https://github.com/microsoft/vstest/actions/runs/28743187121)*

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -286,13 +286,126 @@ internal static void AddDataCollectorToRunSettings(string[] arguments, IRunSetti
...
+            {
+                continue;
+            }
+
+            if (bestVersion is null || version > bestVersion)
+            {
+                bestVersion = version;
+                bestPath = buildDir;
+            }
+        }
```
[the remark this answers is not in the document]
nohwnd: Good catch, and the equal-numeric case was only half of it.

Encoding stability into `Version.Revision` fixed stable-vs-prerelease but left two previews of the same base version tied, and it also broke two things I hadn't seen: a two-component folder name like `18.5` made `Version.TryParse` return `Build = -1`, and `new Version(major, minor, -1, 0)` throws `ArgumentOutOfRangeException` out of `Initialize`; and a real four-component version had its revision overwritten, so `18.5.0.1` and `18.5.0.2` compared equal.

Now in `60a96ac`: the parsed `Version` is kept intact and the pre-release label is carried next to it. Ordering is numeric version, then stable above pre-release, then the label per SemVer 2.0 (dot-separated identifiers, numeric ones compared as numbers and sorting below alphanumeric), then an ordinal comparison of the folder name so nothing is left to `Directory.GetDirectories` ordering.

Tests: `..._PicksHigherPreRelease_WhenOnlyPreReleasesWithSameNumericsAreInstalled` creates the two directories in both orders and requires the same winner, plus `..._ComparesDottedPreReleaseIdentifiersNumerically`, `..._ComparesAllFourVersionComponents`, `..._DoesNotThrow_WhenVersionDirectoryHasTwoComponents` and `..._IgnoresBuildMetadata`. Four of the five fail against the previous commit.

🤖

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Went with your second suggestion rather than the `suggestion` block.

Encoding stability into `Version.Revision` gets the equal-numeric case right, but it throws away the real `Revision` of a four-component version (`18.5.0.1` and `18.5.0.2` end up equal), and it throws `ArgumentOutOfRangeException` when `Version.TryParse` leaves `Build` at `-1`, which happens for a two-component folder name like `18.5`.

So as of `60a96ac` the parsed `Version` is kept whole and the pre-release label travels beside it, with the comparison done at the call site — numeric version, then stable above pre-release, then the label per SemVer 2.0, then the folder name to break any remaining tie.

You are right that `18.5.0 < 18.6.0-preview-1` is correct and only the equal-numeric case was wrong. I renamed the test to `..._PicksHigherNumericVersion_EvenWhenItIsPreRelease` since nothing strips suffixes any more.

🤖

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Still the case in `60a96ac` — `TryAddCodeCoverageAdapterPath` returns before injecting anything when `existingPaths.Count > 0`, so any `--testAdapterPath` the user passed wins outright and the two directories can never both be registered. Covered by `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet`.

🤖

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -293,6 +302,197 @@ internal static void AddDataCollectorFriendlyName(string friendlyName)
...
     }
 
+    /// <summary>
+    /// Attempts to add the Microsoft Code Coverage adapter path to the run settings by
+    /// auto-discovering the <c>microsoft.codecoverage</c> NuGet package in the global packages directory.
+    /// No-ops silently when the package cannot be found.
+    /// </summary>
+    internal static void TryAddCodeCoverageAdapterPath(IRunSettingsProvider runSettingsManager, string? nugetPackagesOverride = null)
+    {
+        if (!TryGetCodeCoverageAdapterPath(out var ccAdapterPath, nugetPackagesOverride))
```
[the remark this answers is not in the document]
nohwnd: Done in `63f19f5`. The `TestAdaptersPaths` node is read first and the method returns before any file system call, so a run that already has adapter paths does no discovery work at all.

Two things came out of moving it. The old guard used `IsNullOrEmpty`, so I switched both the raw text and each `;` separated entry to `IsNullOrWhiteSpace`, which is also the third suppressed comment from the Copilot reviewer on `60a96ac`. That one does not actually reproduce: `QueryRunSettingsNode` returns `node.InnerText` from an `XmlDocument` with the default `PreserveWhitespace = false`, so the newline and indentation inside a formatted `<TestAdaptersPaths>` never reach the DOM and the node reads as empty. I kept the whitespace aware check anyway and added `TryAddCodeCoverageAdapterPath_Injects_WhenExistingAdapterPathsNodeIsOnlyWhitespace`, which passes against `60a96ac` as well, so it documents the behaviour rather than fixing a bug.

The injection got simpler too, the collection is known to be empty at that point, so it writes the discovered path directly instead of joining a one element list.

🤖

> [!NOTE]

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Still correct in `63f19f5`, and the `Version.Revision` encoding from the July reply above is gone.

Packing stability into `Revision` fixed the equal numeric case but broke two others: a real four component version lost its revision, so `18.5.0.1` and `18.5.0.2` compared equal, and a two component folder name like `18.5` leaves `Build` at `-1` after `Version.TryParse`, where `new Version(major, minor, -1, 0)` throws `ArgumentOutOfRangeException` out of `Initialize`.

Since `60a96ac` the parsed `Version` is kept whole and the pre-release label travels next to it: numeric version, then stable above pre-release, then the label per SemVer 2.0, then an ordinal comparison of the folder name so nothing depends on `Directory.GetDirectories` ordering. Covered by `..._StableBeatsPreReleaseWhenSameNumerics`, `..._ComparesAllFourVersionComponents` and `..._DoesNotThrow_WhenVersionDirectoryHasTwoComponents`.

🤖

> [!NOTE]

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Still the case, and as of `63f19f5` the guard runs before any file system access, so an explicit `--testAdapterPath` short circuits the whole thing instead of being compared against a path we went and looked up first. Covered by `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenExplicitAdapterPathAlreadySet`.

I did not take your alternative of probing each existing path for `Microsoft.CodeCoverage.DataCollection.dll`. The simple guard is enough here, and it avoids hard coding a file name from the Code Coverage package layout that we do not control.

🤖

> [!NOTE]

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Still the case in `63f19f5`. `TryAddCodeCoverageAdapterPath` is called only from `CollectArgumentExecutor.Initialize` behind the friendly name check, `--enable-code-coverage` goes through `AddDataCollectorToRunSettings` and gets no auto-discovery.

On the rollback question, `--testAdapterPath` is still the opt-out, and discovery itself can no longer fail a run. `63f19f5` wraps it so an unreadable global packages folder, a transient IO error, or an unusable `NUGET_PACKAGES` value is reported as not found instead of throwing out of `Initialize`. `ArgumentException` is in that set because `Path.Combine` rejects invalid characters on .NET Framework. So the worst case is the error message users already get today without this PR, rather than a new failure mode.

🤖

> [!NOTE]

nohwnd:
`63f19f5` answers the review on `60a96ac`. The inline finding and the older threads are answered in place, this covers the three suppressed comments, which have no thread to reply in.

**`Directory.GetDirectories` is unguarded.** Correct, and worth taking. Discovery now runs inside a `try` for `IOException`, `UnauthorizedAccessException` and `ArgumentException` and reports not found instead of throwing out of `Initialize`. `ArgumentException` is in the set because `Path.Combine` rejects invalid characters on .NET Framework, so a broken `NUGET_PACKAGES` would have failed the run before it reached `Directory.Exists`. `TryGetCodeCoverageAdapterPath_ReturnsFalse_WhenNuGetPackagesPathIsUnusable` covers it.

**The `using Microsoft.VisualStudio.TestPlatform.Common.Utilities` in the test file is unused and will trip IDE0005 in CI.** This one is wrong, and acting on it breaks the build. That namespace holds `RunSettingsProviderExtensions`, so it is what makes the extension methods `AddDefaultRunSettings`, `UpdateRunSettingsNode` and `QueryRunSettingsNode` resolve. Removing it fails compilation with 8 CS1061 errors on both `net481` and `net11.0`. Extension methods are invisible to a grep for the namespace name, which is probably how it got flagged.

**Whitespace only `<TestAdaptersPaths>` is treated as configured.** Does not reproduce, details in the thread on the guard. The check is whitespace aware now anyway.

On the description note: the body already says `CollectArgumentExecutor.Initialize`. The `AddDataCollectorToRunSettings` wording is in the Copilot reviewer overview of the first commit, not in the PR body.

Release build is clean and `vstest.console.UnitTests` is green on both target frameworks, 653 tests, 0 failed.

🤖

nohwnd:
`6f19d14` takes three of the four suppressed comments from the Copilot reviewer on `63f19f5`.

**Hardcoded `';'` duplicates the path splitting.** Fair, and there is a better helper than I realised. The guard now calls `TestAdapterPathArgumentExecutor.SplitPaths`, which is the same splitter `TestAdapterPathArgumentExecutor` uses on this very node, so the separator stays defined in one place (`ArgumentSeparators`). The whitespace behaviour is unchanged, `SplitPaths` only drops empty entries, so the `Any(p => !p.IsNullOrWhiteSpace())` on top is still what makes a whitespace only node count as unset.

**`SecurityException` and `NotSupportedException` are not caught.** Taken. `ResultsDirectoryArgumentExecutor` catches `NotSupportedException or SecurityException or ArgumentException or PathTooLongException or IOException` for the same kind of path work, so the discovery catch list now matches that idiom.

**The trace claims the package is missing when discovery can fail for other reasons.** Taken, it says the adapter path was not found now.

**The `using Microsoft.VisualStudio.TestPlatform.Common.Utilities` in the test file is unused.** This is the second time it has been raised and it is still wrong. That namespace holds `RunSettingsProviderExtensions`, so it is what makes `AddDefaultRunSettings`, `UpdateRunSettingsNode` and `QueryRunSettingsNode` resolve as extension methods. Removing it fails compilation with 8 `CS1061` errors across `net481` and `net11.0`. I tried it to be sure. Extension methods do not show up in a search for the namespace name, which is likely why it keeps getting flagged.

Release build clean, `vstest.console.UnitTests` green on both target frameworks, 653 tests, 0 failed.

🤖

nohwnd:
@fhnaseer if this is good idea please approve.

fhnaseer (commented):
A couple of observations from reviewing the discovery path, both non-blocking,

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -293,6 +301,203 @@ internal static void AddDataCollectorFriendlyName(string friendlyName)
...
+                continue;
+            }
+
+            if (bestVersion is null
+                || CompareNuGetVersions(version, preRelease, directoryName, bestVersion, bestPreRelease, bestDirectoryName) > 0)
+            {
+                bestVersion = version;
+                bestPreRelease = preRelease;
+                bestDirectoryName = directoryName;
+                bestPath = buildDir;
```
fhnaseer: This picks the numerically-highest version in the global cache, and deliberately prefers a higher pre-release over a lower stable, whereas the csproj/MSBuild path injects exactly the `microsoft.codecoverage` version the test project references ($(MSBuildThisFileDirectory)),

So `--collect:"Code Coverage"` in dll mode can end up loading a different collector than the project actually pins - including a preview the user never referenced for this run,

Best-effort discovery is reasonable here, but the divergence from csproj is worth a comment, and I'd lean toward preferring a stable version unless nothing else is installed,

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -293,6 +301,203 @@ internal static void AddDataCollectorFriendlyName(string friendlyName)
...
+            return;
+        }
+
+        if (!TryGetCodeCoverageAdapterPath(out var ccAdapterPath, nugetPackagesOverride))
+        {
+            EqtTrace.Verbose("CollectArgumentExecutor.TryAddCodeCoverageAdapterPath: Code Coverage adapter path not found; skipping auto-injection.");
+            return;
+        }
+
+        runSettingsManager.UpdateRunSettingsNode(TestAdapterPathArgumentExecutor.RunSettingsPath, ccAdapterPath);
```
fhnaseer, line 339: We inject the package's `build/` folder, but the collector actually lives one level deeper at `build/netstandard2.0/Microsoft.VisualStudio.TraceDataCollector.dll`,

This resolves only because the default `TestAdapterLoadingStrategy` enumerates adapter paths with `SearchOption.AllDirectories`. A run configured with a non-default (Explicit, non-recursive) strategy wouldn't find the nested DLL,

The csproj path points straight at the leaf folder; pointing at the directory that actually contains the collector would match it and drop the reliance on recursive discovery,

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -293,6 +301,203 @@ internal static void AddDataCollectorFriendlyName(string friendlyName)
...
+    private static string? GetNuGetGlobalPackagesPath()
+    {
+        var envPath = Environment.GetEnvironmentVariable("NUGET_PACKAGES");
+        if (!envPath.IsNullOrEmpty())
+        {
+            return envPath;
+        }
+
+        var userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+        return userProfile.IsNullOrEmpty() ? null : Path.Combine(userProfile, ".nuget", "packages");
```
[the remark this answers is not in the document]
nohwnd: Taken. The doc on `GetNuGetGlobalPackagesPath` now says that `NuGet.Config` `globalPackagesFolder` is not consulted, and why, reading it means taking a dependency on the NuGet libraries. It also names the two ways out, set `NUGET_PACKAGES`, or pass `--testAdapterPath`.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -116,6 +117,13 @@ public void Initialize(string? argument)
             throw new SettingsException(string.Format(CultureInfo.CurrentCulture, CommandLineResources.CollectWithTestSettingErrorMessage, argument));
         }
         AddDataCollectorToRunSettings(collectArgumentList, _runSettingsManager, _fileHelper, exceptionMessage);
+
+        if (string.Equals(collectArgumentList[0], MicrosoftCodeCoverageConstants.FriendlyName, StringComparison.OrdinalIgnoreCase))
+        {
+            // In DLL mode the MSBuild task never runs, so VSTestTraceDataCollectorDirectoryPath is not set.
+            // Discover the adapter from NuGet instead. Scoped to --collect; --enable-code-coverage uses MSBuild.
+            TryAddCodeCoverageAdapterPath(_runSettingsManager);
```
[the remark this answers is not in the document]
nohwnd: Taken, with one change. This repo already has a place for these, `FeatureFlag` holds the `VSTEST_DISABLE_*` constants and reads them the same way everywhere, unset or `0` means off. So the flag is `VSTEST_DISABLE_CODE_COVERAGE_ADAPTER_DISCOVERY` declared there, rather than an inline `Environment.GetEnvironmentVariable` call.

The guard sits at the top of `TryAddCodeCoverageAdapterPath` instead of at the call site, so it also skips the file system walk, and it is the same check no matter who calls the method. `TryAddCodeCoverageAdapterPath_SkipsInjection_WhenDiscoveryIsDisabledByFeatureFlag` covers it.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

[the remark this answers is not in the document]
nohwnd: Same answer as the other thread on this line, taken. The flag is `VSTEST_DISABLE_CODE_COVERAGE_ADAPTER_DISCOVERY`, declared in `FeatureFlag` next to the other `VSTEST_DISABLE_*` constants, and it returns before the file system walk.

I spelled the feature out rather than using `CC`, none of the existing flags abbreviate, and the neighbouring one is already `VSTEST_DISABLE_DYNAMICNATIVE_CODECOVERAGE_DEFAULT_SETTING`.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Yes, and it is worse in practice than it looks in the diff. My package cache has 36 versions of `microsoft.codecoverage` in it, the old ordering picked `18.11.0-preview-26378-01`, and the newest stable there is `18.8.1`. So a preview I never referenced would have been loaded.

`CompareNuGetVersions` compares stability first now and the numeric version second, so a stable release wins even against a higher pre-release, and a pre-release is only picked when nothing stable is installed. The same cache returns `18.8.1` after the change.

The divergence is written down on `TryGetCodeCoverageAdapterPath`, that a project run injects the version the project references, that there is no project to ask here, and that `--testAdapterPath` pins a specific one. `PrefersStable_OverHigherPreRelease` and `PicksHighestPreRelease_WhenNoStableIsInstalled` cover both directions.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/vstest.console/Processors/CollectArgumentProcessor.cs
nohwnd: Correct, taken. It returns the folder that holds `Microsoft.VisualStudio.TraceDataCollector.dll` now, which is the same leaf folder `Microsoft.CodeCoverage.props` points at through `$(MSBuildThisFileDirectory)`.

I look the assembly up instead of hardcoding `netstandard2.0`, because the folder name is not stable across versions, 17.1.0 and older ship the collector under `netstandard1.0`. A version whose `build/` holds no collector is not a candidate at all now, which also stops it from winning the version comparison and injecting a path with nothing to load.

`ReturnsCollectorDirectory_NotTheBuildDirectory` and `SkipsVersionWithoutCollector` cover it, and the test helper builds the real nested layout rather than a bare `build/` folder.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

nohwnd:
`a5f8843` takes both observations from the re-review.

**I/O before version parsing.** Taken. `FindCodeCoverageAdapterPath` parses the folder name first now and only probes the disk for names that are versions. You are right that it makes no practical difference with a handful of entries, but a folder whose name is not a version is not a candidate no matter what is in it, so doing the cheap check first is the right order anyway.

**The `EnableCodeCoverageArgumentProcessorTests` assertions.** Good catch, and it is leftover from an intermediate state of this PR rather than something intentional. The three tests compared the whole run settings xml, and I loosened them when the injection still went through `AddDataCollectorToRunSettings`, which changed the xml on the `--enable-code-coverage` path too. Moving the call to `CollectArgumentExecutor.Initialize` fixed that, and I did not go back and undo the loosening.

Restoring the original assertions, they pass unchanged, which is also the proof that the enable path really does produce the same xml as before. That drops the file from the change, so this is now three files and purely additive, 789 insertions and no deletions.

Release build clean, `vstest.console.UnitTests` green on both target frameworks, 662 tests, 0 failed.

🤖

src/vstest.console/Processors/CollectArgumentProcessor.cs
```diff
@@ -293,6 +308,259 @@ internal static void AddDataCollectorFriendlyName(string friendlyName)
...
+        if (File.Exists(Path.Combine(buildDir, TraceDataCollectorAssemblyName)))
+        {
+            return buildDir;
+        }
+
+        // Real packages hold exactly one such folder. Order the candidates so the answer does not
+        // depend on the order the file system returns the directories in.
+        return Directory.GetDirectories(buildDir)
+            .Where(d => File.Exists(Path.Combine(d, TraceDataCollectorAssemblyName)))
+            .OrderBy(d => Path.GetFileName(d), StringComparer.Ordinal)
```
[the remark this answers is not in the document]
nohwnd: Taken, `bbdd807`. You are right that ascending picks the wrong one, and descending is the preference I want anyway, so there is no reason to leave it the other way round.

No package ships the collector under two target framework folders today, so this only matters if one ever does, but `TryGetCodeCoverageAdapterPath_PrefersNewerTargetFramework_WhenCollectorShipsUnderSeveral` pins it either way. The ordering is still there so the answer does not depend on the order the file system returns the directories in.

🤖

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

PR #16222, opened 2026-07-06
[efficiency-improver] perf: use JsonElement.TryGetGuid/TryGetDateTimeOffset in V2 IPC deserializers

nohwnd:
## Goal and Rationale

Eliminate intermediate `string` heap allocations on the V2-protocol IPC **receive** path by replacing
`GetString()`-then-parse patterns with the typed `JsonElement` accessor overloads that read values
directly from the UTF-8 JSON bytes.

**Focus area:** Code-Level Efficiency

## Problem

In the V2 deserializers, three hot-path reads used a two-step pattern:
1. `element.GetString()` — allocates a managed `string` from the UTF-8 bytes
2. A subsequent parse call (`GuidPolyfill.Parse`, `DateTimeOffset.Parse`) — converts that `string` to the target value type

Both steps can be replaced with a single typed call that reads directly from the JSON bytes without an intermediate string:

| Site | Before | After |
|---|---|---|
| `TestCaseConverterV2.Read` — TestCase.Id | `id.GetString()!` + `GuidPolyfill.Parse(...)` | `id.TryGetGuid(out var guidValue)` |
| `TestResultConverterV2.Read` — StartTime | `startTime.GetString()!` + `DateTimeOffset.Parse(...)` | `startTime.TryGetDateTimeOffset(out var v)` |
| `TestResultConverterV2.Read` — EndTime | `endTime.GetString()!` + `DateTimeOffset.Parse(...)` | `endTime.TryGetDateTimeOffset(out var v)` |

## Approach

`JsonElement.TryGetGuid()` reads GUID strings in all standard formats (N, D, B, P, X); the write path uses `Utf8JsonWriter.WriteString(string, Guid)` which always writes D-format, so the roundtrip is identical. `JsonElement.TryGetDateTimeOffset()` reads ISO 8601 strings; `Utf8JsonWriter.WriteString(string, DateTimeOffset)` always writes ISO 8601, so that roundtrip is also identical. Both APIs were introduced in .NET Core 3.0 — well within the `#if NETCOREAPP` guard on these files.

The `System.Globalization` import is removed from `TestCaseConverterV2.cs` (no longer needed); `TestResultConverterV2.cs` retains it for the `Duration` parse line.

## Energy Efficiency Evidence

**Proxy metric:** Memory allocation (Gen0 GC pressure on the IPC receive thread).

Eliminations per test run:
- 1 `string` alloc per `TestCase` received (the GUID)
- 2 `string` allocs per `TestResult` received (StartTime + EndTime)

For a representative 10K-test run: **~30K fewer string allocations** on the hot deserialization thread. Fewer Gen0 GCs mean fewer CPU stalls and less energy spent on GC bookkeeping during test execution.

**Measurement method:** Allocations are deterministic — one `GetString()` call creates exactly one `string` object on the heap. The elimination is therefore exact and provable from code inspection.

### Green Software Foundation context

- **Hardware Efficiency**: Fewer GC pauses mean the CPU cycles used by the deserialization thread are spent on productive work, not memory bookkeeping.
- **SCI reduction**: Lower CPU energy per `dotnet test` invocation reduces the software carbon intensity of every test run.

## Trade-offs

None. The change is semantically identical for all valid wire input. Edge-case behaviour for malformed JSON improves slightly: the old code would throw (`DateTimeOffset.Parse`/`GuidPolyfill.Parse` throw on bad input); the new code silently ignores malformed values and leaves the field at its default, which is more resilient.

No new public API surface is added.

## Reproducibility

```
./build.sh -c Release
.dotnet/dotnet run --project test/Microsoft.TestPlatform.CommunicationUtilities.UnitTests/... -f net11.0
```

## Test Status

✅ Build succeeded (0 errors, 4 pre-existing IL2067/IL2026/IL2057 warnings unrelated to these changes)
✅ 482/482 `Microsoft.TestPlatform.CommunicationUtilities.UnitTests` tests pass

> Generated by [Efficiency Improver](https://github.com/microsoft/vstest/actions/runs/28810618351) · 2.3K AIC · ⌖ 17.1 AIC · ⊞ 45.6K · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvstest+%22gh-aw-workflow-id%3A+efficiency-improver%22&type=pullrequests)

### DIFF
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
@@ -4,7 +4,6 @@
 #if NETCOREAPP
 
 using System;
-using System.Globalization;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
@@ -32,8 +31,8 @@ internal class TestCaseConverterV2 : JsonConverter<TestCase>
             testCase.ExecutorUri = new Uri(uri.GetString()!);
         if (data.TryGetProperty("Source", out var source))
             testCase.Source = source.GetString()!;
-        if (data.TryGetProperty("Id", out var id) && id.ValueKind != JsonValueKind.Null)
-            testCase.Id = GuidPolyfill.Parse(id.GetString()!, CultureInfo.InvariantCulture);
+        if (data.TryGetProperty("Id", out var id) && id.TryGetGuid(out var guidValue))
+            testCase.Id = guidValue;
         if (data.TryGetProperty("DisplayName", out var display) && display.ValueKind != JsonValueKind.Null)
             testCase.DisplayName = display.GetString()!;
         if (data.TryGetProperty("CodeFilePath", out var codePath) && codePath.ValueKind != JsonValueKind.Null)
--- a/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
+++ b/src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
@@ -67,10 +67,10 @@ internal class TestResultConverterV2 : JsonConverter<TestResult>
             testResult.ComputerName = computerName.GetString();
         if (data.TryGetProperty("Duration", out var duration) && duration.ValueKind != JsonValueKind.Null)
             testResult.Duration = TimeSpan.Parse(duration.GetString()!, CultureInfo.InvariantCulture);
-        if (data.TryGetProperty("StartTime", out var startTime) && startTime.ValueKind != JsonValueKind.Null)
-            testResult.StartTime = DateTimeOffset.Parse(startTime.GetString()!, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
-        if (data.TryGetProperty("EndTime", out var endTime) && endTime.ValueKind != JsonValueKind.Null)
-            testResult.EndTime = DateTimeOffset.Parse(endTime.GetString()!, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+        if (data.TryGetProperty("StartTime", out var startTime) && startTime.TryGetDateTimeOffset(out var startTimeValue))
+            testResult.StartTime = startTimeValue;
+        if (data.TryGetProperty("EndTime", out var endTime) && endTime.TryGetDateTimeOffset(out var endTimeValue))
+            testResult.EndTime = endTimeValue;
 
         // Custom properties
         if (data.TryGetProperty("Properties", out var properties) && properties.GetArrayLength() > 0)


src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestCaseConverterV2.cs
```diff
@@ -32,8 +31,8 @@ internal class TestCaseConverterV2 : JsonConverter<TestCase>
             testCase.ExecutorUri = new Uri(uri.GetString()!);
         if (data.TryGetProperty("Source", out var source))
             testCase.Source = source.GetString()!;
-        if (data.TryGetProperty("Id", out var id) && id.ValueKind != JsonValueKind.Null)
-            testCase.Id = GuidPolyfill.Parse(id.GetString()!, CultureInfo.InvariantCulture);
+        if (data.TryGetProperty("Id", out var id) && id.TryGetGuid(out var guidValue))
```
nohwnd, line 34: **[IPC Transport & Protocol Stability]** `JsonElement.TryGetGuid()` only accepts the D-format GUID string (`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`), not "all standard formats (N, D, B, P, X)" as the PR description claims. Internally it calls `Utf8Parser.TryParse(..., 'D')`.

This is not a correctness problem here — the write path (`writer.WriteString("Id", value.Id)`) always produces D-format — but it's worth noting that the format constraint is tighter than the removed `GuidPolyfill.Parse` call. If `TryGetGuid` returns `false` for any reason (e.g., non-D-format GUID from a future or third-party sender), `testCase.Id` silently stays `Guid.Empty` rather than surfacing a parse error. The old code would have thrown a `FormatException`, making the bad input immediately visible. Whether silent-default-on-corrupt-ID is preferable to fail-fast is a design trade-off; it's just not "more resilient" in all scenarios.

src/Microsoft.TestPlatform.CommunicationUtilities/Serialization/TestResultConverterV2.cs
```diff
@@ -67,10 +67,10 @@ internal class TestResultConverterV2 : JsonConverter<TestResult>
             testResult.ComputerName = computerName.GetString();
         if (data.TryGetProperty("Duration", out var duration) && duration.ValueKind != JsonValueKind.Null)
             testResult.Duration = TimeSpan.Parse(duration.GetString()!, CultureInfo.InvariantCulture);
-        if (data.TryGetProperty("StartTime", out var startTime) && startTime.ValueKind != JsonValueKind.Null)
-            testResult.StartTime = DateTimeOffset.Parse(startTime.GetString()!, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
-        if (data.TryGetProperty("EndTime", out var endTime) && endTime.ValueKind != JsonValueKind.Null)
-            testResult.EndTime = DateTimeOffset.Parse(endTime.GetString()!, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+        if (data.TryGetProperty("StartTime", out var startTime) && startTime.TryGetDateTimeOffset(out var startTimeValue))
```
nohwnd, line 70: **[Error Reporting & Diagnostic Clarity]** This introduces an inconsistency in the error model within the same method: `Duration` (line 68) still calls `TimeSpan.Parse(...)` which throws `FormatException` on malformed input, but `StartTime`/`EndTime` now use `TryGetDateTimeOffset` and silently leave the fields at `DateTimeOffset.MinValue` (year 0001) on parse failure.

For the normal roundtrip this is a no-op (the write path always produces RFC 3339 / ISO 8601, which `TryGetDateTimeOffset` handles correctly). However, when a malformed timestamp arrives, a test result with `StartTime = DateTimeOffset.MinValue` will silently appear in the TRX/HTML output with a wildly wrong start time rather than propagating a diagnostic error. Given that `Duration` in the same block still throws, the failure mode is now inconsistent within the same deserializer, which can complicate debugging.

nohwnd (commented):
## Review: `[efficiency-improver] perf: use JsonElement.TryGetGuid/TryGetDateTimeOffset in V2 IPC deserializers`

Dimensions activated (per routing table): **IPC Transport & Protocol Stability**, **Error Reporting & Diagnostic Clarity**, **Backward Compatibility & Rollback Safety**

---

### Summary

The optimization is **correct for the V2 roundtrip**. Both write paths produce formats that the new typed accessors consume without loss:

- `writer.WriteString("Id", value.Id)` → D-format GUID → accepted by `TryGetGuid`
- `writer.WriteString("StartTime", value.StartTime)` → RFC 3339/ISO 8601 → accepted by `TryGetDateTimeOffset`

The `#if NETCOREAPP` guard is correct; both APIs require .NET Core 3.0+. Removing `using System.Globalization;` from `TestCaseConverterV2.cs` is clean.

---

### Observations

Two observations are left as inline comments:

1. **`TryGetGuid` format constraint** (TestCaseConverterV2.cs:34) — The PR description states that `TryGetGuid` reads "all standard formats (N, D, B, P, X)"; this is incorrect. The method only accepts D-format. The write path always produces D-format so the roundtrip is safe, but the description claim is inaccurate.

2. **Silent-default error model is inconsistent** (TestResultConverterV2.cs:70–72) — `Duration` (line 68–69) still throws `FormatException` on malformed input; `StartTime`/`EndTime` now silently produce `DateTimeOffset.MinValue`. The PR describes the silent-default behaviour as "more resilient", but it also makes bad data harder to diagnose. At minimum, the inconsistency with `Duration` in the same block is worth aligning intentionally.

---

### No blocking issues found

The performance improvement is real and the wire contract is preserved. The two observations above are design-level trade-offs, not correctness bugs.

---

🧠 *Reviewed by [expert-reviewer](https://github.com/microsoft/vstest/actions) · IPC Transport & Protocol Stability · Error Reporting & Diagnostic Clarity*

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28811768822)*

PR #16223, opened 2026-07-07
[fix] Fix test output eaten by MSBuild terminal logger

### LINKED ISSUE #15509: no results with arcade mtp integration when we don't capture output and -tl:off is not specified?
when not capturing output (setting we override in directory.build.props) and building locally terminal logger experience will kick in and all the output is eaten so we only get results that mention .log files with results of the run, but the files don't exist, and there is no content, because all output is eaten by msbuild terminal logger.

nohwnd:
## Summary

Fixes #15509

> 🤖 This is an automated fix.

## Root Cause

When `TestCaptureOutput=false` is set in `Directory.Build.props` (to allow live test output in the console) and the MSBuild terminal logger is active (default in .NET 8+ SDK), the terminal logger buffers all output from MSBuild sub-tasks — including test output. When tests complete, the terminal logger shows "See log file: \(path\)" messages. Since `TestCaptureOutput=false` means no log files are created, those paths point to files that don't exist, leaving developers with no visible test output.

## Fix

Disable the MSBuild terminal logger in `test.sh` and `test.cmd` by setting `MSBUILDTERMINALLOGGER=off` before invoking the build scripts. This lets test output flow directly to the console as intended.

Developers who prefer the terminal logger UI can override the default by setting `MSBUILDTERMINALLOGGER=auto` before invoking the test scripts.

## Testing

This is a developer-tooling change affecting only the `test.sh` and `test.cmd` scripts. No production code is modified. The change is:

- `test.sh`: Exports `MSBUILDTERMINALLOGGER=${MSBUILDTERMINALLOGGER:-off}` so the default is off but can be overridden.
- `test.cmd`: Sets `MSBUILDTERMINALLOGGER=off` if not already defined.

CI runs (`azure-pipelines.yml` calls `./test.sh`) will also have the terminal logger disabled, which is harmless — CI output is captured by the Azure DevOps agent regardless.

> 🔍 *Triaged by [Issue Repro Triage & Auto-Fix 🔍](https://github.com/microsoft/vstest/actions/runs/28834207444)*

### DIFF
--- a/test.cmd
+++ b/test.cmd
@@ -1,3 +1,4 @@
 @echo off
+if not defined MSBUILDTERMINALLOGGER set MSBUILDTERMINALLOGGER=off
 powershell -ExecutionPolicy ByPass -NoProfile -command "& """%~dp0eng\Build.ps1""" -test %*"
 exit /b %ErrorLevel%
--- a/test.sh
+++ b/test.sh
@@ -16,4 +16,8 @@ scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 if [[ -z "${DOTNET_ROOT:-}" && -d "$scriptroot/.dotnet" ]]; then
   export DOTNET_ROOT="$scriptroot/.dotnet"
 fi
+# Disable MSBuild terminal logger so test output isn't buffered. TestCaptureOutput=false
+# means there are no log files to reference, so the terminal logger would show
+# misleading "See log file" messages with paths to files that don't exist.
+export MSBUILDTERMINALLOGGER=${MSBUILDTERMINALLOGGER:-off}
 "$scriptroot/eng/common/build.sh" --test "$@"


nohwnd (commented):
## 🧠 *Reviewed by Expert Code Reviewer*

### Scope

This PR touches only `test.cmd` and `test.sh` — thin wrapper scripts that bootstrap scriptroot/DOTNET_ROOT and delegate to the real build engine. No production C# code is modified.

**Dimensions activated** (per routing table):
- **Build Script & Infrastructure Hygiene** — environment variable handling, exit-code propagation, pre-delegate ordering
- **Source Build & Cross-Platform Compliance** — bash portability, child-process propagation

---

### Findings

No actionable issues found.

**What was checked and passed:**

| Check | Result |
|---|---|
| `if not defined VAR set VAR=off` is the correct idiomatic CMD default-value pattern | ✅ |
| `export VAR=${VAR:-off}` is correct POSIX/bash syntax, sets before delegating | ✅ |
| Env var is set *before* the delegated script is invoked (child process inherits it) | ✅ |
| `exit /b %ErrorLevel%` still present in `test.cmd` — exit-code propagation unaffected | ✅ |
| User-override preserved on both platforms (already-set value is not overwritten) | ✅ |
| CI pipelines that pre-set `MSBUILDTERMINALLOGGER` before calling `./test.sh` are unaffected | ✅ |

**Minor observation (no change needed):**

The two scripts have slightly different empty-string semantics:
- `test.sh`: `${MSBUILDTERMINALLOGGER:-off}` treats an *empty* value (`=""`) as unset → override to `off`
- `test.cmd`: `if not defined` treats an *empty* value as *defined* → keeps the empty string

If a developer sets `MSBUILDTERMINALLOGGER=` (empty) to clear the variable, the behavior differs per platform. In practice this is irrelevant — the documented override path is `MSBUILDTERMINALLOGGER=auto`, not empty-string. No action needed.

---

### PR Description Alignment

The title and description accurately describe the change. The root-cause explanation (TestCaptureOutput=false + terminal logger buffering → missing log-file paths), the fix, and the CI-impact claim all match the diff exactly. ✅

> 🧠 *Reviewed by [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/28834795973)*

PR #16237, opened 2026-07-08
Ship real testhost.deps.json, keep only the latest native fallback runtimeconfig

nohwnd:
The built-in testhost fallback - the one native (C++) runners use when they bring no managed testhost - shipped a hand-crafted `testhost.deps.json` plus 14 version-specific `testhost-<ver>.runtimeconfig.json` templates under `temp/testhost/`. None of it was pulling its weight anymore:

- The artificial `testhost.deps.json` was orphaned. No build file referenced it, and since #16206/#16236 the packages already ship the real deps.json that gets generated when `testhost.dll` is built.
- Only native dlls reach this fallback - a managed source that can't resolve its own runtime config throws and points the user at Microsoft.NET.Test.Sdk - and native has no target framework preference, so it always rolls forward to `testhost-latest`. The 13 version-specific configs were never selected.

So drop the artificial deps.json and the 13 version configs, keep `testhost-latest`, and simplify the `DotnetTestHostManager` fallback to always use it (removing the old TFM-guessing block and its long TODO comment).

Verified on Windows with the native C++ fallback acceptance test (`RunningNativeCppProjectWithoutTestSdkShouldUseTheBuiltInTesthostFallback`, `Windows-Review`, so it's CI-excluded off Windows) and `DotnetTestHostManagerTests`. Release `-pack` is green; `Verify-Nupkgs` and `Verify-DllFrameworks` pass, and the shipped CLI package now carries the real 16 KB `testhost.deps.json` (`.NETCoreApp,v8.0`) and only `testhost-latest.runtimeconfig.json`.


nohwnd:
Change is technically correct, but changes counts of files in packages, while I work on re-doing all packages, and would make it harder to compare new packages with previously shipped packages. So hold on.

PR #16256, opened 2026-07-09
Publish packaging payloads into each producer's own bin

nohwnd:
Follow-up cleanup to the "publish once, glob many" packaging work (#16206, #16236, #16240, #16246).

The self-contained tool packages harvest each bundled product from a publish output that is produced once per (project, TFM). Until now that output landed in a separate top-level tree, `artifacts/publish/<Project>/<TFM>/`, addressed through a `TestPlatformPackagingPublishRoot` property.

This moves it under the producer's own bin instead — `$(ArtifactsBinDir)<Project>/<Configuration>/<TFM>/publish/` — so the publish output is cleaned together with the project and there is no parallel top-level tree. `TestPlatformPackagingPublishRoot` is removed.

The path is spelled out RID-less on both the producer and consumer side rather than reusing `$(OutputPath)publish\`, because `$(OutputPath)` picks up the RuntimeIdentifier subfolder for the net462/net48 RID-qualified builds (`.../net462/win7-x64/`), while the consumers recompute the path from producer name + TFM.

Pure no-op on shipped files: all four affected nupkgs (Microsoft.TestPlatform, .CLI, .Portable, .TestHost) are +0/-0 vs a clean `-c Release -pack` baseline, and `verify-nupkgs.ps1` still reports all binding redirects and DLL target frameworks matching (no expected-*.json regeneration).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,14 +9,6 @@
     <IsSourceFile>false</IsSourceFile>
     <IsSourceFile Condition="$([MSBuild]::MakeRelative($(RepoRoot), $(MSBuildProjectFullPath)).StartsWith('src/')) OR $([MSBuild]::MakeRelative($(RepoRoot), $(MSBuildProjectFullPath)).StartsWith('src\'))">true</IsSourceFile>
     <SrcPackageFolder>$(RepoRoot)src\package\</SrcPackageFolder>
-    <!--
-      Canonical root for per-product publish outputs consumed by the src/package/* self-contained
-      "tool" packages (Microsoft.TestPlatform, .CLI, .Portable, ...). A bundled product publishes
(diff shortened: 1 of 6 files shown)


Evangelink (approved):
The producer and consumer path formulas are consistent, RID-less placement is intentional, and the shipped package payload remains unchanged. Approved.

One non-shipping follow-up: `playground/TestPlatform.Playground/TestPlatform.Playground.csproj` recursively copies `artifacts\bin\vstest.console\<Configuration>\net8.0\**\*.*` and the equivalent testhost tree. With this relocation, `-pack` builds now include the nested `publish\` closure and duplicate it under the playground's `net\publish\` output. Please exclude `publish\**\*.*` from those two playground globs in a follow-up.

PR #16267, opened 2026-07-13
Remove enable-auto-merge workflow

nohwnd:
Removes `enable-auto-merge.yml`, which has failed on every Maestro dependency PR since it was added (for example #16266).

It calls `enablePullRequestAutoMerge` with the default `GITHUB_TOKEN`, but `main` branch protection restricts pushes to an empty allowlist, so `github-actions[bot]` is rejected with `Pull request User is not authorized for this protected branch`. It is not a required check and never actually enabled auto-merge, so it only leaves a red X on each Maestro PR.

Making auto-merge work would need a branch protection change (allowlist a bot or app with merge rights), not this file.


Youssef1313 (commented):
FYI the same thing works in microsoft/testfx, so there must be some repo configuration difference that should be easy to figure out if you want to restore this back.

PR #16271, opened 2026-07-13
Pass the inferred target platform to the host in run settings

nohwnd:
`DotnetTestHostManager` decided whether to silently force an AnyCPU test host to x64 by reading `RunSettingsHelper.Instance.IsDefaultTargetArchitecture` on the host launch path. That is the shared run settings static state this cleanup is about: in design mode one process serves many requests, and a request that pins a platform (which sets the flag to false) leaves that value behind for the next request that pins none.

vstest.console already knows this fact per request. It is the `IsDefaultTargetArchitecture` on the injected `IRunSettingsHelper` that the `--Platform` executor writes when the user pins an architecture. Instead of letting the host read it back off a process wide singleton, vstest.console now stamps it into the run settings it hands to the host as `<IsTargetPlatformInferred>`, and the host reads it from `RunConfiguration`. The fact travels through the natural per request channel, the run settings, so nothing is shared between requests.

- `TestRequestManager` writes `<IsTargetPlatformInferred>` under `<RunConfiguration>` from the injected helper while it is already patching the run settings for the request.
- `RunConfiguration.FromXml` parses it into a new internal `IsTargetPlatformInferred` property that defaults to true when the element is absent, so synthetic run settings and the direct API keep the previous behavior.
- `DotnetTestHostManager` reads `RunConfiguration.IsTargetPlatformInferred` in `Initialize` and no longer takes an `IRunSettingsHelper` in its constructor.

The marker is a child element under `<RunConfiguration>`, not an attribute, because the run settings parser rejects unknown attributes but skips unknown elements. An older host ignores the element, and a newer host reading run settings that do not carry it falls back to true, which is the behavior before this change.

After this the only remaining `RunSettingsHelper.Instance` references are the composition root defaults, which sets up deleting the singleton in a follow up.

Build, unit tests (ObjectModel, Utilities, CrossPlatEngine, TestHostProvider and vstest.console) and smoke tests verified locally on net11.0 and net481.

Continues the static-state cleanup from #16200/#16205/#16208/#16228/#16243/#16245/#16257/#16258/#16262/#16263/#16268.

### DIFF
--- a/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
+++ b/src/Microsoft.TestPlatform.ObjectModel/RunSettings/RunConfiguration.cs
@@ -97,6 +97,7 @@ public class RunConfiguration : TestRunSettings
         ForwardStandardOutput = !FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_DISABLE_STANDARD_OUTPUT_FORWARDING);
         DisableSharedTestHost = FeatureFlag.Instance.IsSet(FeatureFlag.VSTEST_DISABLE_SHARING_NETFRAMEWORK_TESTHOST);
         CreateNoNewWindow = true;
+        IsTargetPlatformInferred = true;
     }
 
     /// <summary>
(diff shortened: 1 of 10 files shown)


Evangelink (approved):
The per-request marker correctly removes the shared-state dependency, preserves backward-compatible defaults, and reaches both discovery and execution host-launch paths. Approved.

The existing inline comment about reporting `boolValue` instead of the invalid XML text is valid but non-blocking. It would also be useful to add direct `DotnetTestHostManager` coverage proving `<IsTargetPlatformInferred>false</IsTargetPlatformInferred>` prevents silent x64 forcing; current tests primarily cover stamping and parsing.

PR #16272, opened 2026-07-13
Surface out-of-process data collector messages on the MTP execution path

azat-msft:
## Problem
Running a pure Microsoft.Testing.Platform (MTP) app under `vstest.console` with a generic out-of-process data collector (Blame, Event Log, or a custom `/Collect:` collector) completed the run but **silently dropped the collector's run-time and session-end messages and attachments** from the console — only the startup `SessionStarted` line appeared.

> The ~90s *shutdown hang* for Blame under MTP was already fixed by #16235 (`MtpDataCollectionForwarder`). This is the remaining downstream half of that same wiring gap.

## Root cause
On the MTP path there is no testhost, so `MtpProxyExecutionManager` owns the out-of-process data-collection lifecycle directly. The data collector runs in `datacollector.exe` and its log messages reach the console in two hops:

1. **IPC** — during the `AfterTestRunEnd` handshake, `DataCollectionRequestSender.SendAfterTestRunEndAndGetResult` pulls each `DataCollectionMessage` off the socket and buffers it into the in-process `_dataCollectionEventsHandler.Messages` list.
2. **In-process** — draining that list into the run's event handler (the console logger).

The MTP path performed hop 2 **only once, in `BeforeTestRun`** (startup messages). The run-time and session-end batch (`TestHostLaunched`, `TestCaseStarted/Ended`, `SessionEnded`, warnings, `Dispose`) arrives via hop 1 *after* the run, so the buffer filled up and was never flushed. The classic path relays these live via `DataCollectionTestRunEventsHandler`; the MTP path had no live pump and no post-run drain.

## Fix
Extract a `SurfaceDataCollectionMessages` helper and flush the buffer again after `AfterTestRun` completes. General fix for any out-of-process collector — no per-collector special case. The classic testhost path and the working MTP Code Coverage integration are untouched.

## Test
New acceptance test `RunMtpApplicationWithGenericOutOfProcDataCollectorCompletesRun` runs a pure-MTP app with the generic `SampleDataCollector` and asserts the run completes with the normal summary and that the collector lifecycle (`SessionStarted`, `TestHostLaunched`, per-test-case events, `SessionEnded`) and per-test-case attachments are surfaced. Fails on `main` (missing `TestHostLaunched`), passes with this change.

## Validation
- `MtpUnderVstestTests` full suite: **10/10 pass** (incl. Code Coverage path — no regression).
- New test + existing `/Blame` test: **4/4 pass**, run completes in ~1s (no hang).
- Release build of `Microsoft.TestPlatform.CrossPlatEngine` clean.


azat-msft:
Addressed the review feedback (commit 00ac89284):
- `SurfaceDataCollectionMessages` now also flushes and clears the buffered raw (telemetry) messages (`DataCollectionRunEventsHandler.RawMessages`), not just log messages, mirroring `ProxyExecutionManagerWithDataCollection` on the classic path so MTP runs no longer drop datacollector raw messages or retain them for the manager's lifetime.
- The acceptance test now asserts the full `Data collector 'SampleDataCollector' message:` prefix for the `TestCaseStarted`/`TestCaseEnded` checks (no longer matchable by unrelated output), points the collector's source files at a dedicated subdirectory (not the results directory), and asserts the exact per-test-case attachment count while excluding that source directory.

Verified: `RunMtpApplicationWithGenericOutOfProcDataCollectorCompletesRun` passes 2/2 matrix cases.

Note: the two review comments about `BuildPropertyProvider` and the discovery `completed.Wait(...)` refer to the `/TestCaseFilter` code, which is now a separate PR (#16282); both are addressed there.

nohwnd (approved):

PR #16278, opened 2026-07-14
Resolve TODO placeholders in Overview.md and configure.md

azat-msft:
Resolves the long-standing `TODO` / placeholder markers in the docs with source-grounded content, instead of leaving them or simply deleting the open questions. Split out from a larger docs pass so the placeholder-resolution work is reviewable on its own.

**Overview.md**
- Replace the "request/response format" placeholder with the actual message-documentation format (request payload, response payload, wire examples).
- Document protocol version 5 (the source marks it as having no protocol change — `// 5: ???` in `ProtocolVersioning`).
- Complete the "AttachmentSets vs RunAttachments" note: `AttachmentSets` are the per-run attachment sets collected during execution; `RunAttachments` on the enclosing payload holds run-level files such as TRX reports (verified against `TestRunCompletePayload` / `TestRunCompleteEventArgs`).
- Describe the `Package` (test container) property and the TranslationLayer extension points.
- Remove stale author-notes (`<TODO gif>`, "fill in more details", `<TODO example?>`, `// TODO: since?`, "TODO: what is it?") that are already answered by the surrounding text.

**configure.md**
- Fill in the `[mstest config]` and `[nunit config]` reference links (were literally `TODO`), pointing at the current Learn MSTest configuration page and the docs.nunit.org adapter-settings page.


docs/Overview.md
```diff
@@ -87,9 +87,7 @@ Datacollector observes the testhost to collect additional information about the
 
 While the tests execute, the results are reported back to the runner, aggregated, and forwarded to the client.
 
-The client processes the results and shows them in their UI, for example as TestExplorer does it here:
-
-<TODO gif>
```
nohwnd, line 92: Lazy :D

docs/Overview.md
```diff
@@ -1856,10 +1846,10 @@ Additional example of a toy test framework and adapter can be found in <https://
 
 ### TranslationLayer extension points
 
-TODO
+The TranslationLayer exposes client-facing APIs for tools that want to drive discovery,
```
nohwnd: this is supposed to be a docs of the whole thing, not two liner, but good first step.

nohwnd (approved):
Ok as long as you capture the removed todos somewhere and fill them later.

docs/Overview.md
azat-msft: xd
I asked copilot to add an  comment so I don't forget. Seems like smth went wrong

docs/Overview.md
```diff
@@ -1856,10 +1878,20 @@ Additional example of a toy test framework and adapter can be found in <https://
...
+shelling out to `vstest.console.exe`. The entry point is `IVsTestConsoleWrapper` (and its
+async counterpart `IVsTestConsoleWrapperAsync`), which wraps a `vstest.console` process and
+sends protocol messages to it. The main extension points are:
+
+- `StartSession` / `EndSession` - start and tear down the wrapped `vstest.console` process.
+- `InitializeExtensions` - register additional adapter/data-collector extension paths.
+- `DiscoverTests` - discover tests and receive results through an `ITestDiscoveryEventsHandler`.
+- `RunTests` / `RunTestsWithCustomTestHost` - run tests, optionally hosting the testhost through a
+  custom `ITestHostLauncher` (used to attach a debugger or control the testhost process).
+- `ProcessTestRunAttachmentsAsync` (on `IVsTestConsoleWrapperAsync`) - post-process attachments
```
nohwnd: this is still very sparse and does not talk at all about vstest console wrapper extension points. make it go through all public interfaces in the project there are extensions for discovery, run, etc.

docs/Overview.md
azat-msft: the content has been re-reviewed since this comment, so resolving it now to continue the conversation under https://github.com/microsoft/vstest/pull/16278#discussion_r3585722398

docs/Overview.md
```diff
@@ -302,7 +326,7 @@ Versions:
 - 2: Changed serialization from a generic bag that described each property and its type, to explicit properties that are serialized without additional type info.
 - 3: Added AttachDebugger message.
 - 4: Added because version 3 did not update the serialization to use, and it will use v1 serialization (bag) rather than explicit properties. Right side should avoid negotiating 3 and downgrade to 2.
-- 5: Unknown. (TODO)
+- 5: Unknown in the core `ProtocolVersioning` table (the source still marks this version as `// 5: ???`). The TranslationLayer defines a private `MinimumProtocolVersionWithTestSessionSupport = 5` in `VsTestConsoleRequestSender`, but the constant is currently unused, so the exact change associated with v5 is unclear from the current source.
```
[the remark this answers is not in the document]
azat-msft: I'm tired of iterating on this line. The current version lgtm and the re-wording doesn't seem to bring much value, only seems to waste my time and burn tokens

docs/Overview.md
azat-msft: Please take a look at the updated version

docs/Overview.md
```diff
@@ -1856,10 +1884,96 @@ Additional example of a toy test framework and adapter can be found in <https://
 
 ### TranslationLayer extension points
```
nohwnd, line 1885: nit: the firsts part is more of usage rather than extension, you as user of the layer won't extend the IVSTestConsoleWrapper, you will instantiate it and use it. The #### Callback interfaces you implement (the plug-in points) 

are the extension points. So this is probably better to split into usage, and what you need to implement to use it

azat-msft: The usage and extension parts are split, added an explicit mention that `IVSTestConsoleWrapper` is used and not extended.

PR #16282, opened 2026-07-14
Honor /TestCaseFilter on the MTP execution path

azat-msft:
## Problem
On the `vstest.console` path that runs a Microsoft.Testing.Platform (MTP) app as its own testhost (microsoft/vstest#16201), `/TestCaseFilter:<expr>` was **silently ignored during execution** — every test ran regardless of the filter. `/Tests` (specific names) worked, but any filter expression was dropped.

## Root cause
A `/TestCaseFilter` run reaches `MtpProxyExecutionManager` as *sources with no specific tests*. `BuildWork` then yields the source with `null` tests and `RunSource` runs everything. MTP has no notion of the vstest filter expression, and nothing translated it, so the filter had no effect.

## Fix
When a `TestCaseFilter` is present and there are no specific tests, discover the source over MTP, evaluate the expression against the discovered tests with `TestCaseFilterExpression` (traits + boolean operators, same semantics as the classic path), and run only the matching test-node uids. A non-matching filter now runs **zero** tests instead of the whole suite. `/Tests` is unchanged.

## Test
New acceptance test `RunMtpApplicationHonorsTestCaseFilter` runs the MTP app with `/TestCaseFilter:"DisplayName~TestPasses"` and asserts only the two matching tests run (2/0/0) instead of the full 2/1/1 suite.

## Validation
- `RunMtpApplicationHonorsTestCaseFilter`: **2/2 matrix cases pass** (fails on `main` — filter ignored, whole suite runs).
- Release build of `Microsoft.TestPlatform.CrossPlatEngine` and the acceptance project: clean.
- Verified end-to-end against a real TUnit MTP app: `~BasicTests` -> 2 (was 33); non-matching -> 0.


azat-msft:
Addressed the review feedback (commit 140a078d0):
- `BuildPropertyProvider` now exposes every property carried on the test case (by its filter label) plus the `Name`/`DisplayName` aliases and traits, instead of a hard-coded `FullyQualifiedName`/`DisplayName`/`Name` subset. Filters on other properties (e.g. `Source`) previously evaluated to "no value" and silently matched nothing; they are now honored like the classic path (verified: `Source~MtpMSTestProject` matches all tests, `Source~NoSuchSource` matches none).
- `DiscoverSourceTests` now waits for the discovery-completion sentinel honoring the cancellation token and logs a warning on timeout, so a `/TestCaseFilter` is not silently evaluated against an incomplete discovery set.

Verified: `RunMtpApplicationHonorsTestCaseFilter` passes 2/2 matrix cases.

azat-msft:
No more reviews are needed, the PR is a WIP

nohwnd:
feels like we are adding bunch of overhead by doing the discovery and then filtering, lets hold on on this one, ideally we would get official mtp client from MTP (by extracting the one in VS) and use it here, so we don't have to keep fixing the client over an over, working on that one.

nohwnd:
Sorry for the long silence on this one.

First, an update on what you were told to wait for: the official client landed. #16300 retargeted the MTP client onto `Microsoft.Testing.Platform.ServerMode.Client.Sources`, the source only package testfx builds from the MTP server's own protocol and serialization source, and it merged on 14 August. That is the "official mtp client from MTP" I meant in July. #16287, the pipe protocol proof of concept, is superseded by it and no longer carries a rewrite.

The part I got wrong: I assumed the rewrite would bring filtering with it. It did not. I checked `main` today and the gap you found is still there. `BuildWork` still yields the source with no tests when there is no explicit `/Tests` list, `RunSource` then calls `RunTestsAsync` with no filter, and MTP treats that as run everything. So `/TestCaseFilter` on an MTP app still runs the whole suite and reports success.

I filed #16379 to track it, and I wrote your approach into the issue, because that is the part that is expensive to rediscover:

- discovery has to finish first, then the filter is evaluated against the discovered test cases, then only the matching node uids are run
- the property provider has to expose every property on the test case by its filter label, plus the `Name` alias for `DisplayName`, plus traits. That is what makes `Source~`, `TestCategory=` and `Priority=` evaluate instead of only `FullyQualifiedName`, and it is not obvious until someone tries a non FQN filter and gets an empty run
- when nothing matches, the source has to be skipped, not passed an empty uid list, because an empty list omits the MTP filter and runs everything again

Thank you for both of those, the second one especially. That empty run is the same class of bug as the one this PR fixes, and it would have shipped hidden inside the fix if you had not handled it.

I am closing this PR. Not because the work was wrong, it was not, but because the diff cannot be rebased any more, it edits `MtpClientHelpers.cs` and #16300 deleted that file. Redoing it on the current client is less work than untangling the branch. Everything needed is in #16379.

If you want to pick up #16379 on top of the current client, please do, it is yours. If not, that is fine too, it is written down now and it will not get lost again.

🤖

PR #16335, opened 2026-08-03
Address review comments from the TRX astral-plane fix (#16318)

azat-msft:
Follow-up to #16318, which was merged before its review comments were resolved. This addresses all four of @Evangelink's comments.## 1. `Append(str, i, 2)` instead of the chained form> Nit: `builder?.Append(str, i, 2)` says the same thing in one call and makes it obvious the two code units are copied as a unit.Done. The chained `builder?.Append(str[i]).Append(str[i + 1])` was correct — the null-conditional short-circuits the whole chain — but it reads like it might append only the first half when `builder` is null, which is exactly the kind of thing that survives a fast review and then gets "fixed" into a real bug later.## 2. Builder capacity> this capacity hint is guaranteed to be too small … testfx solved the identical problem in `TrxReportEngine.Utilities.cs` with `length + length / 2` plus an overflow guardCorrect — the builder is only ever allocated because something needs escaping, and each escaped code unit expands 1 → 6 chars, so `new StringBuilder(str.Length)` always had to grow at least once. Added `GetInvalidXmlCharBuilderCapacity` matching testfx, and switched to the `StringBuilder(string, int, int, int)` constructor so the valid prefix is copied during construction rather than by a separate `Append`.The capacity is computed lazily at the `??=` site rather than eagerly at the top of the method, so the all-valid fast path still does no work at all.## 3. Unpinned edges + collapse the near-duplicates> two consecutive high surrogates … the inverted order … the empty string … A single `[DataRow]`-driven test would cover them and let you collapse some of the near-duplicate cases above.The six near-identical single-case tests are now two data-driven tests, and all three edges are pinned, plus a few more that came for free (`\udbff\udfff` as the top of the astral range, two consecutive low surrogates, a pair as the entire string).**One thing worth flagging: a lone surrogate cannot be passed through a `[DataRow]`.** My first attempt did exactly that and produced nine tests that looked fine and asserted nothing useful:```Expected: "a\ud800b"But was:  "a��b"```The argument arrives at the test method with the lone surrogate already replaced by U+FFFD. Since U+FFFD is itself a valid XML character, the sanitizer correctly passes it through — so the test fails today, but had the expectations been written against the observed output it would have "passed" forever while testing nothing. I confirmed this is the platform's data-row argument transport and not the C# compiler: a plain string literal round-trips a lone surrogate through metadata intact, and well-formed pairs (`\U0001F389`) survive the data row transport fine — it only breaks on unpaired surrogates, which is the expected UTF-8 behaviour.So the rows spell the lone surrogates as `{H}` and `{L}` and `WithLoneSurrogates` substitutes them inside the test body, after transport. It also asserts the incoming row contains no U+FFFD, so if anyone later inlines a literal lone surrogate the test fails loudly instead of silently going vacuous.## 4. Nothing covered the actual file> every new test targets `XmlPersistence` in isolation, but the premise of the PR is about the bytes that end up in the `.trx` file … Not a blocker for this PR, but I'd like it before this pattern gets copied again.Added `TrxLoggerTests.TrxFileShouldPreserveAstralCharactersInTestNameAndStdOut`. It runs a result with an emoji in both the test name and StdOut through the whole logger via the existing `PopulateTrxFile` harness, then re-reads the `.trx` from disk and asserts:- the raw file text contains the emoji as a real character,- the raw file text does **not** contain the literal `\ud83c`,- the file still parses, and the emoji survives in both an attribute value (`UnitTestResult/@testName`) and element text (`StdOut`).That last pair is what would catch a string reaching the DOM without going through `SaveSimpleData`.## VerificationI verified the tests have teeth rather than assuming it, by sabotaging the fix in each direction and checking what goes red.**Sabotage A — over-strict** (disable the pair branch, i.e. reinstate the original bug):| | ||---|---|| `TrxFileShouldPreserveAstralCharactersInTestNameAndStdOut` | red || 5 valid-pair rows of `SaveObjectShouldNotReplaceValidText` | red || `SaveObjectShouldNotReplaceWellFormedSurrogatePairInAttributeValue` | red || `SaveObjectShouldProduceWellFormedXmlWithAstralCharacters` | red || mixed row of `SaveObjectShouldReplaceLoneSurrogate` | red |**Sabotage B — too permissive** (`IsValidXmlChar` accepts all of `\uD800-\uDFFF`, the naive way to "fix" the astral bug):| | ||---|---|| all 9 rows of `SaveObjectShouldReplaceLoneSurrogate` | red || `TrxFileShouldRemainParseableWhenTestNameContainsLoneSurrogate` | red |Sabotage B is what prompted the second commit. Only one row of `SaveObjectShouldReplaceLoneSurrogate` shows up under sabotage A — that test guards the opposite direction — and nothing was exercising that direction through an actual file. So I added `TrxFileShouldRemainParseableWhenTestNameContainsLoneSurrogate`, which puts a lone high surrogate in a test name and asserts the trx is written, parses, and contains the escaped form. Under sabotage B it fails with exactly the exception the escaping exists to prevent:```System.ArgumentException: The surrogate pair (0xD800, 0x20) is invalid. A high surrogatecharacter (0xD800 - 0xDBFF) must always be paired with a low surrogate character(0xDC00 - 0xDFFF).  at System.Xml.XmlUtf8RawTextWriter.EncodeSurrogate(Char* pSrc, Char* pSrcEnd, Byte* pDst)  at System.Xml.XmlUtf8RawTextWriter.WriteAttributeTextBlock(Char* pSrc, Char* pSrcEnd)```That claim was prose in #16318; it is now pinned by a test. Both regression directions are covered at the file level, not just at the sanitizer.`Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests` — 186/186 passing in Release, 93 on net481 and 93 on net11.0.

## Review round 2

Both suggestions applied in `Apply review suggestions from #16335`.

**The `\ufffd` guard on `SaveObjectShouldNotReplaceValidText`** — fair, and I had the guard on the wrong test. `SaveObjectShouldReplaceLoneSurrogate` compares against hard-coded escaped literals so it cannot go vacuous; this one compares against its own input, so it can. Applied, and verified it fires by temporarily adding a row containing U+FFFD:

```
failed TEMP sabotage - simulates a mangled row
  Assert.DoesNotContain failed. String 'party ? done' does contain string '?'.
  A surrogate pair did not survive the data row transport - this row would assert nothing.
```

**The `CheckCharacters` example** — you are right and I should not have written it. I reproduced your result before removing it:

```
CheckCharacters=True  bytes=110 sha=3C465572775C747C
CheckCharacters=False bytes=110 sha=3C465572775C747C
```

Byte-identical, same hash, for an emoji in an attribute and an escaped lone surrogate in element text. Everything reaching the DOM has already been through `RemoveInvalidXmlChar`, so the writer never sees anything the flag could object to. Dropped from the comment here and from the PR description above, since claiming coverage that does not exist is worse than claiming none.

### DIFF
--- a/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
@@ -1036,9 +1036,8 @@ public class TrxLoggerTests
         // XmlPersistenceTests covers the sanitizer in isolation, but the thing users report is
         // about the bytes that end up in the .trx. This goes through the whole logger - test
         // result in, real file out, re-parsed from disk - so it would also catch a regression
-        // that lives outside the sanitizer, such as XmlWriterSettings.CheckCharacters being
-        // flipped in PopulateTrxFile or a string reaching the DOM without going through
-        // SaveSimpleData.
+        // that lives outside the sanitizer, such as a string reaching the DOM without going
+        // through SaveSimpleData.
         const string testName = "party \U0001F389 done";
         const string stdOut = "output \U0001F389 here";
 
--- a/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
@@ -1068,6 +1068,35 @@ public class TrxLoggerTests
         Assert.AreEqual(stdOut, resultNode.Descendants(ns + "StdOut").First().Value);
     }
 
+    [TestMethod]
+    public void TrxFileShouldRemainParseableWhenTestNameContainsLoneSurrogate()
+    {
+        // The counterpart to the test above, and the reason the fix escapes lone surrogates
+        // instead of simply allowing all of \uD800-\uDFFF through. A lone surrogate is not a
+        // valid Unicode scalar value and XmlWriter throws on one, so escaping it is what keeps
+        // the trx writable and parseable at all. Letting it through would trade a bug that
+        // mangles one string for a bug that breaks every consumer of the file at read time.
+        const string testName = "lone \ud800 surrogate";
+
+        _parameters[TrxLoggerConstants.LogFileNameKey] = "lone-surrogate.trx";
+        _testableTrxLogger.Initialize(_events.Object, _parameters);
+
+        var pass = CreatePassTestResultEventArgsMock(testName);
+        _testableTrxLogger.TestResultHandler(new object(), pass.Object);
+        _testableTrxLogger.TestRunCompleteHandler(new object(), CreateTestRunCompleteEventArgs());
+
+        var trxFile = _testableTrxLogger.TrxFile!;
+        Assert.IsTrue(File.Exists(trxFile), "The trx must still be written when a test name contains a lone surrogate.");
+
+        using FileStream file = File.OpenRead(trxFile);
+        using XmlReader reader = XmlReader.Create(file);
+        XDocument document = XDocument.Load(reader);
+        var ns = document.Root!.GetDefaultNamespace();
+
+        var resultNode = document.Descendants(ns + "UnitTestResult").First();
+        Assert.AreEqual(@"lone \ud800 surrogate", resultNode.Attributes("testName").First().Value);
+    }
+
     private void ValidateTestIdAndNameInTrx()
     {
         TestCase testCase = CreateTestCase("TestCase");
--- a/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
@@ -1030,6 +1030,44 @@ public class TrxLoggerTests
         File.Delete(logger.TrxFile);
     }
 
+    [TestMethod]
+    public void TrxFileShouldPreserveAstralCharactersInTestNameAndStdOut()
+    {
+        // XmlPersistenceTests covers the sanitizer in isolation, but the thing users report is
+        // about the bytes that end up in the .trx. This goes through the whole logger - test
+        // result in, real file out, re-parsed from disk - so it would also catch a regression
+        // that lives outside the sanitizer, such as XmlWriterSettings.CheckCharacters being
+        // flipped in PopulateTrxFile or a string reaching the DOM without going through
+        // SaveSimpleData.
+        const string testName = "party \U0001F389 done";
+        const string stdOut = "output \U0001F389 here";
+
+        _parameters[TrxLoggerConstants.LogFileNameKey] = "astral.trx";
+        _testableTrxLogger.Initialize(_events.Object, _parameters);
+
+        var pass = CreatePassTestResultEventArgsMock(testName, new List<TestResultMessage> { new(TestResultMessage.StandardOutCategory, stdOut) });
+        _testableTrxLogger.TestResultHandler(new object(), pass.Object);
+        _testableTrxLogger.TestRunCompleteHandler(new object(), CreateTestRunCompleteEventArgs());
+
+        var trxFile = _testableTrxLogger.TrxFile!;
+
+        // The emoji must be in the file as a real character, not as the literal text \ud83c\udf89.
+        var rawTrxContent = File.ReadAllText(trxFile);
+        Assert.Contains(testName, rawTrxContent, "The astral character in the test name was mangled on the way into the trx.");
+        Assert.DoesNotContain(@"\ud83c", rawTrxContent, "The astral character was escaped into literal text instead of being written as-is.");
+
+        // And the file must still be parseable, with the character surviving the round trip
+        // through both an attribute value and element text.
+        using FileStream file = File.OpenRead(trxFile);
+        using XmlReader reader = XmlReader.Create(file);
+        XDocument document = XDocument.Load(reader);
+        var ns = document.Root!.GetDefaultNamespace();
+
+        var resultNode = document.Descendants(ns + "UnitTestResult").First();
+        Assert.AreEqual(testName, resultNode.Attributes("testName").First().Value);
+        Assert.AreEqual(stdOut, resultNode.Descendants(ns + "StdOut").First().Value);
+    }
+
     private void ValidateTestIdAndNameInTrx()
     {
         TestCase testCase = CreateTestCase("TestCase");
--- a/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
+++ b/test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
@@ -49,6 +49,11 @@ public class XmlPersistenceTests
         XmlPersistence xmlPersistence = new();
         var node = xmlPersistence.CreateRootElement("TestRun");
 
+        // Unlike the sibling test below, expected here is the input itself, so a row whose
+        // text was mangled in transit would assert sanitize(mangled) == mangled and pass
+        // while proving nothing - U+FFFD is a valid XML character and survives sanitizing.
+        Assert.DoesNotContain("\ufffd", text, "A surrogate pair did not survive the data row transport - this row would assert nothing.");
+
         XmlPersistence.SaveObject(text, node, null, "dummy");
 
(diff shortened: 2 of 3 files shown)


test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
```diff
@@ -34,31 +34,51 @@ public void SaveObjectShouldReplaceInvalidCharacter()
...
-        // unescaped. This case used to be asserted the other way around.
         XmlPersistence xmlPersistence = new();
         var node = xmlPersistence.CreateRootElement("TestRun");
 
-        string validSurrogatePair = new(new[] { (char)0xd800, (char)0xdc00 });
-        XmlPersistence.SaveObject(validSurrogatePair, node, null, "dummy");
+        XmlPersistence.SaveObject(text, node, null, "dummy");
 
-        Assert.AreEqual(validSurrogatePair, node.InnerXml);
+        Assert.AreEqual(text, node.InnerXml);
```
Evangelink, line 59: [Test Coverage Design] nit, but a real one — this is the row set that pins the astral-plane behaviour, and it's the one row set that can't fail loudly.

`expected` here *is* the input, so if the data row transport ever mangles `text` the assertion degenerates into `sanitize(mangled) == mangled`. U+FFFD is a valid XML character, so it passes through `RemoveInvalidXmlChar` untouched and the row goes green while proving nothing. I ran the head implementation against it to be sure:

```
RemoveInvalidXmlChar("party \ufffd done") -> "party \ufffd done"   (unchanged)
```

That is precisely the failure mode `WithLoneSurrogates` exists to prevent 100 lines below — except the guard only sits on the sibling test. `SaveObjectShouldReplaceLoneSurrogate` can't go vacuous anyway, because its `expected` is a hard-coded escaped literal; this one can, and it's the one that guards the actual bug #16318 fixed.

You're right that well-formed pairs survive the transport today. "Today" is what the guard is for, and moving these cases from inline literals into `[DataRow]` is what created the exposure. Same one-liner:

```suggestion
        Assert.DoesNotContain("\ufffd", text, "A surrogate pair did not survive the data row transport - this row would assert nothing.");

        XmlPersistence.SaveObject(text, node, null, "dummy");

        Assert.AreEqual(text, node.InnerXml);
```

Non-blocking.

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
```diff
@@ -1030,6 +1030,73 @@ public void SkipInitializeDictionaryShouldNotFail()
...
 
+    [TestMethod]
+    public void TrxFileShouldPreserveAstralCharactersInTestNameAndStdOut()
+    {
+        // XmlPersistenceTests covers the sanitizer in isolation, but the thing users report is
+        // about the bytes that end up in the .trx. This goes through the whole logger - test
+        // result in, real file out, re-parsed from disk - so it would also catch a regression
+        // that lives outside the sanitizer, such as XmlWriterSettings.CheckCharacters being
+        // flipped in PopulateTrxFile or a string reaching the DOM without going through
+        // SaveSimpleData.
```
Evangelink: [Diagnostic Clarity] The `CheckCharacters` example doesn't hold — neither of the two new tests can catch that flag being flipped.

Every string that reaches the DOM goes through `SaveSimpleData` → `RemoveInvalidXmlChar`; the only other DOM write in `XmlPersistence` is the `SetAttribute` for the integer counters (`XmlPersistence.cs:473`). So by the time `PopulateTrxFile` runs, the writer never sees a character `CheckCharacters` could object to. I rendered the same document both ways to check rather than assuming:

```
CheckCharacters=True  bytes=110 sha=3C465572775C747C
CheckCharacters=False bytes=110 sha=3C465572775C747C
```

Byte-identical, for an emoji in an attribute and an escaped lone surrogate in element text. Flipping it fails nothing.

The second half of the sentence — "a string reaching the DOM without going through `SaveSimpleData`" — is accurate and is the part that carries the weight. Worth dropping the first example so nobody later reads this as coverage that exists:

```suggestion
        // result in, real file out, re-parsed from disk - so it would also catch a regression
        // that lives outside the sanitizer, such as a string reaching the DOM without going
        // through SaveSimpleData.
```

Non-blocking.

Evangelink (approved):
Thanks for coming back and closing these out properly — all four comments are addressed, and the two extra tests in the second commit are the right instinct.

I went through the sanitizer and both test files line by line, and ran the head implementation of `RemoveInvalidXmlChar` / `GetInvalidXmlCharBuilderCapacity` standalone against every one of the 16 data rows plus the three pre-existing cases. All 16 rows produce exactly the asserted values, the three pre-existing expectations are unchanged, and the `ArgumentException` you quoted reproduces exactly (`The surrogate pair (0xD800, 0x20) is invalid...`) when a lone surrogate is written raw, and does not reproduce once it has been through the sanitizer. So `TrxFileShouldRemainParseableWhenTestNameContainsLoneSurrogate` is pinning the thing it claims to pin.

On the pieces I was most likely to push back on:

- `Append(str, i, 2)` — correct, and `StringBuilder.Append(string, int, int)` is available on both netstandard2.0 and net462, so the multi-targeting is fine.
- `new StringBuilder(str, 0, i, capacity)` — the ctor clamps `capacity` up to `length` internally, so a small hint can never truncate the prefix. `GetInvalidXmlCharBuilderCapacity` matches `TrxReportEngine.GetInvalidXmlCharBuilderCapacity` in testfx character for character, and the overflow guard is correct at the boundary: `1431655765 -> 2147483647`, `1431655766 -> 1431655766`, `int.MaxValue -> int.MaxValue`, never negative.
- The `{H}` / `{L}` substitution — good call, and the U+FFFD assertion inside `WithLoneSurrogates` is the right shape for a guard. See the one inline comment about its missing twin.
- No test coverage was lost collapsing the six single-case tests: every removed case maps onto a row, and three previously unpinned edges are now covered.
- No public API surface, no new dependency, no behaviour change beyond what #16318 already shipped — this is refactor + tests, with CI green on Windows, Linux, macOS and source-build.

| Dimension | Status |
|---|---|
| Unicode / astral-plane handling | pass |
| TRX serialization & parse round trip | pass |
| Cross-TFM (netstandard2.0 / net462 / net481 / net11.0) | pass |
| Backward compatibility & rollback | pass (no behaviour change) |
| Public API surface | pass (private static helper) |
| Null / boundary validation | pass (null, empty, end-of-string, int overflow) |
| Performance | pass (lazy alloc preserved, capacity mirrors testfx) |
| Test coverage design | warn (1 finding) |
| Diagnostic clarity | warn (1 finding) |
| Source build & cross-platform | pass (CI green on all legs) |

**Findings:** 2, both minor and non-blocking — one test that can silently go vacuous under the same transport hazard you documented, and one comment that claims coverage the tests don't actually provide. Neither affects shipped behaviour.

Looks good :)

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/XmlPersistenceTests.cs
azat-msft: Applied. You are right that the guard was on the wrong test - `SaveObjectShouldReplaceLoneSurrogate` compares against hard-coded escaped literals so it cannot go vacuous, and this one compares against its own input, so it can. I verified it fires rather than assuming, by temporarily adding a row containing U+FFFD:

```
failed TEMP sabotage - simulates a mangled row
  Assert.DoesNotContain failed. String 'party ? done' does contain string '?'.
  A surrogate pair did not survive the data row transport - this row would assert nothing.
```

Also updated the `WithLoneSurrogates` remarks so the two guards cross-reference each other, rather than reading as if only one test needed one.

test/Microsoft.TestPlatform.Extensions.TrxLogger.UnitTests/TrxLoggerTests.cs
azat-msft: Applied, and thanks for checking rather than just flagging it - you are right and I should not have written that example. I reproduced your result before removing it:

```
CheckCharacters=True  bytes=110 sha=3C465572775C747C
CheckCharacters=False bytes=110 sha=3C465572775C747C
```

Byte-identical, same hash, for an emoji in an attribute and an escaped lone surrogate in element text. Everything reaching the DOM has already been through `RemoveInvalidXmlChar`, so by the time `PopulateTrxFile` runs the writer never sees a character the flag could object to.

Dropped from the comment and also from the PR description, which repeated the same claim. Claiming coverage that does not exist is worse than claiming none.

Issue #15718, opened 2026-04-22
18.5.0 introduced an error Could not load file or assembly 'System.Collections.Immutable, Version=8.0.0.0


mike9900:
> ℹ️ _Update: from maintainers._

The package was unlisted on nuget.org, 18.4.0 should resolve for you. Fix is in 18.5.1 that is currently building and will be released soon - probably in Tuesday there is no significant urgency to risk breaking you over weekend. 

We are trying to push the fix also into VS. And will port it to 18.6, 18.7 etc.

Fails in CI pipelines only with VSTest@2 and distributed run. 

If you are on **cloud** AzDO, update to VSTest@3, VSTest@2 task is soft deprecated, so take it as a nudge to stay up to date.

If you are on on-prem AzDO you can use Test Platform Installer task to set version of VSTest to 18.4.0, or to 18.5.1 once published.

----

System.AggregateException: One or more errors occurred. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Collections.Immutable, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.

It looks like Version 9.0.24.52809 of System.Collections.Immutable is being installed with 18.5.0 of vstest

I'm installing this through Azure DevOps pipeline using the Visual Studio test platform installer with options of Package Feed: Official Nuget and Version: Latest Stable. As soon as it starts test discovery in a Visual Studio Test step it is throwing the above error. 

Using version 18.4.0 does not produce this error.

nohwnd:
Thanks, got this reported from more places. Plesae provide more details how you run the command, no idea why we did not see it in intergration tests, seems like big gap. Unlisting the packgage and backporting the binding redirects verification script from latest.

Sorry for the issue!

nohwnd:
Unlisted.

mike9900:
> Thanks, got this reported from more places. Plesae provide more details how you run the command, no idea why we did not see it in intergration tests, seems like big gap. Unlisting the packgage and backporting the binding redirects verification script from latest.
> 
> Sorry for the issue!

Do you still need additional details?

I use the Azure DevOps step of 'Visual Studio test platform installer' to pull the package, and then use the option Installed by Tools Installer for the Visual Studio Test step

Visual Studio test platform installer has options of Package Feed: Official Nuget and Version: Latest Stable

nohwnd:
yes please more details help. 

[...]

PR #16378, opened 2026-08-18
Add xxHash128 test ids, available but not default

azat-msft:
Adds xxHash128 as an alternative to SHA1 for computing `TestCase.Id`, **available but not the default**. Nothing in this release changes any test id.

Supersedes #15260.

## Why

The test id is an identity, never a security boundary, so a cryptographic hash was always the wrong tool. SHA1 is slower than necessary, trips security scanners, and blocks running under FIPS policies. xxHash128 is the right primitive and is what MSTest already uses.

Changing a test id breaks anything that stored one, and `EqtHash.cs` requires TC-TA sign off for exactly that. Shipping the algorithm separately from the default change means this release carries no compatibility risk and can be taken while the sign off is outstanding.

## What ships

```
VSTEST_DISABLE_XXHASH128_TESTCASE_ID=0    # opt in to xxHash128
```

A feature flag in `FeatureFlag.cs` defaulting to `true` (SHA1) — same shape as `VSTEST_DISABLE_MTP_TESTHOST`. Works from the shell or runsettings, on both the classic and Microsoft.Testing.Platform paths.

**Making xxHash128 the default later is deleting one `DefaultValues` entry.** `1` selects SHA1 and `0` selects xxHash128 both before and after, so nothing anyone writes down changes meaning.

Ids are RFC 9562 version 8 UUIDs carrying the hashing scheme version, ported from MSTest's `VersionedGuidFromHash`, so the algorithm is readable off the id:

```
1ea84a1f-f791-8103-...   leading 1, version 8  → xxHash128
28e7a7ed-8fb9-05b7-...   unstructured          → SHA1
```

`EqtHash.GuidFromStringXxHash128` and `TestIdProviderXxHash128` are public, so adapters that assign ids themselves can adopt them directly.

## Notable decisions

**Vendored xxHash128 rather than the `System.IO.Hashing` package**, which would add a new assembly identity to assets loaded by hosts without binding redirects, needing redirects in all three app.configs. Vendored as `internal`, following the existing `Microsoft.TestPlatform.Filter.Source` pattern; microsoft/testfx vendors the same files for the same reason.

**`System.Memory` becomes explicit on the .NET Framework legs**, because the vendored code uses `Span<T>`; `Microsoft.TestPlatform.AdapterUtilities` previously had no dependencies. Pinned to the version those hosts actually deploy (assembly `4.0.5.0`) — the remedy #15718 and [testfx#9209](https://github.com/microsoft/testfx/pull/9209) both arrived at independently. Normal NuGet consumers are unaffected; hand-assembled layouts are, and are guarded by tests.

## Validation

- Ids are **byte-identical to `main`** by default — the pinned ids in `TestCaseTests` and `TestResultSerializationTests` are untouched.
- Pinned vectors from dotnet/runtime's own suite, plus **MSTest's own published vector**, proving both compute identical ids for identical input. That is the only check that could catch a byte-order error; every other vector came from this implementation.
- Acceptance tests for a runsettings-declared flag reaching the process that computes ids, and for the `System.Memory` binding in net472 hosts without redirects. Each was verified to **fail** when deliberately broken.
- Measured through real adapters: NUnit's ids change; MSTest v3 and v4 are unaffected, because they assign their own.


src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -19,6 +19,31 @@ namespace Microsoft.VisualStudio.TestPlatform.ObjectModel;
...
+    /// for the lifetime of the process.
+    /// </para>
+    /// </remarks>
+    internal const string TestCaseIdAlgorithmEnvironmentVariable = "VSTEST_TESTCASE_ID_ALGORITHM";
+
+    private static readonly bool UseLegacySha1TestIds =
+        string.Equals(
+            Environment.GetEnvironmentVariable(TestCaseIdAlgorithmEnvironmentVariable),
+            "sha1",
+            StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
azat-msft: Good catch — fixed in f3e5013.

The read is now deferred to first use and cached in a `bool?`, so an id still stays stable for the process lifetime but the value is no longer baked in at an arbitrary type-load moment. I went with the lazy read rather than a test-only hook so this fixes the latent behaviour for real users, not just the test brittleness.

Your symptom was real and worse than it looked: with `VSTEST_TESTCASE_ID_ALGORITHM=sha1` exported, `EqtHashTests.TestCase_Id_UsesTheXxHash128Algorithm` also failed, not just the `TestCaseTests` ones — a confusing mismatch that looks like the hashing is broken. Both pinned-id classes now force the default algorithm explicitly, so they assert what they mean regardless of the ambient environment.

Also added `TestCaseIdAlgorithmTests` so the switch is covered rather than merely tolerated: both algorithms, case insensitivity, unrecognized values falling back to the default, that the read really is lazy (touches the type *before* setting the variable), and that the choice is cached once made. Verified the suite is green with the variable unset, `sha1`, `SHA1`, and an unrecognized value.

docs/environment-variables.md
```diff
@@ -254,6 +254,15 @@ This document lists environment variables that are currently handled by VSTest s
 
 ## Discovery Variables
 
+### VSTEST_TESTCASE_ID_ALGORITHM
```
nohwnd: Would prefer this to be a feature flag to opt out from the new XXHASH, rather than an option.

nohwnd: It has  impact on the whole product, and signals better that this is not something you should choose, but rather a new default that can be disabled for given amount of time.

nohwnd (changes requested):
Green but would like to review properly before approving, so wait on me :)

src/Microsoft.TestPlatform.CrossPlatEngine/Client/MTP/MtpProxyExecutionManager.cs
```diff
@@ -429,6 +436,11 @@ private void ApplyRunSettingsEnvironmentVariables(string? runSettings)
             return;
         }
```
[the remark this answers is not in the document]
azat-msft: Addressed in ad66003 — you're right, and this was a real latent bug rather than a theoretical one.

`ApplyRunSettingsEnvironmentVariables` now resolves the declared algorithm **before** the early return, and unconditionally:

```csharp
_testCaseIdAlgorithm = MtpTestNodeConverter.ResolveTestCaseIdAlgorithm(runSettingsEnvironmentVariables);

if (runSettingsEnvironmentVariables is null || runSettingsEnvironmentVariables.Count == 0)
{
    return;
}
```

Resolving from an absent or empty set yields `null`, which is exactly "nothing was declared", so a run that declares nothing now clears what a previous run on the same instance declared instead of inheriting it.

Checked `MtpProxyDiscoveryManager` for the same shape: it already assigns unconditionally at the top of `DiscoverTests`, so it needed no change.

Note the variable named here has since become the `VSTEST_DISABLE_XXHASH128_TESTCASE_ID` feature flag, but the stale-state problem was independent of that.

test/Microsoft.TestPlatform.ObjectModel.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+// Several classes here pin the test id algorithm by setting VSTEST_TESTCASE_ID_ALGORITHM and
+// resetting TestCase's cache of it. Both are process wide, so a class level [DoNotParallelize]
+// does not stop a class that pins a different value from running at the same time. This assembly
+// already runs sequentially because nothing opts it into parallelization; this states that as a
+// requirement rather than leaving it as an accident.
```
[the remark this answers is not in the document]
azat-msft: Fixed in 9e1c296 — the comment was left behind by the switch to the feature flag. It now reads:

```csharp
// Several classes here pin the test id algorithm by setting VSTEST_DISABLE_XXHASH128_TESTCASE_ID and
// resetting the feature flag cache. Both are process wide, so a class level [DoNotParallelize] ...
```

Both the variable name and the reference to `TestCase`'s own cache were stale: the cache is now `FeatureFlag`'s, reset through `TestCase.ResetFeatureFlagCacheForTesting()`. The `[assembly: DoNotParallelize]` itself is unchanged and still needed for the reason the comment gives.

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+// MtpTestNodeConverterTestIdTests pins the test id algorithm by setting
+// VSTEST_TESTCASE_ID_ALGORITHM and resetting TestCase's cache of it. Both are process wide, so a
+// class level [DoNotParallelize] does not stop another class from observing the pinned value. This
+// assembly already runs sequentially because nothing opts it into parallelization; this states that
+// as a requirement rather than leaving it as an accident.
```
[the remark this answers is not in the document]
azat-msft: Fixed in 9e1c296, same as the ObjectModel.UnitTests one. The comment now names `VSTEST_DISABLE_XXHASH128_TESTCASE_ID` and the feature flag cache rather than `TestCase`'s own, which this branch removed.

docs/environment-variables.md
azat-msft: Done — implemented in 648db90.

The switch is now `VSTEST_DISABLE_XXHASH128_TESTCASE_ID`, registered in `FeatureFlag.cs` alongside the other `VSTEST_DISABLE_*` flags, and the string selector is gone.

The one wrinkle was that this release ships xxHash128 *available but not default*, so a plain "disable" flag would have meant disabling something that isn't on. `FeatureFlag` already supports that shape — `VSTEST_DISABLE_MTP_TESTHOST` does the same thing — so it carries `[VSTEST_DISABLE_XXHASH128_TESTCASE_ID] = true` in `DefaultValues` for now, and the release that makes xxHash128 the default deletes that one entry.

Naming the algorithm rather than "the new one" also means the polarity survives that flip: `1` selects SHA1 and `0` selects xxHash128 both before and after, so nothing anyone writes down changes meaning.

Two things fell out of the conversion: `TestCase`'s hand-rolled `Interlocked.CompareExchange` cache is gone, since `FeatureFlag`'s `ConcurrentDictionary.GetOrAdd` already hands every racing caller the value that won the add; and there's a new end-to-end acceptance test proving a runsettings-declared flag actually reaches the process that computes ids on the classic path.

src/Microsoft.TestPlatform.ObjectModel/Utilities/EqtHash.cs
```diff
@@ -39,4 +41,38 @@ public static Guid GuidFromString(string data)
...
+    /// scheme. See TestIdGuid.
+    /// </para>
+    /// <para>
+    /// This deliberately does NOT produce the same value as <see cref="GuidFromString(string)"/>.
+    /// Changing the id of a test is a breaking change for anything that stored it. The warning on
+    /// <see cref="GuidFromString(string)"/> about work item association requiring sign off from the
+    /// TC-TA feature owners applies to adopting this method.
+    /// </para>
+    /// </remarks>
+    public static Guid GuidFromString2(string data)
```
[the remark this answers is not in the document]
azat-msft: Agreed and renamed in cf1086c — `EqtHash.GuidFromString2` is now `EqtHash.GuidFromStringXxHash128`.

The `*2` suffix was carried over from the earlier draft and had the problem you describe: it only means anything while the thing it's numbered against still exists, and it says nothing about what the method does. Naming it after the algorithm fixes both, and leaves room for a future scheme to be named rather than numbered.

The sibling `TestIdProvider2` was renamed to `TestIdProviderXxHash128` in the same commit. `PublicAPI.Unshipped.txt` updated accordingly, so this landed while renaming was still free.

src/Microsoft.TestPlatform.AdapterUtilities/TestIdProvider2.cs
```diff
@@ -0,0 +1,114 @@
...
+/// Used to generate id for tests, using xxHash128.
+/// </summary>
+/// <remarks>
+/// This is the replacement for <see cref="TestIdProvider"/>, which uses SHA1. SHA1 is a
+/// cryptographic hash being used for a non-cryptographic purpose; it is slower than necessary and
+/// its presence trips security tooling. The ids produced here are RFC 9562 version 8 UUIDs that
+/// carry the version of the hashing scheme, so a future change to the algorithm is detectable
+/// from the id itself.
+/// </remarks>
+public class TestIdProvider2
```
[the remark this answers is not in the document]
azat-msft: Agreed and renamed in cf1086c — `TestIdProvider2` is now `TestIdProviderXxHash128`, and the file was renamed to match.

Same reasoning as the `GuidFromString2` thread: a numeric suffix is only meaningful while the type it's numbered against exists, and the SHA1 original is intended to go away eventually. Naming it after the algorithm also makes the pairing with `TestIdProvider` legible at the call site.

Declared in `PublicAPI.Unshipped.txt`, so this happened while the type was still unshipped and the rename cost nothing.

docs/environment-variables.md
```diff
@@ -133,6 +133,15 @@ This document lists environment variables that are currently handled by VSTest s
 - **Values**: Set to `0` to opt in to the experimental MTP testhost; any other value keeps it disabled
 - **Example**: `VSTEST_DISABLE_MTP_TESTHOST=0`
 
+### VSTEST_DISABLE_XXHASH128_TESTCASE_ID
```
nohwnd, line 136: this is kinda huge compared to the other sections here, and spells out a lot of details other entries don't spell out. e.g. taht we set the env variable once and it stays the same etc.

eng/Versions.props
```diff
@@ -70,6 +70,11 @@
     <_MicrosoftVSSDKBuildToolsVersion_>17.14.2119</_MicrosoftVSSDKBuildToolsVersion_>
     <MicrosoftWin32RegistryVersion>5.0.0</MicrosoftWin32RegistryVersion>
     <NewtonsoftJsonVersion>13.0.4</NewtonsoftJsonVersion>
+    <!-- Bumping this requires checking the assembly version this package emits against the
```
nohwnd: why is this relevant to this PR?

test/Microsoft.TestPlatform.ObjectModel.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+// Several classes here pin the test id algorithm by setting VSTEST_DISABLE_XXHASH128_TESTCASE_ID and
+// resetting the feature flag cache. Both are process wide, so a class level [DoNotParallelize]
+// does not stop a class that pins a different value from running at the same time. This assembly
+// already runs sequentially because nothing opts it into parallelization; this states that as a
+// requirement rather than leaving it as an accident.
+[assembly: DoNotParallelize]
```
nohwnd, line 11: okay, but make issue for reverting this, because it feels like for unit tests we should not really set the real env var and instead of that use some IEnvironmentProvider to test it?

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+// MtpTestNodeConverterTestIdTests pins the test id algorithm by setting
+// VSTEST_DISABLE_XXHASH128_TESTCASE_ID and resetting the feature flag cache. Both are process wide,
+// so a class level [DoNotParallelize] does not stop another class from observing the pinned value.
+// This assembly already runs sequentially because nothing opts it into parallelization; this states
+// that as a requirement rather than leaving it as an accident.
+[assembly: DoNotParallelize]
```
nohwnd, line 11: okay, but make issue for reverting this, because it feels like for unit tests we should not really set the real env var and instead of that use some IEnvironmentProvider to test it? 

and then ideally the tests would run in parallel, or we should prove that they cannot for real reasons (like sharing a static class we cannot remove)

test/Microsoft.TestPlatform.ObjectModel.UnitTests/Properties/AssemblyInfo.cs
nohwnd: and then ideally the tests would run in parallel, or we should prove that they cannot for real reasons (like sharing a static class we cannot remove)

eng/Versions.props
azat-msft: removed.

docs/environment-variables.md
azat-msft: shortened.

test/Microsoft.TestPlatform.ObjectModel.UnitTests/Properties/AssemblyInfo.cs
azat-msft: the issue is created: https://github.com/microsoft/vstest/issues/16433

test/Microsoft.TestPlatform.CrossPlatEngine.UnitTests/Properties/AssemblyInfo.cs
azat-msft: the issue is created, see the previous comment for the link

test/Microsoft.TestPlatform.Acceptance.IntegrationTests/TestCaseIdAlgorithmScenarioTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            "available but not default.");
+    }
+
+    /// <summary>
+    /// xxHash128 ids are RFC 9562 version 8 UUIDs carrying the hashing scheme version in the top
+    /// nibble, so the string form starts with the scheme version and its third group starts with the
+    /// UUID version. SHA1 ids are unversioned and match neither.
+    /// </summary>
+    private static bool IsXxHash128Id(string id)
+        => id[0] == '1' && id.Split('-')[2][0] == '8';
```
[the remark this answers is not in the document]
azat-msft: Fixed in 26a9579, though not quite the way you suggested — answering `false` for an unparseable id would have made the failure worse, not better.

The opted-out arm asserts `Assert.IsFalse(IsXxHash128Id(id))`. If a malformed id answered `false`, that arm would have **passed** on garbage. So the test would still have hidden the problem, just silently instead of loudly.

Instead the ids are parsed where they are read, one per `UnitTest` element, with a message that names the bad value:

```
A UnitTest element in '<path>' has id '<value>', which is not a guid, so its test case id cannot be read.
```

`Guid.TryParse` also covers the missing-attribute case, so the separate null check is gone. `IsXxHash128Id` now takes a `Guid` and reads the dashed form, which is fixed width — it cannot throw, and there is no unparseable input left for it to have an opinion about.

Verified: the class passes 5/5 together with the binding-redirect guards.

nohwnd (approved):

Issue #15739, opened 2026-05-08
Microsoft.TestPlatform 18.5.0 net462 Extensions layout ships System.Text.Json without companion dependencies


quasarea:
## Summary

`Microsoft.TestPlatform 18.5.0` installs a `net462` VsTest Extensions layout that contains `System.Text.Json.dll` but not the companion assemblies it depends on for .NET Framework resolution. On hosted Azure DevOps runs using `VisualStudioTestPlatformInstaller@1` plus `VSTest@2` with code coverage, this causes runtime assembly-load failures unless the missing DLLs are staged manually next to the VsTest extensions.

This looks related to, but different from, #15297. That issue is about `Microsoft.CodeCoverage` copying `System.Text.Json.dll` into output/publish layouts. This issue is about the installed VsTest `tools\\net462\\...\\Extensions` runtime layout itself not being a complete dependency closure.

## Repro shape

- Azure DevOps hosted agent (`windows-latest` / Windows Server 2025 repro confirmed)
- `VisualStudioTestPlatformInstaller@1`
- `testPlatformVersion: 18.5.0`
- `VSTest@2`
- code coverage enabled
- native / .NET Framework-oriented test execution path that loads the `net462` VsTest extensions

## What we observed

The installed TestPlatform package contains:

- `tools\\net462\\Common7\\IDE\\Extensions\\TestPlatform\\Extensions\\System.Text.Json.dll`

But it does not contain these companion assemblies in the package/runtime layout:

- `Microsoft.Bcl.AsyncInterfaces.dll`
- `System.Text.Encodings.Web.dll`
- `System.Threading.Tasks.Extensions.dll`

We inspected the extracted packages directly:

### `Microsoft.TestPlatform 18.5.0`

- `System.Text.Json.dll`: present
- `Microsoft.Bcl.AsyncInterfaces.dll`: absent
- `System.Text.Encodings.Web.dll`: absent
- `System.Threading.Tasks.Extensions.dll`: absent

### `Microsoft.CodeCoverage 17.8.0`

- `System.Text.Json.dll`: absent
- `Microsoft.Bcl.AsyncInterfaces.dll`: absent
- `System.Text.Encodings.Web.dll`: absent
- `System.Threading.Tasks.Extensions.dll`: absent

Separately, the `System.Text.Json 6.0.11` nuspec declares these dependencies for `.NETFramework4.6.1`:

- `Microsoft.Bcl.AsyncInterfaces`
- `System.Text.Encodings.Web`
[...]

PR #16383, opened 2026-08-18
Report test extension load failures to the user

### LINKED ISSUE #15577: STJ migration: DesignMode/wrapper communication hangs during test execution

nohwnd:
`TestPluginDiscoverer` only told the user about a failing extension file when `Assembly.Load` threw `FileLoadException`. Two other paths were silent and wrote to `EqtTrace` only: the general `Exception` handler in `GetTestExtensionsFromFiles`, which is the one that catches the `FileNotFoundException` thrown when an extension or one of its dependencies cannot be found, and the `ReflectionTypeLoadException` handler in `GetTestExtensionsFromAssembly`, which carries on with a half-loaded type list.

So an adapter that half-loaded gave the user fewer tests than expected, or a hang, with no way to find out why short of re-running with `/diag`. Both now report through `TestSessionMessageLogger` using the existing, already localised `FailedToLoadAdapaterFile` message, so no `.resx` or `.xlf` change is needed.

Scanning stays best effort. Nothing throws and nothing aborts, a partially loaded assembly is still scanned for every extension type, each file is reported once per run, and the two C++ UWP adapters that are probed speculatively when no extension was found are not reported because they are absent everywhere except UWP.

This is the diagnostic half of #15577. The packaging cause behind the report on that issue, the missing net462 companion assemblies, was fixed separately in #15739 and shipped in 18.9.0, so this does not close #15577 on its own.

Verified: `build.cmd -c Release` compiles clean, `Microsoft.TestPlatform.Common.UnitTests` passes on net481 and net11.0, and a locally built `vstest.console` reproduces both failure modes end to end.

Related to #15577

🤖

🤖

### DIFF
--- a/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
+++ b/src/Microsoft.TestPlatform.Common/ExtensionFramework/TestPluginDiscoverer.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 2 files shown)


nohwnd:
Closing this one: #16382 is the same work and has more of it. Same two files, plus the TestPluginCache change this one is missing.

Both were opened three minutes apart by my automation, which was building a second pull request for work that already had one. That is fixed.

🤖

PR #16397, opened 2026-08-23
docs: fix typo explicitely -> explicitly

vaibhav8a:
One-word typo fix in `docs/analyze.md:135`: "A specific DataCollector can be explicitely..." -> "explicitly".

I left the matching spellings in `docs/RFCs/` and `docs/releases.md` alone, since those are historical records.

### DIFF
--- a/docs/analyze.md
+++ b/docs/analyze.md
@@ -132,7 +132,7 @@ below runsettings:
 </RunSettings>
 ```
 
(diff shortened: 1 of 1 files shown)


azat-msft:
Thank you for noticing the typo!

azat-msft (approved):

PR #16439, opened 2026-09-03
Upgrade gh-aw to v0.88.2 and recompile agentic workflows

azat-msft:
Every agentic workflow in this repo is currently broken, in one of two ways. Both come from the pinned gh-aw compiler (v0.81.6). This upgrades it to v0.88.2 and recompiles.

### 1. Wrong default model — every run that reaches the agent step

The compiled workflows fall back to `claude-sonnet-4.6`, which the Copilot API no longer serves:

```
400 The requested model is not available for integrator "agentic-workflows".
```

The harness retries four times and gives up. Code Simplifier, Malicious Code Scan, Repository Quality Improver, Daily File Diet and the rest all die here. Runs that look green never reached the agent — `pre_activation` gated them out.

v0.88.2 defaults to `auto` and lets the server pick, so the compiled default can no longer go stale.

### 2. Invalid workflow file — PR Iteration and Issue Repro Triage never start

Both generated a step-level condition that reads the `secrets` context:

```yaml
if: ${{ vars.APP_ID != '' && secrets.APP_PRIVATE_KEY != '' }}
```

`secrets` is not available in `if:`, so GitHub rejects the file. That is why these two produce zero-job failed runs on `push` and `pull_request` — events they do not even declare. I confirmed it by reproducing in a fork and watching the failure disappear once the `secrets` reference was removed.

v0.88.2 routes the value through step `env` instead:

```yaml
env:
  GH_AW_IGNORE_IF_MISSING_PRIVATE_KEY: ${{ secrets.APP_PRIVATE_KEY }}
if: ${{ vars.APP_ID != '' && env.GH_AW_IGNORE_IF_MISSING_PRIVATE_KEY != '' }}
```

### Hand-written changes

`pr-expert-reviewer.md` is the only `.md` I touched. v0.88.2 strict mode requires `tools.bash` to be explicit when `github.min-integrity` is `none`, and then requires `tools.cli-proxy` too. Both are set to `false`, which is what the workflow already effectively had — its old lock file granted no bash access.

Everything else is generated output from `gh aw compile`. Note that gh-aw also dropped `merge=ours` from the `.gitattributes` entry for `*.lock.yml`.

### Still broken, not fixed here

Markdown Linter has failed at startup on every scheduled run since mid-August. The file itself is valid and starts fine in a personal fork, and it is the only workflow using a third-party action (`super-linter/super-linter`). It looks like the org allowed-actions policy blocks it, but the policy API returns 403 for me, so someone with org admin needs to confirm.


azat-msft:
Superseded by a branch on the upstream repo so the Expert Code Reviewer workflow can run (it skips fork PRs).

PR #16462, opened 2026-09-09
Fix dotnet host resolution under proot (ARM64/Termux)

### LINKED ISSUE #16446: dotnet test fails on Linux under proot when the current process is not detected as the dotnet muxer (ARM64 'dotnet host' not found)
### Describe the bug

Running `dotnet test` inside an aarch64 Debian rootfs under **proot** (Android Termux) fails with:

```
找不到“ARM64”体系结构的“dotnet”主机。
(Could not find a 'dotnet' host for the 'ARM64' architecture.)
```

`uname -m` = `aarch64`, RID `linux-arm64`. Reproduced with .NET SDK 8.0.424 and 10.0.400.

### Root cause

In `DotnetHostHelper.TryGetDotnetPathByArchitecture` (`src/Microsoft.TestPlatform.CoreUtilities/Helpers/DotnetHostHelper.cs`) the fast path only applies when the current process *is* the dotnet muxer:

```csharp
if (_processHelper.GetCurrentProcessArchitecture() == targetArchitecture)
{
    string currentProcessFileName = _processHelper.GetCurrentProcessFileName()!;
    if (Path.GetFileName(currentProcessFileName) == _muxerName)
    {
        muxerPath = currentProcessFileName;
        ...
        return true;
    }
}
```

On Unix `GetCurrentProcessFileName()` comes from `Process.MainModule.FileName`, which .NET derives from the **first mapping in `/proc/self/maps`**. Under proot an empty placeholder `loader` file is injected at the lowest address of every process, so `MainModule.FileName` returns

`/data/data/com.termux/files/usr/libexec/proot/loader`

even though the real executable is `/usr/share/dotnet/dotnet` (probe from inside the same process):

```
Environment.ProcessPath      = /usr/share/dotnet/dotnet
/proc/self/exe               = /usr/share/dotnet/dotnet
Process.MainModule.FileName  = /data/data/com.termux/files/usr/libexec/proot/loader
maps[0]                      = .../usr/libexec/proot/loader  (injected placeholder)
```

Because the basename is not `dotnet`, resolution falls through to the search branch (`DOTNET_ROOT_ARM64` / `DOTNET_ROOT`, global registration, default installation location). Every candidate is then rejected because `IsValidArchitectureMuxer` only inspects PE headers on Windows and Mach-O on macOS — on Linux it never determines an architecture (`actual ''`; see #4042, closed as By Design). Even a valid, explicitly configured muxer is refused:

```
DotnetHostHelper.IsValidArchitectureMuxer: Incompatible architecture muxer, target architecture 'ARM64', actual ''
DotnetHostHelper: Invalid muxer resolved using env var key 'DOTNET_ROOT' in '/usr/share/dotnet'
```

#4042 was closed as By Design because on regular Linux the current-process fast path always succeeds. This environment breaks that assumption: the process **is** the muxer, but the `/proc/self/maps`-based image-name check reports a different name, so vstest lands in a search branch that can never validate a muxer on Linux.

### Expected behavior / suggested fixes

1. Do not rely on the first `/proc/self/maps` mapping to decide "current process is the muxer" — prefer `Environment.ProcessPath` / `/proc/self/exe` for `GetCurrentProcessFileName()`. (These return the correct `/usr/share/dotnet/dotnet` in the same environment.)
2. Implement ELF `e_machine` inspection in `IsValidArchitectureMuxer` on Linux, mirroring the PE (Windows) and Mach-O (macOS) paths, so the search branch can actually validate real muxers.
3. At minimum, when an arch-specific `DOTNET_ROOT_ARM64` (or `DOTNET_ROOT`) explicitly points to an existing muxer, honour it instead of failing the check on a platform where the validator always reports "incompatible".

### Steps to reproduce

1. On an aarch64 Linux rootfs running inside Termux proot, install the .NET SDK (`linux-arm64`).
2. Create a minimal xUnit project and run `dotnet test -c Release`.
3. Observe the failure and the trace above with `--diag`.

Environment: OS Debian 13 (trixie) aarch64 inside proot; runtime RID `linux-arm64`; .NET SDK 8.0.424 and 10.0.400.

### Workaround

Run tests with the `xunit.runner.console` runner directly (bypasses vstest) — this confirms the failure is isolated to vstest's dotnet-host resolution.

Related: #4042, #4194, #4349, #1749

azat-msft:
Fixes #16446

Under proot, `Process.MainModule.FileName` can return proot's `loader` path instead of the running `dotnet` executable. VSTest then fails to recognize `dotnet` and reports that it cannot find a host for the requested architecture.

Use `Environment.ProcessPath` on .NET to get the executable path. Keep the existing lookup for .NET Framework and as a fallback when `ProcessPath` is null.

### DIFF
--- a/test/vstest.console.UnitTests/ProcessHelperTests.cs
+++ b/test/vstest.console.UnitTests/ProcessHelperTests.cs
@@ -35,16 +35,21 @@ public class ProcessHelperTests
 
         string? currentProcessFileName = processHelper.GetCurrentProcessFileName();
 
-        Assert.IsNotNull(currentProcessFileName, "The running executable must always be identifiable.");
+        using var currentProcess = Process.GetCurrentProcess();
+        string? mainModuleFileName = currentProcess.MainModule?.FileName;
+
 #if NET
         // On .NET we prefer Environment.ProcessPath, because MainModule can report an injected loader
         // rather than the running executable under sandboxes such as proot. See issue #16446.
-        Assert.AreEqual(Environment.ProcessPath, currentProcessFileName);
+        // ProcessPath is nullable, so mirror the fallback the production code performs instead of
+        // assuming it is always set.
+        string? expectedFileName = Environment.ProcessPath ?? mainModuleFileName;
 #else
         // .NET Framework has no Environment.ProcessPath, so the MainModule behavior must be preserved.
-        using var currentProcess = Process.GetCurrentProcess();
-        Assert.AreEqual(currentProcess.MainModule?.FileName, currentProcessFileName);
+        string? expectedFileName = mainModuleFileName;
 #endif
+
+        Assert.AreEqual(expectedFileName, currentProcessFileName);
     }
 
     [TestMethod]
--- a/test/vstest.console.UnitTests/ProcessHelperTests.cs
+++ b/test/vstest.console.UnitTests/ProcessHelperTests.cs
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#if NET
+using System;
+#endif
 using System.Diagnostics;
 using System.Threading;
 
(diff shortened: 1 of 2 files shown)


test/vstest.console.UnitTests/ProcessHelperTests.cs
```diff
@@ -11,17 +14,39 @@
...
+        Assert.IsNotNull(currentProcessFileName, "The running executable must always be identifiable.");
+#if NET
+        // On .NET we prefer Environment.ProcessPath, because MainModule can report an injected loader
+        // rather than the running executable under sandboxes such as proot. See issue #16446.
+        Assert.AreEqual(Environment.ProcessPath, currentProcessFileName);
+#else
+        // .NET Framework has no Environment.ProcessPath, so the MainModule behavior must be preserved.
+        using var currentProcess = Process.GetCurrentProcess();
+        Assert.AreEqual(currentProcess.MainModule?.FileName, currentProcessFileName);
+#endif
```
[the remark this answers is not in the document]
azat-msft: Good catch, fixed in 40107d8e8.

You are right that the assertion contradicted the production fallback. `Environment.ProcessPath` is documented as nullable, and `GetCurrentProcessFileName()` deliberately falls back to `MainModule` when it is null, so a host where `ProcessPath` is null would have failed the test even though the code behaved correctly.

The test now computes the expected value the same way the production code does, instead of assuming `ProcessPath` is set:

```csharp
#if NET
    string? expectedFileName = Environment.ProcessPath ?? mainModuleFileName;
#else
    string? expectedFileName = mainModuleFileName;
#endif

Assert.AreEqual(expectedFileName, currentProcessFileName);
```

I also dropped the separate `Assert.IsNotNull` on the result. If both `ProcessPath` and `MainModule` were null, returning null is the contractually correct behavior, so that assertion could have produced the same class of false failure.

test/vstest.console.UnitTests/ProcessHelperTests.cs
```diff
@@ -11,17 +14,44 @@
...
+
+        using var currentProcess = Process.GetCurrentProcess();
+        string? mainModuleFileName = currentProcess.MainModule?.FileName;
+
+#if NET
+        // On .NET we prefer Environment.ProcessPath, because MainModule can report an injected loader
+        // rather than the running executable under sandboxes such as proot. See issue #16446.
+        // ProcessPath is nullable, so mirror the fallback the production code performs instead of
+        // assuming it is always set.
+        string? expectedFileName = Environment.ProcessPath ?? mainModuleFileName;
```
Evangelink, line 46: This test does not exercise the regression on normal CI hosts. `Environment.ProcessPath` and `Process.MainModule.FileName` normally resolve to the same executable, so the pre-change implementation—which always returned `MainModule.FileName`—also satisfies this assertion. Because the expected value mirrors the production fallback, the test only distinguishes the implementations when it already runs under an environment reproducing the proot discrepancy.

Please add a controllable seam and test distinct values explicitly (prefer `ProcessPath` when available, and fall back to `MainModule` when it is null), or add an automated proot scenario; otherwise this regression can return without the test failing.

azat-msft: I changed the `GetCurrentProcessFileName()` function to have internal parameterized overload, so it can be tested in unit tests. 

Creating a separate arm64 leg that would test the behavior in different infrastructure under proot looks like too big of a change for this PR.

Evangelink (approved):

PR #16476, opened 2026-09-11
[rel/18.10] Isolate code coverage files in published output

azat-msft:
Addresses the publish failures in microsoft/testfx#11178.

Publish coverage tools and their private dependencies under Microsoft.CodeCoverage instead of the application root. Keep those tools out of the WASM bundle and document the one-time cleanup needed for existing publish folders.


azat-msft:
Fixed the PublishDir edge case from the latest review in 1338ece4. The WASM exclusion now uses EnsureTrailingSlash before adding the coverage subfolder. The existing regression tests exercise paths both with and without the final separator; they failed before the fix and pass with it. The matching main change is 225875bf in #16483.

nohwnd (approved):

PR #16477, opened 2026-09-11
[rel/18.10] Preserve application dependencies when publishing coverage

azat-msft:
Addresses the publish failures in microsoft/testfx#11178.

Keep application-selected DLLs when coverage files target the same publish path. Replace stale conflicting output regardless of timestamps and avoid copying private assemblies supplied by the application runtime.


azat-msft:
Different, less invasive approach implemented in https://github.com/microsoft/vstest/pull/16476 was chosen. Closing this PR.

PR #16488, opened 2026-09-15
Use automatic models for agentic workflow subagents

nohwnd:
Remove the `claude-opus-4.6` overrides from the build-failure analyst and MSBuild reviewer prompts so the subagents use automatic model selection. Omit `task.model` because the task tool does not accept `model: "auto"`.

Recompiled the three affected definitions with `gh-aw v0.88.2`. Only the prompt body hashes changed in the generated files. Strict compilation and reproducibility checks pass.

🤖


Evangelink (approved):
Looks good. Automatic subagent model selection keeps Copilot inference on the organization-billed `GITHUB_TOKEN` path through `copilot-requests: write`.

PR #16495, opened 2026-09-17
Allow jq in expert reviewer and document its tool sandbox

### LINKED ISSUE #16494: [aw] Expert Code Reviewer 🧠 timed out
### Workflow Failure

**Workflow:** [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/blob/16492/merge/.github/workflows/pr-expert-reviewer.md)  
**Branch:** dependabot/github_actions/github-actions-5ab709ea60  
**Run:** https://github.com/microsoft/vstest/actions/runs/35223957076  
**Pull Request:** [#16492](https://github.com/microsoft/vstest/pull/16492)

**Agent Timed Out**: The agent job exceeded the maximum allowed execution time (15 minutes).

To increase the timeout, add or update the `timeout-minutes` setting in your workflow's frontmatter:

```yaml
---
timeout-minutes: 25
---
```

### Action Required

**Assign this issue to an agent** to debug and fix the issue.

<details>
<summary>Optimize token consumption</summary>

This failure was triggered by a guardrail limit (max-turns / timeout). Use this prompt with any coding agent (GitHub Copilot, Claude, Gemini, etc.) to analyze token usage and reduce costs:

````
Optimize the agentic workflow token consumption using https://raw.githubusercontent.com/github/gh-aw/main/optimize.md

The workflow run is at https://github.com/microsoft/vstest/actions/runs/35223957076
````

</details>

<details>
<summary>Debug with any coding agent</summary>

Use this prompt with any coding agent (GitHub Copilot, Claude, Gemini, etc.):

````
Debug the agentic workflow failure using https://raw.githubusercontent.com/github/gh-aw/main/debug.md

The failed workflow run is at https://github.com/microsoft/vstest/actions/runs/35223957076
````

</details>

<details>
<summary>Manually invoke the agent</summary>

Debug this workflow failure using your favorite Agent CLI and the `agentic-workflows` prompt.

- Start your agent
- Load the `agentic-workflows` skill from `.github/skills/agentic-workflows/SKILL.md` or <https://github.com/github/gh-aw/blob/main/.github/skills/agentic-workflows/SKILL.md>
- Type `debug the agentic workflow pr-expert-reviewer failure in https://github.com/microsoft/vstest/actions/runs/35223957076`

</details>

> [!TIP]
> <details>
> <summary>Stop reporting this workflow as a failure</summary>
>
> To stop a workflow from creating failure issues, set `report-failure-as-issue: false` in its frontmatter:
> ```yaml
> safe-outputs:
>   report-failure-as-issue: false
> ```
>
> </details>

> Generated from [Expert Code Reviewer 🧠](https://github.com/microsoft/vstest/actions/runs/35223957076) · copilot · 311.4 AIC · [◷](https://github.com/search?q=repo%3Amicrosoft%2Fvstest+is%3Aissue+%22gh-aw-workflow-id%3A+pr-expert-reviewer%22&type=issues)
> - [x] expires  on Sep 24, 2026, 1:16 PM UTC

### PR DESCRIPTION
**Pull request by azat-msft**
Fixes #16494.

The Expert Code Reviewer workflow timed out on #16492 and published nothing. It wasn't a slow review — the agent could not read the diff it was given.

A GitHub MCP response over 512KB (`MCP_GATEWAY_PAYLOAD_SIZE_THRESHOLD`) isn't returned inline. gh-aw spills it to `/tmp/gh-aw/mcp-payloads/.../payload.json` as a **single line** of escaped JSON. #16492's diff is 883KB, so it spilled. The agent reads files by line range, so a one-line file is all-or-nothing — every read came back `1 line read` and told it nothing.

With `tools.bash: false` and `--no-ask-user` there was no permitted way to decode that line and no way to escalate. The agent tried `python3`, `node`, `jq`, `perl`, `sed`, `awk`, `tr`, `xargs`, `find`, `curl` and `gh`, all denied — 45 of ~117 tool calls. It gave up on the payload after ~10 minutes, loaded the reviewer checklist at the 12-minute mark, and was killed just as it was about to start reviewing.

## Changes

- Allow `jq`. That's the fix: `jq -r` decodes the escaped newlines back into real lines, so the payload can be paged normally. `jq` is preinstalled on `ubuntu-latest` and the agent chroots over the runner's `/usr`. It isn't a Copilot CLI stem command, so the bare `shell(jq)` form prefix-matches arguments without a `:*` wildcard.
- Add a **Tools Available to You** section: `jq` is the one granted shell command, a denial is final so don't go looking for another spelling, and how to extract text from a payload file. Since a spilled payload is over 512KB by definition and can still exceed context once readable, it also tells the agent to narrow with `jq` file by file rather than printing the whole blob, and to report `missing_data` rather than quietly reviewing part of a change.
- One-line clarification on Step 6 to use the native `write` tool for the cache-memory files.

`tools.bash: false` stays — switching to a `bash:` allowlist would pull in gh-aw's implicit command grants, which this workflow avoids on purpose.

The `jq` filters were checked locally against four payload shapes: single-element array, multi-element array, bare object, and an object with no `content` envelope.

## Not addressed here

gh-aw has a `GH_AW_MAX_TOOL_DENIALS` guardrail (default 5) that stops inference after repeated denials, but it only applies in SDK driver mode. This workflow runs in CLI mode, which is why 46 denials went unchecked. Worth a separate look.

### CODE REVIEW

**Comment by azat-msft** [comment]
Addressed in 4a23a6e2e.

You're right that the prompt promised utilities that weren't in the generated allowlist — it was leaning on Copilot CLI's built-in defaults, which aren't visible in the lock file and could drift. Fixed by granting them explicitly:

```
shell(cat) shell(ls) shell(grep) shell(head) shell(tail) shell(wc)
shell(git diff:*) shell(git log:*) shell(git show:*) shell(git merge-base:*)
```

The prompt now says the list matches `engine.args` exactly, and the frontmatter comment says to keep them in sync.

Two details in the suppressed comment were off, for the record:

**`git` was not denied.** Read-only git already worked via CLI defaults. In the run that timed out ([35223957076](https://github.com/microsoft/vstest/actions/runs/35223957076)) the agent ran `git diff d4566fa..a467f03 -- .github/aw/actions-lock.json` and got 32 lines back, plus several successful `git show`/`git log`/`git status` calls. The only denied git command was `git fetch` — network, not read-only. It's explicit now regardless, which is the better state.

**`fetch-depth: 1` isn't what governs this.** Line 241 is the activation job's sparse checkout. The agent job re-checks-out and then runs `checkout_pr_branch.cjs`, which sets `fetchDepth = commitCount + 1` with the comment *"+1 to include the merge base"*.

That said, it exposed a real bug in my wording: that guarantees the **merge base**, not `base.sha`. Those coincide only while the base branch hasn't moved. It happened to hold in the run above, so the original text would have worked there and broken later. The fallback now resolves the boundary from `git log` instead:

```bash
git log --oneline            # oldest entry is the merge base
git diff <merge-base>..HEAD
```

and tells the agent that a missing base commit is expected from the shallow fetch — go back to the payload file and `jq`, since `git fetch` is denied.

azat-msft:
Good catch, both parts were real. Fixed in cf79f2a9c.

**`write` vs the restriction.** "Anything else is denied" was meant to scope to shell commands but read as absolute, and it contradicted both the `write` grant and Step 6. Now scoped to "No other **shell command** is permitted", with an explicit paragraph for the native read/write tools, and Step 6 points at `write` rather than leaving a shell redirect as the obvious reading.

**The stale line 319.** That predates this PR and did conflict with the new list. Replaced the "these two wrappers are the only allowed shell commands" claim with a pointer to the tools section, keeping the part that still holds (prefer native MCP tools, don't invoke a shell interpreter or the bridge script directly).

azat-msft:
Fixed in 2fc795740 by removing the git fallback entirely.

Both of your concerns were about the same thing, and it was scope creep on my part. The fix for the timeout is `jq` on the payload file; the git fallback was a convenience I bolted on, and it brought a wildcard grant and an assumption I can't fully guarantee.

- **Wildcard writes.** Correct — `git` is a stem command, so allowing arguments needs `:*`, and `shell(git diff:*)` would also admit `git diff --output=<file>`. Gone. Every remaining grant (`jq`, `cat`, `ls`, `grep`, `head`, `tail`, `wc`) is a non-stem command, so the bare `shell(name)` form prefix-matches arguments without a wildcard, and none of them write.
- **Merge base.** `checkout_pr_branch.cjs` does fetch `commitCount + 1` specifically to include the merge base, so it usually holds. But "usually" isn't good enough for an instruction the agent will follow blind, and PR branches carrying merge commits from main break the assumption. Removed rather than hedged.

In its place the prompt now keeps the agent inside the sandbox when the payload is awkward: ask for per-file contents through the `github` tools, and if that fails report `missing_data` instead of guessing or quietly reviewing part of the change.

Net effect is a smaller PR than before: allow `jq` plus read-only inspection commands, and document the sandbox so the agent stops probing for tools it will never get.

azat-msft:
Partly right — fixed in 9445de3ea.

**The `.[0]` point is correct** and I've fixed it. The filter now iterates every array element and tolerates a missing `content` field.

**The envelope claim isn't.** The payload *is* an MCP `content[]` envelope. Two independent pieces of evidence from run [35223957076](https://github.com/microsoft/vstest/actions/runs/35223957076):

- The gateway log reports `extract_mcp_response: top-level keys=["_meta", "content", "resultType"]`.
- The agent, working blind without `jq`, ended up targeting the payload's literal tail with sed: `","type":"text"}],"resultType":"complete"}]`. Read backwards that is `[ { "content": [ {"text": ..., "type":"text"} ], "resultType":"complete" } ]`.

Rather than argue the point, the guidance is now shape-agnostic. Inspect first:

```bash
jq 'if type == "array" then .[0] else . end | keys' <payload>
```

Extract, iterating all elements and tolerating a missing `content`:

```bash
jq -r '(if type == "array" then .[] else . end) | .content[]?.text' <payload>
```

And a recursive-descent fallback if the envelope ever differs:

```bash
jq -r '[.. | objects | select(.type == "text") | .text] | join("\n")' <payload>
```

I ran these locally against four shapes: single-element array, multi-element array, bare object, and an object with no `content` envelope. The `keys` output matches what the gateway logged, the primary filter returns both parts for a multi-element array, and the fallback recovers text from the no-envelope shape where the primary correctly returns nothing.

azat-msft:
Fair — fixed in b80271741. `safeoutputs` publishes the review comments and verdict, so "everything granted here is read-only" was wrong.

The frontmatter comment now scopes that claim to the inspection commands and calls out `safeoutputs` as write-capable by design, with its writes constrained by the `safe-outputs` config (5 inline comments max, one review, `APPROVE` disallowed).

Also tightened the prompt list so the agent sees the same split: `github` marked read-only, `safeoutputs` marked as the only way to write anything back to the PR.

azat-msft:
Fixed in 5a7154d94 — the comment now says every other *shell* grant is read-only and names the native `write` tool that gh-aw emits for the Step 6 cache-memory files.

azat-msft:
Narrowed the allowlist to `jq` only, in fd24245f6.

Earlier rounds flagged that the prompt named utilities not present in `engine.args`. I resolved that by adding them (`cat`, `ls`, `grep`, `head`, `tail`, `wc`), which was the wrong direction — those commands already worked in the failed run, so granting them unlocked nothing. The sentence that named them was mine, and deleting it was the smaller fix.

So the shell surface is back to the two MCP wrappers plus `jq`, and the prompt no longer claims an inventory of commands this workflow doesn't grant.

I also tightened the rationale now that the mechanism is clearer. The blocker was never "no JSON parser" in the abstract — it's that a spilled payload is a **single line**, and the agent reads by line range, so `1 line read` is all it ever got. `jq -r` restores the newlines, which is what makes the file pageable.

One limitation now stated explicitly rather than glossed over: a spilled payload is over 512KB by definition, and #16492's diff is 883KB, so it can exceed context even once readable. The prompt tells the agent to narrow with `jq` file by file instead of printing the whole blob, and to report `missing_data` rather than silently reviewing part of a change.

0101 (approved):

PR #16496, opened 2026-09-17
Preserve OutputType for Android test projects

jonathanpeppers:
## Summary

Adding `<PackageReference Include="MSTest" Version="4.4.1" />` to an Android test project transitively brings in `Microsoft.NET.Test.Sdk` and its `.targets` file. That file forces `OutputType` to `Exe`, overriding Android’s `Library` output type.

Skip this override when `TargetPlatformIdentifier` is `android`, preserving the existing output type. WinUI/UWP exceptions and entry-point generation remain unchanged.

Found while working on jonathanpeppers/Microsoft.AndroidX.Compose#409, which currently works around the issue using explicit SDK imports and a late override:

```xml
<Project>
  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

  <!-- Existing project properties and package references -->

  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />

  <!-- Restore Android's library output after the test SDK resets it to Exe. -->
  <PropertyGroup>
    <OutputType>Library</OutputType>
  </PropertyGroup>
</Project>
```

This fix removes the need for that workaround.

## Validation

Validated MSBuild behavior for Android libraries and executables, case-insensitive platform matching, standard .NET projects, WinUI, and UWP.


src/package/Microsoft.NET.Test.Sdk/netcoreapp/Microsoft.NET.Test.Sdk.targets
```diff
@@ -16,9 +16,10 @@
     Output type for .NET Core test projects should be exe.
     https://devdiv.visualstudio.com/DevDiv/_workitems?id=375688&_a=edit
     Note: we skip this for WinUI and UWP apps, as they must use 'WinExe'.
+    We also skip Android projects, which may need to remain libraries.
```
jonathanpeppers, line 19: Note that Android is somewhat like Windows, we do something special with `$(OutputType)`. In Android's case, we  always use libraries with no `static void Main()` as Android's entry point is either a special `Activity` or `Application` class.

azat-msft (approved):

azat-msft:
Thank you for you contribution! The change looks fine, I'm merging it.

PR #16502, opened 2026-09-21
Report the test id hashing algorithm on every test case

azat-msft:
## Why

vstest computes `TestCase.Id` by hashing a seed string. That hash has always been SHA1; `main` now also has xxHash128, shipped available but not default behind `VSTEST_DISABLE_XXHASH128_TESTCASE_ID`. A later release flips the default, and at that point the id of every test whose id the platform computes changes.

Visual Studio's Test Explorer caches those ids. When results come back with ids computed a different way it cannot match them against its cache, and the symptom the user sees is that **it keeps showing the previous results** - the new ones are silently dropped. The fix on the VS side is to re-discover, but VS first has to know the algorithm changed. Nothing on the wire told it.

## The contract

Every `TestCase` now carries a property. A companion change in the VS repo codes against this:

| | |
| --- | --- |
| Key | `TestCase.IdAlgorithm` |
| Type | `string` |
| Public API | `TestCaseProperties.IdAlgorithm`, values as constants on `TestCaseIdAlgorithms` |

| Value | Meaning |
| --- | --- |
| `Sha1` | the platform hashed the id with SHA1 |
| `XxHash128` | the platform hashed the id with xxHash128 |
| `SelfAssigned` | the adapter assigned the id; the platform hashed nothing, so this id does not move |
| absent | a vstest that predates this property |

The vocabulary is deliberately the same as the `IdSource` column of the test id report logger, because they answer the same question - except the logger *infers* it by comparing ids, and this is *stated* by the code that produced the id.

## What changed

**The carrier.** An ordinary entry in `TestObject`'s property bag, not a core property. It therefore travels through the mechanism that already carries custom properties (traits, managed names) on every protocol version: no serializer knows about it, no version bump, and an older payload is simply missing it rather than carrying something wrong. Verified round-tripping through `JsonDataSerializer` on versions 1, 2, 3 and 7, which covers both the v1 and the v2+ converter, for a bare `TestCase`, a `TestResult`, and a `TestDiscovery.TestFound` payload.

**Where it is set.** In the `TestCase` constructor, from the algorithm this process will hash with. That is the only moment the answer is knowable without reading `Id` and forcing it to be computed, and it keeps the `Id` getter free of side effects. It is correct because the process that builds a test case is the process that hashes it: on the classic path the testhost, which receives the runsettings environment variables; on the MTP path the runner.

Assigning `Id` downgrades the record to `SelfAssigned`, but **only if this instance recorded something for itself**. That guard is what keeps deserialization honest - a test case built by the serialization constructor records nothing, so restoring its id cannot invent a claim its sender never made. Without it, a payload from an older testhost would arrive at a newer console reading `SelfAssigned`, which is precisely the value that tells a consumer its cached ids are safe.

**The MTP trap.** `MtpTestNodeConverter.ToTestCase` also assigns `Id` explicitly, because the runner has to hash on behalf of an application that is its own host. That is platform-computed with a known algorithm, not self-assignment, so the converter re-records the algorithm it used after assigning. Without that, every MTP test would be reported as an id that never moves and VS would never re-discover on that path.

**Emitted always, not conditionally.** It costs ~150 bytes per test case (+10% on a v1 discovery payload, +25% on v7). Emitting only non-default values is tempting but breaks at the one moment the property exists for: once the default is xxHash128, "omit the default" means platform-computed test cases go from absent to absent and the flip becomes undetectable. Pinning the omitted value to `Sha1` instead would work, but overloads "absent" with three different meanings. One byte-cost for one unambiguous contract.

**Not `[Experimental]`,** unlike the xxHash128 hashing itself. The algorithm choice is experimental; reporting which one was used is not - VS needs it permanently, across the flip and after it.

## Test changes worth a look

Adding a second bag property to test cases that already carry traits exposed that `TestObject`'s bag is a `ConcurrentDictionary`, whose enumeration order is not stable between processes. The pinned payloads were indexing into it positionally and only looked deterministic while there was at most one such property. `SerializationTestHelpers.AssertJsonEqual` now orders property bags before comparing, keeping the converter-written prefix in its written order and sorting the rest; `Minify` is untouched, so payloads fed back in as deserialization input still exercise the order they state.

The pinned payloads themselves gained the new field and nothing else - I checked by stripping the inserted text back out and diffing against `HEAD`.


Evangelink (commented):
I found three issues that can make the reported test ID algorithm inaccurate or leave the default behavior untested.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -85,6 +85,18 @@ public TestCase(string fullyQualifiedName, Uri executorUri, string source)
...
+        // for it: this instance's id has not been assigned, so it will be computed, and it will be
+        // computed with this process's algorithm - which is fixed for the lifetime of the process,
+        // so recording it now cannot disagree with the value Id eventually hashes with. Doing it
+        // here rather than from the Id getter also keeps that getter free of side effects.
+        //
+        // The serialization constructor deliberately does not do this. A test case being
+        // deserialized already carries whatever the sending process recorded, and a test case sent
+        // by a vstest that predates this property must keep carrying nothing rather than have this
+        // process invent an answer on its behalf.
+        SetPropertyValue(TestCaseProperties.IdAlgorithm, IdAlgorithmName(IdAlgorithm));
```
Evangelink, line 99: The public parameterless constructor never initializes `TestCase.IdAlgorithm`. Adapters commonly create `new TestCase()` and populate `FullyQualifiedName`, `ExecutorUri`, and `Source` afterward; when `Id` is then lazily computed, the test case still carries no algorithm metadata. That means consumers cannot detect the SHA1-to-xxHash128 ID transition for this construction path, which defeats the purpose of the property. The same path also misses `SelfAssigned` for adapter-assigned IDs. Please stamp the algorithm when an ID is computed/assigned (while preserving the serializer's ability to distinguish an absent legacy field), and add coverage for computed and assigned IDs on a parameterless instance.

src/Microsoft.TestPlatform.ObjectModel/TestCase.cs
```diff
@@ -113,7 +125,25 @@ public Guid Id
...
+            //
+            // This only ever downgrades a value this instance recorded for itself, which is what
+            // keeps deserialization honest: a test case built by the serialization constructor
+            // carries nothing here, so restoring its id cannot invent a claim its sender never made,
+            // and the recorded value then arrives with the rest of the property bag. The platform
+            // itself assigns an id in one place - the Microsoft.Testing.Platform converter, which
+            // computes the hash in the runner rather than letting this type compute it - and that
+            // is not self assignment, so it records the algorithm it used afterwards.
+            if (GetPropertyValue<string?>(TestCaseProperties.IdAlgorithm, null) is not null)
+            {
```
Evangelink, line 143: `Guid.Empty` is the sentinel that makes the getter compute the platform ID, but this setter records `SelfAssigned` for it. After `testCase.Id = Guid.Empty`, the eventual ID is SHA1/xxHash128-derived while the metadata says it was self-assigned. An assign-then-clear sequence also needs to replace the stale `SelfAssigned` value. Please make the empty assignment restore the active platform algorithm and reserve `SelfAssigned` for non-empty explicit IDs, with tests for both cases.

test/Microsoft.TestPlatform.Library.IntegrationTests/TranslationLayerTests/TestCaseIdAlgorithmPropertyTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        // keeps every other consequence of declaring variables identical across the three arms.
+        var runSettings =
+            "<RunSettings><RunConfiguration><EnvironmentVariables>" +
+            declaration +
+            "</EnvironmentVariables></RunConfiguration></RunSettings>";
+
+        // Clear the flag out of the inherited environment. vstest.console passes its own environment
+        // down to the testhost, so a developer who has the flag exported - exactly the person
+        // evaluating this feature - would otherwise see the not-declared arm pick their value up.
+        var wrapper = GetVsTestConsoleWrapper(new Dictionary<string, string?> { [FeatureFlagName] = null });
```
Evangelink, line 86: This `notDeclared` case does not actually omit the variable. The process helper assigns the dictionary entry to `ProcessStartInfo.EnvironmentVariables`; a null value becomes `VSTEST_DISABLE_XXHASH128_TESTCASE_ID=`. `FeatureFlag` treats an empty-but-present variable as enabled, so this run exercises the SHA1 opt-out again instead of the default xxHash128 behavior. Please add a helper path that removes/omits the key from the child environment and use it here, otherwise a default-selection regression can pass unnoticed.

